PR #14, opened 2019-01-08
Update MOTIVATION.md

romgrk:


bryphe:
Thanks @romgrk ! Good idea to add the link to the ongoing investigation in Neovim 👍

PR #25, opened 2019-02-05
[WIP] Initial status bar state

bryphe:
This starts connecting Neovim's state with Oni2's internal editor model. This picks up the `mode_changed` notification, converts it to Oni2's action, and updates the internal state + re-renders.

Next steps:
- Switch the current 'tab state' to  be from the App state, and not hardcoded in the Tab component
- Start keeping buffer state, so we can actually render a live buffer instead of 'Hello World'
- Track buffer cursor position

### DIFF
--- a/src/editor/Core/State.re
+++ b/src/editor/Core/State.re
@@ -1,0 +1,28 @@
+/*
+ * State.re
+ *
+ * Top-level state of the editor
+ */
+
+module Mode {
+    type t =
+    | Insert
+    | Normal
+    | Other;
+
+    let show = (v) => switch(v) {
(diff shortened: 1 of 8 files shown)


src/editor/Core/State.re
```diff
@@ -0,0 +1,23 @@
...
+  type t =
+    | Insert
+    | Normal
+    | Other;
+
+  let show = v =>
+    switch (v) {
+    | Insert => "insert"
+    | Normal => "normal"
+    | Other => "unknwon"
```
romgrk: typo

PR #32, opened 2019-02-07
Feature/commandline

Akin909:
Add basic implementation of the external commandline

commandline

### Edit
Added [WIP] cursor although currently arrow keys aren't being interpreted so can't actually test if it works

### DIFF
--- a/src/editor/Neovim/Notification.re
+++ b/src/editor/Neovim/Notification.re
@@ -48,7 +48,7 @@ module M = Msgpck;
 let showCommandline = args => {
   /*
      Structure of a cmdline_show response
-     [cmdline_show, [[[attr, inputStr]], position, firstCharacter, indentAmount, index, level]]
+     [cmdline_show, [[[attr, inputStr]], position, firstCharacter, prompt, indentAmount, index, level]]
 
      TODO: handle the attributes being sent
    */
--- a/src/editor/Neovim/Notification.re
+++ b/src/editor/Neovim/Notification.re
@@ -38,13 +38,14 @@ type t =
   | BufferLines(BufferLinesNotification.t)
   | CursorMoved(AutoCommandContext.t)
   | CommandlineShow(Commandline.t)
+  | CommandlineHide(Commandline.t)
   | Ignored;
 
 type commandlineInput = {input: string};
 
 module M = Msgpck;
 
-let parseCommandline = args => {
+let showCommandline = args => {
   /*
      Structure of a cmdline_show response
      [cmdline_show, [[[attr, inputStr]], position, firstCharacter, indentAmount, index, level]]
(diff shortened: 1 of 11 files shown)


src/editor/UI/Commandline.re
```diff
@@ -0,0 +1,51 @@
...
+                marginTop(10),
+                marginLeft(10),
+                fontSize(20),
+                color(Colors.white),
+              ]
+              text={command.firstC ++ command.content}
+            />
+          </View>
+        </View>
+      : <View />
```
Akin909: @bryphe suprised I haven't encountered this yet but how do we return an empty value from a component i.e. render nothing conditionally?

src/editor/Neovim/Notification.re
```diff
@@ -37,13 +37,72 @@ type t =
...
+type commandlineInput = {input: string};
+
 module M = Msgpck;
 
+let showCommandline = args => {
+  /*
+     Structure of a cmdline_show response
+     [cmdline_show, [[[attr, inputStr]], position, firstCharacter, prompt, indentAmount, index, level]]
+
+     TODO: handle the attributes being sent
```
bryphe: Looks like a great start! Thanks for documenting the structure above ☝️

src/editor/UI/Commandline.re
bryphe: Oh, good question! @wokalski might know a better way - but I think one possibility is to use `React.listToElement([])`. We could potentially alias `let empty = React.listToElement([])`, I suppose.

bryphe:
Thanks for the gif - the command line looks great! 👍

src/editor/UI/Commandline.re
wokalski: Yup!

Akin909: The alias sounds like a good idea, I think `ReasonReact` has a `ReasonReact.empty` value that can be used should it be something we add to the project or to the `React` module itself? (seems best in the React module?)

bryphe: Cool! @Akin909 - would you be up for opening a PR on `brisk-reconciler` to add this?

I think the work would be adding a signature here: 
https://github.com/briskml/brisk-reconciler/blob/master/lib/Brisk_reconciler.rei

Like `let empty: syntheticElement;`

And then adding the implementation here: 
https://github.com/briskml/brisk-reconciler

Like `let empty = React.listToElement([]);`

That way we could get @wokalski 's feedback on it! We could also implement it in Revery's react namespace, but I think it seems like it would be useful across the board for brisk-reconciler based apps.

bryphe:
Sorry @Akin909 , I realized you didn't have permissions to merge yourself - fixed now 👍

PR #33, opened 2019-02-08
Remove builds in source from package.json

Akin909:
Remove builds in source (this can cause lsp clients like `reason-language-server` to fail


bryphe:
Thanks @Akin909 ! 🎉

PR #35, opened 2019-02-09
Update README.md

tcoopman:


Akin909:
@tcoopman thanks for adding that :+1:, was thinking we could also point people towards the location of the vendored binaries e.g.
`if you use a Windows, path="absolute/path/to/oni2/vendor/win/...etc"`

bryphe:
Thanks @tcoopman ! I'll bring this in for now, and we can continue to revise. Tracking #40 to remove this requirement completely: https://github.com/onivim/oni2/issues/40

bryphe:
Sorry for the delay on this, @tcoopman ! I guess we won't need this now though with @Akin909 's fix for #40 (no need to set the environment variable now, with the `esy bootstrap` command). 

Thank you for the contribution, though!

Issue #40, opened 2019-02-13
Remove need for ONI2_NEOVIM_PATH environment variable


bryphe:
We require setting an `ONI2_NEOVIM_PATH` variable as described in #35  , just so the binary can figure out where the vendored binaries are. Not ideal and always a pain to set up.

I'd like to add an `esy bootstrap` command that is run after `esy build` - this would generate a JSON or TXT file with the paths to the vendored binaries for development. Then, the executable would pick these up on initial run (our installers would also generate this, or a release build could simply figure out the relative path to the vendored binaries based on our install layout).

Note that we'll need to vendor a couple more binaries:
- `node`
- `ripgrep`?

So whatever solution we come up with will need to be extensible to those binaries too.

Akin909:
@bryphe I started looking at putting together a small shell script to do this since I figured that would be a universal way to do this, without requiring a language dependency like a node script, although since if this is maybe happening when we create builds prior to release would it be safe to assume that node or ocaml are installed (could use ocaml, but then you'd need to build that program into an executable :laughing:).

Would a shell script work on windows? (I really haven't used a windows machine in a *very* long time hence the questions)

bryphe:
Sweet, thanks for looking at this, @Akin909 !

My hope is that - for release - we'll have a determined directory layout (ie, a `vendor` folder at the root), so we can fallback to that layout. In other words, we wouldn't need this for release builds.

For development, though, we'll need some help figuring out where that `vendor` folder is. A shell script should be fine - `esy` comes with a `bash` shell for Windows.

bryphe:
Fixed by #43 ! Thanks @Akin909

PR #43, opened 2019-02-13
Feature/bootstrap-script

Akin909:
Relates to #40, this adds a `bootstrap` command which when called in turn calls the `bootstrap` shell script which creates a platform specific `setup.txt` file that can then be used to derive env vars

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,4 +1,4 @@
-# Set the path for file we are going to generate
+# Set the path for the file we are going to generate
 OUTPUT="$(pwd)/setup.txt"
 
 # Figure out what OS we are on
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,39 @@
+# Set the path for file we are going to generate
+OUTPUT="$(pwd)/setup.txt"
+
+# Figure out what OS we are on
+current_os="$(uname -s)"
+case "${current_os}" in
+  Linux*)     machine=Linux;;
+  Darwin*)    machine=Mac;;
+  CYGWIN*)    machine=Cygwin;;
+  MINGW*)     machine=MinGw;;
+  *)          machine="UNKNOWN:${current_os}"
+esac
+
+# this is a utility to help adding more variables to the generated file
+append_line() {
+  echo $1 >> $OUTPUT
+}
+
+# Based on the operating system get the correct vendor path
+# for the Oni2 binary
+case "${machine}" in
+  Linux)
+    ONI_PATH="/vendor/neovim-0.3.3/nvim-linux64/bin/nvim";;
+  Darwin)
+    ONI_PATH="/vendor/neovim-0.3.3/nvim-osx64/bin/nvim";;
+  *)
+    ONI_PATH="/vendor/neovim-0.3.3/nvim-win64/nvim.exe"
+esac
+
+oni_bin_path="ONI2_PATH=$(pwd)$ONI_PATH"
+
+# create the output file, if it exists remove it first so it is recreated
+if [[ -e $OUTPUT ]]; then
+  rm -f $OUTPUT
+fi
+
+touch $OUTPUT
+
+append_line $oni_bin_path
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -20,14 +20,15 @@ append_line() {
 # for the Oni2 binary
 case "${machine}" in
   Linux)
-    ONI_PATH="/vendor/neovim-0.3.3/nvim-linux64/bin/nvim";;
+      ONI_PATH="$(pwd)/vendor/neovim-0.3.3/nvim-linux64/bin/nvim";;
   Darwin)
-    ONI_PATH="/vendor/neovim-0.3.3/nvim-osx64/bin/nvim";;
+      ONI_PATH="$(pwd)/vendor/neovim-0.3.3/nvim-osx64/bin/nvim";;
   *)
-    ONI_PATH="/vendor/neovim-0.3.3/nvim-win64/nvim.exe"
+      ONI_PATH="$(pwd)/vendor/neovim-0.3.3/nvim-win64/bin/nvim.exe"
+      ONI_PATH="$(cygpath -m $ONI_PATH)";;
 esac
 
-oni_bin_path="ONI2_PATH=$(pwd)$ONI_PATH"
+oni_bin_path="ONI2_PATH=$ONI_PATH"
 
 # create the output file, if it exists remove it first so it is recreated
 if [[ -e $OUTPUT ]]; then
--- a/src/editor/Core/Utility.re
+++ b/src/editor/Core/Utility.re
@@ -12,3 +12,24 @@ let waitForCondition = (~timeout=1.0, f) => {
 
   Thread.join(thread);
 };
+
+let getFileContents = (path, ~handler) => {
+  let contents = ref([]);
+
+  let fileInChannel = Pervasives.open_in(path);
+
+  let fileStream =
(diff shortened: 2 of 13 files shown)


bootstrap.sh
```diff
@@ -0,0 +1,39 @@
+# Set the path for the file we are going to generate
+OUTPUT="$(pwd)/setup.txt"
+
+# Figure out what OS we are on
+current_os="$(uname -s)"
+case "${current_os}" in
```
Akin909: This and the case statement below are pretty identical but I wanted a separate straight forward variable that represented the `os` rather than putting all the os specific logic there as well

Akin909:
Could also look at actually hooking this up so we use it to derive the path in Oni2?

Akin909:
Tried on Linux and this seems to work, should theoretically work on other OSes

bootstrap.sh
bryphe: Makes sense!

bootstrap.sh
```diff
@@ -0,0 +1,39 @@
+# Set the path for the file we are going to generate
```
bryphe: Could we move this file to a folder? Perhaps `build` or `scripts`?

bootstrap.sh
```diff
@@ -0,0 +1,39 @@
...
+OUTPUT="$(pwd)/setup.txt"
```
bryphe: Instead of the `pwd` - could we use the bin folder (where the Oni2 executable ends up)? We can get that with `$cur__bin`.

src/editor/Neovim/NeovimProcess.re
```diff
@@ -13,10 +13,52 @@ let version = (~neovimPath: string) => {
...
+      paths,
+    );
+  path;
+};
+
+let getBuildVariablesFromSetup = (~path="/setup.txt", ()) => {
+  let variableList = ref([]);
+
+  let fileInChannel =
+    Pervasives.open_in(Environment.getWorkingDirectory() ++ path);
```
bryphe: Following from the comment earlier about replacing `pwd` with `cur__bin` - can we use `getExecutingDirectory` here and pick up the setup file next to the binary?

bootstrap.sh
```diff
@@ -0,0 +1,40 @@
...
+# Based on the operating system get the correct vendor path
+# for the Oni2 binary
+case "${machine}" in
+  Linux)
+      ONI_PATH="$(pwd)/vendor/neovim-0.3.3/nvim-linux64/bin/nvim";;
+  Darwin)
+      ONI_PATH="$(pwd)/vendor/neovim-0.3.3/nvim-osx64/bin/nvim";;
+  *)
+      ONI_PATH="$(pwd)/vendor/neovim-0.3.3/nvim-win64/bin/nvim.exe"
+      ONI_PATH="$(cygpath -m $ONI_PATH)";;
```
bryphe: There were a few bugs w/ this on Windows - I tweaked it a bit in https://github.com/onivim/oni2/pull/43/commits/0a3f6ba1dc4a641b1fdf58bdae1e51533eb625b6 👍

bootstrap.sh
```diff
@@ -0,0 +1,40 @@
...
+append_line() {
+  echo $1 >> $OUTPUT
+}
+
+# Based on the operating system get the correct vendor path
+# for the Oni2 binary
+case "${machine}" in
+  Linux)
+      ONI_PATH="$(pwd)/vendor/neovim-0.3.3/nvim-linux64/bin/nvim";;
+  Darwin)
```
bryphe: Based on the case above, seems like this should be `Mac`

src/editor/Neovim/NeovimProcess.re
```diff
@@ -13,10 +13,52 @@ let version = (~neovimPath: string) => {
...
+  fileStream
+  |> Stream.iter(line => {
+       let parts = extractParts(line);
+       variableList := [parts, ...variableList^];
+     });
+
+  variableList^;
+};
+
+let start = (~neovimPath as _: string, ~args: array(string)) => {
```
bryphe: Looks like we could remove ~neovimPath as an argument entirely here!

bootstrap.sh
Akin909: Thanks for trying it out and applying a fix :100:

bootstrap.sh
Akin909: @bryphe I'm not sure where this variable exists? it errors when I add it to the script is it coming from esy? do I need to run a different esy command?

Akin909: Woops it does exist was accessing it incorrectly

Akin909:
@bryphe the `esy x Oni2` command rebuilds/recreates the `cur__build` dir every time its run, not sure how to hook into the command just after its done building, is there an `esy` command for that?

bryphe:
> @bryphe the `esy x Oni2` command rebuilds/recreates the `cur__build` dir every time its run, not sure how to hook into the command just after its done building, is there an `esy` command for that?

Ah, sorry, I forgot about this. Hmm, we could add the `setup.txt` as an 'asset' that always gets copied over, with the following:
- We add a folder `assets\configuration` (there is already a fonts and viml folder)
- We add a `dune` file in that folder that looks like this:
```
(install
    (section bin)
    (package Oni2)
    (files setup.txt))
```
- In `bootstrap.sh`,  we switch `OUTPUT` back to point to `$pwd/assets/configuration/setup.txt`

The main consideration is I want to make sure, at runtime, it's pulling from a deterministic spot (not from current working directory) - I think this would accomplish it 👍 It's similiar to what `esy` does do let its unit tests know where executables are. Hope that helps!

Akin909:
@bryphe thanks that helped a bunch managed to get it working. Also decided to generalize the file reader, I know neovim handles that sort of thing but I'd be surprised if we don't find a use for that functionality elsewhere at some point

bryphe:
Sweet, thanks @Akin909 ! Looks good!

I think we'll need to `esy bootstrap` to our CI config too - looks like the build is complaining it can't find `setup.txt`:
```
info building revery@github:revery-ui/revery#ccac6fc@d41d8cd9: done
File "assets/configuration/dune", line 4, characters 8-17:
4 |  (files setup.txt))
            ^^^^^^^^^
Error: No rule found for assets/configuration/setup.txt
```

Akin909:
@bryphe do you know of a way to make sure the bootstrap command runs before build say a developer runs `esy x Oni2` is there a way to do a `postbuild` or `prebuild` step which is `esy bootstrap`?

bryphe:
> @bryphe do you know of a way to make sure the bootstrap command runs before build say a developer runs `esy x Oni2` is there a way to do a `postbuild` or `prebuild` step which is `esy bootstrap`?

Good question - I'm actually not sure the best way to handle this 🤔 (It's easy for someone testing out the repo to forget!). I'm OK with how it is for now, but we can continue to iterate to see if we can smooth that out (we could potentially add running the bootstrap as part of the build, I suppose - I think that might work).

src/editor/bin/Oni2.re
```diff
@@ -30,12 +28,6 @@ let init = app => {
       "Oni2",
     );
 
-  let neovimPath =
```
bryphe: Thanks for cleaning this up! 👍

src/editor/Core/Utility.re
```diff
@@ -12,3 +12,24 @@ let waitForCondition = (~timeout=1.0, f) => {
 
   Thread.join(thread);
 };
+
+let getFileContents = (path, ~handler) => {
```
bryphe: Thanks for factoring this out 👍

PR #49, opened 2019-02-14
Feature/improve commandline cursor

Akin909:
This PR improves the logic allowing the cursor in the commandline to move using the arrow keys

cmd-cursor

### Note
This has some drawbacks as it also adds line wrapping which the cursor cannot properly follow, because when a text element wraps it does so as a rectangle so if the content of the bottom wrapped section is smaller than the width the cursor pokes out.

Not sure if this should be fixed by revery if not we will have to consider a completely different cursor strategy :confounded:

### DIFF
--- a/src/editor/Core/Actions.re
+++ b/src/editor/Core/Actions.re
@@ -16,6 +16,7 @@ type t =
   | SetEditorSize(EditorSize.t)
   | CommandlineShow(Commandline.t)
   | CommandlineHide(Commandline.t)
+  | CommandlineUpdate((int, int))
   | WildmenuShow(Wildmenu.t)
   | WildmenuHide(Wildmenu.t)
   | WildmenuSelected(int)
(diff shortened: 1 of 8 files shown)


bryphe:
Looks good to me, @Akin909 !

> Not sure if this should be fixed by revery if not we will have to consider a completely different cursor strategy 😖

Some ideas:
- Implement our own wrapping strategy (or reuse the one in Revery - just making it callable externally). If we know the _wrap points_ (the indices where it wraps), we should be able to write a mapping function that goes from the linear cursor position to a (row, column) - this is a lot like what we need to do for our buffer views, too!
- If we have a good API for some of the issues here in revery-ui/revery#203, we might be able to expose a function or parameter that could help (like `cursorPos` that is wrapping-aware, and then is handled by the text input element)

Akin909:
@bryphe I had a similar thought, re adding a prop to the input element that allowed us to control the cursor position and using that component here as that was one of the only reasons we couldn't use an input for Oni1. I'm planning to give the input another looks at some point so will keep that in mind

PR #50, opened 2019-02-15
Feature/buffer enter

### LINKED ISSUE #47: Bug: Edit an existing file doesn't work
__Repro:__ Try and open a file with `:e`, like `:e some-existing-file<CR>`.

Expected: File should be loaded and shown
Actual: No content is rendered

The issue is that we rely on listening to Neovim for buffer updates to get the buffer information via the `nvim_buf_attach` method. When we open a new buffer, we need to call that `nvim_buf_attach` method.

Right now, we only call it once on load for the initial buffer:
https://github.com/onivim/oni2/blob/919cc8f504b8f3482f64944aaa50045b7a799983/src/editor/bin/Oni2.re#L124

We should move this logic so that we don't ever have to manually attach - we could add a listener in the `NeovimProtocol` here:
https://github.com/onivim/oni2/blob/919cc8f504b8f3482f64944aaa50045b7a799983/src/editor/Neovim/NeovimProtocol.re#L59

If we've received a buffer enter event - we should call `nvim_buf_attach` for the new buffer (and detach from existing buffers?)

We'd also need to add that `BufferEnter` notification - which we don't have today:
https://github.com/onivim/oni2/blob/919cc8f504b8f3482f64944aaa50045b7a799983/src/editor/Neovim/Notification.re#L39

and handled parsing it here: 
https://github.com/onivim/oni2/blob/919cc8f504b8f3482f64944aaa50045b7a799983/src/editor/Neovim/Notification.re#L177

We'd need to make sure we get the context we need from it (perhaps just the buffer id)?

Akin909:
This PR will fix #47, by listening to the `BufferEnter` autocommand and attaching to the buffer.

#### Outstanding
- [x] Store and update a list of open buffers using [this](https://neovim.io/doc/user/api.html#nvim_list_bufs()) - how should this be structured cc @bryphe 
  * I'm thinking
```reason
type state = {
  bufferList: list(Buffer.t),
  // activeBuffer: Buffer.t,
  activeBufferId: int, // this already exists, is it redundant
}
```
- [x] Update tabline - this is *sort of* done (it requires a bit more thought i.e. `tabline_update` doesn't give us buffers since there not really a `bufferline` in vim...)

- [x] Detach previous active buffer - this is not necessary as from `nvim` docs

>  When buffer is detached (i.e. updates are disabled). Triggered explicitly by
  |nvim_buf_detach()| or implicitly in these cases:
>  - Buffer was |abandon|ed and 'hidden' is not set.
>  - Buffer was reloaded, e.g. with |:edit| or an external change triggered
    |:checktime| or 'autoread'.
>  - Generally: whenever the buffer contents are unloaded from memory.

tabs


bryphe:
Thanks for working on this, @Akin909 !

So cool to start seeing real buffers around with a minimap and stuff - progress 💯 !
minimap

Regarding the current state - it's great we're keep tracking of open buffers! I see a couple of issues though:
- We're duplicating data between `bufferList` and `activeBuffer`.
- Lookup won't be very fast for the `list(Buffer.t)` for a given buffer - would require iterating across the list. I think we'd want to use a `Map` instead: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Map.Make.html and store the `activeBufferId`. Then activeBuffer is just `Map.find(bufferId, bufferMap)`

src/editor/Core/Types.re
```diff
@@ -58,6 +58,37 @@ module BufferPosition = {
   };
 };
 
+module ViewType = {
```
Akin909: My thinking here is that by adding a Variant to each type of view element then down the line if you were filtering a list that contained multiple types you could pattern match to group them.

I was specifically thinking that if we pursue the route of showing tabs and buffers in a similar way to [vimWinTabs](https://github.com/zefei/vim-wintabs) it might be useful to be able to group these :man_shrugging:

Akin909:
@bryphe I've tweaked the buffers to use a `Map` as you suggested, currently I'm not rendering these yet as I realised that `tabline` events didn't include buffers as I forgot that bufferlines arent natively supported (:laughing: I've had one plugin or another for years I completely forgot) so was going to raise a separate issue about maybe doing something like `VimWinTabs` which shows buffers relevant per tab page.

Also hitting a snag trying to use `nvim_call_atomic` to populate the buffer data anyway so seems like a natural stopping point for this PR (since it was scope creep anyway)

In the meantime the tabline is hooked up, which the screen shot shows. Re. the detachment the docs actually indicate that manually detaching isn't necessary as neovim cleans up for you whenever a buffer is no longer in use.

Akin909:
This PR actually adds a fair bit more than I initially planned

bufferline

It:
* hooks up the buffers to the tabline, and on buf enter this is populated
* hooks up tabs to the tabline, and on `tabline_updated` this is populated
* The tabline reads the `tablineMode` type out of the `configuration` and shows buffers or tabs
* hooks up the `BufEnter` event to show the correct current buffer
* It stores the buffers in a `Map` making retrieval of the `activeBuffer` easier given an `activeBufferId`

src/editor/Neovim/BufferEvents.re
```diff
@@ -0,0 +1,140 @@
+open Oni_Core;
+open Types;
+
```
Akin909: I split out the functionality relating to building up the buffer list as it seemed to me to very quickly fill up `Oni2` and the functionality wasn't really coupled. I wasn't really sure  what to name it, I went with BufferEvents because I imagine there will be similar functionality required for dealing with buffers being closed or renamed etc. so this could basically be a module to deal with data transformation associated with those events.

src/editor/Neovim/Notification.re
```diff
@@ -134,6 +142,59 @@ let updateWildmenu = selected => {
   };
 };
 
+/*
+   Msgpck has a special type for Maps, which returns a tuple
+   consisting of the key and the value, so we pattern
+   match on the name and value and use those to construct
+   the record type we want
+ */
+let parseTabMap = map => {
```
Akin909: Was going to split this out as well into `TabEvents`?

Akin909:
As a followup PR as I got quite carried away here and this exploded in scope.
I've created a mock nvim api so I can test the `NeovimBuffer` functionality.

Since the api is explicitly passed in, I can pass in the mockApi and specify the return values I want for function calls and just test that the processing logic works correctly, I'll do this separately because this is sizeable.

src/editor/Core/Types.re
Akin909: I removed this as I thought it might prove a point of contention 😆 can discuss further when/if its actually necessary

src/editor/Neovim/BufferEvents.re
Akin909: Update renamed this to `NeovimBuffer` as it is essentially a buffer to handle interactions with `neovim` buffers

src/editor/Neovim/NeovimTab.re
```diff
@@ -0,0 +1,50 @@
+open Oni_Core;
+open Types;
+
+module M = Msgpck;
+/*
```
bryphe: Thanks for the detailed comments, @Akin909 !

src/editor/bin/Oni2.re
```diff
@@ -77,7 +76,19 @@ let init = app => {
 
   neovimProtocol.uiAttach();
 
-  let setFont = (fontFamily, fontSize) => {
+  let bufferAttach = bufferId =>
```
bryphe: 🎉 🎉

PR #53, opened 2019-02-16
[WIP] Input: Additional key handling

bryphe:
This adds an `Input` module in `Oni_Neovim` to convert from Revery's `keyEvent` and `keyPressEvent` to neovim input strings.

__TODO:__
- [ ] Hook up `glfwSetCharModsCallback` so we get modifier for keys (and non-printable keys) - we don't need the state machine anymore with this
- [ ] Handle control, alt, shift, super for `keyPressEvent`


src/editor/Neovim/Input.re
```diff
@@ -0,0 +1,50 @@
...
+      | _ => None
+};
+
+let keyPressToString = (~altKey, ~shiftKey, ~ctrlKey, ~superKey, s) => {
+    let s = s == "<" ? "lt" : s;
+    
+    let s = ctrlKey ? "c-" ++ s : s;
+    let s = shiftKey ? "s-" ++ s : s;
+    let s = altKey ? "a-" ++ s : s;
+    let s = superKey ? "m-" ++ s : s;
```
romgrk: Neovim doc says:
```
<S-...>		shift-key			*shift* *<S-*
<C-...>		control-key			*control* *ctrl* *<C-*
<M-...>		alt-key or meta-key		*META* *ALT* *<M-*
<A-...>		same as <M-...>			*<A-*
<D-...>		command-key or "super" key	*<D-*
```
So the last line should be `let s = superKey ? "d-" ++ s : s;`

bryphe: Good catch @romgrk - thank you!

bryphe:
Closing in lieu of #129

PR #72, opened 2019-02-21
Feature/show modified

Akin909:
Show modified Icon when the contents of a buffer has changed

modified

Fixed :+1: ~Was going for [this](https://fontawesome.com/icons/circle?style=solid) icon but for some reason it doesn't render correctly in Oni2~

### DIFF
--- a/src/editor/Core/Buffer.re
+++ b/src/editor/Core/Buffer.re
@@ -64,7 +64,7 @@ let applyUpdate = (lines: array(string), update: BufferUpdate.t) => {
 
 let update = (buf: t, update: BufferUpdate.t) =>
   if (update.version > buf.metadata.version) {
-    let metadata = {...buf.metadata, modified: true, version: update.version};
+    let metadata = {...buf.metadata, version: update.version};
 
     {metadata, lines: applyUpdate(buf.lines, update)};
   } else {
--- a/src/editor/Core/Buffer.re
+++ b/src/editor/Core/Buffer.re
@@ -16,12 +16,11 @@ let ofLines = (lines: array(string)) => {
   lines,
 };
 
-let show = (b: t) => {
+let show = (b: t) =>
   "Buffer ["
   ++ string_of_int(b.metadata.id)
   ++ "]: "
   ++ String.concat("\n", Array.to_list(b.lines));
-};
 
 let ofMetadata = (metadata: BufferMetadata.t) => {metadata, lines: [||]};
 
@@ -63,11 +62,13 @@ let applyUpdate = (lines: array(string), update: BufferUpdate.t) => {
   };
 };
 
-let update = (buf: t, update: BufferUpdate.t) =>
+let update = (buf: t, update: BufferUpdate.t) => {
+  print_endline("Updating!!!");
   if (update.version > buf.metadata.version) {
-    let metadata = {...buf.metadata, version: update.version};
+    let metadata = {...buf.metadata, modified: true, version: update.version};
 
     {metadata, lines: applyUpdate(buf.lines, update)};
   } else {
     buf;
   };
+};
--- a/src/editor/Core/Buffer.re
+++ b/src/editor/Core/Buffer.re
@@ -62,8 +62,7 @@ let applyUpdate = (lines: array(string), update: BufferUpdate.t) => {
   };
 };
 
-let update = (buf: t, update: BufferUpdate.t) => {
-  print_endline("Updating!!!");
+let update = (buf: t, update: BufferUpdate.t) =>
   if (update.version > buf.metadata.version) {
     let metadata = {...buf.metadata, modified: true, version: update.version};
 
@@ -71,4 +70,3 @@ let update = (buf: t, update: BufferUpdate.t) => {
   } else {
     buf;
   };
-};
--- a/src/editor/Core/Reducer.re
+++ b/src/editor/Core/Reducer.re
@@ -40,12 +40,13 @@ let showTablineBuffers = (state: State.t, buffers: list(BufferMetadata.t)) =>
   | _ => state.tabs
   };
 
-let updateTabs = (bufId, tabs: list(State.Tab.t)) =>
-  List.fold_left(
-    (acc, tab: State.Tab.t) =>
-      tab.id === bufId ? [State.Tab.{...tab, modified: true}, ...acc] : acc,
-    [],
-    tabs,
+let updateTabs = (bufId, modified, tabs) =>
+  State.Tab.(
+    List.fold_left(
+      (acc, tab) => tab.id === bufId ? [{...tab, modified}, ...acc] : acc,
+      [],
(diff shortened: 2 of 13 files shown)


src/editor/Core/Buffer.re
```diff
@@ -16,12 +16,11 @@ let ofLines = (lines: array(string)) => {
   lines,
 };
 
-let show = (b: t) => {
+let show = (b: t) =>
```
Akin909: I'm using `refmt` but maybe a new version I just re-installed `reason-cli` recently

bryphe:
Thanks for the work on this, @Akin909 !

I just tried this out and saw a bug (I think it's a regression from `master` behavior):
pull-72-bug

The repro I found is:
- Open Oni2
- Type some text in the empty buffer
- Edit a new file (ie, `src/editor/Core/Types.re`)
- Start typing in the opened file

Expected: Tabs should stay consistent
Actual: Tabs flip order every keystroke

src/editor/Core/Reducer.re
```diff
@@ -67,6 +73,12 @@ let reduce: (State.t, Actions.t) => State.t =
     | ChangeMode(m) =>
       let ret: State.t = {...s, mode: m};
       ret;
+    | BufferWritePost(bs) => {
+        ...s,
+        activeBufferId: bs.bufferId,
```
bryphe: Is it possible to save a buffer if it isn't the active one?

src/editor/UI/Tab.re
```diff
@@ -50,10 +59,25 @@ let createElement =
         color(theme.editorForeground),
       ];
 
+    let modifiedStyles =
+      Style.[
+        color(theme.tabModifiedIconColor),
+        marginHorizontal(5),
+        fontSize(fontPixelSize),
+        fontFamily("FontAwesome5FreeRegular.otf"),
```
bryphe: Found out why the circle wasn't working as expected... Turns out both the solid and hollow circle are the same codepoint (`f111`):
- https://fontawesome.com/icons/circle?style=solid
- https://fontawesome.com/icons/circle?style=regular

We just need to change the font here to `FontAwesome5FreeSolid.otf` instead, to pick up the solid version.

src/editor/Core/Reducer.re
Akin909: you could `writeall`, but you cant target an inactive buffer and save it, or rather I'm pretty sure that's not possible. But this repopulates the bufferlist so we'd find out anyway?

Akin909: actually you can explicitly call `writeall` its only available as `autowriteall` a config option you set in your vimrc

bryphe: Hmm, I was curious why we need to set the `activeBufferId` here - it's possible for the active buffer to change when we're writing it?

src/editor/Core/Theme.re
```diff
@@ -34,9 +35,7 @@ let default: t = {
   editorMenuBackground: Color.hex("#2F3440"),
   editorMenuForeground: Color.hex("#FFFFFF"),
   editorMenuItemSelected: Color.hex("#495162"),
+  tabModifiedIconColor: Color.hex("#737984"),
```
bryphe: I've been trying to stick with the VSCode theme color nomenclature -  documented here: https://code.visualstudio.com/api/references/theme-color 

Just to make it easy to synchronize once we support VSCode themes. Do you know what the equivalent VSCode theme color name is? I looked through the tab ones like `tab.activeForeground` and wasn't sure which one is used for this dirty icon (it seems like the same color is used for the `x` and the dirty icon) 🤔

src/editor/UI/Tab.re
bryphe: Cool, glad it worked! 👍

src/editor/Core/Reducer.re
```diff
@@ -7,48 +7,56 @@
 open Actions;
 open Types;
 
-let truncateFilepath = path => {
+let sortTabsById = tabs =>
```
bryphe: Looks like the bug I saw before is fixed - thanks @Akin909 ! 😍

src/editor/Core/Theme.re
Akin909: @bryphe sorry spent a lot more time in their repository than in the editor I don't know much about that, it looks like its a similar color to the text I can just create a key for the color of the text and use that to style the icon

bryphe:
I saw the first bug I had was fixed (nice!) but there's a new bug I see:

dirty-bug

__Repro:__
- Start Oni2
- Type some characters like `i` + `abc`
- Note the dirty marker
- Open a second file (ie, `:e src/editor/Core/Types.re`)

__Expected:__ The dirty marker for the `[No Name]` buffer should be preserved
__Actual:__ The dirty marker disappears, and then reappears once I start editing the newly opened file

src/editor/Core/Reducer.re
Akin909: I was thinking of cases where if you run an `argadd` command then `bufdo` you can essentially create add a bunch of buffers as arguments then run `bufdo` command and example would be if you run a regex on multiple files at a time, not sure what the autocommand flow would be in those kind of cases. It's possibly/probably unnecessary but we

* have the information as part of the `AutoCommandContext`
* were already setting state
* could catch edge cases

bryphe: Sounds good! Thanks for the info.

src/editor/Core/Theme.re
bryphe: That sounds reasonable. I'm trying to get our themes to match up as much as I can to their theme colors so that its easy to map once we start loading their themes - will save me some time once we hook up VSCode themes!

Akin909:
@bryphe the `no name` doesn't actually exist if you run `:b <Tab>` you'll see it doesn't actually appear in the buffer list its actually a weird then we have thats a mixture I think of artifact on our end since we create it on Oni's side but it doesn't really exist in neovim

If you run the same test with existing actual neovim buffers we don't get that behaviour

Akin909:
Just tested with nvim `u -NONE` in the terminal in that case it opens a similar `[no name]` buffer, but the buffer is unloaded when you navigate away from it, currently in this PR (I've implemented this in a later one) we don't remove unloaded buffers.

Once #77 is in this buffer should be unloaded correctly which is how neovim behaves by default unless a user makes a change to it and decides to save the file

bryphe:
> Just tested with nvim `u -NONE` in the terminal in that case it opens a similar `[no name]` buffer, but the buffer is unloaded when you navigate away from it, currently in this PR (I've implemented this in a later one) we don't remove unloaded buffers.

Ah, cool - thanks for the investigation & explanation!

> Once #77 is in this buffer should be unloaded correctly which is how neovim behaves by default unless a user makes a change to it and decides to save the file

Sounds good! 👌

PR #73, opened 2019-02-21
Run unit tests without an init.vim.

CrossR:
This can cause issues if a developer has a custom config that conflicts.


bryphe:
Thanks @CrossR for fixing this! 💯

PR #74, opened 2019-02-22
open file functionality

Akin909:
Add ability to open buffer by clicking on the tab

### Outstanding
~~- [x] Update buffer correctly on `bufEnter` so content is not duplicated~~
This bug is present on master and relates to the fact that the `changedtick` is not being updated properly not sure exactly why but the check in the `Buffer.update` function always fails. So i'd like to separate a fix for that from this as I'm also looking into `close` functionality for the tabs and I'd like to build off of this PR

### DIFF
--- a/src/editor/Neovim/NeovimProtocol.re
+++ b/src/editor/Neovim/NeovimProtocol.re
@@ -53,12 +53,14 @@ let make = (nvimApi: NeovimApi.t) => {
   let input = (key: string) =>
     nvimApi.requestSync("nvim_input", M.List([M.String(key)])) |> ignore;
 
-  let bufAttach = id =>
-    nvimApi.requestSync(
-      "nvim_buf_attach",
-      M.List([M.Int(id), M.Bool(true), M.Map([])]),
-    )
-    |> ignore;
+  let bufAttach = id => {
+    let _error =
+      nvimApi.requestSync(
+        "nvim_buf_attach",
+        M.List([M.Int(id), M.Bool(true), M.Map([])]),
+      );
+    ();
+  };
 
   Event.subscribe(
     nvimApi.onNotification,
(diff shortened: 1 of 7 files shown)


bryphe:
Thanks for investigating and opening the separate PR to fix 👍

src/editor/Neovim/NeovimProtocol.re
```diff
@@ -27,53 +32,61 @@ let make = (nvimApi: NeovimApi.t) => {
...
+    },
+  )
+  |> ignore;
+
+  let openFile = (~path=?, ~bufferId=?, ~openMethod=Buffer, ()) => {
+    let args =
+      switch (path, bufferId, openMethod) {
+      | (Some(p), None, Buffer) => M.List([M.String("edit" ++ " " ++ p)])
+      | (None, Some(id), Buffer) =>
+        M.List([M.String("edit" ++ " " ++ "#" ++ string_of_int(id))])
```
bryphe: Is there a reason this can't be combined to `"edit #"`?

src/editor/Neovim/NeovimProtocol.re
```diff
@@ -4,17 +4,22 @@
...
 
-/* open Oni_Core; */
+open Oni_Core.Types;
 open Rench;
 
 /*
  * Simple API for interacting with Neovim,
  * abstracted from Msgpck
  */
+module M = Msgpck;
```
bryphe: Thanks, this makes it less verbose! 👍

src/editor/Neovim/NeovimProtocol.re
Akin909: I was trying to be explicit about needing the `#` I didn't realise you needed that to open a buffer by id using `edit` thought some one might overlook it but having it like that makes it clear :thinking: that it was intentional although a comment would do the same and take any guess work out of that

bryphe: A comment sounds good! I think it would serve the same purpose. The concatenation isn't 'free' from a performance perspective, so unless there is a compelling reason to have it, it'd be better just to bake it in.

PR #76, opened 2019-02-22
Bugfix/changed tick

Akin909:
Fix a bug where the `changedtick` of the buffer being edited is not correctly updated.

#### Steps to reproduce
1. Open a buffer with content
1. run `:e<CR>`
1. Content of the buffer is duplicated

On initial investigation it seems the update function is failing to verify that nothing has changed with the buffer, so instead it runs an update of its content

### DIFF
--- a/src/editor/Core/Buffer.re
+++ b/src/editor/Core/Buffer.re
@@ -58,10 +58,22 @@ let applyUpdate = (lines: array(string), update: BufferUpdate.t) => {
 };
 
 let update = (buf: t, update: BufferUpdate.t) =>
-  if (update.version > buf.metadata.version) {
+  switch (update) {
+  /***
+     When a buffer is first attached it emits an update with
+     a startLine of 0 and endLine of -1 in this case we should
+     update the buffer's version but set the content of the buffer
+     rather than update it, which would result in duplication
+   */
+  | {startLine: 0, endLine: (-1), version, _} => {
+      metadata: {
+        ...buf.metadata,
+        version,
+      },
+      lines: Array.of_list(update.lines),
+    }
+  | {version, _} when version > buf.metadata.version =>
     let metadata = {...buf.metadata, version: update.version};
-
     {metadata, lines: applyUpdate(buf.lines, update)};
-  } else {
-    buf;
+  | _ => buf
   };
--- a/src/editor/Core/Buffer.re
+++ b/src/editor/Core/Buffer.re
@@ -6,6 +6,7 @@
 
 open Types;
 
+[@deriving show]
 type t = {
   metadata: BufferMetadata.t,
   lines: array(string),
@@ -15,12 +16,6 @@ let ofLines = (lines: array(string)) => {
   metadata: BufferMetadata.create(),
   lines,
 };
-
-let show = (b: t) =>
-  "Buffer ["
-  ++ string_of_int(b.metadata.id)
-  ++ "]: "
-  ++ String.concat("\n", Array.to_list(b.lines));
 
 let ofMetadata = (metadata: BufferMetadata.t) => {metadata, lines: [||]};
 
--- a/src/editor/Core/BufferMap.re
+++ b/src/editor/Core/BufferMap.re
@@ -62,7 +62,7 @@ let log = (m: t) =>
   Buffers.iter(
     (_, b) =>
       Buffer.show(b)
-      |> (++)("Buffer =======================: \n")
+      |> (++)("Buffer ======================: \n")
       |> print_endline,
     m,
   );
--- a/src/editor/Core/BufferMap.re
+++ b/src/editor/Core/BufferMap.re
@@ -13,6 +13,7 @@ module Buffers =
   });
 
 let empty = Buffers.empty;
+[@deriving show]
 type t = Buffers.t(Buffer.t);
 
 type mapFunction = Buffer.t => Buffer.t;
--- a/src/editor/Core/BufferMap.re
+++ b/src/editor/Core/BufferMap.re
@@ -67,4 +67,7 @@ let log = (m: t) =>
     m,
   );
 
-let getBuffer = (id, map) => Buffers.find(id, map);
+let getBuffer = (id, map) =>
+  try (Buffers.find(id, map)) {
+  | Not_found => Buffer.ofLines([||])
+  };
--- a/src/editor/Core/BufferMap.re
+++ b/src/editor/Core/BufferMap.re
@@ -59,6 +59,12 @@ let updateMetadata = (buffersMap: t, newBuffers: list(BufferMetadata.t)) => {
 };
 
 let log = (m: t) =>
-  Buffers.iter((_, b) => Buffer.show(b) |> print_endline, m);
+  Buffers.iter(
+    (_, b) =>
+      Buffer.show(b)
+      |> (++)("Buffer =======================: \n")
+      |> print_endline,
+    m,
+  );
 
 let getBuffer = (id, map) => Buffers.find(id, map);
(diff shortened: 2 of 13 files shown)


src/editor/Core/dune
```diff
@@ -13,4 +13,4 @@
         yojson
         ppx_deriving_yojson.runtime
     )
-    (preprocess (pps ppx_deriving_yojson)))
+    (preprocess (pps ppx_deriving_yojson ppx_deriving.show)))
```
Akin909: @bryphe I initially had `ppx_deriving.runtime` in the `libraries` section but that made no difference.

The exact error is
```
erMap.cmo -c -impl src/editor/Core/BufferMap.re.pp.ml)
File "_none_", line 1:
Error: Unbound value Buffers.pp
error: command failed: 'refmterr' 'dune' 'build' '--root'
'.' '-j4' (exited with 1)

```

src/editor/Core/BufferMap.re
```diff
@@ -13,13 +13,33 @@ module Buffers =
...
 let map = Buffers.map;
 let update = Buffers.update;
 
+let _unwrap = v =>
+  switch (v) {
+  | Some(v) => v
+  | None => ""
+  };
+
+let log = (m: t) =>
```
Akin909: this is a stop gap whilst I couldn't get `show` to work

src/editor/Core/dune
bryphe: Thanks for pushing up the repro! I tested it out and was able to get it working. You need the following:
- Add `ppx_deriving.runtime` as a lib above
- Add `[@deriving show]` on `Buffer.t`
- Add `[@deriving show]` on `BufferMetadata.t`
- Add `[@deriving show]` on `buftype`

The problem is that, even though `[@deriving show]` is on `BufferMap.t`, when it generates the code, it doesn't know how to `show` the custom types inside - so you have to annotate every type used. Hope that helps!

src/editor/Core/Buffer.re
```diff
@@ -16,12 +16,11 @@ let ofLines = (lines: array(string)) => {
   lines,
 };
 
-let show = (b: t) => {
+let show = (b: t) =>
```
bryphe: I'm glad we're integrating `ppx_deriving.show`, we can get rid of all these random, manual show methods 😍

src/editor/Core/Reducer.re
```diff
@@ -73,10 +73,13 @@ let reduce: (State.t, Actions.t) => State.t =
         buffers: BufferMap.updateMetadata(s.buffers, bs.buffers),
         tabs: showTablineBuffers(s, bs.buffers),
       }
-    | BufferUpdate(bu) => {
+    | BufferUpdate(bu) =>
+      let r = {
         ...s,
         buffers: BufferMap.update(bu.id, applyBufferUpdate(bu), s.buffers),
```
bryphe: Hmm, ya, it looks like when we open a new file it comes in with a buffer update at `start: 0` and `end: -1`. Maybe we aren't handling this case properly in `Buffer.applyUpdate`, if there is existing content in the buffer - might be worth adding a test?

src/editor/Core/BufferMap.re
Akin909: Actually trying to log a full `Map` is a bit involved an from what I can see very messy as it would log the entire thing including its methods :thinking: (not that I actually got it to work) I essentially used show of all the other variables to be able to log the contents here.

Let me know if you happened to get it to work differently, I still encountered the `Buffers.pp` issue having tried what you suggested and it seems it was because the `t` refers to the `Buffers` Module and trying to use `show` on that leads to the problem I described above

src/editor/Core/Reducer.re
```diff
@@ -51,8 +51,7 @@ let updateTabs = (bufId, modified, tabs) =>
     |> sortTabsById
   );
 
-let applyBufferUpdate =
-    (bufferUpdate: BufferUpdate.t, buffer: option(Buffer.t)) =>
+let applyBufferUpdate = (bufferUpdate, buffer) =>
```
Akin909: these are type inferred correctly, I'm never sure whether to annotate for readability? but given how good ocaml is at inference wasn't sure if we should save annotations for when the compiler needs the extra help?

src/editor/Core/Reducer.re
Akin909: Yep seems we were trying to update the buffer in this case, so now we check for that case and if it occurs we set the update lines (aka the buffer content) also added a test case

src/editor/Core/BufferMap.re
```diff
@@ -57,4 +58,16 @@ let updateMetadata = (buffersMap: t, newBuffers: list(BufferMetadata.t)) => {
...
+    (_, b) =>
+      Buffer.show(b)
+      |> (++)("Buffer ======================: \n")
+      |> print_endline,
+    m,
+  );
+
+let getBuffer = (id, map) =>
+  try (Buffers.find(id, map)) {
+  | Not_found => Buffer.ofLines([||])
```
Akin909: This would maybe be a good place in cases like to display a splash/error screen instead of crashing the app

src/editor/Core/Reducer.re
bryphe: I'm ok removing types if they are not needed, inferred! 👍

src/editor/Core/BufferMap.re
bryphe: I think this is a good argument for using methods like `find_opt` instead of `find` - this prompts us that we need to handle _both_ the case where there is no key, and the case where it is.

We should shoot for no crashes, and part of that (and where OCaml / Reason can help a lot) is modelling and reasoning about our state appropriately. If we have code that is throwing / crashing, we should revisit and think about if we could model the state another way to give us better guarantees.

A nice talk that touches on this is the 'Effective ML' video by Yaron Minsky: https://www.youtube.com/watch?time_continue=995&v=-J8YyfrSwTk

test/editor/Core/BufferTests.re
```diff
@@ -22,6 +22,20 @@ describe("Buffer", ({describe, _}) =>
       validateBuffer(expect, updatedBuffer, [|"a"|]);
     });
 
+    test("BufEnter update does not duplicate content", ({expect}) => {
```
bryphe: Nice! Thanks for adding a test here. Perfect time for a test - now, when we update the algorithm / model for the buffer (like in #80 ) - will have this safety net to verify we don't regress this functionality 👍

src/editor/bin/Oni2.re
```diff
@@ -15,6 +15,12 @@ open Oni_Neovim;
 
 module Core = Oni_Core;
 
+/**
+   This allows a stack trace to be printed when exceptions occur
+   TODO: trigger this only if oni is in DEBUG mode
+ */
+Printexc.record_backtrace(true) |> ignore;
```
bryphe: This will hurt performance - can we find a way to conditionally enable this? How about we check an `ONI_DEBUG` environment variable?

Akin909: sounds good didn't realise there'd be a performance hit, was definitely intending that to only run under certain circumstances

src/editor/Core/BufferMap.re
Akin909: @bryphe thanks for the link I'm always interested in hearing about design patterns especially in a new language like this :sparkles:

Akin909: I've updated the affected places to handle the `None` case

bryphe:
Thanks for the update - looks great @Akin909  👍

PR #77, opened 2019-02-23
Feature/add close functionality

Akin909:
Depends on #74, this follows up the open function on tab click with a close functionality on click as well

close

Done ~~I'll rebase this off of `master` once #74 goes in so the diff is clearer~~

### DIFF
--- a/src/editor/Core/Types.re
+++ b/src/editor/Core/Types.re
@@ -123,6 +123,10 @@ module BufferWritePost = {
   };
 };
 
+module BufferDelete = {
+  type t = {buffers: list(BufferMetadata.t)};
+};
+
 module BufferEnter = {
   type t = {
     bufferId: int,
--- a/src/editor/Core/Types.re
+++ b/src/editor/Core/Types.re
@@ -116,18 +116,7 @@ module BufferMetadata = {
   };
 };
 
-module BufferWritePost = {
-  type t = {
-    bufferId: int,
-    buffers: list(BufferMetadata.t),
-  };
-};
-
-module BufferDelete = {
-  type t = {buffers: list(BufferMetadata.t)};
-};
-
-module BufferEnter = {
+module BufferNotification = {
   type t = {
     bufferId: int,
     buffers: list(BufferMetadata.t),
--- a/src/editor/Core/Types.re
+++ b/src/editor/Core/Types.re
@@ -44,6 +44,28 @@ module Mode = {
     };
 };
 
+/**
+   This type module represents the various "views" of vim
+   each of which can be interacted in their own ways
+
+   * A Tab: contains windows
+   * A Window: contains buffers
+   * A buffer: represents a file
+ */
+module Views = {
+  type t =
+    | Window
+    | Tab
+    | Buffer;
+
+  type openMethod =
+    | Buffer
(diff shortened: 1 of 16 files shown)


src/editor/Neovim/NeovimHelpers.re
```diff
@@ -0,0 +1,33 @@
...
+
+module M = Msgpck;
+
+let convertNeovimExtType = (buffer: M.t) =>
+  switch (buffer) {
+  | M.Ext(kind, id) => Some((kind, convertUTF8string(id)))
+  | _ => None
+  };
+
+let getAtomicCallsResponse = response =>
```
Akin909: split out a helper function for dealing with the result of an `atomic_call` this returns `(option(errorList), option(resultList)` so we dont have to figure out the nesting repeatedly which I've found myself doing *a lot*

bryphe: Cool! Saw this used in a few places. Good call on refactoring this 👍

bryphe:
Just tried this out - very nice to have the 'close' functionality now! 👍

src/editor/Core/Utility.re
```diff
@@ -37,8 +37,9 @@ let getFileContents = (path, ~handler) => {
 let convertUTF8string = str =>
   CamomileLibraryDefault.Camomile.(UChar.code(UTF8.get(str, 0)));
 
-let convertNeovimExtType = (buffer: Msgpck.t) =>
-  switch (buffer) {
-  | Msgpck.Ext(kind, id) => Some((kind, convertUTF8string(id)))
-  | _ => None
+let safe_fold_left2 = (fn, accum, list1, list2, ~default) =>
+  try (List.fold_left2(fn, accum, list1, list2)) {
+  | Invalid_argument(reason) =>
```
bryphe: Interesting - is this an expected scenario? When does this fail - different size lists?

src/editor/Core/Types.re
```diff
@@ -44,6 +44,28 @@ module Mode = {
     };
 };
 
+/**
```
bryphe: Following from our convo on Discord - we might have to think about this (as Oni2 will be managing the windows / views) - but not a blocker for now.

src/editor/Core/Utility.re
Akin909: @bryphe it fails in development when I'm messing around its not an expected failure/ doesn't occur when I'm not changing the inputs but it takes down the app which is annoying for development :laughing:

src/editor/Core/Types.re
Akin909: Fair enough :+1:

src/editor/Core/Utility.re
Akin909: Sorry realise that wasn't the specific answer yes different size lists will cause it to fail but technically that shouldn't be possible without messing with the inputs since one request is based on the size of another e.g. I ask for all the buffers get a list of length 3 then I ask for metadata on the buffers -> get a result of list 3 (metadata items)

The way it breaks in development is when I've tweaked this sometimes I get the calls wrong and the app crashes till I fix the broken call but it'd be easier to debug if it returned a default and logged the error

PR #83, opened 2019-02-24
Cursor Element

CrossR:
Add (incredibly) basic cursor location element, and added some missing modes.
For the mode strings, I've made them capitalised for now, but would all caps be best?


src/editor/Core/Types.re
```diff
@@ -32,15 +32,21 @@ module Mode = {
   type t =
     | Insert
     | Normal
+    | Replace
+    | Visual
+    | Operator
     | Commandline
     | Other;
 
   let show = v =>
```
Akin909: @CrossR just a heads up #83 my change in #76 will actually remove this function as we're going to use a package which adds a show function for types, not much to worry about here but if my change is merged first it will conflict with this

Akin909: to clarify only the show function will be removed but the package will still return a string for each type based on what they are called in the type definition so will appear the same

bryphe:
Looks great @CrossR - thank you!

CrossR:
I'd left merging this until after #76 had been merged, so I could update mine, though @bryphe has done that for me! That said it looks like there is an unrelated (?) CI failure, so I'd best wait before I merge this.

Could I get it restarted?

PR #84, opened 2019-02-24
Bugfix/H,M,L Commands

### LINKED ISSUE #79: H, M, L Commands - strange behaviour
Given the new strategy for v2 its a given(:laughing:) that some things don't work and some *won't*, some though similar to `zb`, `zt`, `c-d` etc. are important ways to navigate a buffer without your keyboard.

I think when we encounter cases which mean that a user will have to resort to the mouse, which I think breaks the spirit of `*real* vim bindings` or the goal of avoiding `subpar vim bindings` then we should look into fixes.

Now that the super long intro is over, the issue I believe is that the __High__, __Middle__ and __Low__ commands as I call them don't move a user to the right place in a buffer, it could be that theres some `init.vim` settings missing but I think in this case its more likely that somehow our control of the view layer means that somehow vim isn't aware of what is currently in view so the command doesn't know where to correctly place the cursor.

I think these are good candidates to replace with `Oni` commands, which equate to `H` - position at the top of the view port, `M` - position in the Middle of viewport and `L` position at the bottom

Akin909:
Fix #79, also adds `deriving show` for `Notifications`, this is based off of #76 (since it had all the original `deriving` setup)

This is `WIP` simply because it will need rebasing once #79 goes in.

### DIFF
--- a/src/editor/Core/EditorView.re
+++ b/src/editor/Core/EditorView.re
@@ -138,16 +138,16 @@ let moveCursorToPosition = (~moveCursor, view, lineHeight, position) =>
   switch (position) {
   | Top =>
     let line = getTopVisibleLine(view, lineHeight);
-    moveCursor(~col=0, ~line);
+    moveCursor(~column=0, ~line);
     view;
   | Middle =>
     let topLine = getTopVisibleLine(view, lineHeight);
     let bottomLine = getBottomVisibleLine(view, lineHeight);
-    moveCursor(~col=0, ~line=(bottomLine + topLine) / 2);
+    moveCursor(~column=0, ~line=(bottomLine + topLine) / 2);
     view;
   | Bottom =>
     let line = getBottomVisibleLine(view, lineHeight);
-    moveCursor(~col=0, ~line);
+    moveCursor(~column=0, ~line);
     view;
   };
 
--- a/src/editor/Core/EditorView.re
+++ b/src/editor/Core/EditorView.re
@@ -128,6 +128,29 @@ let recalculate = (view: t, buffer: option(Buffer.t)) =>
   | None => view
   };
 
+type cursorLocation =
+  | Top
+  | Middle
+  | Bottom;
+
+let moveCursorToPosition = (~moveCursor, view, position) =>
+  switch (position) {
+  | Top =>
+    moveCursor(~col=0, ~line=1);
+    view;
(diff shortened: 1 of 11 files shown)


bryphe:
Just tried this it - it doesn't seem to work beyond the initial viewport. For example:

- Open Types.re (`:e src/editor/Core/Types.re`)
- Go to line 100 (`:100`)
- Press `zz` to center
- Press `H`

__Expected:__ In my view, the 84th line is the top, so it should go to that line
__Actual:__ Cursor is moved to line 1

src/editor/Core/EditorView.re
```diff
@@ -125,11 +122,40 @@ let snapToCursorPosition = (view: t, lineHeight: int) => {
...
+  /**
+     Check if there are fewer buffer lines than visibleLines, if so
+     H,M,L should only relate to the bufferlines. If the opposite is
+     true then the commands should only relate to the visible lines
+     of the buffer.
+   */
+  let lines = visibleLines > view.viewLines ? view.viewLines : visibleLines;
+
+  switch (position) {
+  | Top =>
```
bryphe: We should add helper methods to get the 'topVisibleLine' and 'bottomVisibleLine' for the current view. It looks like the current logic doesn't take that into account (ie, if you've scrolled down). Having `topVisibleLine` and `bottomVisibleLine` will make it easy to express the correct behavior for these actions when scrolled.

Akin909: I don't think I can see how based on what we have :thinking:

bryphe: I'll help out with these functions tomorrow if you're still stuck 👍 The main idea is we have the `scrollY` and the `lineHeight` available - the top line at the moment is `scrollY / lineHeight` and the bottom line is (`(scrollY + viewHeight) / lineHeight)`).

Akin909: Thanks @bryphe just needed a hint since I'm wasn't what the variables might map to, that helps a bunch. So scrolling is per line since `scrollY` is a multiplier of lineheight, which I take it means fractional amounts isn't possible not that it'd be desirable or not just checking my understanding

Akin909:
@bryphe thanks for the help, this now works correctly I believe

bryphe:
Works great @Akin909 , thanks for the updates! 👍

PR #88, opened 2019-02-25
Feature/configuration from json

Akin909:
Read the configuration out of a file using `yojson` initial work for having a user editable configuration file

### DIFF
--- a/assets/configuration/configuration.json
+++ b/assets/configuration/configuration.json
@@ -1,0 +1,5 @@
+{
+  "editorLineNumbers": ["relative"],
+  "editorMinimapEnabled": true,
+  "editorTablineMode": ["tabs"]
+}
--- a/assets/configuration/configuration.json
+++ b/assets/configuration/configuration.json
@@ -1,5 +1,5 @@
 {
   "editor.lineNumbers": ["relative"],
-  "editor.minimapEnabled": true,
+  "editor.minimap.enabled": true,
   "editor.tablineMode": ["tabs"]
 }
--- a/assets/configuration/configuration.json
+++ b/assets/configuration/configuration.json
@@ -1,5 +1,5 @@
 {
-  "editorLineNumbers": ["relative"],
-  "editorMinimapEnabled": true,
-  "editorTablineMode": ["tabs"]
+  "editor.lineNumbers": ["relative"],
+  "editor.minimapEnabled": true,
+  "editor.tablineMode": ["tabs"]
 }
--- a/src/editor/Core/Configuration.re
+++ b/src/editor/Core/Configuration.re
@@ -23,14 +23,8 @@ type t = {
 let ofFile = filePath => Yojson.Safe.from_file(filePath) |> of_yojson_exn;
 
 let getConfigPath = () => {
-  let sep = Filename.dir_sep;
-  Revery_Core.Environment.getWorkingDirectory()
-  ++ sep
-  ++ "assets"
-  ++ sep
-  ++ "configuration"
-  ++ sep
-  ++ "configuration.json";
+  let {configPath, _}: Setup.t = Setup.init();
+  configPath;
 };
 
 let create = (~configPath=getConfigPath(), ()) => {
--- a/src/editor/Core/Configuration.re
+++ b/src/editor/Core/Configuration.re
@@ -19,11 +19,14 @@ type t = {
 
 let ofFile = filePath => Yojson.Safe.from_file(filePath) |> of_yojson_exn;
 
-let create = () => {
-  let config =
-    Revery_Core.Environment.getWorkingDirectory()
-    ++ "/assets/configuration/configuration.json"
-    |> ofFile;
+let default = {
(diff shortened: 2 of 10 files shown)


src/editor/Core/Configuration.re
```diff
@@ -4,20 +4,35 @@
...
-  () => {
-    editorLineNumbers: On,
-    editorMinimapEnabled: true,
-    tablineMode: Buffers,
-  };
+let ofFile = filePath => Yojson.Safe.from_file(filePath) |> of_yojson_exn;
+
+let getConfigPath = () => {
+  let sep = Filename.dir_sep;
+  Revery_Core.Environment.getWorkingDirectory()
```
Akin909: figure this will be an issue :laughing: cc @bryphe. I tried using a similar strategy to other paths like `neovimPath` but that's a tough one since those values are pulled in after the app has started whereas currently the config is expected earlier than that, could always refactor things to be an option that is populated after intialization but tbh thought this could be different (I don't recall why we needed the other approach :older_man:)

assets/configuration/configuration.json
```diff
@@ -0,0 +1,5 @@
+{
+  "editorLineNumbers": ["relative"],
```
CrossR: This isn't really related to this PR, but this should probably be `"editor.lineNumbers"` and `"editor.minimap.enabled"`, to mirror the VSCode equivalent. The tabline one doesn't have an equivalent I can think of, so less sure about that one.

src/editor/Core/LineNumber.re
```diff
@@ -4,10 +4,11 @@
...
  */
 
+[@deriving (yojson, show)]
 type setting =
-  | On
-  | Off
-  | Relative;
+  | [@name "on"] On
+  | [@name "off"] Off
+  | [@name "relative"] Relative;
```
CrossR: On a similar vein to my previous comment, what are we going to do about "missing" config options relative to VSCode? Ie, VSCode has an `Interval` option. Are we going to strive to have parity? Should we define all options, even if some are a no-op?

Akin909: I think similarity where it makes sense is one thing but at the end of the day this is a different editor with a smaller team and a similar but different goal for my 2 cents copying them to the letter doesn't make sense to me :man_shrugging:

assets/configuration/configuration.json
Akin909: This is more about reading a `json` file to a record, I think renaming the keys is easy once this bit is done

src/editor/Core/Configuration.re
Akin909: Ignore all of that, tired brain. I now derive the path from the setup but that isn't an issue at all actually :+1:

src/editor/Core/LineNumber.re
bryphe: It would be amazing if we could support the full breadth of VSCode options (eventually!) But ya, it will be an incremental thing for now - I'm okay with supporting a subset and slowly adding them. We should just try and get as much parity as we can (like using the same options / names as VSCode for these initial set of options). We'll also likely have our own vim-specific (like the tabline mode).

@Akin909 - what's the behavior if a user specifies 'Interval' in the config? Does it crash? Or does it use a different default?

Akin909: @bryphe I've tweaked it to print the error but return a default if the user enters the wrong value, but the configuration that is currently being read is intended to be the default so should be correct

I was separately going to look into editing the default and having it save as your own which would mean changing this to be more robust for incorrect user input situations. I would like to do it in a way that uses the tools `yojson` without writing a lot of per case logic.  Such as expecting a string and checking that and generating a Variant which seems like something that wouldn't scale well, might be the only option but will look into it a bit more as part of that.

bryphe: Great! Thanks for investigating this, @Akin909 !

> I would like to do it in a way that uses the tools yojson without writing a lot of per case logic. Such as expecting a string and checking that and generating a Variant which seems like something that wouldn't scale well, might be the only option but will look into it a bit more as part of that.

This sounds like a good next step. I'm wondering if there are any tools in `yojson` or `ppx_deriving_yojson` that can help keep us from writing a lot of boilerplate for every configuration option. 🤔 

I think this also touches on something @CrossR mentioned - in Oni, we started to transition to specifying configuration values separately instead of in one giant store. Perhaps we could investigate a similar approach here? 

We started an API like `Configuration.registerSetting("editor.lineNumbers", configurationOptions)` - it might be worth exploring again? We might be able to bake in some of that parsing in that layer too. I think this is more similar to how VSCode does it (for example - custom parsing or custom UI for modifying it could live in the `configurationOptions`). Just something to think about.

Akin909: @bryphe that's an interesting idea I was actually reading about lenses in ocaml as a way of creating extensible records :thinking: wonder if that pattern could help in constructing it from multiple different pieces

PR #121, opened 2019-03-06
Revert bash change.

CrossR:
I've removed the bash specific syntax, since its actually not needed here, and swapping to bash caused cygwin to not be loaded on systems with the Linux sub-system (I think).

Long term, using nodejs here is probably easier, or writing scripts for each system.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     ]
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


bryphe:
Thanks @CrossR !

PR #126, opened 2019-03-06
Feature/ Add styling per mode to statusbar

Akin909:
statusbar

Add styling based on current vim mode and also separate line/col element in statusbar


bryphe:
Awesome! I've been missing this 😄

src/editor/Core/Theme.re
```diff
@@ -22,6 +22,20 @@ module EditorColors = {
...
+    statusBarItemNormalModeBackground: Color.t,
+    statusBarItemOperatorModeBackground: Color.t,
+    statusBarItemCommandlineModeBackground: Color.t,
+    statusBarItemVisualModeForeground: Color.t,
+    statusBarItemInsertModeForeground: Color.t,
+    statusBarItemReplaceModeForeground: Color.t,
+    statusBarItemNormalModeForeground: Color.t,
+    statusBarItemOperatorModeForeground: Color.t,
+    statusBarItemCommandlineModeForeground: Color.t,
+    statusBarItemForeground: Color.t,
```
bryphe: Should we just use `statusBar.background` and `statusBar.foreground` for these?

Thinking of how they relate here:
https://code.visualstudio.com/api/references/theme-color#status-bar-colors

src/editor/Core/Theme.re
```diff
@@ -22,6 +22,20 @@ module EditorColors = {
     editorMenuForeground: Color.t,
     editorMenuItemSelected: Color.t,
     tabActiveForeground: Color.t,
+    statusBarItemVisualModeBackground: Color.t,
```
bryphe: Thanks for adding the mode colors and defaults! There's no real equivalent in VSCode so I think we will definitely need our own.

I'm wondering if we could generalize this to something like `oniVisualModeBackground`  (using the 'oni' prefix as a convention for theme colors that aren't available in VSCode). 

Seeing this change made me want these colors on top of the active tab (like we did in Oni1), so I was thinking there'd be a case where we might re-use these colors in other parts of the UI, too.

Akin909:
@bryphe addressed your comments btw

PR #127, opened 2019-03-06
Highlight the viewport in the minimap.

CrossR:
Start to highlight the viewport (ie visible lines) in the minimap.

Todo:
 - Mouse over states?
   - VSCode defaults to no highlight, and it enables on mouse over, do we want something similar? Its potentially a lot less useful in an editor where we care mainly about the keyboard, such that always on makes more sense.
 - Config for on/off.
 - Colours/Theme, ie what colour and fixing the now too vibrant cursor line.

image

I assume this is something we'd want and it looked nice and easy to add, so thought I'd do it. If its something we should consider after the performance changes to the minimap, then can leave it for now.


src/editor/Core/Editor.re
```diff
@@ -36,7 +36,7 @@ type viewport = {
   pixelHeight: int,
 };
 
-let getVisibleLines = (view: t) => view.size.pixelWidth / view.lineHeight;
+let getVisibleView = (view: t) => view.size.pixelHeight / view.lineHeight;
```
CrossR: Fixed a bug here, but also thought that 'view' is a more appropriate term here.

Its not really the visible lines, since if I only have 5 lines on screen, this still says 50 since I can see 50 at one time. I should probably add a helper back here to get the number of visible lines as well.

bryphe:
Nice! This is great. It's fun to see the window scroll, and it certainly helps orient the minimap when partially through a doc.
scroll

I did notice with an empty doc it looks a bit strange (perhaps it should be capped to `min(visibleView, bufferLines`):
image

src/editor/Core/Editor.re
bryphe: Makes sense! Thanks for improving this 👍

CrossR:
Think this is okay to be merged for now, the only thing I've not ended up adding is the mouse over handlers.

Its a pretty small change and I think (?) it makes more sense if its added at the same time we make the minimap work as a scroll bar. Otherwise, I can't see why you'd be mousing over the minimap in the first place, at least that is how I use it in VSCode.

bryphe:
Looks great @CrossR - thanks for hooking up the config and everything!

> Its a pretty small change and I think (?) it makes more sense if its added at the same time we make the minimap work as a scroll bar.

Sounds reasonable - tracked as #159 (there's a bunch of places in the UI that we need to add mouse support!)

bryphe:
Looks like there is a formatting change that needs to be fixed (`esy format`) - otherwise looks good to me! 👍

src/editor/UI/Minimap.re
```diff
@@ -62,6 +62,10 @@ let component = React.component("Minimap");
 let absoluteStyle =
   Style.[position(`Absolute), top(0), bottom(0), left(0), right(0)];
 
+let getMinimapSize = (view: Editor.t) =>
+  view.viewLines < Editor.getVisibleView(view)
```
Akin909: is it worth saving the value of `Editor.getVisibleView(view)` in a variable since either side of this ternary means it gets called, but on one side it gets called twice :man_shrugging: 	all in all a very minor issue

Issue #86, opened 2019-02-25
Edit buffer crashes


romgrk:
**Steps**:
 - Run `esy x Oni2`
 - Type `:e Oni.install`

**Result**: Oni2 crashes
**Expected**: buffer is opened

Terminal log:
```
Raw Notification: redraw | [[wildmenu_show,
  [[Oni2.install, Oni2.opam, OniBench.install, OniBench.opam,
    OniUnitTestRunner.install, OniUnitTestRunner.opam, Oni_Core.install,
    Oni_Core.opam, Oni_Neovim.install, Oni_Neovim.opam, Oni_UI.install,
    Oni_UI.opam]]], [wildmenu_select, [0]],
 [cmdline_show, [[[0, e Oni2.install]], 14, :, , 0, 1]], [flush, []]]
[DEBUG - STATE] Mode: commandline editor font measured width: 8 editor font measured height: 16
[DEBUG - STATE] Mode: commandline editor font measured width: 8 editor font measured height: 16
[DEBUG - STATE] Mode: commandline editor font measured width: 8 editor font measured height: 16
[DEBUG - STATE] Mode: commandline editor font measured width: 8 editor font measured height: 16
[DEBUG - STATE] Mode: commandline editor font measured width: 8 editor font measured height: 16
Protocol Notification: unknown
Raw Notification: redraw | [[wildmenu_hide, []], [flush, []]]
Fatal error: exception Failure("hd")
```

Akin909:
@romgrk I can't reproduce this error on master :thinking:, are you seeing or doing anything else also did you run the bootstrap command? What system are you on?

romgrk:
Yes, I've ran `esy install && esy bootstrap && esy build && esy x Oni2` with the master branch just now.

![screenshot from 2019-02-25 17-58-08](https://user-images.githubusercontent.com/1423607/53374660-f399e380-3926-11e9-9e5c-15494f48fe23.png)

Not sure how to debug esy's ocaml, but if you have some tips I'd be happy to.

bryphe:
Thanks for logging the issue, @romgrk ! Usually this exception comes from here:
https://github.com/onivim/oni2/blob/8cf65cdf657c852f7d76f37d985bd7746ebf6c13/src/editor/Neovim/NeovimApi.re#L137

[...]

PR #130, opened 2019-03-07
Bugfix: Add '--noplugin' flag

bryphe:
This adds the `--noplugin` flag for now, so that we get a predictable baseline for nvim integration to start (we'll later explore the way to include VimL plugins). I believe this might be the root cause of #86 crashing.


romgrk:
@bryphe yes it was the root cause, I tried with `-u NORC` and it didn't crash. The delay introduced by plugins went over the 1s timeout.

bryphe:
Thanks for confirming, @romgrk !

PR #131, opened 2019-03-07
Initial release script

bryphe:
This adds an `esy create-release` command that is an initial step towards packaging the application in the `_release/${platform}` folder.


bryphe:
There's still work left to create bundles, etc - but I'll bring those in as separate PRs.

PR #142, opened 2019-03-12
Feature/autogenerate oni2 config dir and files

Akin909:
This PR adds a change that created a user's oni2 directory when they open a config file from the command palette *if* it does not already exist.

It adds a `Filesystem` module (home grown as `fileutils` causes build errors) which uses a monadic? pattern to chain FS operations so that exception and successes can be managed in a concise and predictable fashion

NOTE: I have no idea how this will behave on `windows`


src/editor/Core/Filesystem.re
```diff
@@ -0,0 +1,301 @@
...
+   Infix Operators: These are essentially aliases to the
+   onError and onSuccess functions meaning that both of these can
+   this syntax allows us to use the monadic FS operators like
+   mkdir in a chainable fashion
+
+   mkdir
+    //= error("something went wrong")
+    >>= (_ => return())
+
+   in this example mkdir returns either an Ok(value) or an Error(value)
```
bryphe: Nice work thinking through all this, @Akin909 !

One thing we could consider is bringing in `ppx_let` which adds a helper for this monadic pattern: https://github.com/janestreet/ppx_let

src/editor/Core/Utility.re
```diff
@@ -43,3 +43,13 @@ let safe_fold_left2 = (fn, accum, list1, list2, ~default) =>
     print_endline("fold_left2 failing because: " ++ reason);
     default;
   };
+
+let join = paths => {
```
bryphe: Cool! I like this idea for `Rench`, too (I just made a `Path.join` in that project, but it only took two arguments... I like the list idea).

src/editor/Core/Filesystem.re
```diff
@@ -0,0 +1,301 @@
...
+      )
+    }
+  );
+
+let getOniDirectory = home =>
+  Utility.join([home, ".config", "oni2"]) |> return;
+
+let getHomeDirectory = () =>
+  Unix.(
+    try (getenv("HOME") |> return) {
```
bryphe: Ah, this might not be sufficient on all platforms - here's an example of a check that works for Windows too from `esy`: https://github.com/esy/esy/blob/b0ea114219c77855e0fb36e3cdcfae72adad31b4/esy-lib/Path.re#L36

CrossR: Do we want to store things in `home` for Windows?

The call a few lines down in that file is probably what we want, since `%APPDATA%` is the common store for config and whatnot on Windows. `%APPDATA%/oni` is where Oni1 stored stuff etc.

bryphe: Good point @CrossR . %APPDATA%/oni2 seems like the right spot on Windows for config.

src/editor/Core/Filesystem.re
Akin909: Cool @bryphe, I'm very curious about how these patterns can lead to dependable side-effect code. Although at first glance I have no idea what I'm looking at. I'll read it over and see if I can incorporate it here in a another PR.

src/editor/Core/Filesystem.re
Akin909: :man_facepalming: I just changed it to home, :laughing: I'll have another look at using the data path for windows

Akin909: Fixed :+1:

PR #164, opened 2019-03-17
Reset configuration to buffers mode, add borders to tabs

Akin909:
This change primarily resets the tabline mode to `buffers` as that maps to files and was causing it to seem like the tabs were not rendering properly

It also adds a border between tabs, as an interim measure till we have `inset` shadowing in revery

border-tab

### DIFF
--- a/assets/configuration/configuration.json
+++ b/assets/configuration/configuration.json
@@ -2,5 +2,5 @@
   "editor.lineNumbers": ["relative"],
   "editor.minimap.enabled": true,
   "editor.minimap.showSlider": true,
-  "editor.tablineMode": ["tabs"]
+  "editor.tablineMode": ["buffers"]
 }
--- a/src/editor/UI/Tab.re
+++ b/src/editor/UI/Tab.re
@@ -24,7 +24,7 @@ let horizontalBorderStyles = (tabPosition, numberOfTabs) =>
     | (1, 1) => []
     /* The last tab should also have no borders */
     | (i, l) when i == l => []
-    /* every other tab should have a right border */
+    /* Every other tab should have a right border */
     | (_, _) => [borderRight(~width=1, ~color=Color.rgba(0., 0., 0., 0.1))]
     }
   );
--- a/src/editor/UI/Tab.re
+++ b/src/editor/UI/Tab.re
@@ -17,10 +17,24 @@ let proportion = p => float_of_int(minWidth_) *. p |> int_of_float;
 
 let component = React.component("Tab");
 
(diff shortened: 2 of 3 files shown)


CrossR:
Have we decided if we want to go down the route of all files are tabs? I think the option being there makes sense, but perhaps the default being vim-like makes more sense?

Just thinking of all the issues we hit with people being confused...

The UI looks great tho, I was thinking a few days ago the tabs seemed to melt into one.

bryphe:
> Have we decided if we want to go down the route of all files are tabs? I think the option being there makes sense, but perhaps the default being vim-like makes more sense?

Yes, this was definitely a point of contention with Onivim 1!

I think the default of 'buffers' makes sense, since Onivim2 is controlling window splits. In fact, I'm not sure the 'vim tab' makes sense in the Onivim2 world: I see Onivim 2 as falling into the same model as VSCode/Atom/Sublime, where each split has its own 'buffer line'.

I do like the _idea_ of vim tabs (I use them as 'workspaces' to contain window splits) - I wonder if we could extend that paradigm in Onivim 2? I really like some of the ideas that were discussed in this thread: https://github.com/onivim/oni/issues/1582 , like this plugin: https://atom.io/packages/atom-vim-like-tab . I think something like that could be very useful when jumping between workspaces - not sure if VSCode has an equivalent plugin? If so, it might not be something we have to implement directly in Onivim 2...

bryphe:
The new behavior feels much better to me! I see predictable behavior with the dirty markers now.

tcoopman:
Not sure if this is the best place to discuss, but seeing there is already some discussion, I'm continuing here.

I also use the vim splits/tabs like they are in vim. I love that tabs contain splits. If you call a *workspace* something in a project but with a different scope, for example frontend and backend work, then I do the same thing. I also use tabs for when I want to focus on 1 thing, instead of using `:only`, I do `:tab split` and `:tabclose` when done.

I love this, and every time I use vscode I just don't get how you have to organize my files with tabs in different splits. This is actually one of the main reasons that I prefer to use vim over vscode.

Akin909:
Haha I seems to have unknowingly triggered the old `tabs` and `buffers` debate, tbh this change was more about the fact that the logic is setup to expect a tab is a file so switching to tabs mode but using buffer commands causes weird ui tab behaviour. 

Tbh I've always gotten by with just buffers/single files as its simpler, I do like the idea of `tabs` conceptually as "workspaces", but think we could maybe have an issue to debate it somewhere 😆. I should say though that whatever we would prefer we'd have to take into account that our users might use tab or buffer commands and expect certain things to happen which might not work in Oni2 🤷‍♂️

PR #165, opened 2019-03-17
Swap bootstrap script to JS.

CrossR:
Should fix some of the weirdness, atleast on Windows. Currently it seems like there can be issues with picking up the wrong bash/sh between WSL/System Cygwin/Esy Shell.

I did a quick test on Windows and Linux and the files produced are equal when I diff them, though have not been able to test on Mac.


scripts/bootstrap.js
```diff
@@ -0,0 +1,72 @@
...
+    neovimPath = path.join(vendorPath, "neovim-0.3.3", "nvim-osx64", "bin", "nvim")
+} else if (isLinux) {
+    nodePath = path.join(vendorPath, "node-v10.15.1", "linux-x64", "node")
+    neovimPath = path.join(vendorPath, "neovim-0.3.3", "nvim-linux64", "bin", "nvim")
+} else {
+    console.error("Unknown OS...aborting.");
+    return;
+}
+
+const oniConfig = `{` +
```
Akin909: this could be one `template literal` rather than template literals and concatenation or I think even neater this could be a JS object

```js
const config = {
 neovim: neovimPath
 // ...etc
}

const oniConfig = JSON.stringify(config)
```

bryphe:
Works great for me - thanks for your work on this conversion, @CrossR 👍

PR #166, opened 2019-03-17
Feature/let ppx

Akin909:
Use [`ppx_let`](https://blog.janestreet.com/let-syntax-and-why-you-should-use-it/) to simplify *some* `Filesystem` monadic operations.

Caveats: the `onError` handler and its associated infix operator `//=` are a powerful utility that `let%` currently does not handle (as far as I know) as the ppx only provides `%bind, %map, %switch, and`. The linked article goes into detail re potential use cases for infix vs. for let ppx and suggests using them where there are meaningful variables returned from the operation which is what I've opted for

### DIFF
--- a/src/editor/Core/Filesystem.re
+++ b/src/editor/Core/Filesystem.re
@@ -43,11 +43,15 @@ let _fail = msg => Error(msg);
 
 module Let_syntax = {
   let return = return;
+
   let bind = (t: t('a), ~f: 'a => t('b)) =>
-    fun
+    switch (t) {
     | Ok(x) => f(x)
-    | Error(str) => Error(str);
+    | Error(str) => Error(str)
+    };
+
   let map = (t, ~f) => f(t);
+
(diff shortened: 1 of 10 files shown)


src/editor/Core/Filesystem.re
```diff
@@ -263,55 +299,50 @@ let createOniConfiguration = (~configDir, ~file) => {
...
+let createConfigIfNecessary = (configDir, file) => {
+  open Let_syntax;
+
+  let filepath = Utility.join([configDir, file]);
+
+  let%map dirStats = stat(configDir);
+  switch%bind (dirStats) {
+  | Some(stats) =>
+    isDir(stats)
+    /\/= (
```
Akin909: This operator is similar to bind but rather it specifically responds to `Error(v)` types returned from operators so that the error handling can be sequenced along with the success handling with a lot nested `match` statements

PR #167, opened 2019-03-18
Feature/add quickopen

Akin909:
quickopen

**Update**: with input
quickopen-2

### Outstanding (for a separate PR: see revery-ui/revery#408):
There's still a few kinks to iron out which primarily relate to things in revery
so I'd like to tie those up and then return to this in a separate PR once those changes are in.

- [x] Input should autofocus on open
- [x] Expose the key pressed (in Revery) so that logic based on the specific key can be handled in the onchange - use this to handle pressing escape which should return focus to Oni2
- [x] Fix cursor positioning on start
- [x] Add arrow key behaviour to input
- [ ] Fix backspace not responding

### DIFF
--- a/assets/configuration/keybindings.json
+++ b/assets/configuration/keybindings.json
@@ -1,7 +1,7 @@
 {
     "bindings": [
-        { "key": "<S-C-P>", "command": "quickOpen.open", "when": ["editorTextFocus"] },
-        { "key": "<C-P>", "command": "commandPalette.open", "when": ["editorTextFocus"] },
+        { "key": "<C-P>", "command": "quickOpen.open", "when": ["editorTextFocus"] },
+        { "key": "<S-C-P>", "command": "commandPalette.open", "when": ["editorTextFocus"] },
         { "key": "<ESC>", "command": "menu.close", "when": ["menuFocus"] },
         { "key": "<C-N>", "command": "menu.next", "when": ["menuFocus"] },
         { "key": "<C-P>", "command": "menu.previous", "when": ["menuFocus"] },
--- a/assets/configuration/keybindings.json
+++ b/assets/configuration/keybindings.json
@@ -1,9 +1,10 @@
 {
     "bindings": [
+        { "key": "<S-C-P>", "command": "quickOpen.open", "when": ["editorTextFocus"] },
         { "key": "<C-P>", "command": "commandPalette.open", "when": ["editorTextFocus"] },
-        { "key": "<ESC>", "command": "commandPalette.close", "when": ["commandPaletteFocus"] },
-        { "key": "<C-N>", "command": "commandPalette.next", "when": ["commandPaletteFocus"] },
-        { "key": "<C-P>", "command": "commandPalette.previous", "when": ["commandPaletteFocus"] },
-        { "key": "<CR>", "command": "commandPalette.select", "when": ["commandPaletteFocus"] }
+        { "key": "<ESC>", "command": "menu.close", "when": ["menuFocus"] },
+        { "key": "<C-N>", "command": "menu.next", "when": ["menuFocus"] },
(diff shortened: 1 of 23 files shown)


Akin909:
@bryphe was getting pretty carried away trying to do this all in one PR but there needs to be a fair bit of work on the input component in revery to fix some of the outstanding issues I laid out in the description.

I think what this PR adds is a solid start and the next few things are polishes

assets/configuration/keybindings.json
```diff
@@ -1,9 +1,10 @@
 {
     "bindings": [
+        { "key": "<S-C-P>", "command": "quickOpen.open", "when": ["editorTextFocus"] },
```
bryphe: These bindings - the `<S-C-P>` for quick open, and `<C-P>` for command palette seem reversed to me (compared both to Oni1 or VSCode)

src/editor/Model/Menu.re
```diff
@@ -0,0 +1,62 @@
+open Oni_Core;
+open Types;
+open UiMenu;
+
+let create = (~effects: option(Effects.t)=?, ()) => {
+  menu: Closed,
```
bryphe: It's not clear to me what `menu` is for - where does it get set to something besides `Closed`?

src/editor/Model/QuickOpen.re
```diff
@@ -0,0 +1,24 @@
+open Oni_Core.Types;
+open UiMenu;
+
+let content = (effects: Effects.t) =>
+  effects.getCurrentDir()
```
bryphe: It looks like right now we only pull in the top-level directory - one enhancement we'll need down the road is to also filter through subdirectories. Delegating that to a utility like `ripgrep` might be helpful.

src/editor/bin/Oni2.re
```diff
@@ -172,29 +173,40 @@ let init = app => {
...
-  Reglfw.Glfw.glfwSetCharModsCallback(w.glfwWindow, (_w, codepoint, mods) =>
-    switch (Input.charToCommand(codepoint, mods)) {
-    | None => ()
-    | Some(v) =>
+  let keyEventListener = key =>
+    switch (key, Focus.focused) {
+    | (None, _) => ()
+    | (Some("<ESC>" as v), {contents: Some(_)}) =>
+      /**
+       If a Revery UI element is focused but <ESC> is hit this should
```
bryphe: 👍

src/editor/Core/Filter.re
```diff
@@ -0,0 +1,29 @@
+/**
+   Filter
+
+   Module to handle filtering of items using various strategies
+ */
+
+let menu = (query, items) =>
+  Types.UiMenu.(
```
bryphe: Is this a place where we could use @CrossR 's reasonFuzz library? https://github.com/CrossR/reasonFuzz

Akin909: @bryphe I separated that with exactly that in mind but would rather leave hooking it up to @CrossR this is just place holder rudimentary filtering logic

assets/configuration/keybindings.json
Akin909: yeah they are that's a hang over from me just trying to get it working initially, I'll tweak that

src/editor/Model/Menu.re
Akin909: it was previously called `menuType` it represents the menu's state/type I guess it always has to be something or rather its always theoretically something i.e. `Closed/NoMenu`, `QuickOpen` or `CommandPalette` we change it whenever we open a menu. It tells us its type and we record it in state. The idea being that we can render different UI per menu or use it to specify behaviour per menu type.

It's our only way of knowing what the difference is between the quick open being open or the command palettte

Akin909:
@bryphe switching the event listeners is proving a little trickier than I'd hoped so I'm going to merge this as is and continue working on swapping those out separately

PR #176, opened 2019-03-21
Update Rely.

CrossR:
This bumps the version of Rely we are using, and hooks up the JUnit tests into Azure.

(As an aside, we should perhaps consider adding the lock folders to our `.gitattributes` as generated files. That way, they aren't expanded by default so looking over PR diffs is a bit simpler. The diff is visible if you click through still. If we do that, we could also consider adding the `extensions` folder as vendored code, since it currently comes up as JS code in the repo. Not a bad thing, but also doesn't show of the OCaml/Reason as much!)

### DIFF
--- a/.ci/esy-build-steps.yml
+++ b/.ci/esy-build-steps.yml
@@ -9,13 +9,21 @@ steps:
     displayName: 'Bootstrap Oni2 setup with system specific build variables'
   - script: esy build
     displayName: 'Build: esy build'
-  - script: esy x OniUnitTestRunner
-    displayName: 'Unit Tests: esy x OniUnitTestRunner (round 1)'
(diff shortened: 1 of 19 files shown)


bryphe:
Woah, very cool! 😍 

We get test results in Azure CI now:
image

Thanks for hooking this up, @CrossR !

> As an aside, we should perhaps consider adding the lock folders to our .gitattributes as generated files. 

Agreed 100% - the lockfile changes are noisy now. There's actually an issue logged on `esy` tracking this: https://github.com/esy/esy/issues/825

bryphe:
I'll bring this in so you don't hit a merge conflict with #180 👍 Thanks @CrossR !

PR #177, opened 2019-03-22
Swap home location on Windows.

### LINKED ISSUE #175: Cannot create configuration dir path on Windows - wrong path is used
Oni2 tries to create the directory for storing configuration file, on opening it first-time from command palette. So far, so good.
Unfortunetly, on Windows it fails to do so, with error
`can't create directory 'C:\Users\***\AppData\Local\.config\oni2' because 'No such file or directory`

Causes
1. Creating the oni2 directory doesn't create the parent directories, only the last level. This is reflected in the error msg, so it is the direct cause.
2. The underlying issue is in `src/editor/Core/Filesystem.re :: getOniDirectory` implementation.
It should be also platform specific, as `unsafeFindHome` - using .config is smelly on Windows, Oni used `AppData/Roaming/Oni`, instead of `Appdata/Local/.config/oni2` by the way.

CrossR:
Fix #175.

### DIFF
--- a/src/editor/Core/Filesystem.re
+++ b/src/editor/Core/Filesystem.re
@@ -182,7 +182,7 @@ let unsafeFindHome = () =>
   Revery.(
     switch (Sys.getenv_opt("HOME"), Environment.os) {
     | (Some(dir), _) => dir
-    | (None, Environment.Windows) => Sys.getenv("APPDATA")
+    | (None, Environment.Windows) => Sys.getenv("LOCALAPPDATA")
     | (None, _) => failwith("Could not find HOME dir")
     }
   );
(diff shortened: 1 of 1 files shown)


justinmk:
#175 could be fixed by not adding the .config/ part.

Why change LOCALAPPDATA? Roaming can be problematic, it may change based on Windows network ... stuff.

CrossR:
>Why change LOCALAPPDATA? Roaming can be problematic, it may change based on Windows network ... stuff.

Honestly, I checked the behaviour of Oni1 and some other random programs and they were in Roaming too. If its going to be problematic, I'll swap it back (and nvim is in `%APPDATA%` so helps for consistency.)

bryphe:
Thanks for reviewing this @justinmk and your research / fix @CrossR ! I don't have a strong preference between `%LOCALAPPDATA%` and Roaming - it will be interesting to see if there is any feedback with this (as you mentioned, Oni1 used Roaming - and VSCode does also). 

But I'm OK sticking with `%LOCALAPPDATA%`  since we have the nvim precedent - that's reasonable justification for me.

PR #183, opened 2019-03-24
Bugfix/Menu Input events

Akin909:
Use Window listener rather than Reglfw listener to handle the `keypress` input, this allows the input component to get `backspace`, `arrow` key events etc.

This adds the ability to use `C-N` and `C-P` whilst the input is in focus as well as using `ESC`.

### DIFF
--- a/src/editor/bin/Input.re
+++ b/src/editor/bin/Input.re
@@ -44,15 +44,15 @@ let charToCommand = (codepoint: int, mods: Modifier.t) => {
 let keyPressToCommand =
     ({shiftKey, altKey, ctrlKey, superKey, key, _}: Events.keyEvent) => {
   let keyString =
-    ctrlKey ?
-      /**
+    ctrlKey
+      ? /**
         TODO: currently Revery's toString method returns lower case
         characters which need to be capitalized. Instead we
         should use ?derving show (we will need to format out the KEY_ prefix)
         or convert the return values to uppercase
        */
-      Some(Revery.Key.toString(key) |> String.capitalize_ascii) :
-      (
+        Some(Revery.Key.toString(key) |> String.capitalize_ascii)
+      : (
         switch (key) {
         | KEY_ESCAPE => Some("ESC")
         | KEY_TAB => Some("TAB")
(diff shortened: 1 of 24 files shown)


src/editor/bin/Input.re
```diff
@@ -32,58 +37,68 @@ let charToCommand = (codepoint: int, mods: Modifier.t) => {
...
-  if (buttonState == GLFW_PRESS || buttonState == GLFW_REPEAT) {
-    /* If ctrl is pressed, it's a non printable character, not handled by charMods - so we handle any character */
-    let key =
-      if (Modifier.isControlPressed(mods)) {
+    ({shiftKey, altKey, ctrlKey, superKey, key, _}: Events.keyEvent) => {
+  let keyString =
+    ctrlKey
+      ? /**
+        TODO:
+        Revery's toString method returns lower case
```
bryphe: Is this comment still applicable, now that we are capitalizing the return value?

src/editor/bin/Input.re
```diff
@@ -32,58 +37,68 @@ let charToCommand = (codepoint: int, mods: Modifier.t) => {
...
-      Some(keyToSend);
-    };
-  } else {
-    None;
+    Some((keyPressString, shouldOniListen));
   };
+};
+
+/**
+   Search if any of the matching "when" conditions in the Keybindings.json
```
bryphe: Cool! Thanks for implementing these modes, and matching them up with VSCode's modes 👍

src/editor/bin/Oni2.re
```diff
@@ -192,33 +192,31 @@ let init = app => {
...
-      inputHandler(~state=App.getState(app), v)
+    | (Some((k, true)), {contents: Some(_)})
+    | (Some((k, _)), {contents: None}) =>
+      inputHandler(~state=App.getState(app), k)
       |> List.iter(App.dispatch(app))
+    | (Some((_, false)), {contents: Some(_)}) => ()
     };
 
+  Event.subscribe(w.onKeyDown, keyEvent =>
+    Input.keyPressToCommand(keyEvent) |> keyEventListener
```
bryphe: Thanks for refactoring this 👍

PR #184, opened 2019-03-25
Feature/add ripgrep

Akin909:
use `ripgrep` to search for directories in the `QuickOpen`

### Outstanding
- [x] get binaries and set paths for `mac` and `windows` (windows seems to be ignored by vcs)
- [x] figure out how to handle immense result current `cwd` returns 4900+ results
 which is super laggy ?`Flatlist`

### Future Optimisation
- only download `Rg` based on release version e.g. only mac for mac

### DIFF
--- a/src/editor/Core/Ripgrep.re
+++ b/src/editor/Core/Ripgrep.re
@@ -1,8 +1,6 @@
 let mergeOptions = options =>
   List.fold_left((accum, opt) => accum ++ " " ++ opt, "", options);
 
-let iterateStream = stream => {};
-
 let getPath = () =>
   Revery.Environment.(
     switch (os) {
--- a/src/editor/Core/Ripgrep.re
+++ b/src/editor/Core/Ripgrep.re
@@ -37,7 +37,7 @@ let process = args => {
       }
   );
   close_in(inChannel);
-  Utility.sublist(0, 100, lines^);
+  lines^;
 };
 
 let search = query => process(["--files", "--", query]);
--- a/src/editor/Core/Ripgrep.re
+++ b/src/editor/Core/Ripgrep.re
@@ -1,30 +1,18 @@
+open Rench;
+
 [@deriving show]
-type t = {search: string => list(string)};
+type t = {search: (string, list(string) => unit) => unit};
 
-let mergeOptions = options =>
-  List.fold_left((accum, opt) => accum ++ " " ++ opt, "", options);
+let process = (rgPath, args, callback) => {
+  let cp = ChildProcess.spawn(rgPath, args);
 
-let process = (rgPath, args) => {
-  let lines = ref([]);
-  /**
-     NOTE: this is run in parallel to the main program
-   */
-  let inChannel = mergeOptions([rgPath, ...args]) |> Unix.open_process_in;
-
-  Stream.from(_ =>
-    switch (input_line(inChannel)) {
-    | line => Some(line)
-    | exception End_of_file => None
-    }
+  Event.subscribe(cp.stdout.onData, value =>
+    Bytes.to_string(value)
+    |> String.trim
+    |> String.split_on_char('\n')
+    |> callback
   )
-  |> (
-    stream =>
-      try (Stream.iter(line => lines := [line, ...lines^], stream)) {
-      | _error => close_in(inChannel)
-      }
-  );
-  close_in(inChannel);
-  lines^;
+  |> ignore;
 };
 
 /**
@@ -32,7 +20,7 @@ let process = (rgPath, args) => {
    order of the last time they were accessed, alternative sort order includes
    path, modified, created
  */
-let search = (path, query) =>
-  process(path, ["--files", "--sort", "accessed", "--", query]);
+let search = (path, query, callback) =>
+  process(path, [|"--files", "--sort", "accessed", "--", query|], callback);
 
 let make = path => {search: search(path)};
--- a/src/editor/Core/Ripgrep.re
+++ b/src/editor/Core/Ripgrep.re
@@ -27,9 +27,7 @@ let process = args => {
   Stream.from(_ =>
     switch (input_line(inChannel)) {
     | line => Some(line)
-    | exception End_of_file =>
-      close_in(inChannel);
-      None;
+    | exception End_of_file => None
     }
   )
   |> (
@@ -38,6 +36,7 @@ let process = args => {
       | _error => close_in(inChannel)
       }
   );
+  close_in(inChannel);
   lines^;
 };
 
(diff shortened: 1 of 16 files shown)


CrossR:
>Future Optimisation
>
> only download Rg based on release version e.g. only mac for mac

I think we need that in general for the vendored bits, its just a case of being smart about how we do it. We don't want to end up in the same position we had with Oni1, where we kept consistently hitting GitHub API limits that caused CI to fail. I'm not sure if we ever found a real solution to it though.

It obviously needs to work for both local and CI, so hosting stuff isn't especially an option, unless we have both a normal and a CI setup step. A simple JS download script would work fine locally? Or we can go the route I've seen in other tools and have the binaries in their own repos, since I think cloning is always okay from GitHub.

src/editor/Core/Ripgrep.re
```diff
@@ -0,0 +1,50 @@
...
+          | Browser
+          | Unknown => failwith("not implemented")
+          }
+      )
+    )
+  );
+
+let process = args => {
+  let lines = ref([]);
+  let rgPath = getPath();
```
CrossR: Paths with spaces causing issues for me again...I guess best to find them now rather than later!

I get an error that `F:\User` is not a recoginsed command, since the full path is `F:\User Files\Documents\git\Oni2...`. Think it needs escaping, or quoting.

Akin909: hmm thats odd, the function I'm using comes from revery, I'm otherwise not really doing anything in Oni2 (here) with the actual path. Must mean the function in revery doesn't work on windows. Did the initial error occur in revery or in oni2? either way I'll find a fix but would be good to know the source

CrossR: I'll check when I'm in and see what path it is producing and how it should be formatted. Maybe `revery` should be returning an escaped path?

Out of curiosity, do you know why this way isn't used for the other vendored parts, such as nvim itself, and its instead done with the `bootstrap` script ?

Akin909: @CrossR, @bryphe initially mentioned and discussed this in https://github.com/onivim/oni2/issues/40

I think this pattern I'm using is probably going to need to change to `getExecutingDir` because I *think* if I remember correctly that once we have release build only the build output will be available so things/dependencies should be referenced from that perspective

Akin909: I've added a change to add an escape char to spaces can you let me know if that works for you @CrossR?

src/editor/Core/Ripgrep.re
```diff
@@ -0,0 +1,38 @@
+[@deriving show]
+type t = {search: string => list(string)};
+
+let mergeOptions = options =>
+  List.fold_left((accum, opt) => accum ++ " " ++ opt, "", options);
+
+let process = (rgPath, args) => {
+  let lines = ref([]);
+  /**
+     NOTE: this is run in parallel to the main program
```
bryphe: It looks like, in the current form, ripgrep isn't actually run in parallel - it's blocking until all input lines are picked up. Are there plans to make this run in parallel?

For it to run in parallel - we might need a slightly different API. We might want a callback or a mechanism for it to stream results, and a callback for when it is complete. Then, as we get results, we can plumb them into the menu and apply the fuzzy matching (this is sort of how Oni1 worked)

src/editor/Core/Ripgrep.re
```diff
@@ -0,0 +1,38 @@
...
+   */
+  let inChannel = mergeOptions([rgPath, ...args]) |> Unix.open_process_in;
```
bryphe: We found a couple options to fix the path-in-space issue:
- Use quotes around `rgPath`, ie `"\"" ++ rgPath ++ "\""`
- Use the `Rench` `ChildProcess.spawn` helper (`Unix.create_process` doesn't seem to have this issue)

src/editor/Core/Setup.re
```diff
@@ -20,6 +20,8 @@ type t = {
   configPath: string,
   [@key "keybindings"]
   keybindingsPath: string,
+  [@key "rg"]
+  rgPath: string,
```
bryphe: Thanks for doing all the work to plumb in `ripgrep`! 🎉

src/editor/Core/Ripgrep.re
Akin909: @bryphe the [documentation](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html) for `open_process_in` in the Unix module states that it carries out the request in parallel

> val open_process_in : string -> in_channel

> High-level pipe and process management. This function runs the given command in parallel with the program. The standard output of the command is redirected to a pipe, which can be read via the returned input channel. The command is interpreted by the shell /bin/sh (or cmd.exe on Windows), cf. system.

My understanding was that getting the stream was async, but processing it as I do there is sync, but we do nothing in the processing other than pack it into a list

bryphe: So `open_process_in` does indeed open the process in parallel... but we're _waiting_ on the lines to come in - so the operation as a whole ends up being synchronous. 

This means there is lot of delay when I run `<C-p>` in the Oni2 repo - it's picking up a lot of files and waiting for them all to come in.

Ideally, pressing `<C-p>` should be instant (even if we're still streaming in results in the background), as well as typing to refine the search. 

We can follow up on this in a separate PR, though, if need be.

Akin909: I've switched to use `spawn` and emit an `Update` action with the new content whenever stuff comes through updating the menu content incrementally

Akin909:
@bryphe there is/was actually quite a lot to this and I was initially hoping to work iteratively on this rather than try and solve all the problems in one PR

The escape bug should now be fixed. Re. the filtering and the input, to be clear I saw that as being a separate piece of work, the current filtering is very basic (and is a placeholder). Actually trying to nail that in this PR as well I think is a pretty big scope creep since its a whole other piece of functionality, my aim here was to just get ripgrep working rather than "finish" the quick open experience

bryphe:
> there is/was actually quite a lot to this and I was initially hoping to work iteratively on this rather than try and solve all the problems in one PR

Got it, thanks! I think what would help me is to have the remaining items called out either in a list or as issues (I wasn't sure what was going to be later work, and what was a bug when testing it).

> The escape bug should now be fixed. 

Sweet 👍 

> Re. the filtering and the input, to be clear I saw that as being a separate piece of work, the current filtering is very basic (and is a placeholder). Actually trying to nail that in this PR as well I think is a pretty big scope creep since its a whole other piece of functionality (filtering strategy, handling throttling/or debouncing etc), my aim here was to just get ripgrep working rather than "finish" the quick open experience

That makes sense - the filtering is more broadly applicable to the Menu rather than specific to the QuickOpen. Approved - thanks for your work on this!

PR #192, opened 2019-03-28
Visual mode highlights

bryphe:
__Issue:__ We don't currently show visual mode highlights. This makes visual mode not very useful...

Since we control the rendering of the buffer, we need to also incorporate the visual mode selection.

__Fix:__ We need to send up some context about the current visual range. This is added to our `init.vim` interop story (one thing that could possibly be externalized down the road, too!)

The value we get from that interop is a range of (startLine, startCharacter, endLine, endCharacter). This needs to be rendered differently depending on the mode. This is expressed as a function that takes in a `VisualRange.t` and a `Buffer.t`, and returns a list of `Range.t` that need to be highlighted. The returned `Range.t`s are dependent on the visual mode (the `mode` on `VisualRange.t`, the position of the range, and the length of the buffer lines).

__TODO:__
- [x] Get the latest selection value when mode changes to a visual mode. Right now, we only pick up the selection changes when the cursor _moves_ in visual mode.
- [x] Push `VisualRange.t` updates into the editor view - need an `Action` and to pick this up in the `Editor` reducer.
- [x] Is there a different theme color we should use?
- [x] Use actual visual range values
- [x] Implement strategies in `Selection.getRanges` for each of the visual mode types:
  - [x] `V` Linewise
  - [x] `<C-v>` Blockwise
  - [x] `v` Visual
- [x] Basic test cases
- ~~[ ] Fix background color for text rendered with selection~~
- ~~[ ] Render visual ranges in minimap~~


bryphe:
I'll defer these last two bullet points to separate changes:
- Fix background color for text rendered with selection
- Render visual ranges in minimap

PR #204, opened 2019-03-31
QuickOpen Filter

CrossR:
Started to work on the QuickOpen filter stuff.

This adds a basic filter.
There is a few other things we can do to make it feel better:
 1. Split up the paths into name + description (ie `src/Editor/Core` and `Filter.re`) such that we can match on both and combine the score, and give a higher score for name matches. The logic there might be a bit awkward, but should make it feel nicer.
 2. Hook up index highlights.
 3. Look into "smart" search settings, since there is definitely times I felt the search should have come back with my file, but it was getting a lower than expected score due to the case.


romgrk:
One good matching algorithm to take inspiration from could be [fzy](https://github.com/jhawthorn/fzy) ([explanation here](https://github.com/jhawthorn/fzy/blob/master/ALGORITHM.md)).

bryphe:
It already feels really good @CrossR ! Awesome to see it hooked up with the work @Akin909 did. I believe we'll be able to offer a better experience than Oni1 performance-wise with QuickOpen. 

> Split up the paths into name + description (ie src/Editor/Core and Filter.re) such that we can match on both and combine the score, and give a higher score for name matches. The logic there might be a bit awkward, but should make it feel nicer.

Cool - this sounds like it'll help a lot.

> Hook up index highlights.

Yeah! I miss this too! We don't have good font support for bold/italic at the moment - but we could at least color the matches differently.

> Look into "smart" search settings, since there is definitely times I felt the search should have come back with my file, but it was getting a lower than expected score due to the case.

Makes sense. I really liked the 'refine' functionality you added in QuickOpen for Oni1 too - that's something we could consider adding in the future.

It looks like you'll need to update the `bench` lockfile too (via `esy '@bench' install` - seems like thats why the CI is failing).

CrossR:
With smart search enabled, the search feels a lot better, to me at least.

The two other issues outlined in the first post can be added in a follow up (if that is okay with everyone). I'd like to get his merged so we have some form of QuickOpen since I'm missing having one in general!

`reasonFuzz` does have a fuzzy match alg that will return the indexes for the highlighting, its just waiting for me to write a `compareScores` equivalent for it (`indexCompareScores` etc), which should let that work once the style of the menu is updated accordingly.

The splitting is still potentially useful, but the smart search seems to have fixed the issues I was having in the Oni2 repo at least.

bryphe:
> The two other issues outlined in the first post can be added in a follow up (if that is okay with everyone). I'd like to get his merged so we have some form of QuickOpen since I'm missing having one in general!

Sounds reasonable! I'll merge this in so we can get going with this implementation 👍 Thanks @CrossR !

PR #205, opened 2019-04-01
Feature/add home screen

Akin909:
Updated: 
welcome

### Aim
- [x] Implement minimal (MVP) Home screen primarily to setup the infrastructure that can be built upon

### Out of scope
* proper styling - the styling will be fairly basic to start with till we have a clearer idea what we'd like things to look like
* full set of actions i.e. open folder, previous files etc.

### Considerations

Some editors seem to render the home screen as just another tab, in `atom` it seems to be present *alongside* other editor tabs

In `vscode` it appears as a tab but whenever an editor tab is opened the home screen is *replaced*

We can implement similar-ish behaviour by creating `UiTabs` for the home screen (but potentially in the future for other associated GUI i.e. non editor screens)

NOTE: the simplest and cleanest thing would be to keep these separate (I'm going to try the strategy above but might have to resort to this)


bryphe:
Very cool @Akin909 !

> We can implement similar-ish behaviour by creating UiTabs for the home screen (but potentially in the future for other associated GUI i.e. non editor screens)

> NOTE: the simplest and cleanest thing would be to keep these separate (I'm going to try the strategy above but might have to resort to this)

For now - what if we just show the welcome screen when we have no tabs? I think this would simplify the state management. If we implemented it that way - we wouldn't need any additional state management for the tabs - we could show the welcome screen with the current state we have today.

Akin909:
@bryphe from what I can see you always have tabs, i.e. in a fresh neovim session a `[No Name]` buffer is created and when you close all buffers vim will either quit or open an empty buffer. As far as I can see in neovim there is always a buffer or neovim isn't open.

bryphe:
Ah, got it - I was thinking of #81. Perhaps we could show the Welcome screen when there is _only_ the initial no-name buffer? Then, when there are actual buffers, we'd hide the welcome screen?

Akin909:
:thinking: the no name buffer is essentially a buffer without a "filename" so it gets set to `[No Name]` when the tab is rendered but this could also be the case for a new file but I guess if I closed all the buffers then => welcome is correct.

just wondering if there's a scenario where a user rightly wants a new unnamed buffer they created and is suddenly hit with an unexpected no name buffer

bryphe:
> just wondering if there's a scenario where a user rightly wants a new unnamed buffer they created and is suddenly hit with an unexpected no name buffer

Hmm, good point! I guess we could consider only enabling it on initial load? I'm OK with whatever the simplest option is 👍

Apostolique:
I don't see anything wrong with the welcome screen being a real tab that can be opened alongside other tabs.

vscode works like that too: 
image

The welcome screen can have valuable information anyway.

bryphe:
> I don't see anything wrong with the welcome screen being a real tab that can be opened alongside other tabs.

I'm OK with it being a real tab (just thinking of what the shortest / simplest route to getting this working is). One downside, though, with the current approach is that we could not have the Welcome tab coexist with other tabs 🤔

Akin909:
@bryphe changed tack to allow the welcome to be treated as "just" another tab

* Tabs have a type poorly named `EditorTab` and `MessageTab` (open to suggestions) based on the tab type we render an `EditorView` or the `Home` this can be generalised so we can render Change log screens or some other informational screen like a settings gui screen.

* This works by initially on load we delay attaching to neovim *until* the user has either selected something from the quick open or clicked the open file button. This is done explicitly now because otherwise neovim steals "focus"

* Also tweaked the `activeBufferId` to be an `option` as since now neovim is not always attached it can sometimes be empty. this is needed to correctly give focus to `guiTabs`

* when switching from a neovim buffer to a gui buffer we explicitly detach from the buffer to tell neovim that we are no longer looking at it.

The choices re. attaching and detaching are primarily because if not under the hood neovim is unaware that it isn't "in charge" so it intercepts input responds to commands etc.

### Outstanding
- [ ] handle the case where the home screen is closed. In this case if there are available buffers we need to attach to one of them

src/editor/Neovim/NeovimProtocol.re
```diff
@@ -70,6 +71,17 @@ let make = (nvimApi: NeovimApi.t) => {
     ();
   };
 
+  let bufDetach = id => {
+    let bufId =
+      switch (id) {
+      | Some(id) => id
+      | None => 0
```
Akin909: `0` is equivalent to current

Akin909:
A follow up PR will be needed at some point to prevent the infamous `[no name]` buffer from appearing when the final buffer is closed.

The default behaviour for an attached GUI I think is that when then final buffer is closed the `[no name]` is opened instead.

Rather than this we would like the welcome screen to re-appear.

src/editor/Model/BufferMap.re
```diff
@@ -23,6 +23,14 @@ let map = Buffers.map;
 let update = Buffers.update;
 let remove = Buffers.remove;
 
+let getBuffers = (buffers: t) =>
```
bryphe: This name is a bit confusing - it returns a list of buffer metadatas, right? I would expect that `getBuffer` would give me a single buffer, whereas `getBuffers` would return a list of buffers.

src/editor/Model/State.re
```diff
@@ -19,10 +19,12 @@ module Tab = {
 };
 
 type t = {
+  neovimActive: bool,
```
bryphe: This implementation seems to add a lot of complexity to the state, with these changes (along with the fact we have to maintain an independent set of tab state!). 

Is there a more minimal approach we could take? For example, could we have a special hidden buffer in Neovim represent the 'Home' / Welcome experience?

src/editor/Model/Reducer.re
```diff
@@ -29,20 +29,32 @@ let showTablineTabs = (state: State.t, tabs) =>
...
+          title: filePath |> truncateFilepath,
+          active: false,
           modified,
         },
       buffers,
     )
     |> sortTabsById
-  | _ => state.tabs
+  | (Tabs, _) => state.tabs
+  | (Hybrid, _) => state.tabs
```
bryphe: What does `Hybrid` mean?

src/editor/Store/NeovimStoreConnector.re
```diff
@@ -88,30 +90,82 @@ let start = (executingDirectory, setup: Core.Setup.t, cli: Core.Cli.t) => {
...
+    neovimActive ? Isolinear.Effect.none : attachUIEffect;
+
+  let updater = (state: Model.State.t, action) =>
     switch (action) {
+    | Model.Actions.OpenHome =>
+      let id = state.activeBufferId;
+      (state, detachBufferEffect(id));
+    | Model.Actions.StartEditor =>
+      /**
+      We attach to neovim as a separate action rather than immediately
```
bryphe: This would mean that you can't open the command line in the welcome screen, right? So I couldn't run `:new` from the welcome screen?

src/editor/Model/BufferMap.re
Akin909: this could be `getBufferMetadata` if it seems too confusing?

src/editor/Model/State.re
Akin909: not sure what you mean re independent tab state? do you mean the type? that seemed like something inherent to what we would want i.e. the ability to represent a GUI tab type for certain buffers vs an editor tab type?

src/editor/Store/NeovimStoreConnector.re
Akin909: not necessarily the only thing about this approach is that it requires us to be explicit about attaching to neovim rather than implicitly always being attached. I tried quite a few different approaches and this was the only one that I found that would allow us to have a gui tab independent from neovim without creating a fake neovim tab etc.

src/editor/Model/Reducer.re
Akin909: @bryphe this predates this PR I just took the opportunity to clear with the exhaustiveness check

Akin909:
@bryphe fair enough.

Personally I'd not be very keen to pursue a similar path to the way that the welcome screen worked in Oni1 as I found that very fiddly to implement.

I'm going to leave this problem for now since the solution is unsatisfactory. As I've put in as much work into this as I can.

bryphe:
Sorry, I realized my last message was really poor! Not how I meant to convey at all (In my head I was thinking / planning window splits + integrating the buffer line, but I didn't do a good job framing it at all.)

> Personally I'd not be very keen to pursue a similar path to the way that the welcome screen worked in Oni1 as I found that very fiddly to implement.

True, there were a lot of problems that route too. Esp. the integration with the second neovim instance (would like to avoid that complexity as well!)

> I'm going to leave this problem for now since the solution is unsatisfactory. As I've put in as much work as I'd like to.

Sure thing. We can revisit once we have the tab bars set for the window splits (maybe less unknowns at that point for me). Thank you for your help and time, @Akin909 !

PR #212, opened 2019-04-03
Feature - Extension Host: Document enter event

bryphe:
This adds support for sending the document enter event (`ExtHostDocumentsAndEditors` - `$acceptDocumentAndEditorsDelta`).

### DIFF
--- a/src/editor/Core/Types.re
+++ b/src/editor/Core/Types.re
@@ -157,7 +157,8 @@ module BufferNotification = {
   };
 
   let getBufferMetadata = (id, v: t) => {
-    let result = v.buffers |> List.filter((b: BufferMetadata.t) => b.id == id);
+    let result =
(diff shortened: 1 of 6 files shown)


src/editor/Core/Types.re
```diff
@@ -155,6 +155,10 @@ module BufferNotification = {
     bufferId: int,
     buffers: list(BufferMetadata.t),
   };
+
+  let getBufferMetadata = (id, v: t) => {
+    v.buffers |> List.filter((b: BufferMetadata.t) => b.id == id) |> List.hd;
```
Akin909: @bryphe `hd` can throw a `Failure` which might cause a runtime collapse we could wrap this in a try or not sure if theres an alternative that returns an `option`

bryphe: Good call - I refactored to use `List.nth_opt` instead 👍

PR #227, opened 2019-04-06
Fix #216: add Neovim menu focus to input control mode

### LINKED ISSUE #216: Pressing <C-p> while in command-line opens QuickOpen
Screenshot from 2019-04-03 13-49-52

Akin909:
Add an `inputControlMode` for neovim externalized components so that we can specify their input behaviour. This fixes #216, as previously the input mode was `editorText` so the mapping for the quickopen was still valid whilst the commandline was open, which it shouldn't be.

This also opens the door for allowing oni specific binding for neovim gui externalisations :man_shrugging:

### DIFF
--- a/src/editor/Core/Types.re
+++ b/src/editor/Core/Types.re
@@ -104,13 +104,12 @@ module Range = {
     endPosition,
   };
 
-  let create = (~startLine, ~startCharacter, ~endLine, ~endCharacter, ()) => {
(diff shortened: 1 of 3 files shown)


bryphe:
Thanks for the fix @Akin909 ! Looks good to me 👍 

> This also opens the door for allowing oni specific binding for neovim gui externalisations 🤷‍♂️

I'm game if we can provide a good UX for it 😄 Definitely a lot of possibilities in this thread... https://github.com/neovim/neovim/issues/9421

PR #229, opened 2019-04-07
Rename no name buffer to untitled

Akin909:
Name any buffer's with an empty name `untitled` rather than `[no name]`. This is a bit of a break from what neovim does, I think untitled is more commonly used in other editors and *to my mind* reads better.

Also don't create dummy buffers and dummy state, these aren't necessary as neovim starts with an empty buffer so we don't need to create one beforehand

### DIFF
--- a/src/editor/Model/State.re
+++ b/src/editor/Model/State.re
@@ -46,7 +46,7 @@ let create: unit => t =
     commandline: Commandline.create(),
     wildmenu: Wildmenu.create(),
     activeBufferId: 0,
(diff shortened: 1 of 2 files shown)


src/editor/Model/State.re
```diff
@@ -46,7 +46,7 @@ let create: unit => t =
     commandline: Commandline.create(),
     wildmenu: Wildmenu.create(),
     activeBufferId: 0,
-    buffers: BufferMap.Buffers.add(0, Buffer.ofLines([||]), BufferMap.empty),
+    buffers: BufferMap.empty,
```
bryphe: Thanks for cleaning this up!

PR #230, opened 2019-04-08
fixed lightweight

0xflotus:


Akin909:
@0xflotus thanks for catching this and fixing it :+1:

PR #232, opened 2019-04-08
Add Window manager

Akin909:
:sweat_smile: Initial attempt at a window manager and layout modules, With `Dock` as first split. @bryphe would be good to get your feedback early since I imagine/am guessing you would have had an idea of how you wanted to approach this?

dock

### Outstanding (easier to say what isn't :laughing:)
- [ ] Separator
... tbd

### DIFF
--- a/src/editor/Model/Actions.re
+++ b/src/editor/Model/Actions.re
@@ -50,6 +50,8 @@ type t =
   | CloseFileById(int)
   | OpenFileByPath(string)
   | OpenFileById(int)
+  | AddSplit(EditorSplits.split)
+  | RemoveSplit(int)
   | OpenConfigFile(string)
   | QuickOpen
   | SetLanguageInfo(LanguageInfo.t)
--- a/src/editor/Model/WindowManager.re
+++ b/src/editor/Model/WindowManager.re
@@ -1,0 +1,22 @@
+open Oni_Core;
+open Types.EditorSplits;
+
+type t = {splits};
+
+let empty = IntMap.empty;
+
+let add = IntMap.add;
+
+let remove = IntMap.remove;
+
+let toList = map =>
+  IntMap.fold((_, split, allSplits) => [split, ...allSplits], map, []);
+
+let create = () => {splits: IntMap.empty};
+
+let reduce = (state: t, action: Actions.t) =>
+  switch (action) {
+  | AddSplit(split) => {splits: IntMap.add(split.id, split, state.splits)}
+  | RemoveSplit(id) => {splits: IntMap.remove(id, state.splits)}
+  | _ => state
+  };
--- a/src/editor/UI/Dock.re
+++ b/src/editor/UI/Dock.re
@@ -6,7 +6,7 @@ let component = React.component("Dock");
 
 let dockItemSize = 20;
 
-let iconStyles =
+let icon =
   Style.[fontFamily("FontAwesome5FreeSolid.otf"), fontSize(dockItemSize)];
 
 let button = Style.[marginVertical(8)];
@@ -21,16 +21,11 @@ let createElement = (~children as _, ~state: State.t, ()) =>
           flexGrow(1),
           top(0),
           bottom(0),
-          width(90),
           backgroundColor(bg),
           alignItems(`Center),
         ]>
-        <Clickable style=button>
-          <Text text={||} style=iconStyles />
-        </Clickable>
-        <Clickable style=button>
-          <Text text={||} style=iconStyles />
-        </Clickable>
+        <Clickable style=button> <Text text={||} style=icon /> </Clickable>
+        <Clickable style=button> <Text text={||} style=icon /> </Clickable>
       </View>,
     );
   });
--- a/src/editor/UI/Dock.re
+++ b/src/editor/UI/Dock.re
@@ -1,19 +1,36 @@
-open Revery;
 open Revery_UI;
+open Revery.UI.Components;
+open Oni_Model;
 
 let component = React.component("Dock");
 
-let createElement = (~children as _, ~state as _, ()) =>
-  component(hooks =>
+let dockItemSize = 20;
+
+let iconStyles =
+  Style.[fontFamily("FontAwesome5FreeSolid.otf"), fontSize(dockItemSize)];
+
+let button = Style.[marginVertical(8)];
+
+let createElement = (~children as _, ~state: State.t, ()) =>
+  component(hooks => {
+    let bg = state.theme.colors.editorLineNumberBackground;
     (
       hooks,
       <View
         style=Style.[
+          flexGrow(1),
           top(0),
           bottom(0),
           width(90),
-          backgroundColor(Colors.black),
-        ]
-      />,
-    )
-  );
+          backgroundColor(bg),
+          alignItems(`Center),
+        ]>
+        <Clickable style=button>
+          <Text text={||} style=iconStyles />
+        </Clickable>
+        <Clickable style=button>
+          <Text text={||} style=iconStyles />
+        </Clickable>
+      </View>,
+    );
+  });
--- a/src/editor/UI/Dock.re
+++ b/src/editor/UI/Dock.re
@@ -1,0 +1,19 @@
+open Revery;
+open Revery_UI;
+
+let component = React.component("Dock");
+
+let createElement = (~children as _, ~state as _, ()) =>
+  component(hooks =>
+    (
+      hooks,
+      <View
+        style=Style.[
+          top(0),
+          bottom(0),
+          width(90),
+          backgroundColor(Colors.black),
+        ]
+      />,
+    )
+  );
--- a/src/editor/UI/EditorView.re
+++ b/src/editor/UI/EditorView.re
@@ -45,6 +45,32 @@ let createElement = (~state: State.t, ~children as _, ()) =>
   component(hooks => {
     let theme = state.theme;
     let mode = state.mode;
+    let hooks =
+      React.Hooks.effect(
+        Always,
+        () => {
+          GlobalContext.current().dispatch(
+            AddSplit({
+              id: 0,
+              layout: Full,
+              width: 100,
+              height: 100,
+              component: () => <EditorSurface state />,
+            }),
+          );
+          GlobalContext.current().dispatch(
+            AddSplit({
+              id: 1,
+              layout: VerticalLeft,
+              width: 10,
+              height: 100,
+              component: () => <Dock state />,
+            }),
+          );
+          None;
+        },
+        hooks,
+      );
 
     let tabs = toUiTabs(state.tabs);
     let uiFont = state.uiFont;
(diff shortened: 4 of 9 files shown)


src/editor/Model/Actions.re
```diff
@@ -7,7 +7,7 @@
 open Oni_Core.Types;
 open Oni_Extensions;
 
-type t =
+type t('a) =
```
Akin909: I *really* wish I knew of another way to do this, `State.t` cannot be referenced in `WindowManager.re` because it would cause a circular dependency which ocaml doesn't allow.
so we pass the split type a `'a` and make all the places that use it take a type param of `State.t` :sob:

bryphe:
> @bryphe would be good to get your feedback early since I imagine/am guessing you would have had an idea of how you wanted to approach this?

Sure! Thanks for checking in. I'm glad to see work starting on this, I miss window splits in Onivim 2!

src/editor/Model/Actions.re
bryphe: Hmm, ya, this is tricky! It's rough that the `'a` type sort of propagates through across a bunch of types. But I don't know a great way off the top of my head.

A couple options I was thinking of:

## Option 1: Use `() => React.syntheticElement` instead of `State.t => React.syntheticElement`

This idea essentially uses some _sleight of hand_ to use a `unit => React.syntheticElement` function as our split (while still allowing the `state` to be used!). This is nice because the definition for `WindowManager` is agnostic of the state.

I coded this up here since I thought it might be easier to see via code than to describe:
https://github.com/Akin909/oni2/pull/1/files

I think this approach might be preferable, because we can avoid the avalanche of `'a` types!

## Option 2: Recursive Types

One thing I was thinking about is if we could use recursive types in the `State.re` module, like:
```
type layout = 
| VerticalLeft
| VerticalRight
| HorizontalTop
| HorizontalBottom

type t = {
  mode: Mode.t,
  tabs: list(Tab.t),
  buffers: BufferMap.t,
...
  inputControlMode: Input.controlMode,
  iconTheme: IconTheme.t,
  languageInfo: LanguageInfo.t,
  statusBar: StatusBarModel.t,
  windows: window,
}
and windowSplit = {
    id: int,
    component: t => React.syntheticElement,
    layout,
    width: option(int),
    height: option(int),
}
and window = {
    splits: IntMap.t(windowSplit)
};
```

But the consequence of that is we'd need to move some of the other parts of `WindowManager` there too 🤔 It's kind of a trade-off in either case. But maybe including some of the `WindowManager` methods - like `createSplit`.
Not 100% sure if this works

src/editor/Model/WindowManager.re
```diff
@@ -0,0 +1,48 @@
...
+type layout =
+  | VerticalLeft
+  | VerticalRight
+  | HorizontalTop
+  | HorizontalBottom;
+
+module WindowSplitId =
+  Revery.UniqueId.Make({});
+
+type split('a) = {
```
bryphe: This approach is pretty much on the same page as what I was thinking 👍 Looks good!

src/editor/Model/WindowManager.re
```diff
@@ -0,0 +1,48 @@
+open Revery_UI;
+open Oni_Core;
+
+type layout =
```
bryphe: One concern I have is this layout - it seems like it will be limiting. In particular, it won't allow for nested splits, or multiple split levels.

I think a more natural representation of the splits is a `tree` representation:
```
type splitDirection =
| Horizontal
| Vertical;

type splitTree =
| Parent(direction, list(splitTree))
| Leaf(split)
```

This allows for nesting of splits. You could have a side-by-side layout of two editors, like:
```
Parent(Horizontal, [Leaf(editor1), Leaf(editor2)]
```
And then maybe you decide you want to split editor2 into a vertical split:
```
Parent(Horizontal, [Leaf(editor1), Parent(Vertical, [Leaf(editor2), Leaf(editor3)])
```
etc...

If we use a tree to model the splits - we can open up having nesting.

However, in that case, we might want a separate model for sidebars, like a dock:
```
type dock = list(split)
```

Then our Window Manager might contain:
```
type t = {
   leftDock: dock,
   rightDock: dock,
   windows: splitTree,
};
```

src/editor/UI/Dock.re
```diff
@@ -0,0 +1,31 @@
+open Revery_UI;
```
bryphe: 😍 Thanks for starting this up!

src/editor/UI/Dock.re
```diff
@@ -0,0 +1,31 @@
...
+      hooks,
+      <View
+        style=Style.[
+          flexGrow(1),
+          top(0),
+          bottom(0),
+          backgroundColor(bg),
+          alignItems(`Center),
+        ]>
+        <Clickable style=button> <Text text={||} style=icon /> </Clickable>
```
bryphe: Putting the Unicode character directly code is hard for me to read (it doesn't show up in GH UI, or my default font on Windows).

Would you mind switching to use the `<FontIcon />` component instead? That has an `icon` parameter where you can specify the code directly. An example usage is here: https://github.com/onivim/oni2/blob/c637fc3994c9c4e223f67a51f7d2b9a4df73f469/src/editor/UI/Tab.re#L129

Bonus: You don't need to set a style to pick up the font awesome font, it's already there by default.

Akin909: @bryphe I've converted these, I did see that component but opted against these for a few reasons I think are worth raising.

Currently that component's api requires `icon=int` and the font awesome website gives unicode strings like `f002`, took some digging to find that it needed, `IconTheme.IconDefinition.parseId` to get the `int` value I needed.

You could generalise that but then the api of the component is too permissive either it optionally takes `icon=int` and `iconStr=string` and under the hood does the parsing of the id etc but then it can be passed with a potentially empty value i.e. it no longer demands only one correct thing, since both would have to be options.

Also I guess it does depend on a developer's font but this way means that scanning what icon each of these is rendering is much harder. If we had a group of these and wanted to change one, the only chance you'd have would be to look up the codes? or change stuff till you found the right one :man_shrugging:

bryphe: > Currently that component's api requires icon=int and the font awesome website gives unicode strings like f002, took some digging to find that it needed, IconTheme.IconDefinition.parseId to get the int value I needed.

Ah, you actually shouldn't need anything special - you can just use `0xf002` directly to use it. OCaml supports specifying numbers by their hex notation if you prefix with `0x`: https://caml.inria.fr/pub/docs/manual-ocaml/lex.html

There should be no need for `IconDefinition.parseId` since there's no reason to use a string... you can just use: 
```
<FontIcon icon={0xf002} />
```

> You could generalise that but then the api of the component is too permissive either it optionally takes icon=int and iconStr=string and under the hood does the parsing of the id etc but then it can be passed with a potentially empty value i.e. it no longer demands only one correct thing, since both would have to be options.

I don't see the latest commit, so I may be missing something - but I don't understand why the `<FontIcon />` would need to take either an int or a string.

 One other thing thats related is there is a file that has these values as constants (so you can use `FontAwesome.circle` instead of `0xf111`): https://github.com/onivim/oni2/blob/ad4ca70447c11f1302f2fd1045f0b564f173489d/src/editor/UI/FontAwesome.re#L7

Would be great to add those icons to that list - and then it would be even clearer:
```
<FontIcon icon={FontAwesome.circle} />
```

(as long as the identifier names are in sync with the FontAwesome names)

> Also I guess it does depend on a developer's font but this way means that scanning what icon each of these is rendering is much harder.

Unfortunately I just see this in both GitHub or in the code:

image

image

...not too useful for me ☹️ 

> If we had a group of these and wanted to change one, the only chance you'd have would be to look up the codes? or change stuff till you found the right one 

Does having them stored as constants help here? I think, at least once we have language support hooked up, it could be helpful because we could do a 'find-all references' for `FontAwesome.circle`, etc.

Akin909: Ah I wasn't aware of the constants file or that we could use a `0x` prefix that solves the issue then :+1:

romgrk: I've included the full list of icons in #242, you can pick up that commit if you wish.

bryphe: @romgrk - that's awesome! Thank you!

src/editor/UI/EditorView.re
```diff
@@ -41,10 +42,60 @@ let toUiTabs = (tabs: list(State.Tab.t)) => {
   List.map(f, tabs);
 };
 
+/**
+   We wrap each split component as we have to have a type signature
```
bryphe: Thanks for factoring out and this detailed comment 👍

src/editor/Model/WindowManager.re
```diff
@@ -0,0 +1,115 @@
...
+[@deriving show]
+type dock = {
+  component: componentCreator,
+  width: option(int),
+  height: option(int),
+};
+
+[@deriving show]
+type splitTree =
+  | Parent(direction, int, list(splitTree))
```
bryphe: Thanks for doing the work to factor this into a tree, @Akin909 ! 💯

src/editor/Model/WindowManager.re
```diff
@@ -0,0 +1,115 @@
...
+          newResult,
+          children,
+        )
+    )
+  | Leaf(split) => action(result, split, direction)
+  };
+
+let rec add = (id, split, tree) =>
+  switch (tree) {
+  | Parent(direction, parentId, children) when id == parentId =>
```
bryphe: Do we have to special case if, for example, the Parent has a split direction of `Horizontal`, but we request a `Vertical` split with that parent id? 🤔

Akin909:
### Horizontal Splits:
horizontal-splits

src/editor/Model/WindowManager.re
```diff
@@ -0,0 +1,117 @@
...
+type splitTree =
+  | Parent(direction, int, list(splitTree))
+  | Leaf(split);
+
+[@deriving show({with_path: false})]
+type splits = splitTree;
+
+type t = {
+  windows: splits,
+  leftDock: option(dock),
```
Akin909: Note To Self: Didn't think/realise that these might have to be a list to accomodate the file explorer

src/editor/Model/WindowManager.re
Akin909: that and the `remove` function are still a work in progress, I haven't written down any interim thoughts but was thinking if the new split is changing direction then we'd create a new parent and place it inside but I need to experiment more with the idea. I'll see what I come up with

src/editor/Model/WindowManager.re
```diff
@@ -0,0 +1,117 @@
...
+  | Vertical;
+
+[@deriving show({with_path: false})]
+type split = {
+  id: int,
+  parentId: int,
+  component: componentCreator,
+  direction,
+  /* if omitted the split will grow to occupy whatever space is available */
+  width: option(int),
```
bryphe: Cool! 👍

src/editor/Model/WindowManager.re
bryphe: Ah ok! That plan to add a new parent inside if the direction is different sounds like a good approach to me.

Issue #185, opened 2019-03-27
Using the tab key is unreliable


dannyharding10:
When using the tab key, the cursor moves where it seems it should, but the text entered is far less indented than it should be. You can see the behavior in the attached gif

![bad tabbing](https://user-images.githubusercontent.com/11875246/55049534-a2ab0700-501a-11e9-843c-642b1409b091.gif)

bryphe:
Thanks for logging the issue, @dannyharding10 and for supplying the gif . 

Indeed, we're not handling the 'tab' character correctly here.

PR #233, opened 2019-04-09
Generalize Tokenizer to support additional scenarios

bryphe:
For buffer rendering, we have a `Tokenizer` module that splits out buffer lines into runs of text and positions them. There are a couple of improvements we need for this:
- We currently only split tokens on foreground colors and transition between white-space & non-white-space. We need to also be able to split on whether a token is selected. Needed for #192, to properly split tokens when they cross a selection boundary.
- The current tokenizer logic for positioning is naive, assuming that each character will take up a single column. This isn't the case - for example, depending on tab settings, `/t` might take multiple characters (important for #220 / #185 !)

This splits out the logic into two modules:
- `Tokenizer` is a general module that takes a UTF8 string and will tokenize it, given a split logic and measure function.
- `BufferViewTokenizer` is specific for our buffer view rendering, and provides a split function that splits on whitespace/non-whitespace boundaries, as well as differences in color/background/selection. It also incorporates the indentation settings into the position.


src/editor/Model/Oni_Model.re
```diff
@@ -23,4 +23,5 @@ module StatusBarModel = StatusBarModel;
 module State = State;
 module SyntaxHighlighting = SyntaxHighlighting;
 module Tokenizer = Tokenizer;
+module Tokenizer2 = Tokenizer2;
```
bryphe: Rename `Tokenizer2` -> `Tokenizer`, and rename the old `Tokenizer` to `BufferViewTokenizer`.

src/editor/Model/Tokenizer.re
```diff
@@ -8,118 +8,93 @@ open Oni_Core;
...
-      idx := idx^ + 1;
-    };
-  };
-
-  idx^;
-};
-
-let _moveToNextWhitespace = _moveToNextMatchingToken(_isWhitespace);
-let _moveToNextNonWhitespace = _moveToNextMatchingToken(_isNonWhitespace);
+let tokenize:
```
bryphe: Add benchmark for this; compare with `master` implementation. This is on the rendering critical path.

src/editor/Model/Tokenizer.re
```diff
@@ -8,118 +8,93 @@ open Oni_Core;
...
+      let colorizedToken1 = tokenColorArray[i0];
+      let colorizedToken2 = tokenColorArray[i1];
+      _isWhitespace(c0) != _isWhitespace(c1)
+      || colorizedToken1 !== colorizedToken2;
+    };
 
-    if (startToken < endPos) {
-      let length = endToken - startToken;
-      let text = Zed_utf8.sub(s, startToken, length);
+    let _filterRuns = (r: Tokenizer2.TextRun.t) => {
```
bryphe: Factor out of closure

src/editor/Model/Tokenizer.re
```diff
@@ -8,118 +8,93 @@ open Oni_Core;
...
-  let defaultForegroundColor: Color.t = theme.colors.editorForeground;
-  let defaultBackgroundColor: Color.t = theme.colors.editorBackground;
-
-  let ret: list(t) =
-    switch (tokens) {
-    | [] => _getAllTokens(s, defaultForegroundColor, pos, Zed_utf8.length(s))
-    | [last] =>
-      _getAllTokens(
-        s,
+    let toToken = (r: Tokenizer2.TextRun.t) => {
```
bryphe: Factor out of closure

test/editor/OniUnitTestRunner.re
```diff
@@ -1,4 +1,4 @@
-Oni_Core_Test.TestFramework.cli();
+/* Oni_Core_Test.TestFramework.cli(); */
```
bryphe: Revert these changes!

PR #234, opened 2019-04-09
Add step to build process

rogererens:
I presume one has to enter the directory that contains the package.json file before typing `esy install`


bryphe:
Thanks @rogererens 👍

PR #245, opened 2019-04-13
Fix #236 - Render artifacts due to mismatched background colors

### LINKED ISSUE #236: Visual Selection: Render artifacts due to mismatched background colors
Here's what the visual range selection looks like:
image

We always specify a background color for text to get gamma correction - however, the text is using the wrong background color - it should pick up the selection background color, but it doesn't.

We need to be able to specify the correct background color, accounting for token splitting in selection boundaries. #233 gets us part way there - we need to also plug in the selection range start/end position to 'split' tokens on those. Alternatively, we could populate the array we use to pick up the colors, and override the background color for any cells that intersect the selection range.

bryphe:


src/editor/Model/BufferViewTokenizer.re
```diff
@@ -121,10 +128,11 @@ let tokenize:
     Theme.t,
     list(ColorizedToken.t),
     ColorMap.t,
-    IndentationSettings.t
+    IndentationSettings.t,
+    option(Range.t),
```
bryphe: Add option for defaultBackgroundColor here

PR #252, opened 2019-04-17
Fix #168 - Add max column width for minimap

### LINKED ISSUE #168: Change: Minimap max width
The minimap is currently pretty wide:

image

This is on a 4K monitor with 150% scaling, Windows 10.

I think it probably makes most sense to default to 80 cols, and have a configuration option for more (`"editor.minimap.maxColumn"` in VSCode.) I can't quite tell from the code if this is what is already attempted to be done, so maybe this is only a small change.

bryphe:
__Issue:__ The minimap grows arbitrarily, which isn't very useful past a certain size.

__Fix:__ Add `editor.minimap.maxColumn` configuration setting, to specify the max width of the minimap, with a default value of `120`.


src/editor/Core/Configuration.re
```diff
@@ -45,6 +47,7 @@ type t = {
 let default = {
   editorMinimapEnabled: true,
   editorMinimapShowSlider: true,
+  editorMinimapMaxColumn: 12,
```
bryphe: Fix / revert this

src/editor/Model/Editor.re
```diff
@@ -27,6 +28,7 @@ let create = () => {
     id: 0,
     scrollX: 0.,
     scrollY: 0.,
+    minimapMaxColumnWidth: 12,
```
bryphe: Fix this - use shared constant

PR #271, opened 2019-04-23
Fix #206 - Implement cleanup when quitting

### LINKED ISSUE #206: Error on exit: 'An extension called process.exit() and this was prevented'
We're hitting an error on exit due to the extension host:

```
Error: An extension called process.exit() and this was prevented.
        at process.exit (D:\oni2\src\textmate_service\node_modules\vscode-exthost\out\vs\workbench\services\extensions\node\extensionHostMain.js:18:29)
        at Timeout._onTimeout (D:\oni2\src\textmate_service\node_modules\vscode-exthost\out\bootstrap-fork.js:180:13)
        at ontimeout (timers.js:436:11)
        at tryOnTimeout (timers.js:300:5)
        at listOnTimeout (timers.js:263:5)
        at Timer.processTimers (timers.js:223:10)
```

We need to make sure we properly close down the extension host on close.

bryphe:
__Issue:__ There are several problems / potential crashes on quit, since we currently just kill the host process.

This change improves the situation by allowing features to register handlers on quit, and we execute those prior to actually quitting the app.

In the future, we'd also want to add a 'BeforeQuit' handler hook, so that features can potentially cancel the requested Quit (ie, for an unsaved file).

__Fix:__ This adds `Quit` and a `RegisterQuitCleanup` actions, as well as state to manage it. When `Quit` is dispatched, the `LifecycleStoreConnector` runs through all registered cleanup actions prior to actually quitting.


bryphe:
Future work:

For some operations, we'll want to wait for the cleanup operation to complete (ie, if a process needs to do any bookkeeping prior to close). For now we indiscriminately close the process, but longer-term it'll help to have a hook to handle this more cleanly (the `quitFunction` could take a callback to signify when the operation is complete, and a use case would be waiting for the extension host to close after the `terminate` call)

PR #276, opened 2019-04-24
Feature - Multiple Editor Groups: State management

bryphe:
This adds the state for managing an aggregate of EditorGroups, instead of a singleton. This is infrastructure needed for us to support hosting an EditorGroup-per-Window.

There will still be a bit more work in plumbing the actions correctly (for example, today, every EditorGroup listens for a 'BufferEnter' and spins up a new Editor in response, but we'd want to target this to a specific / active EditorGroup).


src/editor/Model/EditorGroup.re
```diff
@@ -6,7 +6,10 @@
 
 open Oni_Core;
 
+let _latestId = ref(0);
```
Akin909: side note: we have `Revery.UniqueId` so you could use `let editorGroupId = Revery.UniqueId.make({})` then `editorGroupId.getUniqueId` 🤷‍♂️

bryphe: Good point @Akin909 , I'll switch to that 👍

PR #282, opened 2019-04-25
Indent guide fix

### LINKED ISSUE #254: Bug: Indent guides don't extend through blank lines
The indent guides functionality should 'connect' blank lines. Note the gaps in the blank lines in the current implementation:
image

The virtual 'indentation level' of a blank line is a function of the nearest non-blank lines - if the non-blank lines are within 1 indentation level of each other, it can take the larger one. Otherwise, it should take the minimum indentation level.

This is non-local, so it would need to scan up / down to the nearest non-blank line to resolve this.

CrossR:
Fixes #254.

Need to tidy this up, since it feels very messy!

Few things outstanding:

- [x] Performance checking. Should we always recurse until a suitable line is found? Feels expensive, especially when that info is then thrown away and repeated for the next potentially non-blank line.
 - [x] Adding new lines etc doesn't trigger an indent guide to be drawn until insert mode is left. Was that always the case?
- [x] Check some edge cases in both ReasonML and Python (Python ends up being a good thing to check due to the lack of braces meaning syntax objects just stop).


src/editor/UI/IndentLineRenderer.re
```diff
@@ -60,51 +93,69 @@ let render =
     };
 
     incr(l);
+    previousIndentLevel := level;
+
+    if (line == cursorLine) {
+      cursorLineIndentLevel := level;
```
CrossR: Not actually needed for this....was just since we already calculate this, seemed pointless to do it again a few lines later.

CrossR:
One edge case that can't be easily fixed:

image

Line 3 should have the same indent as line 4 (ie take the max). Whereas line 8 should have the same indent as line 9 (ie take the min).

Both of these cases are the same: both lines above and below have none zero indent level, so I'm not sure what we can do. VSCode acts the same way....until you enable the python syntax, in which case the behavior changes:

image
With plain text

image
Set to python

Not something to fix now I don't think.....but to be aware of.

src/editor/UI/IndentLineRenderer.re
```diff
@@ -11,6 +11,53 @@ open Oni_Model;
 
 type bufferPositionToPixel = (int, int) => (float, float);
 
+let rec _getIndentLevel =
```
Akin909: Optional nit: this isn't specific to this PR just happened to mention this time, but I think we should avoid using `_myPrivateVariable` as a way of hiding variables from being exported.

I think if we feel strongly enough that something shouldn't be exported we should use an interface file (`rei`) especially since the compiler reads `_` as things we aren't using.

bryphe: Good point @Akin909 (I need to be better about including `rei` files). 

I wonder if there is a way to lint / check this in our CI to enforce it? I've made this mistake a few times

bryphe:
This is awesome @CrossR , thank you for your work on this!

> Performance checking. Should we always recurse until a suitable line is found? Feels expensive, especially when that info is then thrown away and repeated for the next potentially non-blank line.

I was wondering about this too. A couple potential ideas:
- We can stick with the always-recurse approach for now. We could later move these calculations to something we store as part of the buffer (`cachedIndentLevel` or something), and only update the results on buffer updates. That would mean this expensive work is only doing when the buffer changes, as opposed to every view render.
- We can set some upper bound on how many lines to check (say, 100/500/1000 - something that keeps us from hitting a degenerate case in a very large file).

It might be helpful to put in a quick benchmark for this too.

Thanks for calling out the edge case! I agree - not something we need to fix at the moment, but we can revisit down the road. The behavior you've implemented is an improvement over what we have now! 👍

bryphe:
Merged this up - thank you for the contribution, @CrossR !

PR #286, opened 2019-04-26
Start hooking up missing config options.

### LINKED ISSUE #260: Bug: Changing 'editor.minimap.enabled' has no effect
__Repro:__

- Open Onivim 2
- Press <Control+Shift+P> to open the command palette
- Select 'Edit configuration'
- Change `editor.minimap.enabled` to `false`
- Open command palette again -> Reload Configuration

Expected: Minimap should disappear
Actual: Minimap stays

CrossR:
Fixes #260.

Also hooks up the line option. Going to give a check over of the other options and make sure they are hooked in as well.


CrossR:
I think the main config option that isn't hooked up is the `Tabs`/`Spaces` options.

I don't know if you had any thoughts/ideas on how that should be implemented? I'm fine to try and add that to this, or we can leave this as a small PR to hook up the easy ones and tackle that in a follow up. I assume that is going to require some scaffolding to send VimL commands to set the correct options on load/reload of the config.

CrossR:
Since it looks like the Tabs/Spaces is dealt with elsewhere, I'll give this a once over when I get in from work and check for other config options, but otherwise its complete I think.

bryphe:
Thanks @CrossR ! Looks good to me. I'll merge this up and then bring it once CI is complete.

bryphe:
CI is green! Thanks again for your help with this, @CrossR 👍

PR #290, opened 2019-04-29
[WIP] feat(editor): line wrap - initial scaffolding functions

bryphe:
Currently, in Oni2, there is no way to enable line-wrapping. This is an important feature for any code editor, so ideally we can ship an initial implementation for our June release.

This adds some basic primitives for a naive line-wrap implementation (breaking at characters) - handling at the per-line level, and handling mappings at the per-buffer level. Determining wrap-points is an expensive operation, thus, its important we cache the computation and do a minimum amount of work (the wrap points calculation is a function of `(buffer, wrapOptions, availableWidth)` - if any of those pieces change, we require a recalculation).

Work remaining in this PR:
- [x] Add initial benchmarks for both `LineWrap` and `BufferWrap`
- [x] Instead of `LineWrap` taking a `int` for the column to wrap, we should send it wrap options (ie, `NoWrap`, `WrapColumn(int)`), and forward this from the `BufferWrap` module.
- [x] Implement test case for `bufferPositionToVirtual`, 
- [x] Implement test case for `bufferRangeToVirtual`
- [ ] Implement test case for `update`

In future PRs:
- Cache these calculations in the `Editor` and update at the correct times
- Hook up this infrastructure to the `View` layer (it should get the rendering info from the 'virtual lines' as opposed to directly from the buffer, as it does today)
- Implement additional wrapping strategies (w/o word breaking)
- Update the naive word wrap strategy implemented here to be aware of double-width characters


bryphe:
Superseded by #1969 #1957

Issue #250, opened 2019-04-17
Minimap mouse scrolling input feedback


Apostolique:
Right now, you have to click once to start a dragging behavior. To end the dragging behavior, you need to click a second time.

Instead, the dragging should happen by clicking and holding the click down. On release, the dragging should stop.

____

Also, right now it works on any mouse button. It should be kept on the left button only. 

____

The dragging is also misleading. Instead of moving the minimap position, it moves the scrollbar. I'd expect to be able to click on what I see visually on the minimap so that it brings me there.

Perhaps, put the scrollbar dragging on right click.

Left click -> Scroll minimap.
Right click -> Scroll scrollbar

Edit: I played a bit with the vscode minimap. The behavior is interesting. You can click and drag specifically on the minimap position. It will drag as if it was a scrollbar.

You can single click anywhere else on the minimap to bring the minimap to that visual location. In that case there is no dragging. I will record a video later to showcase the behavior.

KingEdwardI:
I'm working on a fix for this currently. The big issue with it not releasing the click handler had something to do with the upgrade to Revery 0.18.0, it works properly on commits before that. Getting the behavior correct is proving to be trickier.

My plan is to get it to behave like VSCode does. 

@bryphe were there any changes to the way that `Mouse` works between Revery .17 and .18? So far I've figured out how to fix it, but still unsure about the root cause of the problem.

Apostolique:
Doing it like vscode would be better than what I proposed up there. 👍

bryphe:
> @bryphe were there any changes to the way that Mouse works between Revery .17 and .18? So far I've figured out how to fix it, but still unsure about the root cause of the problem.

Sorry I missed this comment, @KingEdwardI ! I don't believe there were any changes to the Mouse event handling, recently

kinglouie:
even after the latest PR the drag behaviour is still a bit funky. Sometimes if you release the left mouse button when dragging over the bounds of the minimap the scrolling is still happening on mouse move and only stop after clicking in the minimap column again.

Phaen:
This issue is still happening, but it's very consistent and easily reproduced. After holding down the left mouse button on the minimap to start a drag event:
 - Release the left mouse button while holding the mouse still: the drag event ends correctly.
 - Release the left mouse button while moving the mouse: the drag event continues with increased sensitivity. Start a new drag event to reset.

[...]

PR #292, opened 2019-04-30
Releasing click stops scrolling

KingEdwardI:
This is a fix for one part of #250

> Right now, you have to click once to start a dragging behavior. To end the dragging behavior, you need to click a second time.


bryphe:
Looks great, @KingEdwardI - thanks for the fix!

PR #293, opened 2019-04-30
Only scroll on left click

KingEdwardI:
This is a fix for a part of #250 

> Also, right now it works on any mouse button. It should be kept on the left button only.

### DIFF
--- a/src/editor/UI/Minimap.re
+++ b/src/editor/UI/Minimap.re
@@ -140,11 +140,13 @@ let createElement =
         Constants.default.minimapCharacterWidth
         + Constants.default.minimapCharacterHeight;
       let linesInMinimap = metrics.pixelHeight / minimapLineSize;
-      GlobalContext.current().editorScroll(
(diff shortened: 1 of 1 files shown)


src/editor/UI/Minimap.re
```diff
@@ -140,11 +140,13 @@ let createElement =
         Constants.default.minimapCharacterWidth
         + Constants.default.minimapCharacterHeight;
       let linesInMinimap = metrics.pixelHeight / minimapLineSize;
-      GlobalContext.current().editorScroll(
-        ~deltaY=scrollTo -. editor.scrollY -. float_of_int(linesInMinimap),
-        (),
-      );
-      setActive(true);
+      if (evt.button == Revery_Core.MouseButton.BUTTON_LEFT) {
```
bryphe: Nice find!

bryphe:
Between this fix, and #292 - big improvement in the minimap experience 👍 Thanks for the help here, @KingEdwardI !

PR #295, opened 2019-05-01
Tidy up test commands, and fix CI test upload.

CrossR:
The `esy test` command wasn't updated, so it failed to run. I've linked that through now, and also removed the CI option from there since we don't especially need it there.

Also updated the CI to upload the JUnit test files.


bryphe:
Ah sorry about that - thanks for catching this @CrossR !

PR #298, opened 2019-05-02
Update EULA v1.0 -> v1.1

### LINKED ISSUE #268: Some questions about the EULA
So I'm super interested in this funding model, and I'm excited to see it succeed :) These are my initial reactions to reading through the EULA, also IANAL but I like to understand what I'm agreeing to 😬

> 1.4 (e) use the Software for creating a product, service, or software that is, directly or indirectly, competitive with or in any way a substitute for services, product or software offered by Company;

This sounds like it means "You can't use Oni as a code editor to edit the code of a competing editor", which is Very Weird. Is that what it means?

> 2.1 (a) Under a Non-Commercial License, Licensee will not use the Software: (a) in conjunction with any other copies or versions of the Software, under any type of License model;

Does this mean that, should Oni implement a collaborative editing feature, users w/ a Non-Commercial license would be unable to collaboratively edit with users w/ a Commercial license?

> 3.1 (b) Licensee shall in respect of the Source Code comply strictly with all other restrictions applying to its use of the Software under this EULA **as well as any other restriction or instruction that is communicated to it by Outrun Labs at any time during the Agreement.**

This "that is communicated" bit seems unreasonably general. Could it say something like "any updates that are made to the EULA" to at least restrict it to instructions given in writing?

> 4.1 Outrun Labs owns and will retain all legal rights in the software

How does this relate to contributors from outside of Outrun labs? Seems like it would be something of a barrier to contribution.

> 6.5 If Licensee wishes to opt out of auto-renewal, then You must email license@outrunlabs.com providing details of the Subscription Licenses which you wish to opt-out

Hopefully this would be updated to some more automatic process in the future?

bryphe:
Fix #268 - loosen a few of the restrictions in the EULA:

- Remove section 1.4e - competitor clause
- Remove section 2.1a - remove non-commercial conjunction clause
- Loosen restriction on 3.1.b 

We actually were able to address almost every question in that #286 , which is great! The only clause we needed to retain from the feedback was 4.1. 

We'll address that particular issue by making the CLA more permissive - allowing contributors to relicense their contributions under MIT, if desired.

### DIFF
--- a/Outrun-Labs-EULA-v1.1.md
+++ b/Outrun-Labs-EULA-v1.1.md
@@ -4,7 +4,7 @@ Copyright 2019 Outrun Labs LLC
 
 Outrun Labs LLC
 END USER LICENSE AGREEMENT
-3/7/2019
+5/2/2019
 
 PLEASE READ CAREFULLY BEFORE USING THIS PRODUCT: This End-User License Agreement ("EULA") is a legal agreement ("Agreement") between (a) you, either an individual or a single entity ("Licensee") and (b) Outrun Labs LLC ("Outrun Labs") that governs your use of software programs ("Software").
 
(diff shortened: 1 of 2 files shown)


bryphe:
Thanks for linking the issue, @CrossR ! I'll bring this in now 👍

PR #302, opened 2019-05-04
File Explorer

Akin909:
This PR adds a File Explorer to Oni2

Files are sourced using `Lwt_unix`, `FileExplorer` functionality is scoped to the module and used in a `FileExplorerConnector` which is triggered via actions in a new `FileExplorerView`.

This change also includes the ability to click to open files or expand a directory node. The explorer can be opened and closed using `C-S-B`. Currently the explorer reuses the same icon theme as the tabs.

### Updated: with icons
explorer

#### Outstanding:
- [x] move state into a `connector`
- [x] click handlers
- [x] populate files and subfolders
- [x] open/close

#### Follow up PR:
- [ ] Find/use vscode file explorer icon theme
- [ ] hide scrollbar till `onHover` implemented
- [ ] Add UI focus to explorer panes
- [ ] Add vim bindings and navigation - this is __?much__ bigger than it sounds as we will need to implement a `VimNavigator` like we had in oni1 (maybe) which was actually powered by a second neovim instance :thinking: :sweat_smile:


src/editor/UI/FileExplorerView.re
```diff
@@ -0,0 +1,43 @@
...
+        hooks,
+      );
+
+    let onNodeClick = (clicked, tree) => {
+      Tree.(
+        switch (clicked) {
+        | Node({data}, _) =>
+          let node = FileExplorer.toFsNode(data);
+          GlobalContext.current().dispatch(SetExplorerTree(tree));
+          GlobalContext.current().dispatch(OpenFileByPath(node.path));
```
CrossR: Do we want to gate this to run only on `node.isDirectory` is false? Else we would be opening folders like files.

src/editor/Store/FileExplorerStoreConnector.re
```diff
@@ -0,0 +1,34 @@
...
+
+  let updater = (state: State.t, action: Actions.t) => {
+    switch (action) {
+    | OpenExplorer(directory) => (
+        state,
+        getExplorerFilesEffect(
+          directory,
+          state.languageInfo,
+          state.iconTheme,
+          ["node_modules", "_esy"],
```
CrossR: Should we make an issue to add in a config option for this?

Ie, hooking up `"files.exclude"` like in VSCode.

Its a decent good first issue as well I think.

Akin909: definitely :+1:, I just got lazy at that point/didn't want to touch anymore files so this PR didn't get too big (which is a bad excuse, was primarily laziness)

Akin909: decided to fix this here rather than create more work later

somebody1234: @CrossR seems like it might be better to read `.gitignore` in this case?

CrossR: Could be worth opening an issue for a follow up for that!

Adding it now doesn't make much sense since I think it would be a bigger task, since we'd need to either load/parse a users' gitignore, or use a tool that does that for us. We get it for free with the QuickOpen, but not here unfortunately.

(Its free for the QuickOpen as that uses ripgrep that automatically uses `.gitignore` files)

Akin909: yeah, doing `.gitignore` stuff is definitely out of scope for the original implementation

somebody1234: Hmm. then... I think specialcasing those two isn't a great idea (seems like you're still specialcasing them in `default`), plus it might be a good performance test

CrossR: VSCode does come with a default list:

```json
"files.exclude": {
  "**/.git": true,
  "**/.svn": true,
  "**/.hg": true,
  "**/CVS": true,
  "**/.DS_Store": true
}
```

`node_modules` is usually excluded as well, and since this is a reason project `_esy` makes sense too.

We can always add a performance benchmark as part of the follow up tasks. Would probably fit well with the checks around listing the home folder etc (ie making sure we bail out if we start recursing too far).

This all seems to make sense for the base implementation though!

Akin909:
Managed to add the `depth` concept to a branch based off of this one, so I pass in a `maxDepth` so the traversal only goes so deep, then on clicking on a node that was at the limit we fetch the children for that node going down another `maxDepth` amount.

Seems to work pretty nicely, but might save that branch for after this goes in :man_shrugging:

src/editor/Model/FileExplorer.re
```diff
@@ -0,0 +1,164 @@
...
+   to resolves its subfolders and files. We do this concurrently using
+   Lwt_list.map_p
+
+   TODO: Add the concept of depth to this function so we can terminate
+   the file resolution early and avoid choking on large directories
+ */
+let rec getFilesAndFolders = (cwd, getIcon, ~ignored) => {
+  /* Wrap the operation in a try%lwt which will catch all async exceptions */
+  try%lwt (
+    Lwt_unix.files_of_directory(cwd)
```
bryphe: Neat! Good idea to use `Lwt` here 👍

bryphe: Will be nice for perf because the async I/O can be run in parallel with other threads (like the rendering).

src/editor/Model/UiTree.re
```diff
@@ -0,0 +1,21 @@
+include Revery.UI.Components.Tree;
```
bryphe: Great to be able to use this here  👍

bryphe:
Woah just tried this out - so cool @Akin909 ! 😍 

> Files are sourced using Lwt_unix, FileExplorer functionality is scoped to the module and used in a FileExplorerConnector which is triggered via actions in a new FileExplorerView.

Sounds good to me 👍  Neat to see how you were able to leverage the Tree work you did in Revery for rendering the view.

> This change also includes the ability to click to open files or expand a directory node. The explorer can be opened and closed using C-S-B. Currently the explorer reuses the same icon theme as the tabs.

Very cool to see the icons in the Explorer. 

In terms of next steps: 
>  Add UI focus to explorer panes

This would be great! We need a story in general for navigating between the window splits / panes we manage. So that you could `<C-w-h>` over to the explorer view from a buffer (and back to the buffers with `<C-w-l>`).

> Add vim bindings and navigation - this is ?much bigger than it sounds as we will need to implement a VimNavigator like we had in oni1 (maybe) which was actually powered by a second neovim instance 

For sure - this is a big work item. On Discord, I'm thinking that the `VimNavigator` powered by the second neovim was perhaps overly complicated for what we got from it (in retrospect 🤔 - this was my fault). Perhaps there is a different / simpler paradigm, or a better abstraction, we could explore for the vim bindings / navigation.

Anyway awesome work @Akin909 , thank you - looks great to me!

src/editor/Model/FileExplorer.re
```diff
@@ -0,0 +1,186 @@
...
+        Lwt.bind(promise, files =>
+          Lwt_list.map_p(
+            file => {
+              let path = Filename.concat(cwd, file);
+              let isDirectory = isDir(path);
+              let nextDepth = depth + 1;
+              let notMaximumDepth = nextDepth < maxDepth;
+
+              let%lwt children =
+                isDirectory && notMaximumDepth
```
Akin909: I've added the depth concept to this function so we can control how far this function recurses.

If run from a very deep directory we hit `EMFILE` error which is a unix error that says that our process has too many files open, this limit I think can be increased and I've seen a similar issue on vscode on my system.

With this though we only got 10 levels deep and if a user is trying to go further then when they expand a node at depth level 10, we fetch its contents (pretty seamlessly perceived performance wise. I didn't actually believe it was working so tested it a bunch, its super fast and imperceptible)

Issue #263, opened 2019-04-19
Bug - Keybinding: Command Palette on OSX 


bryphe:
`<Cmd+Shift+P`> should open command palette on Mac. However, you need `<Control+Shift+P>`, right now. 

This will be jarring and unintuitive for users coming from other editors (ie, VSCode and Onivim 1!), so we should fix this binding to use `Cmd` on Mac instead of `Control`. We'll want to check out keybindings and make sure we're handling this correctly - I believe VSCode, for example, has faculty for saying something like `CmdOrControl` in the key bindings.

maxdietrich:
Hey, I'd like to look into this if no one is working on this right now.

bryphe:
That would be great @maxdietrich - thank you for the help!

maxdietrich:
Okay, I had a look into this. And with my limited context, this looks like it needs to be addressed upstream in `reason-glfw` first. The callback defined [here](https://github.com/onivim/oni2/blob/master/src/editor/bin/Oni2.re#L160) has a `mods` argument of the type defined [here](https://github.com/bryphe/reason-glfw/blob/793be75dc868bf3834c03811501f4e1b0c4f8943/src/glfw.re#L71). There is a handling for Shift, Ctrl, Alt and Super, but not for Cmd yet. Without receiving the info whether the `Cmd` modifier is pressed, I don't see a way of fixing this Issue inside `oni2`.

CrossR:
I think (not 100%) that super is `Cmd` on Mac. It seems to change name a bunch, but going off this old GLFW reference (https://www.glfw.org/GLFWReference278.pdf) it seems to be.

```
GLFW_KEY_LSUPER                         Left super key, WinKey, or command key
GLFW_KEY_RSUPER                         Right super key, WinKey, or command key
```

Since it can be the Win key on Windows, you'd need to make sure to gate the command to being Mac only.

maxdietrich:
@CrossR you're right! I've got a version of this running, but there is one issue left. For now, I used `Sys.unix` to decide which Command Modifier Key to use. This would change it so that Linux users would also have to press their super key instead of Ctrl, which I personally would expect as a Linux user. I guess you have more experience in Reason and can maybe point me to a way to explicitly check for MacOS and not just unix generally? I have a hard time finding something that does that. Is there maybe a way integrating JS can solve that issue?

CrossR:
[...]

PR #303, opened 2019-05-04
Change command modifier for MacOS to Cmd

maxdietrich:
This addresses https://github.com/onivim/oni2/issues/263

I had a hard time writing test cases for this, mainly because I didn't really know how to create a Key-Event object to pass to the corresponding method. Since the rest of `Input` isn't tested so far, I assume, others had similar issues?

I'd be happy for feedback either way, since this is my first contact with real world Reason.

### DIFF
--- a/assets/configuration/keybindings.json
+++ b/assets/configuration/keybindings.json
@@ -1,10 +1,14 @@
 {
     "bindings": [
         { "key": "<C-P>", "command": "quickOpen.open", "when": [["editorTextFocus"]] },
+        { "key": "<D-P>", "command": "quickOpen.open", "when": [["editorTextFocus"]] },
         { "key": "<S-C-P>", "command": "commandPalette.open", "when": [["editorTextFocus"]] },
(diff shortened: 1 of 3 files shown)


CrossR:
Thanks for the PR!

I think leaving the tests for this should be fine, we can add tests for all of the Input in one go.

It looks like there is a CI failure, but you should be able to fix that by running `esy format` and then pushing up the changes.

I've stuck one comment on the PR as well.

bryphe:
Thank you for the contribution @maxdietrich (and thanks @CrossR for the review)!

Just tried it out on Mac, Win - works great! Glad to have a precedent for this.

> I had a hard time writing test cases for this, mainly because I didn't really know how to create a Key-Event object to pass to the corresponding method. Since the rest of Input isn't tested so far, I assume, others had similar issues?

Agree - thank you for thinking about this. IMO we should add an API to Revery to make it easy to generate key events - logged an issue on Revery here: https://github.com/revery-ui/revery/issues/482 

Cheers & thank you again - bringing this in now 👍

PR #305, opened 2019-05-07
Add error handling to Explorer

Akin909:
This PR adds some error handling to the calls to get the folders and files for the file explorer, if an error is encountered trying to resolve subdirs we abort returning an empty children list and log the error. This allows the function to populate as much as possible.


bryphe:
Thanks for thinking about this @Akin909 ! Looks great 👍

PR #306, opened 2019-05-07
Bugfix: Clicking the 'x' doesn't close editors

bryphe:
__Issue:__ Clicking the 'x' on an editor does not work

__Defect:__ We weren't piping in the 'editor id' - the id used for selecting / closing editors. There was a translation in the middle that would default to an id of 1 if there was no buffer associated, and otherwise, it would use the buffer id, which isn't guaranteed to line up with the editor id.

__Fix:__ Use the editor metadata directly for populating the Ui tabs. Remove unused code.


src/editor/Core/Utility.re
```diff
@@ -89,3 +89,17 @@ let escapeSpaces = str => {
   let whitespace = Str.regexp(" ");
   Str.global_replace(whitespace, "\\ ", str);
 };
+
+let filterMap = (f, l) => {
```
Akin909: I actually found out recently that there is a `List.filter_map` in the standard lib

Issue #311, opened 2019-05-08
Performance: Elements 'snap-in' on load


bryphe:
It looks like layout is being blocked while the File Explorer loads:

![elements-snap-in-on-load](https://user-images.githubusercontent.com/13532591/57390749-e5134800-7171-11e9-803b-7f7c8bbe2ff2.gif)

This adds some 'jank' to the loading experience. 

I wonder if there are any blocking portions of loading the explorer?

Akin909:
@bryphe just tried opening it repeatedly on my machine and I can't replicate this on my machine unless I up the recurse depth, but in that case it does do something similar.

Could be that I've missed a trick in the `docs`(https://ocsigen.org/lwt/4.1.0/manual/manual). Two things of note is that when `Lwt_main.run` is called it waits for the promise to be fufilled, I don't quite know how `Isolinear` works but the `run` is called in an effect there so maybe waiting for that promise stops things temporarily. I also just realised that although it talks about a separate thread a lot I think you have to use `Lwt_preemptive.detach` to actually run the promise in a seperate thread

Akin909:
Scratch that `lwt_preemptive.detach` is intended for long running *non*-lwt functions

Akin909:
I might have jumped in too heavily on the `Lwt` aspect of things I guess something to also consider is rendering/processing the tree

Akin909:
Ok I'll stop spamming this thread I'll look into it (unless someone else has already), it could be that I'm misusing `Lwt` in parts :thinking:

Akin909:
Dropping the `maxDepth` to `5` or so completely resolves the issue in large directories on my machine, although before implementing that I'm going to look into if/how I can get `Lwt` to handle this differently.

bryphe:
Cool, thank you for investigating this @Akin909 ! 💯 

> Dropping the maxDepth to 5 or so completely resolves the issue in large directories on my machine, 

This seems reasonable - we don't expand directories by default, I don't think, so would it be OK to just use a `maxDepth` of `1` for the initial load - and then go deeper when the user expands a folder? 🤔

Akin909:
@bryphe tbh that's definitely doable its probably what I'd have done from the outset if I'd implemented it with the depth concept at the beginning as all the folders are closed on load. I actually tested with one when implementing the depth, there's no real difference to the user :man_shrugging:.

But yeah it would be nice for `Lwt` to handle all the processing of to the side and only report back when the tree is loaded

[...]

PR #320, opened 2019-05-13
Performance - Startup: Limit initial population of FileExplorer to 1 level

bryphe:
__Issue:__ When opening Oni2 in some folders, the startup time would be quite long (ie, opening the root directory of the drive).

__Fix:__ The population logic was traversing up to 5 levels deep, which is expensive in a big folder. Since we only show a single level for now, we can restrict to a single level for the initial load. Verified that expanded later worked as expected.

### DIFF
--- a/src/editor/Core/Constants.re
+++ b/src/editor/Core/Constants.re
@@ -37,6 +37,6 @@ let default: t = {
   minimapLineSpacing: 1,
   scrollBarThickness: 15,
   minimapMaxColumn: 120,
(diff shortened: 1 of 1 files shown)


bryphe:
Fixes #311

Akin909:
@bryphe looks good :+1:, sorry was meaning to make a PR for this but was still fiddling with `lwt`

bryphe:
Thanks for reviewing @Akin909 ! 💯

PR #323, opened 2019-06-06
Contributing: make installation instructions in codeblocks rather than in list

vinitkumar:
cc @bryphe


bryphe:
Looks great @vinitkumar ! Thank you for taking the time to improve this 👍

PR #326, opened 2019-06-07
Initial libvim integration

bryphe:
This is the initial work to hook up [libvim](https://github.com/onivim/libvim) and [reason-libvim](https://github.com/onivim/reason-libvim) to the project. 

This embeds and links in a Vim implementation that we can run directly w/o RPC - simplifying our architecture in terms of the way we interact with it. In this new model, input is treated as being handled synchronously - and we can always check and introspect the state of Vim after a key input. (Note - even though the input is synchronous, we'll be able to parallelize it with our UI rendering - the plan is eventually to move the entire Store update logic to a thread).

The motivation for this was an intermittent crash we experienced in our RPC handling - since we've dealt with issues in the past, my preference was to remove that complexity completely and go with a direct linking model.

Still a lot work to stabilize - I'll have some more details out in a blog post / update in the next week or so.

__EDIT:__ Blog post here: https://www.patreon.com/posts/onivim-2-update-27614608


bryphe:
This is getting close to being mergeable! With the latest iteration, it should be buildable, and some operations basically usable.

Some blockers that I would like to address before merging:
- [ ] The `c` change operator hangs - needs to be fixed in `libvim`: <https://github.com/onivim/libvim/pull/76>
- [ ] The window commands `zz`, `zb`, etc need to be hooked up again

And then some work I'll take after this is merged:
- Re-integrating commandline completion
- Integrate bracket-pair matching
- Integrate search highlights
- Removing the pre-built `nvim` binaries and setup collateral for now

bryphe:
Still some bugs to fix - but at a point where we can start getting feedback and help testing 👍

PR #332, opened 2019-06-19
fix(docs): Fix `npm install` typo

billyvg:
I'm assuming this should be `npm install` instead of `node install`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ esy build
 
 ```sh
 cd src/textmate_service
-node install
+npm install
 npm run build
 ```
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for the PR @billyvg !

Actually, our step here is a bit non-standard - we need to run the `install.js` instead of npm install. I added some details here in the document: https://github.com/onivim/oni2/blob/master/CONTRIBUTING.md#build-the-textmate-service

I hope that helps clarify! Let me know if you have any suggestions - in the meantime I'll close this out.

billyvg:
Thanks for the clarification!

PR #346, opened 2019-06-22
Extensions: Add xml/yaml syntax extensions from VSCode

bryphe:


mhanberg:
Would these be better as git submodules? Or is there value in having them vendored?

(I am also not sure if it is possible to submodule them)

bryphe:
Yes, good point - I think we'll bundle up the VSCode extensions in a separate package and bring them over, potentially.

bryphe:
Will close this for now - but we can bring in the XML or YAML extensions separately in the meantime

PR #362, opened 2019-06-26
Fix #360 - Add a few additional keys to ignore on the Vim side

### LINKED ISSUE #360: Pressing shift inserts junk data into buffer (linux) (Arch)
1. Build and launch Oni2 on Arch
2. Press the shift key
3. On the first press, the string "FT>" gets inserted into the buffer, on every successive press "<SHIFT>" is inserted instead

Not confirmed/checked on other distros yet

bryphe:
__Issue:__ Several keys aren't properly handled by `libvim` - we prune those out so we don't send them. This manifests in a few ways: 
- `<SHIFT>` gets interpreted literally, like in #360 
- `<S-C->` puts `libvim` in a blocking mode and crashes Onivim 2

__Fix:__ We should make `libvim` itself more robust in the face of these inputs - but for now, we'll sanitize these on the Onivim 2 side.

Fixes #360

### DIFF
--- a/src/editor/Store/VimStoreConnector.re
+++ b/src/editor/Store/VimStoreConnector.re
@@ -175,7 +175,7 @@ let start = () => {
         if (!String.equal(key, "<S-SHIFT>")
             && !String.equal(key, "<C->")
             && !String.equal(key, "<A-C->")
-            && !String.equal(key, "<SHIFT>") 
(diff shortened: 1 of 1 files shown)


bryphe:
(I've force-pushed the build/arch-build-override branch with this fix, too)

Issue #366, opened 2019-07-01
Configuration - Auto-reload configuration when saving


bryphe:
__Issue:__ Today, when changing `configuration.json`, there is no auto-reload of the configuration. After saving, you have to manually open the command palette and run `Control+Shift+P` -> `Reload configuration`.

Here's an example repro:
![reload-configuration-manual](https://user-images.githubusercontent.com/13532591/60404248-e5b9d080-9b5b-11e9-843e-539af77d206d.gif)

We could make this smoother by auto-reloading the configuration when we know it has been saved. The most reliable way to make this work would be to implement a file-watcher to watch the configuration file, but I don't think we have a cross-platform one at the moment easily available via reason.

An alternative, to at least make this a little better today, would be to watch for BufferWritePost autocommands, and check if the buffer path matches the configuration file.

We have these pieces available today - we can listen to autocommands from `reason-libvim` via the [Autocommands](https://github.com/onivim/reason-libvim/blob/master/src/AutoCommands.re) module (we use this in our [`VimStoreConnector`](https://github.com/onivim/oni2/blob/master/src/editor/Store/VimStoreConnector.re))

In addition, we just need to dispatch an `Actions.ConfigurationReload` to pick up the configuration changes: https://github.com/onivim/oni2/blob/a765ed93fd8ac0dd9489852cf01a38c643fc364e/src/editor/Store/ConfigurationStoreConnector.re#L42

So the idea would be to wire these up - listen to the buffer write autocmd, check if matches the configuration file, and if it does, dispatch the `Actions.ConfigurationReload` action.

tatchi:
Hey @bryphe, 

I would like to work on this one if not taken yet 😀

tcoopman:
This one can be closed now?

bryphe:
Indeed, thanks @tcoopman - fixed by @tatchi 's PR in #371 👍

PR #371, opened 2019-07-02
Configuration: Auto-reload when saving

tatchi:
My attempt to implement  #366

It should work but I'm not sure I've done it the best way so don't hesitate if you see any improvement.

### DIFF
--- a/src/editor/Store/ConfigurationStoreConnector.re
+++ b/src/editor/Store/ConfigurationStoreConnector.re
@@ -8,12 +8,28 @@ open Oni_Core;
 open Oni_Model;
 
 let start = () => {
+  let configurationFileName = "configuration.json";
+  let reloadConfigOnWritePost = (~configPath, dispatch) => {
+    Vim.AutoCommands.onDispatch((cmd, buffer) => {
+      let bufferFileName =
+        switch (Vim.Buffer.getFilename(buffer)) {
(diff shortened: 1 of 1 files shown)


bryphe:
Looks great @tatchi !

Just tried it out on Windows and Mac:
configuration-reload

So much nicer than before 😍 Thanks for the contribution!

bryphe:
Looks like there was a warning during build which caused the CI to fail:
```
Error: This expression has type unit -> unit
       but an expression was expected of type unit
       because it is in the left-hand side of a sequence
       Hint: Did you forget to provide `()' as argument?
```

(The `Vim.AutoCommands.onDispatch` returns a function that you can call if you want to unsubscribe - which we don't need to worry about this case, we'll subscribe for the lifetime of the app. But I added a `let _ = ` to suppress the error)

I pushed that change up - will bring this in once it's green -

tatchi:
Thanks,

The CI still fails but I don't really understand why and if it's related to the code.

Do you have any idea?

bryphe:
@tatchi - I just checked it out, it looked unrelated to the code - seems there were some '502' errors coming from NPM when installing packages (maybe related to the cloudflare outage yesterday).

Kicked it off again and it passed now - will bring it in now. Thank you for the contribution!

Issue #364, opened 2019-07-01
Bug - Explorer: Explorer not sorted by files / folders


bryphe:
__Issue:__ The sorting for the explorer, at the moment, sorts only alphabetically by name, and does not differentiate between files and folders.

This is a bit non-standard, as usually the folders are grouped at the top (this is the case for Oni1, and for other editors, like Sublime and VSCode). 

This is what Onivim 2 looks like right now:
![image](https://user-images.githubusercontent.com/13532591/60404028-501d4180-9b59-11e9-80ed-24115b552d1d.png)

__Expected:__ Folders like `.git`, `assets`, `bench, `bench.esy.lock`, etc should be grouped at the top of the `Oni2` folder - and then files beneath that.

A particular problem with the experience we have right now is that, when expanding a folder, it can be tough to know if a file is part of that expanded folder, or not - see `azure-pipelines.yml` in the below screenshot:
![image](https://user-images.githubusercontent.com/13532591/60404041-86f35780-9b59-11e9-93a9-6dabdfa86bb5.png)

__Where to start:__

The code to load the directory tree is here: https://github.com/onivim/oni2/blob/c2c46a9d60a1cca575e72102a25e0e5fc789f938/src/editor/Model/FileExplorer.re#L156

And the code that converts the list to tree nodes is here:
https://github.com/onivim/oni2/blob/c2c46a9d60a1cca575e72102a25e0e5fc789f938/src/editor/Model/FileExplorer.re#L128

I think we need to apply a sorting strategy to the `children` list created in the `listToTree` - based on whether the item is a file or a directory. Alternatively, we could potentially insert some sorting where we get the list of files and folders: https://github.com/onivim/oni2/blob/c2c46a9d60a1cca575e72102a25e0e5fc789f938/src/editor/Model/FileExplorer.re#L89

r281GQ:
I'm happy to take that one.

tatchi:
@bryphe this one can be closed right?

r281GQ:
I think so.

bryphe:
Indeed, thank you for the fix @r281GQ !

PR #377, opened 2019-07-03
Add sorting by folder && name

r281GQ:
For issue #364 . I just added this sorting to FileExplorer.re.  But there is a slight issue. As you can see on the pictures, it sorts as expected expect the first item goes as the last.

file_explorer

 ```reason
let rec traverse =
              (tree: Oni_Model__.UiTree.tree(Oni_Model__.UiTree.treeItem)) => {
        UiTree.(
          switch (tree) {
          | Node({data: FileSystemNode({displayName, _}), _}, list)
              when List.length(list) === 0 =>
            displayName |> Console.log
          | Node({data: FileSystemNode({displayName, _}), _}, list) =>
            List.fold_left((a, b) => b |> traverse, (), list)
          | Empty => "empty" |> Console.log
          }
        );
      };
```

Now I made this noob traverse to to see what is going on in the [FileExplorerStoreConnector.re](https://github.com/onivim/oni2/blob/283304d4f9e8f492c4f44b5263d5ec00ecc3b58d/src/editor/Store/FileExplorerStoreConnector.re#L15) just before the tree got dispatched to the global store. It has the right format, so first item is the first there. (.ci goes before .git)

So that made me think, that something is going on in the renderer, but as I checked Tree is part of Revery. 

I looked at it, maybe [this](https://github.com/revery-ui/revery/blob/de0ecfd6e3c902215790a94be8b4c2f6beefb1ad/src/UI_Components/Tree.re#L107) has to do with it?

### CODE REVIEW

**Comment by bryphe** [comment]
Change looks great @r281GQ !

Thanks for investigating into that trailing `.ci` issue - good idea to log out the node hierarchy before it makes it to the `<Tree />` component. 

> I looked at it, maybe this has to do with it?

Yes, nice catch! I believe you're right, the logic you highlighted doesn't seem correct - seems like that block is putting `firstChild` at the very end of the tree.

File: src/editor/Model/FileExplorer.re [hunk 1]
```diff
@@ -148,7 +169,22 @@ let rec listToTree = (~status, nodes, parent) => {
         Node({id, data: node, status: Closed}, descendants);
       },
       nodes,
-    );
+    )
+    |> List.sort((a, b) =>
+         UiTree.(
+           switch (a, b) {
+           | (
+               Node({data: FileSystemNode({isDirectory: true, _}), _}, _),
```
bryphe: Nice 👍

bryphe:
I tested out your theory in https://github.com/revery-ui/revery/pull/519 and linked it locally by adding a resolution in Oni2's root `package.json`:
```
"revery": "link:/revery"
```

And built with your changes - and indeed the order looks correct!

image

Great find & investigation @r281GQ . Everything looks correct in this PR, so we can bring this in - and then patch + upgrade Revery 👍

PR #381, opened 2019-07-04
change sorting of the file explorer

tcoopman:
This is just a different implementation of the same thing (#377) Not sure if it's worth merging, but to me it is a bit more readable.

Feel free to close if it doesn't add any value.


bryphe:
Thank you @tcoopman ! I do think putting the `|>` pipes sequentially and factoring out the logic does improve readability.

PR #393, opened 2019-07-10
Fix #139 - Don't render parts of lines that are offscreen

### LINKED ISSUE #139: Performance: Long Lines
There's a bug right now where we are drawing _every token of a line_ - you can imagine when a file has very long lines - or a single long line - this would be a performance killer!

We need to optimize both the minimap and surface to pick out only the renderable tokens, and skip other ones.

In addition, we should disable syntax highlighting for very long lines (as VSCode does)

bryphe:
__Issue:__ When there are long lines on a file, even the parts that are offscreen are being rendered - this is wasteful and can incur a tangible performance penalty on rendering as described in #139 

__Defect:__ We just tokenize the entire line, even parts that are offscreen.

__Fix:__ Only tokenize + render the part of the line that is on-screen.

### DIFF
--- a/bench/lib/TokenizerBench.re
+++ b/bench/lib/TokenizerBench.re
@@ -28,12 +28,12 @@ let tokenizeWithoutSplits = () => {
 };
 
 let tokenizeWithSplitsSub = () => {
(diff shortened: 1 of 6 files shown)


bryphe:
Added a benchmark for the core tokenization algorithm, to compare with getting a subset of a line vs the entire large line - big improvement!

image

PR #402, opened 2019-07-11
Fix #374 - Change editorLineNumber default to 'on'

### LINKED ISSUE #374: Line number isn't displaying correctly when scrolling
So I notice this behaviour when scrolling up and down.

given a file with 10 lines and the cursor is on line 5, the line numbers are displayed like this:

```
1 2 3 4   5    6 7 8 9 10
4 3 2 1 ->5<-  1 2 3 4 5 
```

Gif:

bryphe:
__Issue:__ The `editor.lineNumber` default of `Relative` is confusing, especially if you've never encountered it in Vim before. Even Vim doesn't default to this setting. It's an opinionated remnant of my config.

__Fix:__ Change the default to `On` which is more in-line with other editors. In addition, I noticed `off` wasn't being picked up - so I added a fix for that as well.

Fixes #374

### DIFF
--- a/src/editor/Core/ConfigurationValues.re
+++ b/src/editor/Core/ConfigurationValues.re
@@ -35,7 +35,7 @@ let default = {
   editorMinimapEnabled: true,
   editorMinimapShowSlider: true,
   editorMinimapMaxColumn: Constants.default.minimapMaxColumn,
-  editorLineNumbers: Relative,
+  editorLineNumbers: On,
   editorInsertSpaces: false,
   editorIndentSize: 4,
   editorTabSize: 4,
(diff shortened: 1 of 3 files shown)


CrossR:
I think this actually isn't fixed (I've stuck a longer discussion of the fix over in the linked issue above).

Currently since we write out the default config by copying the file `assets/configuration/configuration.json` to `~/.config/oni2/` (or Windows equiv), the user will still have `Relative`.

I've updated that issue to instead be about auto-generating the JSON from the defaults, which is both easier from a duplication of info PoV, but also fixes the relative file copy issue that the issue was originally about.

PR #404, opened 2019-07-11
Hook up onDirectoryChanged to Explorer.

CrossR:
Tiny PR to pipe this through and update the explorer.

The quick open seems to just use the `cwd` already, so this isn't necessary for that.


bryphe:
Awesome! Been missing this. Thanks @CrossR - looks great.

PR #413, opened 2019-07-12
Offset highlights when scrolling.

### LINKED ISSUE #363: Backgroung highlight pos. is wrong, if horizontal scroll pos. ≠ "LeftMost"
There is some column indexing error in highlighting in Visual Mode.

Long lines doesn't get highlighted correctly when horizontal scrollbar is is not in its leftmost position: background coloring doesn't match cursor position, and characters gets this weird inner highlighting.
The Glith width/offset is proportional to the scrollbar position -> it dissapears completly if scrollbar returns to 0 ( horizontal scrollbar can't be seen as of 54a7a66)

Screenshot shows highlight started from line ending:
highlight glitch

CrossR:
Fix #363.

If we are scrolled, then take that from the final offset.

`getCharacterPositionAndWidth` is used in 3 places, and I've updated 2 places to use it. The final place is in the calculation of the cursor offset and cursor char width. As far as I can tell, that is fine so I've left it be for now.


bryphe:
Looks great @CrossR - thank you! Glad these highlights are in the right spot now - it was jarring when a search was active and they wouldn't move when scrolling horizontal. 🎉 

Sorry about the CLA signature being required again - I fixed a typo in it (Contributon -> Contribution) and forgot it invalidates them

PR #421, opened 2019-07-13
Only increment or decrement the selected item, when items exist.

### LINKED ISSUE #419: Pressing tab during command completion with no completion options crashes editor
Try to cd somewhere, type some text that can't match any subfolder of current folder, press the tab key

Expected result: nothing happens, maybe message that no matching options were found
Result: editor crashes and exits

CrossR:
Fix #419.


bryphe:
Nice! Thanks for the quick fix, @CrossR 😍 Looks great!

PR #432, opened 2019-07-14
Fix Buffer Walking

### LINKED ISSUE #424: viewport walks when changing splits
1. Open two files side by side in a split
2. do "G" in one file to go to the bottom
3. use ctrl+w and then h or l to switch between splits

observed result: the pane where viewport was moved to the bottom of the file advances the cursor and the viewport up by one line every time I switch to that split

CrossR:
Fix #424

Change to fix buffer walking.

I've made this WIP for now, since I want to give it a more thorough check over with a clean pair of eyes in a day or so. Its pretty confusing looking over the various types of One vs Zero indexed parts, and working out if bits should be +1 or not...

This does fix the issue, but just want to ensure it makes sense and doesn't change anything else.

### DIFF
--- a/src/editor/Model/Actions.re
+++ b/src/editor/Model/Actions.re
@@ -81,8 +81,8 @@ and editor = {
   bufferId: int,
   scrollX: float,
   scrollY: float,
-  lastTopLine: int,
-  lastLeftCol: int,
+  lastTopLine: Index.t,
+  lastLeftCol: Index.t,
   minimapMaxColumnWidth: int,
   minimapScrollY: float,
   /*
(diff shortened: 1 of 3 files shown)


bryphe:
Awesome, thanks @CrossR !

The 1-index/0-index is confusing (coming from libvim). It's inconsistent in that the lines are 1-based but the columns are 0-based 😬 Sorry about that... Think eventually that needs to be fixed in `libvim` -> `reason-libvim`.

But the behavior is much better now with your change! Thank you for the investigation and fix!

bryphe:
I'll pull this in so we have it for release builds tomorrow -

Issue #465, opened 2019-07-24
Feature: Zen Mode


bryphe:
'Zen-mode' or 'distraction-free' mode is a way to temporarily remove extraneous UI, and just focus on the editor or document being modified.

This feature is adding a first, basic implementation to toggle 'zen-mode' on and off.

There would be a few pieces necessary to implement this:

__Part 1: Add a 'zen-mode' flag to our state__

We'd need to add a `zenMode` flag to our state, here: https://github.com/onivim/oni2/blob/29085a2f859f631937f182ffc3d81044c78c87d7/src/editor/Model/State.re#L10

This could simply be a `bool` for now - `true` if we are in zen mode, `false` otherwise.

__Part 2: Define Actions to toggle zen mode__

We'd need to add an Action to toggle zen mode, like `ToggleZenMode` in our Actions - this would be needed here: https://github.com/onivim/oni2/blob/29085a2f859f631937f182ffc3d81044c78c87d7/src/editor/Model/Actions.re#L11

__Part 3: Add a reducer to update state__

We'd need to add a reducer to update our `zenMode` flag, based on the `ToggleZenMode` action - this could happen in the root reducer, here: https://github.com/onivim/oni2/blob/29085a2f859f631937f182ffc3d81044c78c87d7/src/editor/Model/Reducer.re#L28

__Part 4: Add a command in the command palette for toggling zenMode__

As a simple initial entry point for this feature, we could add an entry in the command palette (accessible via `Control+Shift+p` or `Command+Shift+p` on Mac).

We can add an entry here: https://github.com/onivim/oni2/blob/917d70f0bc93b75ef6d4627de6566cde0304d91e/src/editor/Model/CommandPalette.re#L15 to associate it with the `ToggleZenMode` action.

__Part 5: Update the UI when zenMode is enabled__

In our Root UI component - https://github.com/onivim/oni2/blob/master/src/editor/UI/Root.re#L53 - we'd want to update the `statusBarVisible` flag to also take into configuration our `state.zenMode` flag.

Finally, we'd want to ignore all other window splits and only render the active split.

There are three components that come into play:
- The `<EditorView />` is the top-level view, used by the `<Root />` component
- The `<EditorLayoutView />` renders all the splits / windows (ie, sidebars as well)
- The `<EditorGroupView />` renders a single EditorGroup (a container of tabs + active editor).

[...]

PR #466, opened 2019-07-24
implement zen mode

tcoopman:
First stab at a #465

### DIFF
--- a/src/editor/UI/EditorView.re
+++ b/src/editor/UI/EditorView.re
@@ -30,5 +30,18 @@ let createElement = (~state: State.t, ~children as _, ()) =>
     let style =
       editorViewStyle(theme.colors.background, theme.colors.foreground);
 
-    (hooks, <View style> <EditorLayoutView state /> </View>);
+    if (state.zenMode) {
+      (
+        hooks,
+        <View style>
+          <EditorGroupView
+            state
+            windowId=0
+            editorGroupId={state.editorGroups.activeId}
+          />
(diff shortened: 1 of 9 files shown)


src/editor/UI/EditorView.re
```diff
@@ -30,5 +30,18 @@ let createElement = (~state: State.t, ~children as _, ()) =>
...
       editorViewStyle(theme.colors.background, theme.colors.foreground);
 
-    (hooks, <View style> <EditorLayoutView state /> </View>);
+    if (state.zenMode) {
+      (
+        hooks,
+        <View style>
+          <EditorGroupView
+            state
+            windowId=0
```
tcoopman: Not sure if this is correct?

bryphe:
Looks awesome @tcoopman ! Thank you!

Tested it out, and it worked great:
toggle-zen-mode

src/editor/UI/EditorView.re
bryphe: Ah, we should be setting it to the `state.windowManager.activeWindowId` in this case.

It's unlikely to cause problems at the moment, since it is only used for the mouse down handler (to decide which window to focus) - but it'd be good to have the right ID hooked up to prevent issues in the future.

tcoopman:
I've also implemented to hide the tabs by default

bryphe:
Sweet! Looks great, @tcoopman - thank you for the contribution!

PR #486, opened 2019-07-27
[Windows] File Associations

### LINKED ISSUE #468: Windows - Setup: Add file association registry keys
__Issue:__ When right-clicking on a file in Windows, after having set up Onivim 2 via the installer, there is no option for Onivim 2 to open a file.

__Defect:__ We're not setting the registry keys in the installer.

This was implemented for Onivim 1 in https://github.com/onivim/oni/pull/613 and https://github.com/onivim/oni/pull/1769 - we need to port over that logic for our Onivim 2 installer.

CrossR:
Fix #468.

This adds the file associations.

There was also a few other issues:

 - Reg keys were pointing at "Onivim2.exe" rather than "Oni2.exe" which meant it didn't work even with the associations.
 - Was pointing to the old app icon location (I couldn't see a real .ico file anywhere, so you may need to add one yourself @bryphe, since currently there is no App Icon).
 - The setup in Oni wasn't setup to deal with arrays of extensions very well, so I've swapped that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,16 @@
         ]
       },
       {
+        "name": "Javascript source",
+        "role": "Editor",
+        "ext": [
+          ".js",
+          ".es6",
+          ".mjs",
+          ".pac"
+        ]
+      },
+      {
         "name": "JSON",
         "role": "Editor",
         "ext": [
--- a/package.json
+++ b/package.json
@@ -21,16 +21,16 @@
         "name": "C++ source",
         "role": "Editor",
         "ext": [
-          ".cpp",
-          ".cc",
-          ".cxx",
-          ".hpp",
-          ".hh",
-          ".hxx",
-          ".h",
-          ".ino",
-          ".inl",
-          ".ipp"
+          "cpp",
+          "cc",
+          "cxx",
+          "hpp",
+          "hh",
+          "hxx",
+          "h",
+          "ino",
+          "inl",
+          "ipp"
         ]
       },
       {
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -24,9 +176,8 @@
   "dependencies": {
     "ocaml": "~4.7.0",
     "@esy-ocaml/reason": "3.4.0",
-    "isolinear": "*",
```
CrossR: A few lines down we have an explicit version specified so I removed this duplicate. (If it should be the specific one that is removed and this one kept, just shout up).

bryphe:
This is awesome, thanks @CrossR !

> Was pointing to the old app icon location (I couldn't see a real .ico file anywhere, so you may need to add one yourself @bryphe, since currently there is no App Icon).

Nice catch... I'll create one shortly and bundle it 👍

bryphe:
Weird output for the 'verify windows release build' in the latest job:
```
Mode                LastWriteTime         Length Name                                                                  
----                -------------         ------ ----                                                                  
-a----        7/28/2019   4:58 PM       38570199 Onivim2-deb40a4-x64.exe                                               
-a----        7/28/2019   4:58 PM       39338393 Onivim2-deb40a4.zip                                                   
** Validating .zip package **
Expand-Archive : The path 'D:\a\1\a/Release_Windows/Onivim2-b0cb229.zip' either does not exist or is not a valid file 
system path.
```

It seems like it's looking for `Onivim2-b0cb229.zip`, but it actually should be `Onivim2-deb40a4.zip` (matching the commit id)

PR #494, opened 2019-07-29
Packaging - Windows: Add ICO file

bryphe:
This adds a `.ico` file following up from @CrossR 's work in #486


scripts/windows/BuildSetupTemplate.js
```diff
@@ -36,6 +36,7 @@ const valuesToReplace = {
     AppExecutableName: executableName,
     AppSetupExecutableName: `${prodName}-${version}-${buildFolderPrefix}win`,
     Version: version,
+    SetupIconFile: path.join(__dirname, "..", "..", "assets", "images", "oni2.ico"),
```
CrossR: The "Icon"/"DefaultIcon" bits below will also need updating with this new path, since I just put a temp `{app}\\oni2.ico` one in.

bryphe: Thanks @CrossR for reviewing!

I _think_ it should be OK - I added a line to copy the `oni2.ico` to the root application directory (the change above in `scripts/release.js` - so that `{app}\\oni2.cio` should be valid.

Seems like it works (although we'll need a better icon at some point!)
image

PR #512, opened 2019-07-31
Docs: Tweak installation steps

bryphe:
Due to issues with #501 , and the notarization, we'll be using the archive `Onivim2.tar.gz` on Mac (and later - pivot to a `zip`)

### DIFF
--- a/docs/docs/getting-started/installation.md
+++ b/docs/docs/getting-started/installation.md
@@ -8,9 +8,10 @@ While we're developing Onivim 2, the latest builds are available at our [early a
 
 ## OSX (10.13+)
 
-1. Download `Onivim2.dmg`.
-2. Double-click on the downloaded DMG to expand.
+1. Download `Onivim2` for macOS.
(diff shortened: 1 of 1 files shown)


lpalmes:
Hey @bryphe, i've just got an early version build of oni2, congrats on the early builds! Looking good!

I've downloaded the early build, and moved the Onivim2.app to my Applications folder, what i'm looking to do is to be able to launch it from the terminal, I've did this but it doesn't seem to work.
```
ln -s /Applications/Onivim2.App/Contents/MacOS/Oni2 /usr/local/bin/oni2
```

Do you know how that could be done? If you do it would be nice to add it to the installation steps as an optional step.

Aside from that congrats! Looking great!

bryphe:
Hey @lpalmes - nice to see you here! Thanks for taking it for a test run!

Ah ya, we need an easy way to launch the app - there's a tracking issue with some ideas here: https://github.com/onivim/oni2/issues/518

Issue #514, opened 2019-07-31
Packaging - Ubuntu 18.04: Crash on startup


bryphe:
Have reports from Discord & Twitter that the packaged app is crashing.

So far - seems exclusive to Ubuntu. Trying to track down:

- Which Ubuntu versions are affected?
- Are both the `AppImage` and `tar.gz` affected?

bryphe:
One report is from 18.04 (on twitter: https://twitter.com/koutselakis_m/status/1156645449622835200)

thangngoc89:
```
➜ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.2 LTS
Release:        18.04
Codename:       bionic
```

both `AppImage` and `tar.gz`

Note: I can run and build from source for master branch

bryphe:
Thanks @thangngoc89 ! So far, seems like 18.04 is hitting it. Will see if I can get a Ubuntu 18.04 environment set up to test. That's interesting that build-from-source works but packaging fails.

CrossR:
I tested on Ubuntu 18.10 and everything worked perfectly. I can try to get a 18.04 build setup as well later if useful.

bryphe:
@CrossR - that would be awesome!

I wonder if it is related to a certain lib not being installed - I was a bit concerned with the GTK packaging. Would be helpful to run the binary directly and see if there is any logging (ie, via the ONI2_DEBUG=1 environment variable), or if its crashing before the dylibs are loaded.

bryphe:
Reported on 16.04 as well -

koutselakismanos:
Maybe because of this?
![Screenshot from 2019-07-31 23-18-11](https://user-images.githubusercontent.com/23614722/62245128-8e707f80-b3e9-11e9-9add-72fe3c217bba.png)

vrotaru:
I have the same issue on Linux Mint, crash on startup. A windows appear for a short moment then nothing.

And the is the following in `.xsession-erorrs`:

`Cinnamon warning: Buggy client sent a _NET_ACTIVE_WINDOW message with a timestamp of 0 for 0x7a00007 (Oni2)`

CrossR:
So the issue here is the following: 

```
[...]

PR #521, opened 2019-07-31
Packaging - Linux: Bundle harfbuzz / freetype symbols

bryphe:
This is an attempt to workaround the `harfbuzz` issue in #514 - where Oni2 is trying to load a newer version of the library, but the symbol is not available. This fixes it by including a harfbuzz locally with the version we need (it overrides the exclude list used by `linuxdeploy`).


bryphe:
Unfortunately, doesn't seem like this helps with #514 - we get a version mismatch in pango instead: `/tmp/.mount_OnivimPkvcwA/usr/bin/Oni2: symbol lookup error: /usr/lib/x86_64-linux-gnu/libpangoft2-1.0.so.0: undefined symbol: hb_font_funcs_set_variation_glyph_func`

PR #527, opened 2019-08-01
#Fix 514 - Remove usage of harfbuzz 1.7.7+ API

bryphe:
This picks up a change in `reason-fontkit` (https://github.com/revery-ui/reason-fontkit/pull/27) to remove the use of `hb_blob_create_from_file` for better compatibility with system-installed harfbuzz.

In #514  , it seems that issues are at least partially stemming from our use of that API, which isn't available in the 1.7.2-ish version we saw.

It sounds like based on @CrossR 's trial that this fixed the issue (although, there was potential still an error with copying configuration)

Based on conversation in Discord - sounds like this fixes it!


awilkins:
Is there a way to clear the build cache? Have pulled tip of master to try this out (should build on 18.04 now, right?) but still getting the same error, it's coming up so fast it's likely from cached artifacts.

bryphe:
Hmm, if you're building from source - there may be a different error. The build cache should get invalidated when it sees the new dependency for `reason-fontkit` (you can force it by deleting the `_esy` folder at the root of the project, though)!

Would you mind sharing out the error message you are seeing?

awilkins:
Trashed my entire `~/.esy` folder - must have had the desired effect because things taking much longer... to get to the next error :-)

```$ esy build
info esy build 0.5.8 (using package.json)
File "assets/configuration/dune", line 4, characters 8-18:
4 |  (files setup.json configuration.json keybindings.json))
            ^^^^^^^^^^
Error: No rule found for assets/configuration/setup.json
error: command failed: 'refmterr' 'dune' 'build' '-p' 'Oni2' '-j4' (exited with 1)
error: build failed with exit code: 1
  
```

bryphe:
Ah, that's promising! You'll need to run `esy bootstrap` now to fix that error 👍

awilkins:
Huzzah! Built.

```
esy run
```

... as soon as I open a file with syntax highlighting it dies though.

```
$ esy test
error: project is not installed, run `esy install`
  
awilkins@ryoko:~/oni/oni2 (master=)
$ esy install 
info install 0.5.8 (using package.json)
awilkins@ryoko:~/oni/oni2 (master=)
$ esy test
error: project is not installed, run `esy install`
  
```

That seems a bit odd too.

bryphe:
Ah, you'll also need to build the textmate service: https://onivim.github.io/docs/for-developers/building#build-the-textmate-service

(That same doc also has instructions for running tests / benchmarks)

awilkins:
Amazing what you can do when you actually read the docs... can confirm that all builds and runs nicely on Ubuntu 18.04 for me now.

bryphe:
@awilkins - awesome! Glad its working for you 👍

Issue #356, opened 2019-06-25
MacOS: configuration directory isn't automatically created


vincentjorgensen:
commit: 63913658814299af9a1ab6ef511e491e43e5a3c3
OS: MacOS

When starting with `esy run` for the first time, the config directory isn't automatically generated, throwing the following error

```
Error loading configuration file: Failed to copy from /Volumes/Projects/oni2/src/assets/configuration/configuration.json to /Users/vincent/.config/oni2/configuration.json, encountered No such file or directory, whilst calling open with /Volumes/Projects/oni2/src/assets/configuration/configuration.json
```

CrossR:
The configuration directory shouldn't need making at all, its part of the repo (`oni2/assets/configuration`).

Have you run `esy bootstrap`? If you have, its possible the recent changes to the CLI startup have caused an issue where we are looking in the wrong location for that file, since its looking in `oni2/src/assets/` not `oni2/assets/`.

vincentjorgensen:
Per instructions, I ran:
```
git clone https://github.com/onivim/oni2
cd oni2
esy install
esy bootstrap
esy build
cd src/textmate_service
node install.js
npm run build
cd ..
esy run
```

When I created the directory on the command line, the error went away.

CrossR:
Ah!

On the initial run, we copy the bundled config by using `GetExecutingDirectory()` joined with `configuration.json`. However, since you were running from `src`, it looks in there rather than the root of the repo.

We should probably update the build instructions to say to do `npm run build; cd ../..`, and then also actually fix the code to always point to the repo root.

Running `esy run` from `oni2/` works as expected and does copy the config for me, so we just need to make the finding of the root more robust. Perhaps with a `GetGitBaseDirectory()`, not the current directory.

CrossR:
Thinking about this again, what this actually needs is the ability to go from the `default` in the `ConfigurationValues.re` to JSON. That JSON can then be written out using YoJSON after being prettied.

That lets us skip the issue of pulling in configuration from a relative directory, as well as stopping the current issue of a duplication of configuration values (which causes issues like #402 to be more confusing to fix).

The snag with that is we need to translate the property names in `default` to the actual JSON setting name, ie `editorTabSize` -> `editor.tabSize`. Not sure if we could achieve that with the current config parser, so may need an extra one just for the converting of a property name to a JSON field.

tristil:
I'm planning to take a crack at the suggested solution this weekend FYI

PR #546, opened 2019-08-01
Config Generation

### LINKED ISSUE #524: Oni2 cannot copy default config to user folder
When I launch Oni2.exe under windows with -f, it gives me this log:
image
Once I copy configuration.json manually from Oni2 folder to AppData/Local/Oni2, it no longer complains.

### LINKED ISSUE #525: The default config has a comma at the end
configuration.json in Oni2 installation directory looks like this:
```json
{
  "editor.minimap.enabled": true,
  "editor.insertSpaces": false,
  "editor.indentSize": 4,
  "editor.tabSize": 4,
}
```
And I think that comma before closing brace makes it fail to parse it.

### LINKED ISSUE #535: "Open configuration file" does not work/respond
I am working on Win64, with onivim2 Ver.  2.0.2.0
The problem:
While "CTRL+SHIFT+P" does bring up the Preferences option, selecting
Preferences: Open Configuration file
and then hitting either the "Right" key or the "Enter" key leads to no response (the same happens when selecting "Open keybindings file")

Possibly related:
typing
:set
or
:f 
leads to no visible response.

CrossR:
Fixes #524, Fixes #525, Fixes #535.

#356 is a better approach, but is going to take some more time (mainly around the converting from ReasonML names to the JSON names we want ie `workbenchEditorShowTabs` -> `workbench.editor.showTab`).

Generate the config from a string.

To do:

 - [x] Test case to compare the parsed default config against the default configuration object.
 - [x] Tidy up the code (Move the strings to their own file I think would be a bit cleaner).
 - [x] Error handling around where we write the file out.


src/editor/Core/ConfigurationDefaults.re
```diff
@@ -1,3 +1,45 @@
...
+  "editor.renderIndentGuides": true,
+  "editor.highlightActiveIndentGuide": true,
+  "editor.renderWhitespace": "all",
+  "workbench.activityBar.visible": true,
+  "workbench.editor.showTabs": true,
+  "workbench.sideBar.visible": true,
+  "workbench.statusBar.visible": true,
+  "workbench.iconTheme": "vs-seti",
+  "files.exclude": ["_esy", "node_modules"],
+  "editor.zenMode.hideTabs": true
```
CrossR: For now, I've extended this to the full config, mainly to make the test comparison much easier.
It does also make discoverability a little easier whilst the config story is pretty basic.

I can change it to a smaller one if needed though?

src/editor/Core/ConfigurationValues.re
```diff
@@ -57,10 +57,3 @@ let default = {
   filesExclude: ["node_modules", "_esy"],
   zenModeHideTabs: true,
 };
-
-let getBundledConfigPath = () => {
```
CrossR: This wasn't being used anyways.

test/editor/Extensions/TextmateTokenizerTests.re
```diff
@@ -221,7 +221,6 @@ describe("Textmate Service", ({test, _}) => {
 
         switch (tokenizeResult, colorMap^) {
         | (Some(v), Some(cm)) =>
-          Console.error(v.colors);
```
CrossR: Removed this whilst I was running the tests since it was printing whilst the test run. Can undo if its needed to get extra information out during CI?

bryphe: Oh, no, accidentally left this in - thanks for removing it!

src/editor/Core/ConfigurationValues.re
bryphe: Awesome! Thanks for removing it - one less thing to think about 😄

src/editor/Core/ConfigurationDefaults.re
bryphe: This sounds good! We can tweak it down the road, and as you mentioned, it helps discoverability for now.

assets/configuration/dune
```diff
@@ -1,4 +1,4 @@
 (install
  (section bin)
  (package Oni2)
- (files setup.json configuration.json keybindings.json))
```
bryphe: Thanks for handling this for keybindings.json as well - it's great to not have to worry about these files

src/editor/Core/Filesystem.re
```diff
@@ -255,13 +255,21 @@ let copy = (source, dest) =>
   );
 
 let createOniConfiguration = (~configDir, ~file) => {
-  open Utility;
+  let userConfigPath = Utility.join([configDir, file]);
```
bryphe: I can't remember - do we create the `configDir` somewhere? Or do we need to check if that folder exists, and create it here?

PR #562, opened 2019-08-02
fix(defaults): remove duplicate default settings and sort them

benox3:


CrossR:
Thanks for de-duplicating this and fixing it up!

It looks like you are getting a CI failure at the mo since it is failing the parse the config. I think that's due to the final comma in the config causing issues.

src/editor/Core/ConfigurationDefaults.re
```diff
@@ -10,30 +10,26 @@ let getDefaultConfigString = configName =>
...
-  "editor.highlightActiveIndentGuide": true,
   "editor.renderWhitespace": "all",
+  "editor.tabSize": 4,
+  "editor.zenMode.hideTabs": true,
+  "files.exclude": ["_esy", "node_modules"],
   "workbench.activityBar.visible": true,
   "workbench.editor.showTabs": true,
+  "workbench.iconTheme": "vs-seti",
   "workbench.sideBar.visible": true,
   "workbench.statusBar.visible": true,
```
CrossR: The parser we are using doesn't like/we've not set it up to like a trailing comma, so I'm assuming this is what is causing the CI failure.

benox3:
> Thanks for de-duplicating this and fixing it up!
> 
> It looks like you are getting a CI failure at the mo since it is failing the parse the config. I think that's due to the final comma in the config causing issues.

Thanks for the review! Should be fixed now.

bryphe:
Awesome, thank you @benox3 for the contribution (and thanks @CrossR for reviewing!)

Issue #540, opened 2019-08-01
cmd + p crashing app


zacharymacke:
On a fresh install first time open, cmd + p bringing up the file explorer crashes the application. Sometimes it will stay open fine but crashes upon typing first letter. OSX 10.14.5

nolondil:
As I have experienced it too: it seems it tries to list too many items or one item is forbidden for Oni2. When I launch Oni2 directly in a directory (through the terminal) I don't run into any issue with cmd + p.

zacharymacke:
I will try and run through terminal and see if that fixes it!

nolondil:
(JIC) I give the path of the project folder as an argument to the oni2 command.

CrossR:
>As I have experienced it too: it seems it tries to list too many items or one item is forbidden for Oni2. When I launch Oni2 directly in a directory (through the terminal) I don't run into any issue with cmd + p.

I expect this is the case, since MacOS is pretty protective over certain folders, so it wouldn't suprirse me if we run ripgrep over some places we shouldn't and hit some issues.

anisjonischkeit:
Looks like the same issue as #467

bryphe:
This should be fixed now on OSX with #467

jpinnix:
I am still experiencing this with the last several nightly builds.

bryphe:
Thanks for the heads up, @jpinnix - will take a look.

bryphe:
This should be fixed now (no longer crashes in large directories). 

The main fix for this was in #693 - which time-slices our processing of the results, so that we can stay on top of processing output from Ripgrep.

There are still improvements we can make to improve the speed of processing (important for large directories) - but the crash should no longer reproduce. I'll close this, but please let me know if you're still seeing crashes with Ripgrep on latest builds 👍

raguay:
01/21/2020

macOS 10.15.2
Oni2 just built on 01/21/2020

## Actions taken

1) <cmd>+P, while the files are listing type the name of one of them and press enter.

## What should happen

It should open a buffer in that file

## What happened

Total crash of Oni2 with a crash log of:
```
Fontkit.FontKitRenderGlyphException("[ERROR]: Unable to render character at FT_Load_Char\n"):
```

Not sure if it’s related to this issue, but this one seemed to be the closest match.

bryphe:
I believe this particular issue should be fixed now - this code-path: 
```
[...]

PR #563, opened 2019-08-02
Fix #467 / #540 - Command+P crashes in large directory

### LINKED ISSUE #467: OSX - Command+P crashes in default working directory of installed app
This issue is specific to our packaged builds - repro:

- Install build from DMG
- Open it via Applications launcher
- Open quickopen via Command Palette (<Cmd+P>)

EXPECTED: Fuzzy finder should open and run
ACTUAL: Crash

It seems there are really two issues here:
1) Why is the fuzzy finder crashing at all?
2) We should have a better default directory when opening for the first time (at least, the user directory)

bryphe:
This is the start of a fix for #467 / #540 - the issue is that we are doing too much work in the `ripgrep` thread, allowing the output buffer to get full and eventually crash.

This initial fix 'offloads' the work to the main thread (via a new Revery API: https://github.com/revery-ui/revery/pull/530) - but it comes with its own issues, as now it blocks input. These fixes are relevant not just for the `/` directory but for any large directory.

There's a few additional pieces we need:
- [x] Experiment with offloading the conversion of `Bytes.t` -> `string list` to a thread, prior to sending it to the main thread.
- [x] Instead of trying to filter the _entire_ set of input (which is a lot in the case of #467 / #540 !), we should amortize the filtering across multiple frames using a `Job.t`
- [x] ~~We should only start filtering when a character has been pressed, to assist by minimizing results, to an extent~~ I took this out since we don't have anything to show for now - we could add some helper text like 'Start typing to filter...' or show recent files.
- [x] Fix bugs in `rench` around high-frequency process closing / opening
- [x] Fix up bug with 'selected' behavior - the `selectedItem` is not always in sync with the filtered UI, and sometimes `List.nth` can crash due to this.
- [x] Regression test case so we know that this works build-over-build
      - Integration test case that runs QuickOpen on a large directory, and waits for it to complete
      - Integration test case that runs QuickOpen on a large directory, and changes query with high frequency (there was a bug in `rench` that was causing us to leak file descriptors in this case)


bryphe:
There's still a lot of room for improvement in the QuickOpen experience, but at least this unblocks the crash, now -

PR #569, opened 2019-08-02
Fix #513 - Cursor doesn't keep up with fast typing

### LINKED ISSUE #513: Bug: In the command palette, the cursor can't keep up with fast typing
Hi guys! I just downloaded Onivim2 from the early access portal and wanted to report a bug I noticed. In the command palette, if you type too fast the cursor starts to trail the text.

**Operating System:** Windows

bryphe:
__Issue:__ #513  - when typing quickly in the QuickOpen, the cursor will sometimes lose position. Very annoying!

__Defect:__ The `<Input />` control in Revery updates the the cursor position and text at different frequencies - the text update happens immediately, but the cursor position gets updated based on render frequency (it occurs in the effect hook). 

__Fix:__ I started a fix in Revery in https://github.com/revery-ui/revery/pull/513 , but it was difficult to handle both the uncontrolled and controlled cases. For Onivim's use of the text input, we only need to handle the uncontrolled case (which is simpler), so I'm vendoring the text input control to implement the uncontrolled case, and handle both cursor position and text together in the reducer.


tcoopman:
If fixing both controlled and uncontrolled inputs is making them too complex, couldn't you explicitly split them in 2 components?

bryphe:
@tcoopman - I'd be OK with splitting them in two components!

Most of the complexity is in the controlled component (the uncontrolled, which we use in Onivim 2, is the simpler strategy code-wise). The problem with the controlled component is there is some degree of synchronization logic needed.

Since we don't need the controlled component for Onivim 2 - I'm going to hold off on that piece of the implementation so I can focus on some of the other main blockers that have come up.

PR #572, opened 2019-08-02
Fix #506 - Part 1 - Insert-mode paste

### LINKED ISSUE #506: Clipboard Integration
We should have a sane, simple, out-of-box interaction with clipboard. Right now, there is no clipboard interaction at all, which is pretty broken.

My proposal would be the following (similar to what we had in Onivim 1):

- __Insert Mode__ + `Control+V`/`Command+V`- paste from clipboard as if typed
- ~~__Normal Mode__ + `Control+V`/`Command+V` - paste from clipboard as if entering insert mode and running `Control+V`, and return to normal mode with cursor at the end of the text~~ 
-~~ __Visual Mode__ + `Control+C`/`Command+C` copy visual selection to clipboard, and exit visual mode.~~

I think this would be a reasonable start to unblock simple cases.

We should also wire up the `"+` register to pull from the system clipboard by setting up a callback in `libvim`.

bryphe:
Part 1 of #506 

This implements the insert-mode paste gesture:
- Add a bindable command: `editor.action.clipboardPasteAction`
- Implement the command in `VimStoreConnector`
- Plumb through the glfwGetClipboardString API from `Oni2_editor` -> `StoreThread` -> `VimStoreConnector`

__TODO:__
- [x] Test case
- [x] Add 'insertMode' condition


src/editor/Core/ConfigurationDefaults.re
```diff
@@ -43,6 +43,8 @@ let getDefaultConfigString = configName =>
 {
     "bindings": [
         { "key": "<C-P>", "command": "quickOpen.open", "when": [["editorTextFocus"]] },
+        { "key": "<C-V>", "command": "editor.action.clipboardPasteAction", "when": [["editorTextFocus"]] },
```
CrossR: Does this need an additional clause to prevent the breaking of normal mode visual block mode?

bryphe: Good catch @CrossR - indeed we should have a `when` condition for this.

bryphe: I put together an `insertMode` condition, but really we need more fidelity (something to match the VSCode expression language - ands/ors, string conditions, boolean conditions)

Issue #460, opened 2019-07-24
Configuration: Add 'editor.rulers' configuration option


bryphe:
We should add support for the `editor.rulers` configuration option, similar to what VSCode provides.

This configuration option takes an array of numbers, like:
```
"editor.rulers": [80, 120]
```

And draws a vertical rule on the editor surface.

There'd be a few things we'd need to add:

__Part 1 - Add a Configuration Setting__

We'd need to add the configuration definition to [Core/ConfigurationValues.re](https://github.com/onivim/oni2/blob/29085a2f859f631937f182ffc3d81044c78c87d7/src/editor/Core/ConfigurationValues.re#L14) and a default value to the same file [here](https://github.com/onivim/oni2/blob/29085a2f859f631937f182ffc3d81044c78c87d7/src/editor/Core/ConfigurationValues.re#L37).

We'd also need to add a `ConfigurationParser` that knows how to handle parsing an array of numbers in [Core/ConfigurationParser.re](https://github.com/onivim/oni2/blob/master/src/editor/Core/ConfigurationParser.re) - it'd be similar to the `parseStringList` configuration parser in that file.

__Part 2 - Pick up configuration setting and render__

For the actual rendering of the editor rulers, we'd want to draw a 1-pixel wide rectangle on the editor surface, at the proper column. We'll need to take into account the line number and the font size.

We can read from the configuration and get the `editorRulers` like we do for other configuration values, like line numbers: https://github.com/onivim/oni2/blob/29085a2f859f631937f182ffc3d81044c78c87d7/src/editor/UI/EditorSurface.re#L206

This rendering could take place in [`UI/EditorSurface.re`](https://github.com/onivim/oni2/blob/29085a2f859f631937f182ffc3d81044c78c87d7/src/editor/UI/EditorSurface.re#L498) right around where we render the background for the cursor line (we'd probably want to render these rulers _after_ the cursor line, though).

For each `editor.ruler`, we could call the [`bufferPositionToPixel`](https://github.com/onivim/oni2/blob/29085a2f859f631937f182ffc3d81044c78c87d7/src/editor/UI/EditorSurface.re#L251) helper method for each value. The `line` value we could ignore and pass in something like 0 or  1 - we care about the `char` value, which would correspond to a value in the editorRulers array. This would give us the _pixel position_ of the editor ruler, taking into account line numbers, scroll, and the font width.

[...]

PR #573, opened 2019-08-03
Editor Rulers (Part 1 - Add a Configuration Setting)

KevinSjoberg:
This is my first attempt at tackling part one of #460. I'm totally new to Reason and the project in general so any feedback is much appreciated.

A few questions that arose as I implemented this:

- Is the preferred way to run the tests using `npm test`?

- When I finished the implementation I ran `npm run format`. It threw an error (containing a diff) on me. Rerunning the command did not throw and error and terminated successfully. Is that expected?

- I tried to add a test case for parsing a list of integers. I noticed that the parsed values come back flipped. I assume it's due to the fact that we do ` [v, ...accum]` instead of ` [...accum, v]`. I didn't want to change this behaviour as I had a feeling it might be due to performance reasons. Is that the case?

### DIFF
--- a/src/editor/Core/ConfigurationDefaults.re
+++ b/src/editor/Core/ConfigurationDefaults.re
@@ -27,6 +27,7 @@ let getDefaultConfigString = configName =>
   "editor.renderIndentGuides": true,
   "editor.highlightActiveIndentGuide": true,
   "editor.renderWhitespace": "all",
+  "editor.rulers": [],
   "workbench.activityBar.visible": true,
   "workbench.editor.showTabs": true,
   "workbench.sideBar.visible": true,
--- a/src/editor/Core/ConfigurationParser.re
+++ b/src/editor/Core/ConfigurationParser.re
@@ -34,6 +34,22 @@ let parseStringList = json => {
   };
 };
 
+let parseIntList = json => {
+  switch (json) {
+  | `List(items) =>
+    List.fold_left(
+      (accum, item) =>
+        switch (item) {
+        | `Int(v) => [v, ...accum]
+        | _ => accum
+        },
+      [],
+      items,
+    )
+  | _ => []
+  };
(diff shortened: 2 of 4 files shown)


bryphe:
Oh, and missed your questions in the comments:

> Is the preferred way to run the tests using npm test?

For most of the commands, we actually use `esy` instead of `npm` (and we use a feature called [sandboxes](https://esy.sh/docs/en/multiple-sandboxes.html) that let us namespace commands).

For running unit tests, you'd need to do:
- `esy '@test' install`
- `esy '@test' run`

And integration tests are similar:
- `esy '@integrationtest' install`
- `esy '@integrationtest' run`

> When I finished the implementation I ran npm run format. It threw an error (containing a diff) on me. Rerunning the command did not throw and error and terminated successfully. Is that expected?

Ah, you might want to try `esy format` instead (the `esy` command provides some environment context that is needed for the build tools)

> I tried to add a test case for parsing a list of integers. I noticed that the parsed values come back flipped. I assume it's due to the fact that we do [v, ...accum] instead of [...accum, v]. I didn't want to change this behaviour as I had a feeling it might be due to performance reasons. Is that the case?

Nice observation - yes, this is actually due to performance reasons. The `list` type in OCaml is a singly-linked list, so insertions at the head of the list are cheap (just take the new node, and point it to the remainder of the list), but immutable insertions at the tail are expensive (you have to create a new list, copy it, and tack the item on at the end).

To preserve order, the canonical approach I've seen is to use the `[v, ...accum]` method which inserts in reverse, but then, if the order is important, use `List.rev` to flip it back to the correct order. This is still an extra copy, but cheaper to do it once than for every node. 

I don't think it is necessary in this case, but it is important for others - like this bug, where typing events for the main editor could come out of order: https://github.com/onivim/oni2/pull/451

PR #574, opened 2019-08-03
Editor Rulers (Part 2 - Pick up configuration setting and render)

KevinSjoberg:
Given any rulers present in `editor.rulers` they will be rendered in the viewport. The color is currently hardcoded and will be fixed in part three of #460 

This pull request depends on the work done in #573, so be sure to merge that first.

### DIFF
--- a/src/editor/UI/EditorSurface.re
+++ b/src/editor/UI/EditorSurface.re
@@ -205,6 +205,9 @@ let createElement =
     let bufferId = Buffer.getId(buffer);
     let lineCount = Buffer.getNumberOfLines(buffer);
 
+    let rulers =
+      Configuration.getValue(c => c.editorRulers, state.configuration);
+
     let showLineNumbers =
       Configuration.getValue(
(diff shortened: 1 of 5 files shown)


KevinSjoberg:
Looking through the codebase I can see that other features such as rendering indent guides are using `activeBuffer`. Is that something I should take into account as well?

src/editor/UI/EditorSurface.re
```diff
@@ -547,6 +550,20 @@ let createElement =
                 (),
               );
 
+              /* Draw configured rulers */
+              let renderRuler = ruler =>
+                Shapes.drawRect(
+                  ~transform,
+                  ~x=fst(bufferPositionToPixel(0, ruler)),
```
bryphe: Nice find on the `fst` method - I forget about this sometimes... 👍

KevinSjoberg: Thanks! 🙂

PR #576, opened 2019-08-03
Add documentation for `editor.rulers` option

KevinSjoberg:
Add documentation for the `editor.rulers` configuration option. Was not entirely sure how we want the type signature to be. I could see `list(int)`, `[int]` and `list of int` and in the end I settled on the latter.


bryphe:
Nice! I should include documentation in a PR checklist - thank you for thinking about this and adding it!

PR #578, opened 2019-08-03
Launch app in user directory on Linux and MacOS

praveenperera:
Problem: 
- When you open the app via the app icon, `Sys.getcwd`, which is used to find the `workingDirectory` returns `"/"` on MacOS.
   - I am not sure how this works on linux, but I am assuming its the same.

Possible Solution:
- Instead of defaulting to workingDirectory, this change looks for the `$HOME` env variable which is usually set on MacOS and Linux. It uses that as the default directory
- It only uses this directory if a file or directory was not passed in as an argument. 
- If the `$HOME` env var is not found, it will default to the working directory 

This is first crack at this codebase so feedback is definitely welcome.

### DIFF
--- a/docs/website/core/Footer.js
+++ b/docs/website/core/Footer.js
@@ -52,10 +52,10 @@ class Footer extends React.Component {
           </div>
           <div>
             <h5>Community</h5>
-            <a href="https://discordapp.com/">Reddit</a>
-            <a href="https://discordapp.com/">Discord</a>
+            <a href="https://reddit.com/r/onivim">Reddit</a>
+            <a href="https://discord.gg/7maEAxV">Discord</a>
             <a
-              href="https://twitter.com/"
+              href="https://twitter.com/oni_vim"
               target="_blank"
               rel="noreferrer noopener">
               Twitter


CrossR:
Code changes look good!

I think we just need to work out what we want the default behaviour to be.

This change would mean that me doing `oni2` in `~/git/oni2/` would now open `~/` instead, which isn't necessarily a bad thing, since then I'd need to be explicit and do `oni2 .`. Is that the behaviour we want? The other option here would be to check that if the `cwd` is the install location, and default to home in just that occasion, but to use the `cwd` in any other case.

(This isn't 100% directed at you @praveenperera! Just thinking out loud, and if anyone has any opinions it could be useful!)

Eventually, I think we should aim to replicate the code behaviour and have a state of no folders open, and prompting the user to pick between some recents and an opening a new folder.

praveenperera:
> This change would mean that me doing `oni2` in `~/git/oni2/` would now open `~/` instead, which isn't necessarily a bad thing, since then I'd need to be explicit and do `oni2 .`. Is that the behaviour we want? The other option here would be to check that if the `cwd` is the install location, and default to home in just that occasion, but to use the `cwd` in any other case.

Thanks for the feedback, I had the same thought. In fact that's what the first version of my code did. But I settled with this because the code was a bit simpler and I think there so some precedence there. But ya someone will have to decide what the default behaviour should be. 

Here is the other version of the code I had initially

```reason
  let folder =
    switch (directories, filesToOpen, workingDirectory) {
    | ([hd, ..._], _, _) => hd
    | ([], [hd, ..._], _) => Rench.Path.dirname(hd)
    | ([], [], "/") => homeOrWorkingDirectory
    | ([], [], workingDirectory) => workingDirectory
    };

```

PR #580, opened 2019-08-04
Fix root path explicit open on MacOS

### LINKED ISSUE #579: Can't open with root directory on MacOS
Can't explicitly open on root path, reproduction: 

`/Applications/Onivim2.App/Contents/MacOS/Oni2 /`

From source with debug enabled

```bash
ONI2_DEBUG=1 esy run -f "/"
[ERROR] Exception Invalid_argument("\"\": invalid path"):
Raised at file "stdlib.ml", line 34, characters 20-45
Called from file "src/Path.re", line 4, characters 2-12
Called from file "src/editor/Core/Cli.re", line 97, characters 26-48
Called from file "src/editor/bin_editor/Oni2_editor.re", line 19, characters 17-33
```

praveenperera:
**ONLY Tested on MacOS**

Closes #579

### DIFF
--- a/src/editor/Core/Cli.re
+++ b/src/editor/Core/Cli.re
@@ -44,7 +44,7 @@ let parse = () => {
 
   let stripTrailingPathCharacter = s => {
     let len = String.length(s);
-    if (len >= 1 && s.[len - 1] == '/') {
+    if (len > 1 && s.[len - 1] == '/') {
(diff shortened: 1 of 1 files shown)


bryphe:
Thank you @praveenperera ! 

Nice catch on the bug, and the fix looks great! 💯 Much appreciated -

PR #588, opened 2019-08-04
Add app icon to Windows explorer.

CrossR:
It looks like we needed to specify the icon directly in the `Icons` section (which isn't actually where icons live, its more to define shortcuts such as in the Start menu etc).

This fixes the app icon in the start menu and similar, but the reason we need to do this versus Oni1 where we did not is because the `exe` does not have an icon associated with it, whereas it did then, assumedly as part of the `electron-builder` process.

I'm going to assume we can't edit the actual building of the `exe` since its from `esy`, so I would expect our only "fix" here is to use a CLI tool to update the icon as part of the publish process.


scripts/windows/setup.template.iss
```diff
@@ -12,7 +12,7 @@ AppName={{AppName}}
 AppVersion={{Version}}
 DefaultDirName={pf}\{{AppName}}
 DefaultGroupName={{AppName}}
-UninstallDisplayIcon={app}\{{AppExecutableName}}
+UninstallDisplayIcon={{AppIcon}}
```
CrossR: I think due to the same issue, this was pointing at `Oni2.exe` which has no icon, so the uninstaller would have no icon.

bryphe: Thanks for the details, @CrossR !

bryphe:
Thank you for investigating and fixing this @CrossR  - this will be a major improvement for our Windows experience!

I tested out the PR and I'm not seeing the icon get installed (or maybe I'm looking in the wrong places):

image

I tried the following:
- Cleaned and rebuilt (I also validated the fix is in the generated `setup.iss` in the `_release` fodler)
- Uninstalled previous versions
- Restarted explorer

Unfortunately still didn't see the icon, though. Is there something else I should try?

CrossR:
Ah! I use a custom start menu, and since it was fixed in there I assumed it was fixed in the default one: 
image

The control panel -> Uninstall option should also have an icon.

It looks like the default start menu is based on the Icon of the app itself, rather than the shortcut, so for that I think we are going to need to add a step that uses some form of resource editor to edit the `exe` to include the default icon.

There is a few different ones out there, including this one from the `electron` team : https://github.com/electron/node-rcedit. I think we can then use that to update the `Oni2.exe` to have a default icon, such that we can then just link against the executable, as in Oni1. Probably makes sense to just build that into the existing `release.js` script.

CrossR:
I've undone some of the changes I had before, and swapped to updating the exe directly to have an embedded icon. This has fixed it for me in the same places as before (Start menu, control panel) and also in the explorer if you navigate to the Oni2 install location.

The only one that hasn't updated is the start menu one....though if I "View File Location" its updated, and if I "Pin to start" the start tile is updated. I'll spin up a VM later today and try it out there, its possible its some thumbnail cache or something (though I did empty that...).

scripts/release.js
```diff
@@ -193,4 +204,11 @@ if (process.platform == "linux") {
   fs.copySync(extensionsSourceDirectory, extensionsDestDirectory, {deference: true});
   fs.copySync(textmateServiceSourceDirectory, textmateServiceDestDirectory, {deference: true});
   fs.removeSync(path.join(platformReleaseDirectory, "setup.json"));
+
+  // Now that we've copied set the app icon up correctly.
+  // rcedit is imported here as its a big import, only needed for windows.
+  // The image path here does not need to be in the release dir, as its just to do the embed.
+  const rcedit = require("rcedit");
```
bryphe: Sweet! Thanks for finding this and hooking it up to our release script, @CrossR 👍

bryphe:
Just tried it and it worked great! I had the same issue with the Start Menu, but I'm assuming it's like you mentioned - a thumbnail cache (I had a previous install).  Thanks for finding the `rcedit` tool - binary also looks better in the `_release` folder now too 😄

PR #589, opened 2019-08-04
Fix adding to PATH on Windows.

CrossR:
We added some code to the Inno setup file to remove the old Oni path on Windows, and that code was still around and removing the new and real Oni2 env var.

For context, Oni1 we added the default install location, say `C:\Program Files\Oni` and then later on, we split the CLIs up, which meant we had to remove `C:\Program Files\Oni` and add a new custom CLI location. To do that removal, we had an install step that would check and remove `C:\Program Files\Oni` (to clean up after ourselves) but now the `{app}` variable pointed to the Oni2 install, so it removed that instead 🤦‍♂.

The uninstall steps work fine, so I've left them, since we do want to remove when we uninstall.


bryphe:
> For context, Oni1 we added the default install location, say C:\Program Files\Oni and then later on, we split the CLIs up, which meant we had to remove C:\Program Files\Oni and add a new custom CLI location. 

Forgot about all this - thanks for the details! Nice find @CrossR & thanks for fixing this! 💯

PR #598, opened 2019-08-05
Fix #549 - Part 1: Allow configuration of fontSize

### LINKED ISSUE #549: Configuration: Set editor font family and size
We're missing customization options for setting the editor font-family and size. 

We should integrate with https://github.com/revery-ui/reason-font-manager to be able to resolve a font from a family name, and use that instead of requiring a full font file path.

This should be specified by JSON configuration options:
- `"editor.fontFamily"`
- `"editor.fontSize"`

bryphe:
Part 1 of a fix for #549 

This adds two configuration options:
- `editor.fontFamily`
- `editor.fontSize`

(This PR only hooks up the `editor.fontSize` for now, though)

And hooks up a new store connector, `FontStoreConnector`, that listens for configuration changes and loads the new fonts. The `editor.fontSize` option is mostly hooked up, but the `editor.fontFamily` piece still needs to be set up.

The main challenge with `editor.fontFamily` is that Revery doesn't really have a concept of font families... it just cares about the absolute path to some font file. This isn't very ergonomic though, especially if you just want to pick up a system font or installed font.

I'm working on a reason library for font discovery here: https://github.com/revery-ui/reason-font-manager - part of bringing in `editor.fontFamily` will be wiring up that discovery library and integrating it with Onivim 2. This would let us specify a font family via `editor.fontFamily`, resolve a path to an actual font file, and use that path internally.

__Deferred to part 2:__
- Hook up reason-font-manager
- Add documentation for settings
- Add health check for resolving system fonts
- Add integration test


remi-deltombe:
So happy to see this feature arriving!
Just one bug :  Font size is not applied correctly when the app boot.
Demonstration in this video : https://youtu.be/2UZWt40pPvQ
OS : Arch

bryphe:
> So happy to see this feature arriving!

Me too! Very basic & critical feature 😄 

> Just one bug : Font size is not applied correctly when the app boot.
> Demonstration in this video : https://youtu.be/2UZWt40pPvQ
> OS : Arch

Great catch @remi-deltombe ! Thank you for trying it out.

It turns out, we'd load the config, and then _revert_ the font back to size 14 - this would cause the measurements to be out-of-date (we'd _render_ the configured font, but use the metrics associated with the default font - which causes those artifacts).

Believe it should be fixed now.

PR #604, opened 2019-08-06
Copy Ruby extension from vscode

mhanberg:
There are a couple of test files here, I'm not sure if they should be removed or not.

Thanks!


bryphe:
Thanks for porting this over, @mhanberg ! Tried it with the included fixture and worked great.

> There are a couple of test files here, I'm not sure if they should be removed or not.

Longer-term - I plan on making a fork of VSCode just for building and packaging the extensions (some have compiled typescript or other code) - the tests will be stripped as part of that build process. But for now its fine 👍

Issue #439, opened 2019-07-15
Splits: Handle Rotation


szbergeron:
Example: ctrl-w r
Expected result: splits rotate
Actual result: nothing happens

Related actions: ctrl-w L, ctrl-w J, etc

bryphe:
We're still missing the rotation actions from the `libvim` layer (so that the Vim keybindings would trigger this behavior), but we could still potentially start on the logic here, by also implementing it as a command available in the command palette:

__Part 1: Add a `rotateForward` and `rotateBackward` method to `WindowTree`__

The first part of this work would be figuring out how to implement the rotation logic, given a target split id (`int`) and a current window tree [`WindowTree.t`](https://github.com/onivim/oni2/blob/master/src/editor/Model/WindowTree.re)

Could be called as follows:
```
// Find the parent split container containing split 1, and rotate it -> move the head to the tail
let newWindowTree = WindowTree.rotateBackward(1, windowTree)
```

The `WindowTree.t` type is defined here: https://github.com/onivim/oni2/blob/8037062c3edbd818b817ef0543f8bc7beb342a8b/src/editor/Model/WindowTree.re#L18

And we'd need to implement a recursive method that finds the `Parent(direction, children)` where `children` contains the target window split ID. When we find that ID, we'd implement the rotation logic on its children.

__Part 2:__ Add a command to the Command Palette and hook up to reducer

Next, once we have that logic, we'd need a way to invoke it. We could add a command to the `CommandPalette` like: https://github.com/onivim/oni2/blob/master/src/editor/Model/CommandPalette.re

- we could use commands like `Actions.Command("view.rotateForward")` and `Actions.Command("view.rotateBackward")`

With the commands and the logic implemented in Part 1, we could wire it up in the window reducer here: 
https://github.com/onivim/oni2/blob/8037062c3edbd818b817ef0543f8bc7beb342a8b/src/editor/Store/WindowsStoreConnector.re#L94

The `windowManager` state contains both the windowTree (`windowManager.windowTree`)and the currently selected split (`windowManager.activeWindowId`) - so we'd have all the context we need to pass to the functions we implemented in Part 1.

Once this is wired up, we should be able to open the command palette and rotate splits forward or back.

bryphe:
(scoping this issue to just the rotation aspect for now)

KevinSjoberg:
I’d be happy to take a stab at this. 🙂

bryphe:
[...]

PR #605, opened 2019-08-06
Splits (Rotation) - Part 1

KevinSjoberg:
This was a lot more challenging than I initially thought. I had to wrap my head around the `WindowTree` and the type `t` and the data structure it represents. Thinking about the problem from a functional perspective didn't help. 🙂

In #439, I said you can't rotate a split that is adjacent a Parent. While this is true, I could not come up with a good reason for why we shouldn't allow it. So the implementation allow it.

I've added two functions to the `WindowTree`:

- **rotateForward** - Given an id and the current tree, it will rotate all splits, confined within the same `Parent`, one step to the right, for the matching split. If no matching split is found, it returns the current tree.

- **rotateBackward** - Given an id and the current tree, it will rotate all splits, confined within the same `Parent`, one step to the left, for the matching split. If no matching split is found, it returns the current tree.

I can honestly say, that performance has not been a priority. Any suggestions regarding performance or the code in general is much appreciated. ❤️

This solves Part 1 of #439.

### DIFF
--- a/src/editor/Model/WindowTree.re
+++ b/src/editor/Model/WindowTree.re
@@ -1,3 +1,5 @@
+open Oni_Core;
+
 module WindowSplitId =
   Revery.UniqueId.Make({});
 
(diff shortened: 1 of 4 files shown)


src/editor/Model/WindowTree.re
```diff
@@ -1,3 +1,5 @@
+open Oni_Core;
```
KevinSjoberg: I was not too sure if we want expose everything from `OniCore` here or do something like `module Utility = Oni_Core__Utility` and then prefix `last` and `dropLast`.

test/editor/Core/UtilityTests.re
```diff
@@ -0,0 +1,33 @@
+open TestFramework;
+
+module Utility = Oni_Core__Utility;
+open Utility;
```
KevinSjoberg: Perhaps, it's possible to do `open Oni_Core__Utility` directly?

src/editor/Model/WindowTree.re
bryphe: Good question! I've been using the `open Oni_Core` style a lot, but I'm not sure its canonical Reason / OCaml (which prefer less pollution in the module as far as I can tell). This is reasonable for now though and consistent with the current styling of the code 👍

test/editor/Model/WindowTreeTests.re
```diff
@@ -80,3 +80,161 @@ describe("WindowTreeTests", ({describe, _}) =>
     });
   })
 );
+
+describe("rotateForward", ({test, _}) => {
```
bryphe: Nice set of test cases! Thank you for adding these.

bryphe: Gives me a lot of confidence in the functionality 👍

test/editor/Core/UtilityTests.re
bryphe: Ah, good question... We should perhaps expose `Utility` on the `Oni_Core` module, like we do for the other ones: https://github.com/onivim/oni2/blob/6899c1af8556adc37650405499a588c96c4bcc20/src/editor/Core/Oni_Core.re#L31

This would let us do `module Utility = Oni_Core.Utility` or just `open Oni_Core`.

KevinSjoberg: I think it’s already exposed. 🙂

https://github.com/onivim/oni2/blob/6899c1af8556adc37650405499a588c96c4bcc20/src/editor/Core/Oni_Core.re#L21

bryphe: Oh geez totally missed it, was looking at `Uri` and thought they were arranged alphabetically... oops 😄 

In that case, I think at least the `Oni_Core.Utility` is preferred (the double-underscored  `Oni_Core__Utility` is I believe intended to be hidden / private by the `dune` build tool)

PR #606, opened 2019-08-06
Add resolution for delete key.

CrossR:
Delete doesn't work for me (and some other people on Discord) at the moment, and adding a resolution here seems to work.

One thing I'm not 100% on is if I should be using `"DELETE"` here, or if I need some other representation? I.e, I see that `KEY_BACKSPACE` returns `Some("C-h")`, which is the terminal equivalent as far as I know.


bryphe:
Looks great @CrossR - thank you!

It looks like `<DELETE>` is acceptable to `libvim` (this gets passed verbatim). Good point about the backspace case... I believe `<BS>` would actually work for that as well.

PR #607, opened 2019-08-06
Splits (Rotation) - Part 2

KevinSjoberg:
Depends on the work done #605. Solves part 2 of #439 and by extension solves the issue in its entirety. 🎉


bryphe:
Awesome! Looks great @KevinSjoberg - thank you for the contribution! 😍

PR #612, opened 2019-08-07
Copy HTML extension from vscode

mhanberg:
cc/ @bryphe

edit: Wait a second, I may have messed something up.


mhanberg:
I realized I didn't see that this extension was already integrated...

Closing

PR #613, opened 2019-08-07
Fix #498 - Viewport scrolls back to last position

### LINKED ISSUE #498: Mouse click at bottom of 300 line file puts cursor there, but the viewport at the last position.
Open a large file (300 lines in my case), set the cursor at the top, scroll to the bottom and mouse click the line. The cursor point will be placed properly, but the viewport scrolls back to the last position of the cursor. Very reproducible.

bryphe:
__Issue:__ When scrolling with the mouse, and clicking in the viewport to place the cursor, the viewport will snap back to the previous position. This is very disorienting and confusing.

__Defect:__ As part of our `Editor.t` type, we store the last top line position that came from Vim. This is independent of the actual scroll using the mouse (this allows us to temporarily decouple the cursor position from the scroll window).

However, when we click to place the cursor, we never get a new topline from Vim - so we end up reverting to the previous topline from the last cursor position - this is out-of-date and usually the wrong thing to do.

__Fix:__ Prior to setting the cursor position, let vim know what our current topline is based on the top visible line. Then, place the cursor - if the topline changes again, Vim will set us an update (https://github.com/onivim/reason-libvim/pull/59).

This logic required some fixes on the `libvim` side - it turns out the `w_botline` was out-of-sync so setting the topline initially behaved unpredictably, but this was addressed in https://github.com/onivim/libvim/pull/157


bryphe:
This fixes _most_ of #498 - however, there is a separate issue blocking the `<c-f>` key - I'm going to move that to a separate, more scoped issue to address.

bryphe:
This fixes #498 - there is a visible 'shift' though in some cases when the viewport gets aligned to the current line, but it is much better behavior than before. A further improvement would be not to shift the viewport if it is a subline-change to avoid that slight shift.

Issue #565, opened 2019-08-02
Commands: Window movement


bryphe:
In the same vein as #564 , we should have bindable commands for the window splits:

Commands to be referenced in `keybindings.json`:
- `window.moveLeft`
- `window.moveRight`
- `window.moveUp`
- `window.moveDown`

These won't need defaults, per-se, since we'd inherit the Vim behavior for these movements via `<C-w>` bindings. But at least we'd have an ability to re-bind these.

Use case: Binding `<c-h>, <c-j>, <c-l>, <c-k>` for window motions

KevinSjoberg:
I'd be happy to take a stab at this one. 🙂

KevinSjoberg:
@bryphe I guess we can close this one. 🙂

bryphe:
@KevinSjoberg - oh, ya - thank you for fixing this! 🎉

PR #617, opened 2019-08-07
Setup window movement commands

KevinSjoberg:
Add the ability to setup keybindings for moving between windows. I've tested this by setting defaults since we're not currently reading the `keybindings.json` file.

Solves https://github.com/onivim/oni2/issues/565.

### DIFF
--- a/src/editor/Store/CommandStoreConnector.re
+++ b/src/editor/Store/CommandStoreConnector.re
@@ -61,7 +61,6 @@ let start = _ => {
   };
 
   let windowMoveEffect = (state: State.t, direction, _) => {
-    Log.info("WINDOW MOVE EFFECT");
     Isolinear.Effect.createWithDispatch(~name="window.move", dispatch => {
       let windowId = WindowManager.move(direction, state.windowManager);
       let maybeEditorGroupId =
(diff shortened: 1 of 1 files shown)


src/editor/Store/CommandStoreConnector.re
```diff
@@ -60,6 +60,23 @@ let start = _ => {
     });
   };
 
+  let windowMoveEffect = (state: State.t, direction, _) => {
+    Isolinear.Effect.createWithDispatch(~name="window.move", dispatch => {
```
KevinSjoberg: If we want to get picky, I could stringify the `direction` and append that to the effect name.

bryphe: Cool idea, thanks for thinking about it @KevinSjoberg !

The main idea with the `name` is to have it show up in our logs when we run effects (logging when an effect begins / ends and how long it takes, etc). If it ends up that we need more granular logging - we could add the direction as you suggest to the name, or log the direction in the effect - but I'm OK leaving as-is for now until we need it 👍

bryphe:
Looks great - thank you for the contribution, @KevinSjoberg !

PR #620, opened 2019-08-07
Map Vim window rotation types to actions

KevinSjoberg:
Depends on the work done in https://github.com/onivim/reason-libvim/pull/63. This pull request add the ability to trigger window rotation using `c-w r` and `c-w R`. 🎉

### DIFF
--- a/src/editor/Store/VimStoreConnector.re
+++ b/src/editor/Store/VimStoreConnector.re
@@ -154,7 +154,7 @@ let start = (getState: unit => Model.State.t, getClipboardText) => {
       Log.info("Vim.Window.onMovement");
       let currentState = getState();
 
-      let move = (moveFunc, state) => {
+      let move = (moveFunc) => {
         let windowId = moveFunc(currentState.windowManager);
         let maybeEditorGroupId =
           Model.WindowTree.getEditorGroupIdFromSplitId(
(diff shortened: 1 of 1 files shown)


src/editor/Store/VimStoreConnector.re
```diff
@@ -154,28 +154,35 @@ let start = (getState: unit => Model.State.t, getClipboardText) => {
...
-        Model.WindowTree.getEditorGroupIdFromSplitId(
-          v,
-          currentState.windowManager.windowTree,
-        );
-      switch (editorId) {
-      | Some(ed) => dispatch(Model.Actions.WindowSetActive(v, ed))
-      | None => ()
+      switch (movementType) {
+      | FullLeft
+      | OneLeft => move(Model.WindowManager.moveLeft, currentState)
```
KevinSjoberg: I was not entirely sure what's best practice here. I don't technically have to pass `currentState` to `move` due to lexical scope. `currentState` is defined in the outer scope. However, passing `currentState` makes the dependency obvious.

Thoughts?

tcoopman: You are not using `state` but `currentState` in your move function, so now the argument makes no sense at all.
I probably wouldn't pass the state in, but not sure what the best practices are for this.

KevinSjoberg: Oh, your're right. Nice catch! 🙂 Well, given that the `move` function do use the `state` parameter, does it make sense?

tcoopman: I don't think it matters much, and think it's more of a style thing.

bryphe:
Awesome! Just tested it out and it works great 👍 Thanks for the work on this @KevinSjoberg , and the help reviewing @tcoopman . 

Change looks good to me - just needs an `esy format` (and might need to be merged w/ `master` to pick up the latest `reason-libvim`) - and should be green after that.

KevinSjoberg:
@bryphe rebased on latest master and fixed formatting.

bryphe:
@KevinSjoberg - awesome! Bringing this in now - thank you!

PR #622, opened 2019-08-08
Populate Mac version based on package.json

CrossR:
Could be being dumb, or is there a nicer way to do the package.json loading?

That said, I've found that running `esy create-release` in any folder that isn't the root causes me issues already.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -1,6 +1,7 @@
 const cp = require("child_process");
 const fs = require("fs-extra");
 const path = require("path");
+const package = require("./../package.json")
 
 let curBin = process.env["cur__bin"];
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for catching and fixing this @CrossR !

> Could be being dumb, or is there a nicer way to do the package.json loading?

I think the way you did it is the easiest way 👍 

> That said, I've found that running esy create-release in any folder that isn't the root causes me issues already.

Yeah, unfortunately this is the case... probably need to replace some of the `process.cwd()` with `__dirname`'s... but as you mentioned this doesn't impose any new constraints.

bryphe:
Looks great, thanks for the contribution @CrossR !

PR #627, opened 2019-08-08
add zenModeSingleFile

### LINKED ISSUE #495: Zen Mode: Single-file mode
This is actually @manuhornung 's idea:

---

Apart from full IDE experience, there's also a use case for quickly opening, editing and closing a single file. It always felt unnecessary and overwhelming to me to show that file in a full-fledged IDE window. I'd rather have a simple buffer view without a file explorer or any other sidebar, tabs etc. For that case. Also opening it as a new buffer in an existing window seems confusing to me.

---

I think this would be a pretty cool setting to have. Onivim 2 starts up fast compared to my other editors that it makes sense to use it in this single-file capacity.

#### Part 1 - Add Configuration Value

We'd need to add a `zenModeSingleFile` setting to our configuration value type here:
https://github.com/onivim/oni2/blob/19de6c6fa1a4b7aff19272eaa0bab0b64624f8e0/src/editor/Core/ConfigurationValues.re#L15

I think probably we'd want to default this to `false` - but I'm open to ideas.

#### Part 2 - Add Configuration Parser

Next, we'll need to a parser to pick up the value of `zenMode.singleFile` from the JSON configuration. That can be added here:

https://github.com/onivim/oni2/blob/19de6c6fa1a4b7aff19272eaa0bab0b64624f8e0/src/editor/Core/ConfigurationParser.re#L132

#### Part 3 - Hook up setting

The trickiest part is the 'wiring' up of this setting. We need to know:
1) The value of  `zenMode.singleFile`
2) The number of files the user specified

And then, when we load the configuration, check the value of `zenMode.singleFile` + the number of files, and turn on Zen mode if necessary.

We parse the CLI early here:
https://github.com/onivim/oni2/blob/ac6951508f3555ce9420ad14ba5ee651533ddf50/src/editor/bin_editor/Oni2_editor.re#L19

And we load the configuration here:
https://github.com/onivim/oni2/blob/ac6951508f3555ce9420ad14ba5ee651533ddf50/src/editor/Store/ConfigurationStoreConnector.re#L47

I think the best place to handle this logic would be in the `ConfigurationStoreConnector`. We'll need to plumb the CLI information from that `Oni2_editor` -> `StoreThread` -> `ConfigurationStoreConnector`. Then, on the _first_ load of the configuration, we check that `zenModeSingleFile` value against the number of files specified - and if we need to actiave zen mode, we can dispatch the `ToggleZenMode` action.

despairblue:
Fixes #495 

I've been pretty much following the instructions from the issue. There are a couple of things I'd like to fix, but I'm not sure how:

* opening Oni like that still loads a folder, so using the file switcher is possible and one can open more files which does not disable zen mode, that may be confusing.
* when starting Oni with one file, disabling zen mode and then changing the config file will re-enable zen mode every time since the `reloadConfigurationEffect` runs every time the configuration changes

Let me know what you think.

### DIFF
--- a/integration_test/lib/Oni_IntegrationTestLib.re
+++ b/integration_test/lib/Oni_IntegrationTestLib.re
@@ -39,6 +39,7 @@ let runTest = (~name="AnonymousTest", test: testCallback) => {
       ~getClipboardText=() => _currentClipboard^,
       ~executingDirectory=Revery.Environment.getExecutingDirectory(),
       ~onStateChanged,
+      ~cliOptions=None,
       (),
     );
 
--- a/src/editor/Core/ConfigurationParser.re
+++ b/src/editor/Core/ConfigurationParser.re
@@ -162,6 +162,10 @@ let configurationParsers: list(configurationTuple) = [
     "editor.zenMode.hideTabs",
     (s, v) => {...s, zenModeHideTabs: parseBool(v)},
   ),
+  (
+    "editor.zenMode.singleFile",
+    (s, v) => {...s, zenModeSingleFile: parseBool(v)},
+  ),
 ];
 
 let keyToParser: Hashtbl.t(string, parseFunction) =
--- a/src/editor/Store/ConfigurationStoreConnector.re
+++ b/src/editor/Store/ConfigurationStoreConnector.re
@@ -40,19 +40,22 @@ let start = (~cliOptions: option(Cli.t)) => {
 
   let initConfigurationEffect =
     Isolinear.Effect.createWithDispatch(~name="configuration.init", dispatch => {
-      dispatch(Actions.ConfigurationReload);
       let configPath =
         Filesystem.getOrCreateConfigFile(configurationFileName);
       switch (configPath) {
       | Ok(configPathAsString) =>
         switch (ConfigurationParser.ofFile(configPathAsString), cliOptions) {
         | (Ok(configuration), Some(cliOptions)) =>
+          dispatch(Actions.ConfigurationSet(configuration));
+
           let zenModeSingleFile =
             Configuration.getValue(c => c.zenModeSingleFile, configuration);
 
           if (zenModeSingleFile && List.length(cliOptions.filesToOpen) == 1) {
             dispatch(Actions.ToggleZenMode);
           };
+        | (Error(err), _) =>
+          Log.error("Error loading configuration file: " ++ err)
         | _ => ()
         };
         reloadConfigOnWritePost(~configPath=configPathAsString, dispatch);
--- a/src/editor/Store/ConfigurationStoreConnector.re
+++ b/src/editor/Store/ConfigurationStoreConnector.re
@@ -7,7 +7,7 @@
 open Oni_Core;
 open Oni_Model;
 
-let start = (~cliOptions: Cli.t) => {
+let start = (~cliOptions: option(Cli.t)) => {
   let configurationFileName = "configuration.json";
   let reloadConfigOnWritePost = (~configPath, dispatch) => {
     let _ =
@@ -34,12 +34,17 @@ let start = (~cliOptions: Cli.t) => {
         | Ok(v) =>
           dispatch(Actions.ConfigurationSet(v));
 
-          let zenModeSingleFile =
-            Configuration.getValue(c => c.zenModeSingleFile, v);
+          switch (cliOptions) {
+          | Some(cliOptions) =>
+            let zenModeSingleFile =
+              Configuration.getValue(c => c.zenModeSingleFile, v);
 
-          if (zenModeSingleFile && List.length(cliOptions.filesToOpen) == 1) {
-            dispatch(Actions.ToggleZenMode);
+            if (zenModeSingleFile && List.length(cliOptions.filesToOpen) == 1) {
+              dispatch(Actions.ToggleZenMode);
+            };
+          | None => ()
           };
+
         | Error(err) => Log.error("Error loading configuration file: " ++ err)
         }
       | Error(err) => Log.error("Error loading configuration file: " ++ err)
--- a/src/editor/Store/ConfigurationStoreConnector.re
+++ b/src/editor/Store/ConfigurationStoreConnector.re
@@ -31,20 +31,7 @@ let start = (~cliOptions: option(Cli.t)) => {
       switch (configPath) {
       | Ok(configPathAsString) =>
         switch (ConfigurationParser.ofFile(configPathAsString)) {
-        | Ok(v) =>
-          dispatch(Actions.ConfigurationSet(v));
-
-          switch (cliOptions) {
-          | Some(cliOptions) =>
-            let zenModeSingleFile =
-              Configuration.getValue(c => c.zenModeSingleFile, v);
-
-            if (zenModeSingleFile && List.length(cliOptions.filesToOpen) == 1) {
-              dispatch(Actions.ToggleZenMode);
-            };
-          | None => ()
-          };
-
+        | Ok(v) => dispatch(Actions.ConfigurationSet(v))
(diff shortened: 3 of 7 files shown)


despairblue:
The changes break the tests. I'm trying to fix this by making `cliOptions` optional on `StoreThread.start()` and `ConfigurationStoreConnector.start()` but the type checker just goes crazy. Either telling me that `ConfigurationStoreConnector.start()` doesn't want an option, even though it does. When I then try not to pass the optional value at all it all of the sudden tells me `Oni_Model.State.t` :confused:

despairblue:
Mhh, ok, seems the type checker gets mighty confused when doing something like that 
`let start = (~cliOptions: option(Cli.t)=?, ()) => {`
where this would work
`let start = (~cliOptions: option(Cli.t), ()) => {`

tcoopman:
I'm noticing some weird behaviour while playing with the configuration option when the editor is already open.
To reproduce: open with `singleFile: true`. set `singeFile: false`, set `singleFile: true` now the editor jumps out of zenMode.

My guess is that these things happen because `start` is called on configuration reload as well?

src/editor/Store/ConfigurationStoreConnector.re
```diff
@@ -31,7 +31,20 @@ let start = () => {
...
-        | Ok(v) => dispatch(Actions.ConfigurationSet(v))
+        | Ok(v) =>
+          dispatch(Actions.ConfigurationSet(v));
+
+          switch (cliOptions) {
+          | Some(cliOptions) =>
+            let zenModeSingleFile =
+              Configuration.getValue(c => c.zenModeSingleFile, v);
+
+            if (zenModeSingleFile && List.length(cliOptions.filesToOpen) == 1) {
```
tcoopman: This will run on every configuration reload, so also when you have a lot of files open. The toggle should only be done on startup.

src/editor/Store/ConfigurationStoreConnector.re
```diff
@@ -31,7 +31,20 @@ let start = () => {
...
+              dispatch(Actions.ToggleZenMode);
```
tcoopman: I'm wondering if we want to add a `bool` to ToggleZenMode to be explicit that we enable it (or make it 2 actions EnableZenMode/DisableZenMode). Because now this really is a toggle, and we're assuming that we are the first to execute this action.
Making it explicit will make it more clear, and less prone to strange behaviour (because of the reload)

src/editor/Store/ConfigurationStoreConnector.re
despairblue: Well, yeah, that's what I mentioned in the description :smile: 

The question is how? Should I add a ref(boolean) to the closure and flip it?

src/editor/Store/ConfigurationStoreConnector.re
despairblue: Good idea, @bryphe should I change this in this PR?

despairblue:
> I'm noticing some weird behaviour while playing with the configuration option when the editor is already open.
> To reproduce: open with `singleFile: true`. set `singeFile: false`, set `singleFile: true` now the editor jumps out of zenMode.
> 
> My guess is that these things happen because `start` is called on configuration reload as well?
> I guess that when you reload and you started with 1 file (but maybe have a lot open now) you activate zenMode again. Maybe you can check if this happened at startup or check the number of open files?

That's probably related to the second point in the PR description and your comment [here](https://github.com/onivim/oni2/pull/627#discussion_r312371609). Hopefully that's fixed when I fixed that. Once I know what the preferred way would be. :thinking:

src/editor/Store/ConfigurationStoreConnector.re
tcoopman: why not do this in the `initConfigurationEffect`?

despairblue:
By the way, thanks for taking a look @tcoopman :)

src/editor/Core/ConfigurationParser.re
```diff
@@ -162,6 +162,10 @@ let configurationParsers: list(configurationTuple) = [
     "editor.zenMode.hideTabs",
     (s, v) => {...s, zenModeHideTabs: parseBool(v)},
   ),
+  (
+    "editor.zenMode.singleFile",
```
CrossR: You'll want to add this configuration option to the `ConfigurationDefaults.re` file as well, since that string is currently used to generate the config file on peoples system when they open it for the first time.

I had added a test case to try and alert people when they forgot to update it, but it looks like since I'm comparing the parsed config, it always has every option in it as parsing adds back any missing values... Guess I need to update that test to use the non-parsed config string which will be a pain.

Sorry about that!

src/editor/Store/ConfigurationStoreConnector.re
despairblue: I would need to read and parse the config file there as well. So then we parse it twice unless I thread it through to `reloadConfigurationEffect` by parameterizing `Actions.ConfigurationReload` with and `option(Configuration.t)` :thinking:

despairblue: I moved it, currently we just read and parse the config file twice, on the other hand. Maybe I don't need to ... nevermind found a solution. If I dispatch `Actions.ConfigurationSet(configuration)` in `initConfigurationEffect` instead of `Actions.ConfigurationReload` we don't load the config twice.

src/editor/Core/ConfigurationParser.re
despairblue: I added it, thanks for pointing that out :)

src/editor/Store/ConfigurationStoreConnector.re
despairblue: Let me know what you think.

src/editor/Store/ConfigurationStoreConnector.re
despairblue: Maybe we should open an issue about this discussing it first?

integration_test/lib/Oni_IntegrationTestLib.re
```diff
@@ -1,105 +1,106 @@
...
+
+  logInit("Starting store...");
+
+  let (dispatch, runEffects) =
+    Store.StoreThread.start(
+      ~setup,
+      ~getClipboardText=() => _currentClipboard^,
+      ~executingDirectory=Revery.Environment.getExecutingDirectory(),
+      ~onStateChanged,
+      ~cliOptions=None,
```
bryphe: Makes sense to pass this into the `Store` - thanks for plumbing it through 👍

src/editor/Store/ConfigurationStoreConnector.re
bryphe: Nice catch @tcoopman , and thanks for investigating a fix @despairblue. Latest iteration looks good - it seems like it only uses the `zenModeSingleFile` on first load (in combination with the cli options passed in). Looks good to me! 👍

src/editor/Store/ConfigurationStoreConnector.re
bryphe: I like the idea of switching from `ToggleZenMode` to `EnableZenMode`/`DisableZenMode! Agree with @tcoopman that it would be a bit safer - instead of assuming we're the first ones to call toggle, just calling `EnableZenMode` means this logic will always work regardless of if it was called previously.

It would be a little clearer in the command palette too if we could show the right one - showing "Enable zen mode" and "Disable zen mode" conditionally is clearer than just "Toggle zen mode". 

I'm ok tracking it another issue, though 👍

despairblue: Opened an [issue](https://github.com/onivim/oni2/issues/634) to track it.

bryphe:
Bringing this in now - thank you for the PR and implementing this feature, @despairblue - nicely done! 🎉

PR #628, opened 2019-08-09
Dependency: reason-fontkit -> 2.5.0

bryphe:
This picks up the latest version of `reason-fontkit` which has a fix for building on Linux when there is a system-installed version of `harfbuzz` > 2.0


tcoopman:
confirmed that this fixes building oni on archlinux for me

bryphe:
Thanks for confirming, @tcoopman !

PR #633, opened 2019-08-09
Change the indent size in the tree explorer

BenoitPingris:
#### New entry in the configuration

Added a new entry to handle the intent size of tree explorer
```json
{
"explorer.indentSize": 4
}
```

Before this, the indent size was hardcoded (2) and the tree explorer could look too narrowed for some users.

The `itemRenderer` (`TreeView.re`) now receives the state so it can get the explorer indent size from the configuration and it simply uses it instead of the old hardcoded value.

### DIFF
--- a/src/editor/Core/ConfigurationDefaults.re
+++ b/src/editor/Core/ConfigurationDefaults.re
@@ -31,7 +31,7 @@ let getDefaultConfigString = configName =>
   "workbench.iconTheme": "vs-seti",
   "workbench.sideBar.visible": true,
   "workbench.statusBar.visible": true,
-  "explorer.indentSize": 4,
+  "explorer.indentSize": 2,
 }
 |},
     )
--- a/src/editor/Core/ConfigurationDefaults.re
+++ b/src/editor/Core/ConfigurationDefaults.re
@@ -31,7 +31,7 @@ let getDefaultConfigString = configName =>
   "workbench.iconTheme": "vs-seti",
   "workbench.sideBar.visible": true,
   "workbench.statusBar.visible": true,
-  "workbench.tree.indent": 2,
+  "workbench.tree.indent": 2
 }
 |},
     )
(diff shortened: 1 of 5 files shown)


src/editor/Core/ConfigurationDefaults.re
```diff
@@ -30,7 +30,8 @@ let getDefaultConfigString = configName =>
   "workbench.editor.showTabs": true,
   "workbench.iconTheme": "vs-seti",
   "workbench.sideBar.visible": true,
-  "workbench.statusBar.visible": true
+  "workbench.statusBar.visible": true,
+  "explorer.indentSize": 2,
```
bryphe: Could we rename this setting to `"workbench.tree.indent"` please?

Two reasons I'm thinking:
- VSCode uses `workbench.tree.indent` - that would help us keep our configurations compatible
- We'll likely include other tree views - both as part of the extension work, and for other features (like, showing marks in the sidebar) - it would make sense for this indent setting to be generalized across those.

Sorry to ask you to rename it everywhere! But I think it would help keep people from having to update their configuration if we had to change it down the road 😄

src/editor/UI/TreeView.re
```diff
@@ -65,7 +66,8 @@ let itemRenderer =
       backgroundColor(bgColor),
     ];
 
-  let indentStr = String.make(indent * 2, ' ');
+  let explorerIndent = Core.Configuration.getValue(c => c.explorerIndentSize, state.configuration)
```
bryphe: This strategy looks great to me! Thanks for diving in and figuring out how to wire it through to the renderer 👍

src/editor/Core/ConfigurationDefaults.re
```diff
@@ -30,7 +30,8 @@ let getDefaultConfigString = configName =>
   "workbench.editor.showTabs": true,
   "workbench.iconTheme": "vs-seti",
   "workbench.sideBar.visible": true,
-  "workbench.statusBar.visible": true
+  "workbench.statusBar.visible": true,
+  "workbench.tree.indent": 2,
```
bryphe: It looks like the configuration-defaults test was failing - it didn't like this trailing comma. I went ahead and removed it 👍

bryphe:
Thanks for making the changes, @BenoitPingris ! 

I saw the CI was failing so I fixed a couple of issues:
- Configuration defaults test case was failing when parsing (via `esy '@test' run`) - trailing comma was causing a problem
- Format check was failing - ran `esy format`
- Merged with `master`

Thanks for the updates & tweaking the configuration value name - looks great!

PR #635, opened 2019-08-10
add EnableZenMode and DisableZenMode actions

### LINKED ISSUE #634: Split `ToggleZenMode` into `EnableZenMode` and `DisableZenMode`
As discussed [here](https://github.com/onivim/oni2/pull/627#discussion_r312372447) we may want to have actions specific for enabling or disabling zen mode instead of only having an action to toggle it which may lead to bugs coming from race conditions or more difficult code where we'd need to check the current state before dispatching the toggle action.

despairblue:
Adding EnableZenMode and DisableZenMode.

I left `ToggleZenMode` in there for now. Probably we should remove it though. Not sure. Any opinions?

Fixes #634 

/cc @tcoopman as this was your idea :heart:

### DIFF
--- a/src/editor/Model/Actions.re
+++ b/src/editor/Model/Actions.re
@@ -78,7 +78,6 @@ type t =
   | StatusBarDisposeItem(int)
   | ViewCloseEditor(int)
   | ViewSetActiveEditor(int)
-  | ToggleZenMode
   | EnableZenMode
   | DisableZenMode
   | CopyActiveFilepathToClipboard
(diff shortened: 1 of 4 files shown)


tcoopman:
Yeah, let's remove the toggle.

despairblue:
I rebased the branch on master and removed `Actions.ToggleZenMode`.

bryphe:
Looks great, @despairblue ! Thank you for the contribution!

Issue #601, opened 2019-08-06
Feature: Go syntax highlighting


cnguy:
Similarly to https://github.com/onivim/oni2/issues/473, would it be possible to port over a Go syntax extension?

bryphe:
Should be resolved now with @NinoFurger 's PR #636

PR #636, opened 2019-08-11
Added syntax highlighting for golang

NinoFurger:
As requested here: #601 
I tried it a bit and the syntax highlighting seems to work for me.

### DIFF
--- a/extensions/go/language-configuration.json
+++ b/extensions/go/language-configuration.json
@@ -1,0 +1,31 @@
+{
+	"comments": {
+		"lineComment": "//",
+		"blockComment": [ "/*", "*/" ]
+	},
+	"brackets": [
+		["{", "}"],
+		["[", "]"],
+		["(", ")"]
+	],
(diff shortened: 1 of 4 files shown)


bryphe:
Thanks for bringing this in, @NinoFurger ! Looking into the build failure now - doesn't seem like it's a result of your changes.

I'll kick off the build again just to check-

bryphe:
Green this time around 😄 Bringing this in now - thanks again @NinoFurger !

cnguy:
Wanted to mention this: thank you all! I'll be using this. :) I already use Onivim for Python-related projects.

bryphe:
@cnguy - awesome! 😄 

It looks like there are tree-sitter grammars for Python and Go:
- https://github.com/tree-sitter/tree-sitter-python
- https://github.com/tree-sitter/tree-sitter-go

Will be nice if we can get those integrated w/ the tree-sitter work that is happening

PR #641, opened 2019-08-11
add CopyActiveFilepathToClipboard action

despairblue:
I'm currently using atom and this is one of my most used commands. Whenever I work on a test and I want to run that test only I use that command to the the path of the file I'm currently editing. Or basically if I want to do anything on the terminal with the file I use this.

I hope this is useful for other's as well.


bryphe:
@despairblue - looks great! 

Really appreciate that you added the integration test as well (gives a lot of confidence that it works cross-platform too 👍 )

Thank you for the contribution!

Issue #161, opened 2019-03-16
Native Syntax Highlighting


bryphe:
We currently use the [`vscode-textmate`](https://github.com/microsoft/vscode-textmate) library to facilitate syntax highlighting and theming. However, because it's `node`, that necessitates we run a separate node process for this, and we have to maintain duplicate state for the buffers, which is an expensive use of memory.

Ideally, we could run this in-proc in our ReasonML front-end. If we had a version of this library that was in reasonml or ocaml instead of `node`, we could facilitate this easily.

The work required is to create a reasonml version of the library, like `reason-textmate` that has a similar API. The library itself is a wrapper around the [`oniguruma`](https://github.com/kkos/oniguruma) regex library, so we'd probably want to preserve that and use the C FFI to talk to it.

Mike-Now:
What about TreeScript?  It's more robust than TextMate tokenization and has already bindings to several languages. All that's required is creating bindings to Reason through FFI
http://tree-sitter.github.io/tree-sitter/
https://github.com/tree-sitter/tree-sitter

Akin909:
@Mike-Now I'm mega excited about tree sitter and have been thinking about writing reason bindings for it for a while now, just gotta get some free time to do it. Hopefully if we have a permissive enough syntax highlighting api we can integrate it a separate option for users

CrossR:
I think for VSCode extension compatibility, we are always going to have textmate highlighting around, so a ReasonML version is going to be useful regardless, to help those extensions where tree-sitter doesn't cover yet etc.

That isn't to say that for a lot of languages Tree-Sitter couldn't be much better, more that a ReasonML textmate highlighter would still be useful alongside it. I know personally there are a few weird languages I use that I doubt will ever get a tree-sitter syntax (unless I write one I guess) but they already have TextMate ones.

Mike-Now:
Sure. My reasoning was it could be much easier and future-proof endeavour to integrate TreeSitter than to rewrite vscode-textmate extension given that preparing oni2 beta release will take considerable time.

bryphe:
@Apos mentioned this project: https://github.com/georgewfraser/vscode-tree-sitter - looks interesting!

bryphe:
> My reasoning was it could be much easier and future-proof endeavour to integrate TreeSitter than to rewrite vscode-textmate extension given that preparing oni2 beta release would take considerable time.

[...]

PR #650, opened 2019-08-14
Native Syntax Highlighting - Part 1: TextMate Scope Selector Resolution

bryphe:
In either native syntax highlight option - whether it is using `tree-sitter` or using a native implementation of textmate grammars, we'll eventually need to take textmate scopes and turn them into styles based on the theme (we want to support VSCode themes - and VSCode themes are always specified in terms of textmate selectors).

This might seem odd with `tree-sitter`, but it turns out the syntax tree you get from `tree-sitter` - for example, parsing `[1, "a"]` yields this syntax tree: ` (value (array (number) (string (string content))))` isn't usable for theming directly. 

There actually aren't any syntax theming solutions that will style against this representation, as far as I could find. Looking at Atom, it turns out they have an intermediate layer, that specifies how to take the `tree-sitter` syntax tree and convert it into a theme - an example for JSON is here: https://github.com/atom/language-json/blob/04f1fbd5eb3aabcfc91b30a2c091a9fc657438ee/grammars/tree-sitter-json.cson#L49

It's still open whether we'll support `tree-sitter` or not (still investigating feasibility), but I know that whichever native solution we go, we'll need the ability to take textmate scopes -> resolve into textmate styles. (With `tree-sitter`, we'll need that intermediate layer to map the syntax tree -> textmate scopes, and then use this logic. With a native `textmate` strategy - we'll need the grammar matching to produce textmate scopes from the code file).

This selector-matching logic is inspired by VSCode's selector matching: https://code.visualstudio.com/blogs/2017/02/08/syntax-highlighting-optimizations - we use a similar representation (a Trie) to represent the scopes.


bryphe:
Related to #161

PR #652, opened 2019-08-14
Fix #506 - Part 3/3 - Paste from clipboard

### LINKED ISSUE #506: Clipboard Integration

bryphe:
This is the final piece of #506 to enable clipboard integration as documented in that issue and here: https://onivim.github.io/docs/using-onivim/configuration#vim

This uses the new `vimSetClipboardGetCallback` API from `libvim`/`reason-libvim` to hook into the clipboard behavior.

In the case where `+` or `*` are used, we always respond with the current status of the clipboard.
In the case where the unnamed register is used, we'll send the current clipboard if `vim.useSystemClipboard` contains `"paste"`, otherwise we won't impact that behavior.
In the case where a named register is used - we always defer to Vim behavior.


bryphe:
Fixes #506

Issue #655, opened 2019-08-15
Feature: Fuzzy-finder for opened files


bryphe:
In Onivim v1, we had a fuzzy finder accessible by `Control+Tab` / `Command+Tab` that was similar to the QuickOpen fuzzy finder accessible by `Control+P` - except it only searched through actively opened Windows. It'd be great to have this in v2!

## Part 1: Create a command for this new file finder

The commands are registered in this file:
https://github.com/onivim/oni2/blob/60f2a85ce068bd0c3d29e4a9c8501a4f3dd0e7ed/src/editor/Store/CommandStoreConnector.re#L89

We'd want to add a new command, perhaps something like `quickOpen.files`, and create an effect to handle this - similar to some of the other actions in that file:

https://github.com/onivim/oni2/blob/60f2a85ce068bd0c3d29e4a9c8501a4f3dd0e7ed/src/editor/Store/CommandStoreConnector.re#L127

For now, a good way to validate the pieces are hooked up is to add a `print_endline("Hello, world!")` inside the effect.

## Part 2: Create an entry point to trigger the command

Later, we'll add a keybinding to the command - but for now, we can simply trigger this via the command palette. We just need to add an entry that matches our above command here: https://github.com/onivim/oni2/blob/60f2a85ce068bd0c3d29e4a9c8501a4f3dd0e7ed/src/editor/Model/CommandPalette.re#L2

A good test at this point is to run the application with `esy run -f` (the `-f` gives debugging output), and verify that we can see the logging for our effect when we run our command from the command palette (`Command+Shift+P` on Mac, `Control+Shift+P` on Windows/Linux).

## Part 3: Create our menu

The command palette file above is actually a good example of a menu! The `create` method is simply a menu population function.

We'll want to do something similar for this menu - except populate it based on active files, instead of commands.

Our effect should take in the `state` as a parameter - which we'll need to access to the current opened files via the `state.buffers` here: https://github.com/onivim/oni2/blob/60f2a85ce068bd0c3d29e4a9c8501a4f3dd0e7ed/src/editor/Model/State.re#L13

Inside our effect, we'd want to dispatch the `MenuCreate` action, which takes a menu creator function - something like this:
```
// The create method gets some extra parameters passed for advanced scenarios,
// but we only care about `setItems`...
let quickOpenFilesCreate = (setItems, _, _) => {
   let activeFiles = figureOutActiveFilesFromState(state);
   setItems(activeFiles);
};
```

[...]

PR #661, opened 2019-08-17
Feature - Add quick open files

r281GQ:
For the issue #655 .


bryphe:
Looks like great progress @r281GQ ! Thanks for your work on this 👍

src/editor/Store/CommandStoreConnector.re
```diff
@@ -87,6 +87,64 @@ let start = _ => {
...
+          Str.replace_first(re, "", fullPath);
+        };
+
+        multipleActionEffect([
+          MenuOpen(
+            (setItems, _, _) => {
+              let commands =
+                state.Oni_Model.State.buffers
+                |> IntMap.to_seq
+                |> List.of_seq
```
tcoopman: I would guess that using [Seq](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Seq.html) directly would be a tiny bit more efficient. Probably doesn't matter much, but seems a bit cleaner to me as well.
Also, because of the duplication on `getMetadata` it might be usefull to use `Seq.filter_map`.

r281GQ: Thanks @tcoopman. Ill make those changes.

r281GQ:
Thanks @bryphe. I was able to try this out from a temp command, but i could not get it to work through shortcut. Seems like my build doesnt pick up the changes from the default config. I dont know enough about the build process to tweak it. :(

CrossR:
At a guess @r281GQ, you've only updated the config defaults string.

That is, that string is used to generate the default config, so if you already have that keybinds file, its entirely possible that value isn't being loaded. You might need to open it manually via the command menu and check it.

We should probably think about what the best way of updating that file is over time...

src/editor/Store/CommandStoreConnector.re
```diff
@@ -87,6 +87,54 @@ let start = _ => {
...
+                     switch (Buffer.getMetadata(buffer).filePath) {
+                     | Some(path) =>
+                       Some({
+                         category: None,
+                         name: getDisplayPath(path, currentDirectory),
+                         command: () => {
+                           Oni_Model.Actions.OpenFileByPath(path, None);
+                         },
+                         icon:
+                           Oni_Model.FileExplorer.getFileIcon(
```
bryphe: Thanks for figuring out how to set up the icon 👍

src/editor/Core/ConfigurationDefaults.re
```diff
@@ -43,6 +43,7 @@ let getDefaultConfigString = configName =>
 {
     "bindings": [
         { "key": "<C-P>", "command": "quickOpen.open", "when": [["editorTextFocus"]] },
+        { "key": "<C-TAB>", "command": "quickOpen.openFiles", "when": [["editorTextFocus"]] },
```
bryphe: When I tested locally, I had some problems with `"<C-TAB>"` getting picked up correctly - due to bugs in our input code. Working on a fix for those.

bryphe:
Thanks @r281GQ , looks good!

I tested this locally and it wasn't working at first... but it seems it was due to a bug with resolving `<C-TAB>`. The Revery `Key.toString` method was returning an empty string for `KEY_TAB` (fixing here: https://github.com/revery-ui/revery/pull/539) and even with that... we weren't correctly picking up `<C-TAB>` - so I added a small fix to our input resolution method:https://github.com/onivim/oni2/pull/669

Neither of those bugs are related to this PR though - so I'll go ahead and bring it in (and when I tested with those fixes locally applied, it worked great!) 

Nice work @r281GQ , thank you for the contribution! 👍

PR #667, opened 2019-08-19
Fix #665 - Part 1 - Menu items are not clickable

### LINKED ISSUE #665: Unable to click in fuzzy finder or command menu
If you open the menu, and attempt to click an item, it will actually click through the menu and click whatever is behind it. I noticed this when it opened a file that was in the sidebar.

click-through

bryphe:
This improves the mouse behavior in the Menu (there's still a lot missing - ie, a scrollbar) - but this fixes makes the items clickable again.

This is dependent on a fix in Revery: https://github.com/revery-ui/revery/pull/536

### DIFF
--- a/src/editor/UI/EditorGroupView.re
+++ b/src/editor/UI/EditorGroupView.re
@@ -91,6 +91,7 @@ let createElement =
         left(0),
         right(0),
         bottom(0),
+        pointerEvents(`Ignore),
         backgroundColor(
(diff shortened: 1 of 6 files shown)


bryphe:
We're still missing several basic mouse interactions:
- Scrollbars / scrollwheel
- Click-away from menu

But this fixes the most egregious gap called out in #665

bryphe:
The command line is actually a different control, so that will require a different fix (splitting this into Part 1 to track the menu, and part 2 to track the command line fix)

PR #668, opened 2019-08-20
[WIP] Perf - QuickOpen: Partially offload ripgrep output to thread

bryphe:
There were still cases where we'd hang when processing Ripgrep's output, just because of the volume of it (splitting the strings takes a significant amount of time!). This keeps the processing of the strings offloaded onto a thread, and only delegates it back to the main thread once they've been parsed into lines.

From there, the main thread logic only works on chunks of these lines at once.


bryphe:
Closing in lieu of #693

PR #694, opened 2019-08-24
Update to make filters consistent.

### LINKED ISSUE #691: ⌘+⇧+p Fuzzy-finder is case-sensitive
## What?
When searching the command palette the current results of the fuzzy finder are case-sensitive, yielding unexpected behavior where, as you complete a word, the correct match disappears.
## Example
**Correct casing**
Screen Shot 2019-08-24 at 11 22 44
**Somewhat matching initial segment (wrong-case)**
Screen Shot 2019-08-24 at 11 23 08
**False-negative initial segment (wrong case)**
Screen Shot 2019-08-24 at 11 22 56

## Environment
Onivim v0.2.0
macOS 10.14.6

## Comments
I don't know if this is considered a feature. From a new-user UX standpoint I find it counter-intuitive. For a comparison, VSC searches the command palette in a case-insensitive manner.

CrossR:
Fix #691.

Updates the pre-filter in the MenuJob and the ranking filter in Filter to be consistent:

 - Both should use category + name when available.
 - Both make the full name lowercase when a lower case search term is used.

I know there is some performance concerns around here, so hopefully I'm not doing anything to cause further issues.


bryphe:
Thanks for fixing this, @CrossR ! Sorry that I regressed it while working on the perf.

I ran the benchmarks against this PR and `master`:
__PR:__
```
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  MenuJob: doWork atom                                               |  1000       |  0.931620836258    |  1314      |  0         |  344295688    |  1327491   |  1327491
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
```
__master:__
```
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  MenuJob: doWork atom                                               |  1000       |  0.843745708466     |  1306      |  0         |  342294955    |  1323673   |  1323673      |
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
```

So it looks like there is a slight decrease in performance - but I think this should be OK (it shouldn't affect typing input, as we amortize this work across frames). I did notice from these benchmarks though that it seems like `reasonFuzz` has some heavy allocations (and incurs some GC) - might be something we could look at down the road!

Change looks great, thank you for fixing it 👍

CrossR:
>I did notice from these benchmarks though that it seems like reasonFuzz has some heavy allocations (and incurs some GC) - might be something we could look at down the road!

Wouldn't surprise me...I spent some time on speeding it up with some simple changes, but I haven't spent a lot of time looking at allocations and more yet. Something to look at for sure though.

bryphe:
Since I broke this, I figured I should add a test case to cover it 👍

bryphe:
Will bring this in now - thanks for the fix, @CrossR ! 💯

PR #695, opened 2019-08-24
[WIP] Loading spinner for QuickOpen

bryphe:
Currently, QuickOpen gives no indication that it is actively searching - this change adds an animated loading spinner to show that something is happening.

__TODO:__
- [x] Fix up spinner (rotation is off-center) (used placeholder)
- [x] Use Job progress API to show progress once MenuJob is complete
- [x] Port over fix when running multiple iterations of job in `tick`
- [x] Investigate why sometimes `setLoading(false)` doesn't make it when Ripgrep exits quickly
- [x] Add `deltaTime` and `totalTime` to `Tick` action
- [x] Store menu creation time
- [x] Store menu animation time to fade-in spinner for less jarring in cases when it is only briefly displayed
- [ ] Test if there are any remaining issues for #540 to close


bryphe:
Tested the dev build - it was able to search to completion on root directory on OSX. Need to check with a compiled and code-signed build, though.

Issue #686, opened 2019-08-23
Zen-Mode: Exit zen mode when creating a split


bryphe:
__Issue:__ When in Zen-mode, if the user explicitly creates a split, it's a jarring experience - no split is created. This feels broken.

__REPRO:__
- Open a file in Onivim 2
- Press `Control+Shift+P` or `Command+Shift+P` to open command palette
- Select 'Enable Zen Mode'
- Run `:vsp<CR>` to create a vertical split

__Expected:__ The user should be taken out of Zen mode
__Actual:__ The user is still in Zen mode, and nothing changed

__Implementation:__

The logic for controlling ZenMode is here (`state.zenMode`): https://github.com/onivim/oni2/blob/460232a518d9ed134e0ecbf017047dcc2da3e79e/src/editor/Model/Reducer.re#L38

We'd want to add some additional behavior - the `AddSplit` (https://github.com/onivim/oni2/blob/460232a518d9ed134e0ecbf017047dcc2da3e79e/src/editor/Model/Actions.re#L70) and `RemoveSplit` (https://github.com/onivim/oni2/blob/460232a518d9ed134e0ecbf017047dcc2da3e79e/src/editor/Model/Actions.re#L70) actions are dispatched when a new split is created. We should update our reducer to set `zenMode: false` whenever those actions are dispatched.

__Testing:__

We should have an integration test for this to verify we don't regress this behavior.

A good example test case is:
https://github.com/onivim/oni2/blob/master/integration_test/AddRemoveSplitTest.re

We'd want a similar test, just prior to creating a split - we'd dispatch the `EnableZenMode` action, and validate that `state.zenMode` is true. Then, we'd create a split, and validate that `state.zenMode` switches back to false.

Note that we'd need to add two entries in the `dune` file for the new test: https://github.com/onivim/oni2/blob/master/integration_test/dune

r281GQ:
Hi @bryphe

I can start investigating and create test for this issue, if not done already.

bryphe:
Awesome! Thanks for the help @r281GQ !

bryphe:
Fixed by #713

PR #713, opened 2019-08-27
Disable ZenMode when split is created

r281GQ:
Hey dudes! Sorry for the delay!

This is for #686 .


bryphe:
Awesome @r281GQ , very excited for this! I've been using the 'single-file' mode but hit this frequently.

integration_test/ZenModeSplitTest.re
```diff
@@ -0,0 +1,16 @@
+open Oni_Model;
+open Oni_IntegrationTestLib;
+
+runTest(~name="ZenModeSplitTest", (dispatch, wait, _) => {
```
bryphe: Thanks for adding the test!

src/editor/Store/WindowsStoreConnector.re
```diff
@@ -134,6 +134,7 @@ let start = getState => {
       }
     | AddSplit(direction, split) => {
```
bryphe: I tested this locally, and I think it is actually breaking the 'single-file' mode case.

When we load for the first time - we actually dispatch an `AddSplit` for the first editor group. So it seems like zen-mode is being disabled on launch, even when `editor.zenMode.singleFile` is true. 

I think we'll have to add a condition here to check if there is more than one window split - in other words, we don't want to disable zen mode if this is the first split being added.

bryphe: Fixed in https://github.com/onivim/oni2/pull/713/commits/8b634aa2ee1781b454fa549bd6810a95bb01cd70

PR #716, opened 2019-08-28
Update faq.md to fix broken link to early access portal.

abernix:
The `i` was missing in the URL here!

(Also, I'm excited about the direction this project is going!)


bryphe:
Ah nice catch, thank you @abernix ! Really appreciate the PR 👍

PR #725, opened 2019-08-31
[WIP] Initial merlin integration / diagnostics rendering

bryphe:
In order to exercise some of the language integration features (rendering diagnostics, hover, go to definition, completion), this PR includes merlin discovery for OCaml / Reason, and integrates with it to provide live diagnostics.

This will help us set up the 'front end' for language support - the state management and UI views we require for it, so that we can plugin in the VSCode extensions into the same 'front end'.

### DIFF
--- a/src/editor/Core/Range.re
+++ b/src/editor/Core/Range.re
@@ -11,6 +11,8 @@ let createFromPositions = (~startPosition, ~endPosition, ()) => {
   endPosition,
 };
 
+let ofPositions = createFromPositions;
+
 let create = (~startLine, ~startCharacter, ~endLine, ~endCharacter, ()) =>
   createFromPositions(
     ~startPosition=Position.create(startLine, startCharacter),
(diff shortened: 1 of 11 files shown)


yunti:
I thought those features, completion, go to definition etc... would be handled by the language server plugin - eg jared's reason language server?

CrossR:
I think since we can run `merlin` directly, we are doing here for performance. It also lets us prototype the UI etc without needing to setup the whole VSCode interface parts. When we do, the UI etc should just be able to be re-used and would then work for the lang server etc.

That could be wrong, but I think thats what the PR implies.

bryphe:
> I thought those features, completion, go to definition etc... would be handled by the language server plugin - eg jared's reason language server?

This is the long-term plan! This is more of a 'tactical' fix to unblock basic language support for ReasonML while we continue working on those pieces. This lets us build out the completion and diagnostic UX in parallel while we do that work - so that when the VSCode exthost is in, we already have that functionality to just plug-in.

And also - unfortunately all of the reason language servers I've tested still have issues on Windows. So it's not only about getting the VSCode exthost support... but also fixing those language plugins to support Windows (merlin actually mostly works on Windows). This helps scaffold up some basic language support in the meantime so that Onivim 2 can be usable for developing Onivim 2 😄

PR #735, opened 2019-09-04
Feature: Add user extension discovery

bryphe:
This adds a search path for extensions in the user's `$HOME/.config/Oni2/extensions` folder - this enables manually downloading VSCode extensions and extracting them to that location.

Currently, we only support the textmate grammars and other bits of language info - but as we continue to build out the extension host integration, we'll support more and more from them.


src/editor/Core/Filesystem.re
```diff
@@ -302,6 +302,11 @@ let createConfigIfNecessary = (configDir, file) =>
       )
   );
 
+let getExtensionsFolder = () =>
+  getHomeDirectory()
+  >>= getOniDirectory
+  >>= (dir => getPath(dir, "extensions"));
+
 let getOrCreateConfigFile = filename =>
```
CrossR: Looks like we need an equivalent of this now, `getOrCreateConfigFolder`.

CrossR: I've found myself wanting this now, to add some random languages that don't make much sense to add to the main source here.

This is what I ended up using:

```rust
let getOrCreateConfigFolder = configDir =>
  stat(configDir)
  >>= (
    fun
    | Some(dirStats) =>
      /*
        Check the stats, if its a folder we can stop.
        If its not, we need to make a folder.
       */
      isDir(dirStats)
      /\/= (_ => mkdir(configDir, ()))
      >>= (() => return(configDir))
    | None =>
      /*
       No folder was present, lets make one.
       */
      mkdir(configDir, ()) >>= (() => return(configDir))
  );

let getExtensionsFolder = () =>
  getHomeDirectory()
  >>= getOniDirectory
  >>= (dir => getPath(dir, "extensions"))
  >>= getOrCreateConfigFolder;
```

bryphe: Thanks @CrossR ! Brought this in 👍

PR #742, opened 2019-09-05
Fix #304 - Idle CPU usage increases every time QuickOpen is used

### LINKED ISSUE #304: Idle Rendering Issue
It looks like there is a bug that stops the Idle rendering loop kicking in, in some fashion.

Opening Oni2 and not using QuickOpen means that when you unfocus the editor or don't make any changes to the buffer, the CPU and GPU usage drop to 0%.

However, once you open the QuickOpen menu (not even selecting an item), the idle usage jumps up to match the focused/changed usage.

For example, here I've opened 4 files and then unfocused Oni2:
image

I then left it for a few mins (just to check it wasn't a random issue), then opened the QuickOpen menu. The usage jumps up instantly as the Menu is opened, and doesn't drop back down even after closing or selecting an item.
image

The actual usage seems to vary: I've seen between 1-10% GPU usage, and it seems to be somewhat related to the buffer contents. It is higher when the editor is open, but no changes are being made, whereas before opening the QuickOpen the usage will stay at 0% until a change is made.

Ie this is the unfocused usage of the editor being visible, but not focused: 
image

Where it should be 0% and 0%.

It feels like this is probably something to do with the ripgrep process. When the quick open is actually open I do see a `rg.exe` process, so I think that is being closed properly since it goes away, but its possible the process used to open the process isn't closed?

bryphe:
__Issue:__  Everytime QuickOpen is opened, the idle cpu usage increases. 

__Defect:__ Initially, I suspected the ripgrep process was sticking around, or some of the threads we use for I/O were not being stopped. However, the actual cause turned out to be the timer we use for cursor blinking for input was never being disposed - so every time we created a quickopen, we'd actually trigger a timer that would cause a re-reconciliation, re-layout, and re-render of the UI. The root cause is actually in brisk-reconciler here: https://github.com/briskml/brisk-reconciler/pull/40

__Fix:__ For now, we'll turn off the cursor blinking since it degrades performance. However, we'll need to fix https://github.com/briskml/brisk-reconciler/pull/40 at some point!

### DIFF
--- a/src/editor/UI/OniInput.re
+++ b/src/editor/UI/OniInput.re
@@ -6,13 +6,11 @@ type state = {
   isFocused: bool,
   internalValue: string,
   cursorPosition: int,
-  cursorTimer: Time.t,
 };
 
 type action =
   | CursorLeft
   | CursorRight
-  | CursorTimer
   | SetFocus(bool)
   | DeleteCharacter
   | DeleteLine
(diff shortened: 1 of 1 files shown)


bryphe:
Fixes #304

PR #750, opened 2019-09-09
Experimental Feature: Native TextMate highlighting

bryphe:
Related to #161 - this picks up the [`reason-textmate`](https://github.com/onivim/reason-textmate) library we've been working on and applies to syntax highlighting.

The benefit of this is that we no longer need a standalone node-process to process syntax highlighting - we can process them synchronously and incrementally, in a time-sliced fashion, in our main render loop. There was significant overhead to parsing the syntax highlights in the separate node process; this is a significant streamlining.

In addition, integrating themes was much more complicated in the separate-node-process world, as it had a separate concept of themes, independent from our front-end. With the syntax highlighting integrating directly, we have a single place for themes, so integrating with VSCode themes becomes much simpler.


bryphe:
Still a few bugs to fix + performance improvements to incorporate; but I'll bring this in gated behind an `experimental.nativeTextMate` flag for now

bryphe:
This requires the `experimental.nativeTextMate` flag to be set to `true` in the configuration, but I plan on enabling it by default soon. 

Most grammars I've tested work well with it, and have parity with the previous implementation:
- Reason
- OCaml
- HTML
- Markdown
- JavaScript
- TypeScript
- JSON
- C

Issue #721, opened 2019-08-29
CentOS 7: Unable to start Onivim 2


bryphe:
Some information from Discord thread:
```
Running with the -f and --checkhealth options produces the following output:
/tmp/.mount_OnivimyqQcEG/usr/bin/Oni2: /lib64/libstdc++.so.6: version CXXABI_1.3.8' not found (required by /tmp/.mount_OnivimyqQcEG/usr/lib/libmirclient.so.9)
/tmp/.mount_OnivimyqQcEG/usr/bin/Oni2: /lib64/libstdc++.so.6: versionCXXABI_1.3.9' not found (required by /tmp/.mount_OnivimyqQcEG/usr/lib/libmirclient.so.9)
/tmp/.mount_OnivimyqQcEG/usr/bin/Oni2: /lib64/libstdc++.so.6: version GLIBCXX_3.4.21' not found (required by /tmp/.mount_OnivimyqQcEG/usr/lib/libmirclient.so.9)
/tmp/.mount_OnivimyqQcEG/usr/bin/Oni2: /lib64/libstdc++.so.6: versionGLIBCXX_3.4.20' not found (required by /tmp/.mount_OnivimyqQcEG/usr/lib/libmircommon.so.7)
/tmp/.mount_OnivimyqQcEG/usr/bin/Oni2: /lib64/libstdc++.so.6: version CXXABI_1.3.9' not found (required by /tmp/.mount_OnivimyqQcEG/usr/lib/libmircommon.so.7)
/tmp/.mount_OnivimyqQcEG/usr/bin/Oni2: /lib64/libstdc++.so.6: versionGLIBCXX_3.4.21' not found (required by /tmp/.mount_OnivimyqQcEG/usr/lib/libmircommon.so.7)
Plus more lines that are very similar. I can't paste all of them as I exceed the message size limit.
```

It looks like that CentOS 7 bundles GCC 4.8, but some of the dependencies bundled by the AppImage kit require libstdc++ 6 >= 5.2. For example, `libmirclient9` seems to require `libstdc++6` >= 5.2: https://packages.ubuntu.com/xenial/libmirclient9

Checking distrowatch - it seems that Ubuntu / Arch / Mint / Manjaro all ship with newer versions of GCC, which is why we haven't hit this issue before.

I saw some similiar issues with other AppImages:
- https://github.com/shadowsocks/shadowsocks-qt5/issues/534
- https://github.com/shadowsocks/shadowsocks-qt5/issues/687

[...]

PR #783, opened 2019-09-26
[Experiment] #721 - Downgrade gcc on Ubuntu machine to 4.8

bryphe:


bryphe:
Looks like this is a no-go, since it's still using a higher `libstdc++` version for dependencies - even with gcc downgraded.

PR #797, opened 2019-10-01
A few improvements to the Tabs component

glennsl:
- Enables using the mousewheel to scroll through the tabs
- Brings the active tab into view when it changes
- Includes the icon in the clickable area of the tab
- Simplifies `tabInfo` by adding `editorId` and removing the properties that were precomputed based on it, `active`, `onClick` and `onClose`.

Depends on https://github.com/revery-ui/revery/issues/579 to fix tabs overflowing


src/UI/EditorGroupView.re
```diff
@@ -55,11 +55,9 @@ let toUiTabs = (editorGroup: Model.EditorGroup.t, buffers: Model.Buffers.t) => {
       let (modified, title) =
         Model.Buffers.getBuffer(v.bufferId, buffers) |> getBufferMetadata;
       let ret: Tabs.tabInfo = {
+        editorId: v.editorId,
         title,
         modified,
-        active: EditorGroup.isActiveEditor(editorGroup, v.editorId),
```
bryphe: Thanks for cleaning this up! 🎉

src/UI/Tabs.re
```diff
@@ -10,45 +10,171 @@ open Rench;
...
     mode
-    onClick={t.onClick}
-    onClose={t.onClose}
+    onClick={() => GlobalContext.current().openEditorById(t.editorId)}
+    onClose={() => GlobalContext.current().closeEditorById(t.editorId)}
   />;
 
-let viewStyle = Style.[flexDirection(`Row)];
+let measureWidth: option(node) => int = fun
+  | Some(outer) => outer#measurements().width
```
bryphe: I didn't realize you could use the `fun` keyword like this - cool!

src/UI/Tabs.re
```diff
@@ -10,45 +10,171 @@ open Rench;
...
+let findIndex = (predicate, list) => {
+  let rec loop = i => fun
+    | [] => None
+    | [head, ...tail] when predicate(head) => Some(i)
+    | [_, ...tail] => loop(i + 1, tail);
+  loop(0, list);
+};
+
+let isPendingRender: option(node) => bool = fun
+  | Some(outer) => outer#firstChild()#firstChild()#measurements().width < 0
```
bryphe: Wow! Nice find on how to figure out if a render is pending. 

I wonder if we could provide a helper API on the Revery side to make this more ergonomic?

src/UI/Tabs.re
```diff
@@ -10,45 +10,171 @@ open Rench;
...
+    let innerStyle =
+      Style.[
+        flexDirection(`Row),
+        transform(innerViewTransform),
+      ];
+      
+    (
+      hooks,
+      <View
+        onMouseWheel=scroll
```
bryphe: Nice! 🎉

src/UI/Tabs.re
glennsl: I hope so!

This is basically just what i tried to do, and found to return garbage data in some circumstances. It worries me a bit that such a widely used API can just return garbage like that though. It kind of makes the point of using a safe language a bit moot. Should it be considered a bug in revery?

The easy fix is probably just to have measurements return an `option`, have it detect `MIN_INT` on some measure and return `None` then. but that will still require this `onDimensionsChange` hack everywhere this can happen. Perhaps we could have measurements take a callback instead, that's called when rendering is done?

bryphe: >  It worries me a bit that such a widely used API can just return garbage like that. It kind of makes the point of using a safe language a bit moot. Should it be considered a bug in revery?

Agree 100% ! Good observation. The garbage is a problem - we should be leveraging the language to make it safe and easy to create UI (like we do for things like hooks).

>  Perhaps we could have measurements() take a callback instead, that's called when rendering is done?

I like this idea!

bryphe:
Hi @glennsl ,

Sorry - realized the CI was busted for this change! I'll merge up `master` for this and re-kickoff the CI.

So excited to have this in 🎉

bryphe:
Oh - looks like I need to bring in the overflow Revery fix to it. Looking into that.

bryphe:
Found and fixed an additional overflow issue: https://github.com/revery-ui/revery/pull/599

Brought it in; I'll merge up and this should be good-to-go! 👍

bryphe:
Green now - thanks @glennsl for the incredible contribution! 🎉

PR #799, opened 2019-10-02
Escape path to allow copying into paths with spaces.

CrossR:
Also need to have the following 2 PRs: https://github.com/revery-ui/revery/pull/580


bryphe:
Nice catch @CrossR - thank you for finding and fixing this!

I wish there were a way we could have a 'path-with-spaces' as the default build directory on the Azure CI machines, so we could catch this before they make it in.

PR #811, opened 2019-10-04
Fix typo in docs

glennsl:


bryphe:
Nice catch! Thank you @glennsl 

CI failure was unrelated; will overide and bring this in 👍

PR #823, opened 2019-10-06
Fix typos in MOTIVATION.md

glennsl:
Also, this is a bit outdated so consider it a nudge as well ;) Very interesting and enlightening though!


glennsl:
Crap, I forgot to test the ***NO_CI*** thing....

bryphe:
Thanks a lot for the help cleaning this up, @glennsl !

> Also, this is a bit outdated so consider it a nudge as well ;) Very interesting and enlightening though!

It would be interesting to revisit and see how we're tracking based on the doc - I kind of kept it around for historical reasons 😄 Actually, quite a bit has changed since then (like `libvim`, `SDL2`, etc).

PR #837, opened 2019-10-11
Change Treesitter to tree-sitter

imbsky:


bryphe:
Thanks for fixing this, @imbsky !

Looking at the Tree-sitter docs: http://tree-sitter.github.io/tree-sitter/ - it looks like 'Tree-sitter' is the proper naming. Just tweaked to match 👍

imbsky:
👍

PR #839, opened 2019-10-11
Fix of: Search highlight doesn't move with edits

### LINKED ISSUE #554: Search highlight doesn't move with edits.
If you search for text, go to an instance, and edit the line so that the searched for text moves position, then the highlight will stay at the original text position. If you scroll to the top and then press `n`, you will go to the line but the highlight will be off of the searched for text. The cursor operates correctly. Just the highlight isn't moving.

dycu:
Fixes #554 

I've added refreshing of the highlights after each edit. Little refactor to reuse the common code. Just starting with Reason so it probably needs some polishing.

### DIFF
--- a/src/Model/SearchHighlights.re
+++ b/src/Model/SearchHighlights.re
@@ -45,7 +45,7 @@ let highlightRangesToMap = (ranges: list(Range.t)) => {
 let reduce = (action: Actions.t, state: t) => {
   switch (action) {
   | SearchSetMatchingPair(bid, startPos, endPos) =>
-    Log.info("Handling SearchSetMatchingPair");
+    Log.debug("Handling SearchSetMatchingPair");
     IntMap.update(
       bid,
       oldHighlights =>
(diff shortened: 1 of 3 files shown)


src/Model/SearchHighlights.re
```diff
@@ -45,6 +45,7 @@ let highlightRangesToMap = (ranges: list(Range.t)) => {
 let reduce = (action: Actions.t, state: t) => {
   switch (action) {
   | SearchSetMatchingPair(bid, startPos, endPos) =>
+    Log.debug("Handling SearchSetMatchingPair");
```
tcoopman: should these be removed again?

src/Store/VimStoreConnector.re
```diff
@@ -13,6 +13,27 @@ module Model = Oni_Model;
 module Log = Core.Log;
 module Zed_utf8 = Core.ZedBundled;
 
+let findHighlightList = () => {
```
tcoopman: did you run the formatter?

src/Model/SearchHighlights.re
dycu: were these here before? I've found them helpful so thought I would leave on debug

src/Store/VimStoreConnector.re
dycu: can you hint me how to do that?

CrossR: You can use `esy format` and it will run the default formatter tool over your code.

That should then let it pass the bit the CI failed on.

bryphe:
Woohoo, green! 🎉 Thanks for the fix here - big improvement in the experience when searching!

PR #844, opened 2019-10-12
UI: Initial 'hover' component

bryphe:
This is an initial, basic hover component that will follow the cursor. For now, we can show the following:
- Syntax tokens (per token provider)

And then extend it with other information, as we build out support
- Errors / diagnostics
- QuickInfo
- Links

__TODO:__
- [x] Switch to editor font for diagnostics
- [x] Fix layout / presentation issues
- [x] Configuration: `editor.hover.delay` - how long the hover should wait to show
- [x] Configuration: `editor.hover.enabled` - whether or not we should the hover UX (`true`, `"mouse"`, `false`)
- [x] Theme Color: Background?
- [x] Theme Color: Border?
- ~~Drop shadow~~ - there's some bug with our boxshadow component - holding off for now for further investigation.


glennsl:
Is there a difference between this and a potential Tooltip component?

bryphe:
Good question!

> Edit: I guess a tooltip would typically be anchored to some other UI component.

Right, the `<Hover />` will be based on the cursor position in the buffer (and some logic to determine whether it needs to be above/below depending on the position)

One other difference is I think there are a few types of data we'll be showing in the hover:
- QuickInfo (text / markdown)
- Errors / diagnostics
- Definition location

I believe a `<ToolTip />` will actually be simpler - just a string to show as the tooltip with the border. The interesting part is detecting the hover and deciding when to show it. It's possible we might be able to reuse some pieces.

I also need to check the theme colors - if VSCode themes specify different colors for tooltips & quickinfo (for borders, background, etc)

PR #856, opened 2019-10-15
Bugfix: Check that build is notarized before releasing

bryphe:
It looks like the latest build released to the early access portal for OSX was not notarized - which triggers a gatekeeper warning:

image

If there is a failure during notarization, we should not release the build. This updates the gatekeeper check to occur at the end of the script, so that if it fails, we'll fail the build.


bryphe:
Better fix in #857

Issue #818, opened 2019-10-05
Initial results in ´Ctrl+P´ menu not deterministic.


m-kru:
~~1. When I navigate with down arrow key below the position of last entry in the list I can't navigate with arrow keys anymore.~~ Edit: Fixed in #894

2. I have pressed `CTRL + P` few times in the Oni2 repository, I have get different results. It looks like the behavior is not deterministic. The order or result looks like it is random. I did not realize any period.

![p6](https://user-images.githubusercontent.com/26227737/66256669-8093fd80-e790-11e9-9a7a-718e32804418.png)
![p5](https://user-images.githubusercontent.com/26227737/66256640-3f035280-e790-11e9-9ac5-c422b4270620.png)
![p4](https://user-images.githubusercontent.com/26227737/66256641-3f035280-e790-11e9-8e19-f12e4f1b6430.png)
![p3](https://user-images.githubusercontent.com/26227737/66256642-3f035280-e790-11e9-8d75-63c8258ffd67.png)
![p2](https://user-images.githubusercontent.com/26227737/66256643-3f9be900-e790-11e9-917f-6d8db779f22b.png)
![p1](https://user-images.githubusercontent.com/26227737/66256644-3f9be900-e790-11e9-8d40-e26336c6cd65.png)

glennsl:
FYI: I'm working on this as part of the menu renovation project.

glennsl:
The first issue was fixed in #894, and the second will be fixed in #914 I believe. But I'll close this out anyway as a duplicate of #911, since that's more specific.

glennsl:
Reopening, since this is actually a separate issue.

There seems to be two causes for this:

1. Ripgrep isn't asked to sort the results. This might be because sorting will be slower because it disables parallelism. This trade-off makes sense for workspace search, I think, but not so much for the Ctrl+P menu.

2. Ripgrep results are processed using a FIFO list, which means results will be reordered depending on how many "blocks" ripgrep puts in before the processing job pulls it out. I see no reason for not using a LIFO queue here instead.

Issue #758, opened 2019-09-12
[Feature Request] Command Prompt should expand based on input


historybuffjb:
I was just trying to cd to a nested directory and the command prompt does not expand or provide any other way to see the different directory choices after the cutoff. Either the Command Prompt should expand when the text fills horizontally or another method should be provided to see the dropdown choices.

historybuffjb:
A possible easy solution for this would be to show relative paths instead of full paths or even just shift the drop down options to show right to left so that the most nested directory is always visible.

glennsl:
FYI: I'm working on this as part of the menu renovation project.

glennsl:
I think this needs to be considered as part of the overall question of how to handle overflow in the menus. Having though about it for a bit (and, perhaps more importantly, looked at what vscode does), I propose the following:

* Handle overflow of the command/input itself by scrolling the visible text. OniInput basically just ought to behave like a standard input and scroll to follow the cursor.

* For menu items, separate and emphasize the most significant part of the label, which for a path would typically be the last segment, the filename if looking at open buffers and the directory name currently being entered when using `:cd`. The less significant parts can be shown as dimmed subtext, and just let it overflow. Not entirely sure how to represent this internally with regards to highlighting though.

* If the most significant part of the label still overflows, we can either just let it overflow or try to expand the width of the menu. Since it seems like this would be a very infrequent occurrence, I say we just let it overflow. At least for now.

Issue #749, opened 2019-09-09
Question: A way to navigate command palette with vim (j,k) keys instead of arrow keys?


Mungoid:
I know they are letter keys and need to be used as such, but I find using the arrow keys in the command palette throws off the 'vim' experience for me. I think it would feel more natural if there were a way to use the _j_ and _k_ keys without them typing into the palette. 

Was thinking something like if you are holding down a modifier of some sort and pressing _j_ or _k_ while the palette was open, it would allow you to move up and down. Not sure how that would impact other uses of those keys though. 

This is definitely not critical by any means. More just curious if something like that could be possible?

CrossR:
For reference, we do have `Ctrl-p`/`Ctrl-n` bound to scroll through them, which is the default for the autocomplete etc in vim to my knowledge.

I would guess that we could use `j` `k` for this as well (with a modifier), and I would expect that you could just add that to your keybinds config file already by duplicating the existing `Ctrl-p/n` ones that are there.

That said, I feel like there might be something missing with linking up the keybinds, though I can test this as I'm away from my Windows PC at the moment and my Mac has a bunch of config things that use Ctrl-j etc already, so I've unbound them from other apps...

jrwrigh:
I would generally say that `Ctrl-p` and `Ctrl-n` are much more standard for the "vim experience" rather than using `mod+[j/k]`. 

My two cents: leave it with `Ctrl-p` and `Ctrl-n`. This leaves `mod+[j/k]` open to do other keybindings. If users want to use `mod+[j/k]` to move through lists, then they can make it a keybinding. That might be the conclusion y'all already came to, but I couldn't quite tell.

Issue #469, opened 2019-07-24
Suggestion list doesn't scroll when tabbing off end


szbergeron:
Steps to reproduce: 

1. use ctrl-p or :e in a folder with a lot of files
2. press tab several times

when the highlighted item reaches the last visible item, and then tab is pressed one more time, the highlighted entry is no longer within the dropdown viewport and no feedback is given for what the selected item is

Expected behavior: either contents of the dropdown scroll to contain the currently highlighted/selected entry, or no entry outside the visible dropdown viewport is allowed to be selectable (personal preference for the first option)

glennsl:
FYI: I'm working on this as part of the menu renovation project.

PR #858, opened 2019-10-15
[WIP] Menu unification and renovation

### LINKED ISSUE #815: Can't navigate around suggestions with arrow keys, only with tab that always goes down.
sugg

I would like to be able to navigate with up and down arrow keys. Right now, I can only hit tab, and it always gets down, so it looks like there is no option to go up.

glennsl:
I believe this is now in a state where it's ready for a review. There are still more features to implement and bugs to fix (see below), and I expect numerous changes from review, but the overarching design is fairly settled I hope.

Fixes #815, #818 (the first point, not sure about the second), #469, #321, #314

The [WIP commit](https://github.com/onivim/oni2/pull/858/commits/44d56c339741b560b5dacf9994487055f9baca1c) itself can safely be ignored, unless you want to consider handling input entirely in the QuickmenuView component. The [revamp commit](https://github.com/onivim/oni2/pull/858/commits/45762f9b485f3a1923d8a8809126d9f561a6f779) is the big one, with most of the unification and a lot of refactoring. The rest of the commits are fairly small and contained.

The [subscription infrastructure commit](https://github.com/onivim/oni2/pull/858/commits/d6dd702b70084510045c8978869386c79dcfff21), as well as the following two subscription provider commits, are the most interesting ones. I think this could be used to simplify a lot of the functionality in the StoreConnectors.

#### Remaining Work

- [x] Scrollbar (#219)
- [x] Search highlight
- [x] Handle prop changes in AnimatedView
- [x] Generalize MenuJob as FilteringJob
- [ ] Optional: Open file without closing menu if right arrow is pressed when cursor is at the end of input (mimics vscode)
- ~~Expand menu based on input (#758)~~ (Will be handled separately)

#### Known Issues
- [x] Wildmenu selection out of sync. Since it is an auto-completion menu that updates its commandline based on selection and therefore needs to not have an item selected initially. Requires making selected an `option`.
- [ ] Using up/down arrow "commits" Wildmenu selection, removing other possible completions from the menu. This is caused by using `NotifyKeyPressed` which also sends the key press to vim.
- [x] ESC closes the menu, but does not exit "Commandline" mode.
- [x] Keybindings do not work in QuickOpen menus, only in Commandline/Wildmenu (#749)
- [ ] Loading animations are currently not very smooth. Also unsure about having two separate animations.
- [ ] `onMouseOver` on menu items no longer works after adding scrollbar, because the menu items  are now wrapped in another `View`. **Probably a bug in revery, @bryphe looking**
- [x] MenuJob never completing
- [x] Menu unable to follow selection if it's updated very very rapidly (ref @szbergeron)

#### Code Quality Issues
- [ ] Terminology: "Focus" vs "Select" (still an issue, see discussion below)
- [x] Terminology: `menuCommand` -> `menuItem`
- [ ] Terminology: Need a unified term to refer to "Commandline" and "Wildmenu" combined.
- [ ] `NotifyKeyPressed` is used to "capture" the up/down arrow keys for menu navigation as it doesn't seem appropriate to use ordinary keybindings for this. It works, but is not great long-term as it's stringly-typed and changes to its original use case might accidentally break this. Consider using some sort of "hard-coded keybinding" instead.
- [ ] Consider abstracting away `Ripgrep` from `RipgrepSubscription` as there's no need to hard-code this to a specific technology. Also might be an idea to specialize it as directory search, hard-code the connection to `FilterSubscription` which then allows hiding the awkward `itemStream` connection.


src/Store/MenuStoreConnector.re
```diff
@@ -9,177 +9,270 @@ module Model = Oni_Model;
...
-      )
-    | MenuSearch(query) => (
+          selected: max(0, min(count, index)) // TODO: Could use a clamp function
+        }
+      }, state),
+      Isolinear.Effect.none
+    )
+
+    | NotifyKeyPressed(_, "<UP>")
+    | MenuFocusPrevious => (
```
glennsl: This would be much less fragile if `NotifyKeyPressed` had carried a strongy typed key, but doing so would involve normalizing `keyEventListener` in `InputStoreHandler` on `keyEvent` instead of on a string.

src/Store/MenuStoreConnector.re
```diff
@@ -9,177 +9,270 @@ module Model = Oni_Model;
...
+              0
+            } else {
+              (state.selected + 1) mod count
+            }
+        }
+      }, state),
+      Isolinear.Effect.none
+    )
+
     | MenuSelect =>
```
glennsl: Originally there were multiple different meanings used for the word "select". I've introduced "focus" to indicate the highlighted item in the `Actions.t` type, but haven't followed up on the use elsewhere. It would be easier to just call `MenuSelect` something else, but what...?

glennsl:
#### On naming

There are many names here that aren't very descriptive or accurate, but I've avoided changing most of them to make it easier to follow the progression of the code base. Instead I'll list them here for consideration:

- **Menu -> Quickmenu**: This one I've already renamed because it's pretty much been completely rewritten anyway, but there are still plenty of remnants like in the the `Menu*` constructors of the `Actions.t` variant, the `MenuStoreConnector` etc. The rationale for renaming is that there might be many different kinds of menus, and it will eventually become unclear which menu these refer to.

- **MenuJob -> FilteringJob**: `MenuJob` doesn't really say much about what it does. It should also be possible to generalize this so it's not hard-coded to work on `menuCommand`s, in which case I think `FilteringJob` would be a better name. Or if it's still specialized then `MenuFilteringJob`.

- **RipgrepSubscription -> ?**: There's no need to tie this to a specific technology, and it's not a very descriptive name anyway, but I've failed to think of a better name since all it really does is recursively list the files of a certain directory. `DirectorySearch` for example isn't really accurate since it doesn't actually search...

- **quickOpen.open/quickOpen.openFiles -> quickOpen.workspaceFiles/quickOpen.buffers**: Already renamed because I found it massively confusing, `openFiles` referred to buffers and `.open` referred to workspace files.

- **menuCommand/commands -> menuItem/items**: I find `menuCommand`/`commands` to be too specific and often think it's referring only to, or is specific to, the `CommandPalette`.

- **Select/Focus**: As mentioned in a comment above, "select" is overloaded with several different meanings. For example, `state.quickmenu.selected` refers to the currently highlighted menu item, while the `MenuSelect` action refers to the act of choosing/executing the currently "selected" item. I've introduced the term "focus" to refer to the former meaning, but have not gone all in on it since it's much more work to change the name of this meaning than it would be to rename `MenuSelect`. But I have no idea what alternative term to use for the latter. It also doesn't help that "focus" also has different meanings, referring to keyboard/input focus in revery I believe.

- **Wildmenu/Commandline**: Currently the vim command menu is referred to as both `Commandline` and `Wildmenu` in different places. For example, the view is called `CommandlineView` but the `menuVariant` constructor is called `Wildmenu`. From what I understand these are terms from vim and there refer to separate concepts, but it would be good to find a common name for it to use in Oni. Or at least some principle makes their use more consistent.

- **Job.mapw**: I can easily see what this does by looking at its definition. But then I forgot and get just as confused the next time I see it. I'm not sure why this even exists, as it mostly just seems to obscure what would be really clear if just substituted with its definition: `job |> Job.map(...) |> Job.doWork`.

tcoopman:
> * **RipgrepSubscription -> ?**: There's no need to tie this to a specific technology, and it's not a very descriptive name anyway, but I've failed to think of a better name since all it really does is recursively list the files of a certain directory. `DirectorySearch` for example isn't really accurate since it doesn't actually search...

RecursiveDirectoryList or RecursiveListDirectory maybe from what you described

src/UI/FlatList.re
```diff
@@ -82,23 +120,54 @@ let createElement =
...
     );
+    (
+      hooks,
+      <View
+        style=Styles.container(~width, ~height=min(menuHeight, count * rowHeight))
+        onMouseWheel=scroll >
+
+        <View style=Styles.viewport(~isScrollbarVisible=scrollbar==React.empty)>
+          ...items
+        </View>
```
glennsl: Wrapping items in another `View` cause `onMouseOver` to stop triggering. Seems like a bug in revery though.

glennsl:
@tcoopman Seems accurate, but doesn't exactly roll off the tongue :laughing: I'm wondering of it shouldn't just use MenuJobSubscription directly though, thereby actually becoming `DirectorySearch` and removing the need for the awkward `itemStream` to connect them together in "userland".

glennsl:
Just fixed an issue caused by 'VimStoreConnector.updater' being called before `MenuStoreConnector.updater`, because the former depend on state being set by the latter. This would not have happened if the communication with vim went through a subscription, since subscriptions by definition react to the state changing, not to actions. This seems very doable at least for the completion aspect, but seems out of scope for this PR. Food for thought though.

src/Store/VimStoreConnector.re
```diff
@@ -631,15 +632,21 @@ let start =
         state,
         pasteIntoEditorAction,
       )
-    | Model.Actions.WildmenuNext
-    | Model.Actions.WildmenuPrevious
-    | Model.Actions.WildmenuSelect(_) =>
+    | Model.Actions.MenuFocusNext
+    | Model.Actions.MenuFocusPrevious
+    | Model.Actions.MenuFocus(_) =>
+      // IFFY: Depends on the ordering of "updater"s
```
bryphe: This 'update-ordering' dependency is tricky for sure. I wonder if there is a better way for us to model this? Does Elm have the same problem? 🤔

src/UI/QuickmenuView.re
```diff
@@ -60,83 +56,103 @@ module Styles = {
...
+
+    // TODO: Cam this be removed in favor of NotifyKeyPressed in MenuStoreConnector?
     let handleKeyDown = (event: NodeEvents.keyEventParams) =>
       switch (event) {
       | {key: Revery.Key.KEY_DOWN, _} =>
-        GlobalContext.current().dispatch(MenuNextItem)
+        GlobalContext.current().dispatch(MenuFocusNext)
       | {key: Revery.Key.KEY_UP, _} =>
-        GlobalContext.current().dispatch(MenuPreviousItem)
+        GlobalContext.current().dispatch(MenuFocusPrevious)
```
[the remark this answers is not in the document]
bryphe: Thanks for streamlining this!

src/UI/QuickmenuView.re
```diff
@@ -60,83 +56,103 @@ module Styles = {
...
-    let loadingOpacityAnimation = Animation.getValue(loadingAnimation);
     let loadingSpinner =
-      Animation.isActive(loadingAnimation)
-        ? <View style=Style.[height(40), width(Constants.menuWidth)]>
-            <Center>
-              <View
-                style=Style.[
-                  transform(
-                    Transform.[RotateY(Math.Angle.Radians(time *. 2.))],
+      if (jobProgress == 0.) {
```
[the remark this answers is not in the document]
bryphe: Ya, sorry, it is more complicated than it should be! There was a bug previously with `OnMount` effects: https://github.com/briskml/brisk-reconciler/pull/40 - this would cause the `OnMount` cleanup function to not be called.... which would leave animations hanging (triggering re-renders, and also leaking memory).

But now that this is fixed - we could go back to a declarative approach and animate the components and remove the animations from being handled.

There were two animations that I used for the 'quickmenu`:
- A progress bar at the bottom of the menu, while we were discovering files.
- A spinner while we were filtering files.

src/UI/WildmenuView.re
```diff
@@ -1,44 +0,0 @@
-open Revery.UI;
```
bryphe: Thanks for consolidating this! 🎉 🎉

src/Store/MenuStoreConnector.re
```diff
@@ -9,177 +9,270 @@ module Model = Oni_Model;
...
+module Quickmenu = Model.Quickmenu;
 module MenuJob = Model.MenuJob;
 
-let start = () => {
-  let (stream, dispatch) = Isolinear.Stream.create();
 
-  let position = (selectedItem, change, count) => {
-    let nextIndex = selectedItem + change;
+// TODO: Remove after 4.08 upgrade
+module Option = {
```
bryphe: I'm looking forward to putting those monadic binding operators to use 👍

src/Core/Types.re
```diff
@@ -193,7 +185,6 @@ module Input = {
     | [@name "menuFocus"] MenuFocus
     | [@name "textInputFocus"] TextInputFocus
     | [@name "editorTextFocus"] EditorTextFocus
-    | [@name "commandLineFocus"] CommandLineFocus
```
bryphe: Thanks for simplifying this!

src/Store/MenuStoreConnector.re
bryphe: Hmm, `MenuPick`? `MenuChoose`? One thing we'll have to consider as well is the 'keyboard' focus vs the mouse hover.

src/UI/QuickmenuView.re
```diff
@@ -0,0 +1,243 @@
...
+                  ),
+                ),
+                backgroundColor(theme.oniNormalModeBackground),
+              ]
+            />
+          </View>
+        </Opacity>
+      };
+
+    let renderItem = index => {
```
bryphe: Cool! I can use this component in the completion UI too to render highlights for matching completion items 🎉

src/UI/FlatList.re
bryphe: Interesting - I can follow up on this part 👍

src/UI/QuickmenuView.re
```diff
@@ -0,0 +1,243 @@
...
+      let item = items[index];
+
+      let labelView = {
+        let font = GlobalContext.current().state.uiFont.fontFile;
+        let normalStyle = Styles.label(~font, ~fg=theme.editorMenuForeground);
```
bryphe: One quirk about our text rendering right now (pre-Skia...) is that we actually need to know the background color to get full fidelity (gamma corrected font). If we add the background color - the text will render a bit crisper.

Once we have Skia integrated, it won't be necessary to do this anymore

szbergeron:
Tried using this a bit ago, if you use the up and down keys and let them repeat very quickly the selected item can actually go off the end of the visible viewport. This is fixed by waiting and pressing either up or down again.

bryphe:
Thanks for the thinking around naming, @glennsl !

A couple thoughts: 

> MenuJob -> FilteringJob: MenuJob doesn't really say much about what it does. It should also be possible to generalize this so it's not hard-coded to work on menuCommands, in which case I think FilteringJob would be a better name. Or if it's still specialized then MenuFilteringJob.

Generalizing makes sense. I think we could potentially use this for completion, for example.

> Job.mapw: I can easily see what this does by looking at its definition. But then I forgot and get just as confused the next time I see it. I'm not sure why this even exists, as it mostly just seems to obscure what would be really clear if just substituted with its definition: job |> Job.map(...) |> Job.doWork.

I'm fine with remove the `mapw` and replacing it with the more explicit version (it isn't saving much!)

glennsl:
@szbergeron Hmm, unable to reproduce this right now. Could be that I have accidentally fixed it. I did make a a tweak [here](https://github.com/onivim/oni2/pull/858/commits/6b0f21e616cf30d607c15be32d79fa6dc330241c#diff-43c905447085a1d30623f77a3d2881b7R100-R105) yesterday to ensure it's not scrolled past the items when they suddenly disappear as they're filtered out. That might have affected it.

Let me know if you find it still occurring though. Also which menu it occurs in and if it also happens with `Ctrl+N/P`. I'm trying hard to get them all to behave the same, but they're still handled differently in some cases.

src/UI/QuickmenuView.re
glennsl: Ah, so that's why. I couldn't see any difference, but didn't look _that_ closely. Will fix!

src/UI/QuickmenuView.re
glennsl: As it is it's probably a bit bulky for that, but I'm sure it can be generalized!

src/Store/VimStoreConnector.re
glennsl: Elm code typically isn't split up this way, for several reasons: 1) Because it uses "subscriptions" to start and stop continuous processes, which of course is why I've tried to replicate that here. And 2) because using the wildcard pattern is discouraged and exhaustiveness checking on variants to driving development is an important part of the Elm workflow.

I think Evan's talk on "The Life of a File" goes through this workflow, and Richard Feldman's tour of `elm-spa-example` goes into more high-level choices. But the workflow can roughly be summed up as first thinking very hard about the model to make sure you get it right, what states should be representable and which should be impossible to represent (Yaron Minsky has a good talk on this), and what state transitions (ie. actions) are needed. Then just run the compiler until it stops complaining. The types and exhaustiveness checking makes sure you handle all the possible actions and initialize all the values correctly. All that's left then is to render the UI and wire up the events.

If a file seems to get too big to handle (a line of thinking that Evan is very much against, mind you) You'll typically split it up by defining a sub-action type and giving it its own constructor in the main action variant. So just like you have `state.quickmenu`, you might have an action `Quickmenu(Quickmenu.action)`.

To avoid the resulting tedious boilerplate when creation actions, like `dispatch(Quickmenu(MenuSelect(42))`, Elm provides `map` functions, like `Html.map` to map `Html` elements from one message type to another. In F#'s elmish, which also uses a dispatch function, [this is very easy to do with just function composition](https://stackoverflow.com/a/53678466/7943564). Unfortunately Reason does not have a built-in function composition operator or first-class variant constructors.

src/Store/MenuStoreConnector.re
glennsl: Yeah, those seem decent. I'll go with one of those. We can revisit later. Disambiguating select lets me revert the use of focus, so that shouldn't be ambiguous then either

src/UI/QuickmenuView.re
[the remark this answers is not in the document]
glennsl: Two animations still seem needlessly complicated though. I understand they say different things of course, but I don't think end-users do, much less which process they refer to. We should either try to unify them into one animation, or just go with a basic "working" animation.

szbergeron:
@glennsl just pulled the latest changes and was able to reproduce, I can reproduce in both Ctrl+p and Ctrl+Shift+p. This happens both with arrow keys and with Ctrl+N and Ctrl+P. This is with key repeat delay of 25ms (40hz); it might not happen with slower repeat rates.

src/Core/ConfigurationDefaults.re
```diff
@@ -44,8 +44,8 @@ let getDefaultConfigString = configName =>
       {|
 {
     "bindings": [
-        { "key": "<C-P>", "command": "quickOpen.open", "when": [["editorTextFocus"]] },
-        { "key": "<C-TAB>", "command": "quickOpen.openFiles", "when": [["editorTextFocus"]] },
+        { "key": "<C-P>", "command": "quickOpen.workspaceFiles", "when": [["editorTextFocus"]] },
+        { "key": "<C-TAB>", "command": "quickOpen.buffers", "when": [["editorTextFocus"]] },
         { "key": "<C-V>", "command": "editor.action.clipboardPasteAction", "when": [["insertMode"]] },
         { "key": "<D-V>", "command": "editor.action.clipboardPasteAction", "when": [["insertMode"]] },
         { "key": "<D-P>", "command": "quickOpen.open", "when": [["editorTextFocus"]] },
```
CrossR: This line needs updating to `quickOpen.workspaceFiles` as well.
We have CMD-P bound on Mac, the same as VSCode.

glennsl: Thanks! Fixed locally.

src/Store/MenuStoreConnector.re
glennsl: Hmm, there's a keybinding command too, "menu.select". "menu.choose" certainly doesn't seem as intuitive, and having "menu.select" map to `MenuChoose`, with `MenuSelect` having a different meaning isn't ideal...

glennsl:
I've updated the PR description with a number of code quality issues. I think the TODO list is now pretty exhaustive. Feedback on the opinionated parts of the remaining items would be appreciated.

Also, @bryphe, I could use your input in particular on these issues:

1. I could use a rundown of how `Job` works/is supposed to work. I thought it would just be a matter of running `Job.tick` until completed, but it doesn't seem like that's sufficient. Currently the filtering job for `quickOpen.workspaceFiles` never completes despite `job.tick` being run as often as possible. Also, when removing `Job.mapw` I also removed a few calls to `Job.doWork` that seemed unnecessary, but then weird(er) things started happening. So clearly I do not understand how this works...

2. Animations are currently not very smooth. How do I debug rendering issues like this?

src/Store/VimStoreConnector.re
bryphe: Thanks for the resources and details here! I'll check out the talk and Richard Feldman's example.

> If a file seems to get too big to handle (a line of thinking that Evan is very much against, mind you) You'll typically split it up by defining a sub-action type and giving it its own constructor in the main action variant. So just like you have state.quickmenu, you might have an action Quickmenu(Quickmenu.action).

Interesting. I like this - I could see this being more understandable than our uber `Actions.t`.

> Unfortunately Reason does not have a built-in function composition operator or first-class variant constructors.

This is a bummer. I keep hoping the first-class variant constructors will come with the OCaml compiler.

bryphe:
> How does Job work, or how is it supposed to work?

The motivator behind introducing the `Job` at all was when we had very, very large lists (for example, opening quickopen at the root `/` like in #467) - when we'd 'filter' after a keystroke, that could take a an appreciable amount of time, and there'd be very noticeable input latency. It was introduced in #563 to help manage that.

The idea is, instead of trying to filter _everything_ (because we'd always be able to find a list long enough to cause it to block), we'd time-slice the filtering - we'd set a budget of X ms, and filter as much as we can in that timeframe. And keep doing that until we complete. This keeps the input unblocked and puts an upper-bound on the amount of per-frame work we're doing.

Also - I actually had some test cases for it too: https://github.com/onivim/oni2/blob/master/test/Model/MenuJobTest.re - might be worth looking at those & comparing the behavior with the new filter job with those.

>  I thought it would just be a matter of running Job.tick until completed, but it doesn't seem like that's sufficient.

This _should_ be how it works -we call `let job  = Job.tick(job)` until eventually `Job.isComplete(job)` is `true`.

There's a difference in this new behavior vs the old behavior though - it looks like in the `FilterSubscription` we're calling `updateQuery` constantly - this seems to be resetting the completed state. Is it by design to call `updateQuery` even when the query hasn't changed? I'll add a comment in-line specific to the place I'm seeing the issue. 

> Animations are currently not very smooth. How do I debug rendering issues like this?

Hmm, I'm not seeing the animations at all right now. It's possible we're still doing a lot of per-frame work that is causing us not to render at 60 hz? I sometimes log out the animation values from render, or set the `REVERY_PERF=1` flag to get a sense of the render timings.

src/Core/Subscription.re
```diff
@@ -0,0 +1,98 @@
...
+
+  let start : (~id: string, ~params: params, ~dispatch: action => unit) => unit;
+  let update : (~id: string, ~params: params, ~dispatch: action => unit) => unit;
+  let dispose : (~id: string) => unit;
+};
+
+type provider('action, 'params) =
+  (module Provider with type action = 'action and type params = 'params);
+
+// We need this to be a GADT because we want to "hide" the `'params` type
```
bryphe: Thanks for the detailed comment here!

src/Store/FilterSubscription.re
```diff
@@ -0,0 +1,118 @@
...
+        disposeMessagePump();
+      };
+
+      Hashtbl.add(jobs, id, {job, dispose});
+    };
+
+    let update = (~id, ~params as {query}, ~dispatch) =>
+      switch (Hashtbl.find_opt(jobs, id)) {
+        | Some({ job } as state) =>
+          // `MenuJob.updateQuery` checks if `query` has changed, so we don't need to. It would make the commented out log message below less annoying though
```
bryphe: I believe this actually might be necessary, as if I check the `query`, the job actually gets completed: ```
          if (!String.equal(Job.getPendingWork(job).filter, query)) {
          let job = Job.map(FilterJob.updateQuery(query), job) |> Job.doWork;
          Hashtbl.replace(jobs, id, { ...state, job });
          }
```

One other thought - `FilterJob.updateQuery` _always resets the completed flag to `false`_ - it looks like we're calling this even if the job is complete... I think that's really the root cause. But just wrapping this in a block like this:
```
if (!Job.isComplete(job)) {
          let job = Job.map(FilterJob.updateQuery(query), job) |> Job.doWork;
          Hashtbl.replace(jobs, id, { ...state, job });
}
```

isn't the right fix - because we do want to reset the completion state to `false` if the query changed (for example - if I backspace - it'll need to filter a new set of results)

**Comment by bryphe** [comment]
I wonder if it'd be possible to split out some parts of the PR? I'm seeing a few potentially 'orthogonal' pieces - I always prefer smaller changes if possible, just in case there is a regression - it makes it easier for me to track down! 

Potential splits I see:
- Menu highlights - (CommandPalette.re, Filter.re, MenuJob changes pre-functorization)
- Menu scrollbar + ImmediateList
- Menu refactor to support 'subscription' model
- Consolidation WildMenu <-> Commandline

I think, at least the first two, could potentially go in immediately (but perhaps there are dependencies I'm missing on the other parts). I also secretly want the menu highlights so I can incorporate them into the completion experience 😉 

Thanks again for all your work on this, @glennsl !

File: src/Core/Subscription.re [hunk 15]
```diff
@@ -0,0 +1,98 @@
+module type Provider = {
+  type action;
+  type params;
+
+  let start : (~id: string, ~params: params, ~dispatch: action => unit) => unit;
+  let update : (~id: string, ~params: params, ~dispatch: action => unit) => unit;
+  let dispose : (~id: string) => unit;
+};
+
+type provider('action, 'params) =
```
bryphe: I didn't realize you could assign a type like this - cool!

src/Store/StoreThread.re
```diff
@@ -125,17 +123,24 @@ let start =
...
     let (newState, effect) = storeDispatch(action);
     accumulatedEffects := [effect, ...accumulatedEffects^];
     latestState := newState;
 
     if (newState !== lastState) {
       onStateChanged(newState);
+
+      // TODO: Wire this up properly
+      let menuSubs = menuSubscriptionsUpdater(newState)
+      MenuSubscriptionRunner.run(dispatch, menuSubs);
```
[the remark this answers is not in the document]
bryphe: Cool! This seems like the missing architecture piece we need from TEA. I wonder if it makes sense to integrate it into Isolinear, which is where most of the state / action stuff lives at the moment.

Maybe it makes sense to add a `subscriptions` argument to `Isolinear.Store.create`? And then - whenever there is a state change - we'd ask for an updated set of subscriptions?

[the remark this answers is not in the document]
glennsl: Yeah, that sounds like a good idea. I haven't looked at `Isolinear.Store` at all so far, but just from how it's used in Oni it seems good. Would you want that done before this is merged, or can we merge it as is and look at it later? It's so contained that it shouldn't take much to rip it out later.

src/Store/FilterSubscription.re
glennsl: Yeah, that was exactly it. An assumption made based on an oversimplified impression of `updateQuery` that was then forgotten about... :/ Fixed by actually checking whether the query has changed.

Thanks!

It also seems like I can now revert https://github.com/onivim/oni2/pull/858/commits/6276a8306d5bdf36d72dce6a0c72a92d2134d0e8 without strange things happening!

glennsl:
Thanks for the explanation @bryphe! My assumption about `updateQuery` itself checking whether the query had changed had thrown me off quite a bit as it was behaving strangely in more ways than just never completing. Like if I added items without calling `doWork` they'd never be picked up it seems. But after fixing that, everything now works as you've explained, and as I originally expected!

> I'm not seeing the animations at all right now
Hmm, I had them running all the time before I fixed the menu job. Now I'm not getting it either though. But this is how it looks for me:

loading

> I sometimes log out the animation values from render, or set the REVERY_PERF=1 flag to get a sense of the render timings.

I've tried logging out the animation values, and they seem smooth. `REVERY_PERF=1` doesn't give me any more logging.

> I wonder if it'd be possible to split out some parts of the PR?

Yeah, sorry about that. The PR seems to have snowballed a bit. Because I didn't know what was needed to unify them, I wasn't sure how I would implement some of the features in a way that wouldn't just make the unification harder. Particularly with regards to input handling. I also ended up trying out both approaches anyway since I chose wrong.

Anyway, I can easily pull out the first few commits, and some later commits like `Subscription` and `RipgrepSubscription` but they don't contain all that much interesting. The rest I'm unsure how to untangle without causing a mess when trying to rebase this, because there's been a lot of renaming and refactoring along the way, especially in the latest commits. I worry that I'd have to rebuild a lot of it, which I can of course do, and would lead to much cleaner and easier to review commits anyway, but it'll take a significant amount of extra time I think, time, especially since PRs will depend on each other and need to be reviewed and merged in sequence,

I'm not against it though, I definitely prefer neat and well-defined commits too.

src/Store/FilterSubscription.re
bryphe: Woohoo! Glad that was the issue 👍 I suspect that was maybe related to the weirdness described here too:
>  I also removed a few calls to Job.doWork that seemed unnecessary, but then weird(er) things started happening. 
(Since the `Job.doWork` would flip it back to completed). Anyway great news!

bryphe:
> I've tried logging out the animation values, and they seem smooth. REVERY_PERF=1 doesn't give me any more logging.

Interesting. It's possible the frames themselves are taking a long time to render? It'd be nice if we could output the result of `REVERY_PERF` into a performance trace we could load in Chrome (https://github.com/revery-ui/revery/issues?utf8=%E2%9C%93&q=is%3Aissue+profile) - could give a nice visualization of the frametimes.

> Anyway, I can easily pull out the first few commits, and some later commits like Subscription and RipgrepSubscription but they don't contain all that much interesting. The rest I'm unsure how to untangle without causing a mess when trying to rebase this, because there's been a lot of renaming and refactoring along the way, especially in the latest commits. 

No worries! It was just a thought I had while looking through the PR, so wanted to run it by you - but you're in the best position to weigh the trade-offs so I'm happy to defer the decision to you 👍

glennsl:
I'll give it a shot, and see how far I can go. Could be an interesting learning experience in rebasing at the very least 😁

src/Store/VimStoreConnector.re
glennsl: > I keep hoping the first-class variant constructors will come with the OCaml compiler.

As do I, but I wouldn't hold my breath. It was a conscious design decision back in the day, and I think the confusingly ambiguous syntax that resulted makes it difficult to implement in both a non-breaking and non-awkward way.

PR #859, opened 2019-10-15
Revert "Fix of: Search highlight doesn't move with edits (#839)"

bryphe:
This reverts commit 68dc5c00dd1b77b4390c8d06ddfc32acfefefdd1, as unfortunately there
are new crashes appearing on the master branch after this change. 

I haven't isolated an exact repro; but on Windows I would see crashes when switching buffers. When I reverted this change, the crash went away. I suspect there may be cases where `let highlights = Vim.Search.getHighlights();` is having problems, depending on the state of the buffer. 

For example, we're running it on buffers that have never had a search applied, or have a first-time update - it could be that we're hitting a crash here.

In addition - I also saw cases where highlights would 'come back' after running `:nohl`.


dycu:
I'll try to reproduce that and fix if possible.

dycu:
@bryphe do you think that it can be windows specific? I'm trying to run it on mac and it does not crash in any case.

bryphe:
Hey @dycu !

> I'll try to reproduce that and fix if possible.

Thanks a lot for the help! 🎉 

> @bryphe do you think that it can be windows specific? I'm trying to run it on mac and it does not crash in any case.

It's possible. The crash occurred when switching buffers after doing some searching. I didn't have a reliable repro, though - sometimes it would happen on the first buffer switch, and sometimes after working for a while. Some cases I would try:
- Search in a buffer, go to a new buffer
- Search in buffer1, use a new search query in buffer2, switch back to buffer1 and modify

I also noticed a couple other issues I didn't get a chance to investigate more deeply:
- In some cases, I'd get highlights re-appearing after running `:nohl` and updating the buffer.
- For large files, there'd be noticeable perf degradation with large searches - for example, opening a file like this: https://github.com/onivim/libvim/blob/master/src/ex_cmds.c - searching for `/a`, and then modified the buffer.

It's totally possible the bug is in the `libvim` side! We _should_ be able to always call `Vim.Search.getHighlights` and trust that it isn't going to crash, so it may unfortunately be a bug deeper in [`libvim`](https://github.com/onivim/libvim) or [`reason-libvim`](https://github.com/onivim/reason-libvim). 

One thing we could try is to limit when we call `Vim.Search.getHighlights` - as a first step, we could keep track of if the buffer has an active search, and _only_ call it if it does have an active search query.

In terms of addressing the large-file-perf, we could actually run `Vim.Search.getHighlights` only on the changed region (it takes an optional start/end line), and shift the other search highlight ranges in the un-modified part.

PR #867, opened 2019-10-18
Initial Completion UX

bryphe:
Very basic, simple completion UX, hooked up to merlin for now

basic completion

> __NOTE:__ This doesn't have general support for language servers yet - this is to test the UX. But hopefully we'll have it wired up to the VSCode exthost soon! 🤞 

__TODO:__
- [x] Hook up theme colors
- [x] Add icons in completion menu
- [x] Add details
- [x] Fix hang found by @szbergeron (`Tab.re` -> line 8 -> `o` -> `o`)

__NEXT STEPS:__
- Fuzzy-filtering / sorting
- Selection (only show details on selection)


src/Model/Completions.re
```diff
@@ -21,101 +21,98 @@ let default: t = {
...
-  | [_hd, ..._tail] => true
-  | _ => false
-  }
-  }
+  | Some(_) =>
+    switch (v.filteredCompletions) {
+    | [_hd, ..._tail] => true
+    | _ => false
+    }
+  };
```
glennsl: I think this can be reduced to `v.meet != None && v.filteredCompletions != []`

src/Merlin/MerlinProtocolConverter.re
```diff
@@ -32,3 +33,16 @@ let toModelDiagnostics = (errors: MerlinProtocol.errorResult) => {
...
 };
+
+let toModelCompletions = (completions: MerlinProtocol.completionResult) => {
+  let f = (cmp: MerlinProtocol.completionResultItem) => {
+    let ret: Model.Actions.completionItem = {
+      completionKind: CompletionKind.Function,
+      completionLabel: cmp.name,
+      completionDetail: Some(cmp.desc),
+    };
+    ret;
```
glennsl: I've seen this pattern a few times, where a seemingly unnecessary variable is created and then immediately returned. I think I've even removed an instance of it in the menu PR, but I'm starting to wonder if there's some stylistic purpose to this that I don't get?

src/Model/Completions.re
bryphe: Good call!

src/Merlin/MerlinProtocolConverter.re
bryphe: Ah, I was using this just for the type annotation... but I realized there is a more concise syntax recently (trying to break the habit though!). I've also left a few where I had logging in the middle and then forgot to revert back. Wish there was a linter rule that could help me catch it!

I've fixed this instance. Thanks for looking the PR over!

glennsl: > Ah, I was using this just for the type annotation... but I realized there is a more concise syntax recently

To get the record fields into scope you mean? There are lots of ways. In case you don't know all of them:

```rust
let f = x => M.{ foo: x, bar: 42 }

let f = x => { M.foo: "", bar: 42 }

let f = x => {
  open M;
  { foo: "", bar: 42 }
}

let f = x => ({ foo: "", bar: 42 }: M.t)

let f = (x: string): M.t => { foo: "", bar: 42 }
```

bryphe: I did not know about all these! This is excellent - thanks for the tips / reference 👍 

I need to use this:
```
let f = x => M.{ foo: x, bar: 42 }
```

and this:
```
let f = x => ({ foo: x, bar: 42 }: M.t)
```

more often to avoid those unnecessary variables.

src/UI/CompletionsView.re
```diff
@@ -85,17 +155,60 @@ let createElement =
...
+
+            let detailElem =
+              switch (curr.completionDetail) {
+              | None => React.empty
+              | Some(v) when String.length(v) > 0 =>
+                let detailWidth =
+                  Types.EditorFont.measure(~text=v, editorFont)
+                  +. 0.5
+                  |> int_of_float;
+                let detailWidth = min(remainingWidth, detailWidth);
```
glennsl: Very nitty, but you could just add `|> min(remainingWidth)`  here instead of a new binding shadowing the old.

src/UI/CompletionsView.re
```diff
@@ -85,17 +155,60 @@ let createElement =
               |> int_of_float;
             let height =
               Types.EditorFont.getHeight(editorFont) +. 0.5 |> int_of_float;
+            let remainingWidth = 400 - width;
+
+            let detailElem =
+              switch (curr.completionDetail) {
+              | None => React.empty
+              | Some(v) when String.length(v) > 0 =>
```
glennsl: If you call `v` here `text` instead, it's both more descriptive and you can use punning with both `measure(~text, ...)` and `<Text text ... />` later.

src/UI/CompletionsView.re
bryphe: Good catch, clearer than shadowing the name.

src/UI/CompletionsView.re
bryphe: Nice!

PR #875, opened 2019-10-20
Load keybinds file.

CrossR:
We weren't actually loading the `keybindings.json`....which would explain why my changes there did nothing!

Just need to update this to auto reload now as well.


bryphe:
Thanks for adding this, @CrossR !

bryphe:
One thing I'm thinking about though - I wonder how we'll handle upgrade? For example - when we need to change some of the 'default' keybindings or add new commands.

I think what we might want to do is have a few 'sample' key bindings in this file - and then a default set. We'd also want to have a way to remove the default behavior - like VSCode has: https://monosnap.com/file/928VV6lRxslP6LzpdIR83qCGx1BFMo

PR #894, opened 2019-10-24
Menu Renovation - Part 1: Easy pickings

### LINKED ISSUE #219: QuickOpen: Add scrollbar to menu
In Oni1, we had a scrollbar for our Menu / quickopen:
oni1-scroll-bar

Our current QuickOpen experience doesn't have this in Oni2 - would be nice to bring this back!

### LINKED ISSUE #314: Menu does not cycle from first to last item
The menu (`quickopen` and `commandPalette`) cycles on `MenunextItem` from last to first. But does not cycle from first to last on `MenuPreviousItem`.

If this is something that should be added I already can open a PR with the fix.

### LINKED ISSUE #321: Menu (FlatList) viewport doesn't follow the selected item
When there are more items than the Menu/FlatList shows it doesn't follow the `selectedItem`.

You can see here: https://www.youtube.com/watch?v=xPhhNNhfVkM

_I changed my menu config form \<C-N\>/\<C-P\> to \<C-J\>/\<C-K\>_.

I'd like to give it a shot at fixing it.

---

Here is what I believe should be the behavior for the menu

**List shorter than view**:

Current:

current-short-list

Expected? (This is how VS Code menu works)

expected-short-list

**List is longer than view**

Current

current-longer-list

Expected

expected-longer-list

### LINKED ISSUE #469: Suggestion list doesn't scroll when tabbing off end

glennsl:
Split out from #858 

Fixes the first point of #818
Fixes #469 
Fixes #321 
Fixes #314 
Fixes #219

### DIFF
--- a/src/UI/FlatList.re
+++ b/src/UI/FlatList.re
@@ -10,9 +10,25 @@ open Revery_UI_Components;
 
 module Utility = Oni_Core.Utility;
 
+// TODO: Remove after 4.08 upgrade
+module Option = {
+  let map = f =>
+    fun
+    | Some(x) => Some(f(x))
+    | None => None;
+
+  let value = (~default) =>
+    fun
+    | Some(x) => x
+    | None => default;
+
+  let some = x => Some(x);
+};
+
 type renderFunction = int => React.syntheticElement;
 
 module Constants = {
+  let scrollWheelMultiplier = 25;
   let additionalRowsToRender = 1;
   let scrollBarThickness = 6;
   let scrollTrackColor = Color.rgba(0.0, 0.0, 0.0, 0.4);
@@ -84,6 +100,10 @@ let render = (~menuHeight, ~rowHeight, ~count, ~scrollTop, ~renderItem) =>
     indicesToRender |> List.map(itemView) |> List.rev;
   };
 
+type action =
+  | SelectedChanged
+  | SetScrollTop(int);
+
 let createElement =
     (
       ~height as menuHeight,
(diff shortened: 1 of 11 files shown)


glennsl:
I think the CI failure is unrelated to my screwups. "Publish JUnit file" fails with the error "Service unavailable".

bryphe:
Thanks for splitting this out, @glennsl !

> I think the CI failure is unrelated to my screwups. "Publish JUnit file" fails with the error "Service unavailable".

Indeed - I re-ran and its looking good now!

I was trying this PR out, with the new scrollbar - I was getting a runtime exception: 
```
Fatal error: exception Invalid_argument("List.init")
```

It's unfortunately easy to hit on Windows - we should try and find a fix for it prior to this going in:
crash

src/UI/FlatList.re
```diff
@@ -4,106 +4,179 @@
...
+    [];
+  } else {
+    let startRow = scrollTop / rowHeight;
+    let startY = scrollTop mod rowHeight;
+    let rowsToRender =
+      min(
+        menuHeight / rowHeight + Constants.additionalRowsToRender,
+        count - startRow,
+      );
+    let indicesToRender = List.init(rowsToRender, i => i + startRow);
```
bryphe: I think the crash I saw is related to this. If I log out the `rowsToRender` value, I see that sometimes it can become negative:
```
[INFO] rowsToRender: -3
Fatal error: exception Invalid_argument("List.init")
```

Might be sufficient to set it up so that if `rowsToRender` is less than or equal to zero, we just return an empty list?

glennsl: Ah, right, there's some flawed assumptions here, such as the selected index always being valid, which doesn't hold because the items array might change without the selection being adjusted. I also think `selected` might be negative to indicate no selection, and that this causes problems somewhere.

It's an easy fix, but there might be a few more of these to weed out, that I've fixed in a later commit in the original PR.

glennsl: Should be fixed now. Also backported the other fixes I could find in FlatList

src/UI/FlatList.re
```diff
@@ -4,106 +4,207 @@
...
+      bottom(0),
+      width(Constants.scrollBarThickness),
+    ];
+
+  let viewport = (~isScrollbarVisible) =>
+    Style.[
+      position(`Absolute),
+      top(0),
+      left(0),
+      right(isScrollbarVisible ? 0 : Constants.scrollBarThickness),
```
bryphe: I spent a little time troubleshooting the 'hover' issue - if we put `bottom(0)` here, it seems to fix it. It looks like this was treated as having a smaller height than expected, causing it and the children to not get the expected mouse events.

glennsl: Huh, too simple... But awesome!

PR #897, opened 2019-10-24
Menu Renovation - Part 2: Subscribption Infrastructure

glennsl:
Split out from #858. Based (ie. depends) on #894

This (mostly) just adds the subscription infrastructure, but does not actually employ it for anything. That comes later.


bryphe:
Awesome, thank you for splitting this out @glennsl - looks great! 🎉 

I think the 'subscription' concept was the missing architectural piece for us, so I'm looking forward to exploring this more.

PR #899, opened 2019-10-24
CI: Upgrade to node 10.x

bryphe:
...and trying to install it again is giving us errors. Unfortunately the default node 6.x isn't sufficent, though.

Related issues:
https://github.com/microsoft/azure-pipelines-tasks/issues/11627
https://github.com/microsoft/azure-pipelines-tasks/issues/11626
https://github.com/nodejs/build/issues/1993

### DIFF
--- a/.ci/use-node.yml
+++ b/.ci/use-node.yml
@@ -1,7 +1,7 @@
 steps:
   - task: NodeTool@0
-    displayName: 'Use Node 8.x'
+    displayName: 'Use Node 10.x'
     inputs:
-      versionSpec: '8.x'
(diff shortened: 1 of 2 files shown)


bryphe:
Looks like this isn't needed anymore, once the downloads are fixed in NodeJS.

fanantoxa:
@bryphe Hello.
JFYI Node v12 finally become LTS version few days ago (it pretty stable since development started in Apr 2019)
A lot performance updates and new JS features here.
Maybe make since to update to it since you're on this page already?)
https://nodejs.org/en/about/releases/

PR #905, opened 2019-10-24
Input: 'when' clause parser

bryphe:
Switch to proper handling of 'when' clauses for input - so we can handle constructs like:
- `&&`
- `||`
- `!`
and precedence via `(` and `)`

This is almost ready to go!

A few remaining items:
- [x] Documentation: We should add documentation of the `when` clauses and the current supported values (so when #875 + this lands - we'll have info for customizing!)
- [x] Error handling: We should show a notification when we fail to load these bindings
- [x] Backwards compatibility: We should handle supporting the previous array syntax, too. Needs a test.


src/Input/Expression.re
```diff
@@ -0,0 +1,33 @@
...
+let rec show = (v: t) =>
+  switch (v) {
+  | Variable(s) => Printf.sprintf("Variable(%s)", s)
+  | And(v1, v2) => Printf.sprintf("And(%s, %s)", show(v1), show(v2))
+  | Or(v1, v2) => Printf.sprintf("Or(%s, %s)", show(v1), show(v2))
+  | Not(v) => "!" ++ show(v)
+  };
+
+let evaluate = (v: t, getValue) => {
+  let rec f = v =>
```
glennsl: `loop` `aux` and `helper` are common names for internal recursive function that basically just do the work. `evaluate'` might also be a possibility, but `f` is a bit too non-descriptive I think 😆

src/Core/ConfigurationDefaults.re
```diff
@@ -44,30 +44,30 @@ let getDefaultConfigString = configName =>
...
+        { "key": "<C-P>", "command": "menu.previous", "when": "menuFocus || textInputFocus" },
+        { "key": "<D-N>", "command": "menu.next", "when": "menuFocus || textInputFocus" },
+        { "key": "<D-P>", "command": "menu.previous", "when": "menuFocus || textInputFocus" },
+        { "key": "<C-N>", "command": "wildmenu.next", "when": "commandLineFocus" },
+        { "key": "<C-P>", "command": "wildmenu.previous", "when": "commandLineFocus" },
+        { "key": "<D-N>", "command": "wildmenu.next", "when": "commandLineFocus" },
+        { "key": "<D-P>", "command": "wildmenu.previous", "when": "commandLineFocus" },
+        { "key": "<TAB>", "command": "wildmenu.next", "when": "commandLineFocus" },
+        { "key": "<S-TAB>", "command": "wildmenu.previous", "when": "commandLineFocus" },
+        { "key": "<CR>", "command": "menu.select", "when": "menuFocus || textInputFocus" },
```
glennsl: Nice!

src/Input/Expression.re
bryphe: Thanks for the feedback! Makes sense - I went with `eval` which hopefully is a bit more readable.

Issue #907, opened 2019-10-29
Zoom: setting invalid values causes editor to be unusable


lf-:
![image](https://user-images.githubusercontent.com/6652840/67743476-0136ca00-f9dc-11e9-92fc-f735e1c228ef.png)

If you accidentally set your `ui.zoom` setting to `1` rather than `1.0` or another thing that's parsed as a float, Oni2 turns blue.

This was reproduced on a source-built 7b37658 on Windows 10.

It should also be noted that the location of the config file is not listed in the documentation, so I had to do some recursive `Get-CurrentItem` magic to find it. I intend to write a pull request to rectify this documentation oversight.

lf-:
I'd also like to note that I am very affected by revery-ui/revery#598, which makes oni2 unusable without very substantial squinting at my monitor (and which prompted playing around with the setting; 2.0 has huge UI elements and 1.0 is not usable).

bryphe:
Thanks for all the details @lf- (and for adding the docs - huge help!). I created a PR to fix this so that we don't accept invalid values, and to make our parsing more forgiving.

> I'd also like to note that I am very affected by revery-ui/revery#598, which makes oni2 unusable without very substantial squinting at my monitor 

Interesting... on Windows 8+, we should be grabbing the scale factor for the associated window:

![image](https://user-images.githubusercontent.com/13532591/67760392-8607ff00-fa41-11e9-8874-2f18b093f46f.png)

Would you mind sharing your setting for that? Might give us some clues. Also - if you run with `esy run -f` there will be some logging at the beginning that will show us the scale factor we're picking up from the system - it might be useful to see if that is lining up with your display settings. 

Do you happen to have multiple displays with different scale factors?

PR #908, opened 2019-10-29
Document where the config file is on the config documentation page

lf-:
I was testing zoom values and broke the editor (#907), so I had to figure out where the config file was in order to fix my editor. This pull request documents its location.


bryphe:
Thanks for adding this, @lf- - really appreciate it!

PR #909, opened 2019-10-29
Fix #907 - Don't allow invalid values for ui.zoom setting

### LINKED ISSUE #907: Zoom: setting invalid values causes editor to be unusable

bryphe:
There were two problems:
- We were allowing invalid values like `ui.zoom` of `0`, which made the editor unusable - this is very frustrating because there is no entry point to fix or update the editor.
- We were only accepting float values like `1.0` for the setting... so you could hit the above case by using a value of `1` which is invalid.

This issue updates the store connector to make sure the zoom isn't set to a lower value than `1`, as well as our configuration parsing for float values - there is no reason not to accept integers as well.

### DIFF
--- a/src/Core/ConfigurationParser.re
+++ b/src/Core/ConfigurationParser.re
@@ -20,6 +20,7 @@ let parseInt = json =>
 
 let parseFloat = json =>
   switch (json) {
+  | `Int(v) => float_of_int(v)
   | `Float(v) => v
   | _ => 0.
   };
--- a/src/Store/ConfigurationStoreConnector.re
+++ b/src/Store/ConfigurationStoreConnector.re
@@ -136,7 +136,8 @@ let start =
   let vsync = ref(Revery.Vsync.Immediate);
   let synchronizeConfigurationEffect = configuration =>
     Isolinear.Effect.create(~name="configuration.synchronize", () => {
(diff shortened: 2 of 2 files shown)


CrossR:
>This issue updates the store connector to make sure the zoom isn't set to a lower value than 1

Is there any cases where a user would want to scale the UI down, so a value of less than `1.0` would be valid? I don't know how the zoom deals with values less than 1, so that use case might have to be dealt with else where.

bryphe:
> Is there any cases where a user would want to scale the UI down, so a value of less than 1.0 would be valid? I don't know how the zoom deals with values less than 1, so that use case might have to be dealt with else where.

Good question! I wasn't sure if there was a use-case for this, or not. Unfortunately we don't handle fractional scaling well (currently - hopefully Skia can help us with that!), so I think with that limitation `1.0` is an OK default. But we can see if there is a use case for < 1.0.

PR #914, opened 2019-10-30
Menu Renovation - Part 4: Unification

### LINKED ISSUE #749: Question: A way to navigate command palette with vim (j,k) keys instead of arrow keys?

### LINKED ISSUE #815: Can't navigate around suggestions with arrow keys, only with tab that always goes down.

### LINKED ISSUE #911: QuickOpen Fuzzy Search Unreliable
The current fuzzy search is unreliable, I think due to the streaming of results:

https://gfycat.com/weightyarcticasiaticgreaterfreshwaterclam

As can be seen here, the same search query gives different results when run whlist results are still loading, and then after the results are fully loaded.

The first time, the file I want hasn't been reached yet, so I don't see it, and instead get the second result as the top result. The second time I get the file I actually wanted, as the full file list is available.

The scoring is being applied, as we do get the "best" result as the selected one, its just from a different, non-complete file list. Ideally, as soon as the new files are streamed in and scored, we'd update the full list to include any better matches.

glennsl:
Split out from #858. This is the big one, and probably the final one. It currently compiles and works, but there's still remaining work, bugs and code quality issues to iron out. See below.

Fixes #749 
Fixes #815
Fixes #911

**Remaining work:**
- [x] Fix/Simplify/Unify loading animations
- [x] Generalize MenuJob as FilteringJob
- [x] Add "hard-coded" keybindings instead of using `NotifyKeyPressed`
- [x] Optional: Open file without closing menu if right arrow is pressed when cursor is at the end of input (mimics vscode)

**Known issues:**
- [x] Jerky loading animations.

**Code quality issues:**
- [x]  Terminology: "Focus" vs "Select"
- [x] Terminology: "Menu" -> "Quickmenu" or something else? "Smartmenu"?
- [x] Terminology: `menuCommand` -> `menuItem`
- [x] Terminology: Unified name for "Commandline"/"Wildmenu" Edit: Settled on just "Wildmenu"
- [x] Using up/down arrow "commits" Wildmenu selection, removing other possible completions from the menu. This is caused by using NotifyKeyPressed which also sends the key press to vim.
- ~~Consider abstracting away Ripgrep from RipgrepSubscription as there's no need to hard-code this to a specific technology. Also might be an idea to specialize it as directory search, hard-code the connection to FilterSubscription which then allows hiding the awkward itemStream connection.~~ Edit: Let's wait and see how this will work with workspace search.

**Observations:**
- Ctrl+P menu's listing of files is non-deterministic. Discussion on Discord has suggested that this might be expected behaviour because sorting would disable parallelism.  I think this makes sense for workspace search, but it seems like the wrong trade-off for the Ctrl+P menu. And it would be an easy fix I think, just add `--sort path` to `ripgrep` and use a proper LIFO queue instead of a list for the processing queue. See #818 
- There seems to be quite a bit more "Upshifting into active state" now. Might be because an animation hook is used instead of updating the model and re-rendering everything on `Actions.Tick`. Unsure if this indicates a problem. Would also be nice to have a more powerful logging framework with proper namespacing and filtering abilities.


glennsl:
I believe all that remains now is a couple of minor terminology issues:

- **"Focus" vs "select"** - Both terms are currently overloaded. "Select" is used both for choosing an item, and partly for the currently highlighted item. "Focus" is used partly for the currently highlighted item and for input focus. Using a term like "choose" instead of "select" would be the easiest way to remove this overloading, but as this term is currently exposed to the user through the "menu.select" keybinding/command it would be a breaking change and should be considered carefully to avoid confusion.

- **"Menu" vs "Quickmenu" vs "Smartmenu" vs something else?** - It's currently just called "Menu", which is incredibly vague. I called it "Quickmenu" in #858, after the QuickOpen menus, but don't think that's a great term. It should ideally be more descriptive, and since it's also exposed to the user through the "menu.*" keybindings/command, ought to be considered carefully as well. Also note that "Wildmenu" is a term used internally, as one kind of "Menu", and is of course also an established vim term. We should take care to avoid confusing these terms as well.

bryphe:
Thanks for all the work here, @glennsl ! I just tried the latest branch and it works great 🎉 

Regarding the nomenclature - one constraint that might help us is that, at least for __keybinding commands__ and __theme colors__ - my goal is to have compatibility with VSCode. And they tend to have differentations for these:
```
list.activeSelectionBackground: List/Tree background color for the selected item when the list/tree is active.
list.activeSelectionForeground: List/Tree foreground color for the selected item when the list/tree is active.
list.dropBackground: List/Tree drag and drop background when moving items around using the mouse.
list.focusBackground: List/Tree background color for the focused item when the list/tree is active.
list.focusForeground: List/Tree foreground color for the focused item when the list/tree is active. An active list/tree has keyboard focus, an inactive does not.
list.highlightForeground: List/Tree foreground color of the match highlights when searching inside the list/tree.
list.hoverBackground: List/Tree background when hovering over items using the mouse.
list.hoverForeground: List/Tree foreground when hovering over items using the mouse
```
https://code.visualstudio.com/api/references/theme-color#lists-and-trees

(And parallels for their command set).

So in terms of these: 

> "Select" is used both for choosing an item, and partly for the currently highlighted item.

The VSCode themes seem to use `selection` to refer to a selected item. This is important in the case of _multi-select_ lists / menus - where you can have multiple items in a 'selection'. Whereas 'select' is used to refer to either:
- In a multi-selectable list, adding an item to the 'selection'
- In a non-multi-selectable list, 'choosing' the item and running the associated command

> "Focus" is used partly for the currently highlighted item and for input focus.

IMO, I think it'd reasonable to use this just for input focus. We'd have to consider . a

> Using a term like "choose" instead of "select" would be the easiest way to remove this overloading, but as this term is currently exposed to the user through the "menu.select" keybinding/command it would be a breaking change and should be considered carefully to avoid confusion.

Given the constraint of sticking with VSCode's command conventions - we'll need to stick with `"menu.select"` - so we'll have to figure out another change to get consistent nomenclature.

> "Menu" vs "Quickmenu" vs "Smartmenu" vs something else?

I believe the `menu.*` commands / theme colors apply to the equivalent VSCode concept, so I'd lean towards sticking with `menu`. VSCode calls the Control+P experience 'QuickOpen', which is a type of `menu`, as well as other features like theme picker. So switching seems to take us further away from parity with those concepts.

Hope that helps. Thanks for all your work on this, @glennsl ! The experience has really been leveled up thanks to all you've done!

glennsl:
Thanks for the input @bryphe!

> The VSCode themes seem to use selection to refer to a selected item. This is important in the case of multi-select lists / menus - where you can have multiple items in a 'selection'.

That is a very good point. But in the case of multi-selection `focus` is often used to refer to a specific item in the selection that can be moved with keyboard or pointer input, and will expand or contract the selection relative to the previously focused item (See for example the [HTML Selection interface](), which also has the concept of an `anchor`, referring to the where the selection started). So in a single-selection list, is the "selected" item not effectively just indicating the current input focus?

> Given the constraint of sticking with VSCode's command conventions - we'll need to stick with "menu.select" - so we'll have to figure out another change to get consistent nomenclature.

👍 

> I believe the menu.* commands / theme colors apply to the equivalent VSCode concept, so I'd lean towards sticking with menu. VSCode calls the Control+P experience 'QuickOpen', which is a type of menu, as well as other features like theme picker. So switching seems to take us further away from parity with those concepts.

Does vscode have any `menu.*` commands? I haven't been able to find any.

I think it makes sense for theme colors to apply consistently to all kinds of menus, as it does in vscode on Linux at least. The QuickOpen menu, context menus, menu bar menus and toolbar "overflow" menus are all styled the same on Linux. On macOS everything but QuickOpen uses native menus however.

I also think it makes sense to use `menu.*` for commands that should apply to all kinds of menus. But should that be the case here? VSCode behaves differently on Linux and macOS. On macOS you can use `Ctrl+P`/`N` on both native and QuickOpen menus, but not on Linux.`Ctrl+P´ can be used in the QuickOpen menu to scroll down, but will "select" the "focus"ed item when `Ctrl` is _released_, which is interesting behaviour.

But even if we just use `menu.*` in keybindings/commands and theme colors, we still need to be more specific internally I think, since we're probably going to have at least the same four kinds of menus vscode has.

bryphe:
> So in a single-selection list, is the "selected" item not effectively just indicating the current input focus?

I think this is accurate - and consistent with the actions in VSCode bindings (although, I was wrong about `menu.select` - it doesn't look like there is a binding by default for VSCode, but the equivalent action is `workbench.action.acceptSelectedQuickOpenItem`).
image

For reference - there's a richer set of actions for `list`s (things like the file explorer in VSCode are actually controlled by these `list` actions:

image

> Does vscode have any menu.* commands? I haven't been able to find any.

You're right! I was mistaken. There are a set of commands for quickOpen (`workbench.action.*quickOpen*`) - we should be using those instead 👍 

> I think it makes sense for theme colors to apply consistently to all kinds of menus, as it does in vscode on Linux at least. 

Makes sense - I think this is in parity with the theming convention of VSCode (which has `menu.*` theme colors that apply to all of these, as far as I can tell). Although, there are some separate / additional colors for `menubar.*`: https://code.visualstudio.com/api/references/theme-color#menu-bar-colors

> But even if we just use menu.* in keybindings/commands and theme colors, we still need to be more specific internally I think, since we're probably going to have at least the same four kinds of menus vscode has.

Right, and seems like VSCode is too. Here's an attempt at it (inspired from your suggestion + VSCode's differentiation):
- `QuickOpen` for the `Control+P` experience (and related ones like theme picker) - what we call `Menu` today 😬 
- `ContextMenu` for the mouse-right-click contextual menu
- `MenuBar` for the top-level menu (we don't have this yet)
- `Suggest` - the auto-complete 'menu'
- `OverflowMenu`- we might be able to re-use context menu for this? But this would correspond to the toolbar overflow menu. I don't see any keybindings or styling specific to this type of menu.

glennsl:
> `workbench.action.acceptSelectedQuickOpenItem`

Wow that's a mouthful!

> There are a set of commands for quickOpen (`workbench.action.*quickOpen*`) - we should be using those instead.

Should we also aim to implement all the `list` actions? I think that'll be a part 5 if so.

> `QuickOpen` for the `Control+P` experience (and related ones like theme picker) - what we call `Menu` today.

That would make `WIldmenu` a kind of `QuickOpen` menu. Does that seem intuitive? I'm not entirely sure. Otherwise I think the other suggestions are good.

glennsl:
Looking more closely at the vscode keybindings, it doesn't seem like most `workbench.action.*quickOpen*` commands are bound. Instead it probably uses the `list.*` commands, including `list.select` instead of the mouthful. Not sure why the `workbench.action.*` commands are needed then. Perhaps they're legacy commands and deprecated?

I also see some different terminology in the `when` clauses, where it uses `inFIlesPicker`, `inRecentFilesPicker` and `inEditorsPicker` where I've used `Buffers` and `WorkspaceFiles`. Not idea what `RecentFiles` is since it's not bound and `quickOpen` already order by most recent. Perhaps that's legacy/deprecated too.

glennsl:
Ok, I think this is done!

bryphe:
So stoked to have this in, @glennsl ! Thank you for all the work on this! Very excited to see it land 😎 

> Ctrl+P menu's listing of files is non-deterministic. Discussion on Discord has suggested that this might be expected behaviour because sorting would disable parallelism. I think this makes sense for workspace search, but it seems like the wrong trade-off for the Ctrl+P menu. And it would be an easy fix I think, just add --sort path to ripgrep and use a proper LIFO queue instead of a list for the processing queue. See #818

Thanks for thinking about this! My only concern is that the `--sort` flag makes `ripgrep` much slower - but I'm open to exploring it. There might be ways to improve the experience, for example, run it when a folder is opened so that we have a cached list ready to go (I was thinking about having a separate 'indexer' process - like what SublimeText does). 

> There seems to be quite a bit more "Upshifting into active state" now. Might be because an animation hook is used instead of updating the model and re-rendering everything on Actions.Tick. Unsure if this indicates a problem. 

Appreciate you calling this out. It does seem like there is some state that happens to be changing when the menu is 'idle' - the downside is that this will consume CPU and cause a re-render every frame. We might want to turn off any animations once the search is complete.

> Unsure if this indicates a problem. 

It's a bug we'll want to fix - it could either be an animation going in the background, or the tick function causing the state to change (which would trigger a re-render). But we can track this separately.

> Would also be nice to have a more powerful logging framework with proper namespacing and filtering abilities.

Definitely! Maybe it would make sense to have this in Revery (or even a stand-alone project)

tcoopman:
I've just pulled the latest master and see some strange behaviour, not sure if this is related though.
So this is the problem. I'm opening a file with `ctrl+p` and no navigation/changing modes (basically I can't do anything in the opened file) works at this point for me. To make it work again, I need to open the `ctrl+p` menu again and then close it again. After this, navigation works again.

glennsl:
@tcoopman Sorry about that. Fixed in #937

PR #866, opened 2019-10-18
Experiment - Performance: Run textmate highlighting in OCaml thread

bryphe:

PR #922, opened 2019-11-11
Performance: Async Syntax Highlighting

bryphe:
This is a follow-up to #866 - the issue being that, even though our textmate syntax highlighting ([`reason-textmate`](https://github.com/onivim/reason-textmate) is _faster_ than the JS syntax highlighting strategy when run in isolation: https://github.com/onivim/reason-textmate#performance ) - it's actually _slower_ in wall-clock time when run in the editor.

The problem is that we are 'time-slicing' our syntax highlighting - giving it a budget of 2ms per frame. This means we are running it at 1/8 the potential speed (assuming a frame time of 16ms) - essentially 8x slower.

#866 was an attempt to move this to an OCaml thread to run in parallel, but unfortunately this is where the lack of multicore is problematic - most the of operations are CPU bound (the processing of the regex), so it causes delays in the rendering / responsiveness. When multicore is in place - threading will be a viable option, and we can revisit!

However, without OCaml multicore, we have to go back to running the syntax highlighting in a separate process - this will allow us to still get parallelism. The implementation can be simpler and more performant than our previous node-textmate strategy, which used JSON-RPC, because we can use OCaml's efficient [`Marshal`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Marshal.html) module as the bridge, and we can strongly-type both ends of the protocol in Reason, and leverage the native [`reason-textmate`](https://github.com/onivim/reason-textmate) parsing (and [`reason-tree-sitter`](https://github.com/onivim/reason-tree-sitter). 

__TODO:__
- [x] Send language info to client
- [x] Send theme to client
- [x] Create type for managing the state in `Oni_Syntax_Server`
    - [x] Track buffers / buffer updates in state
    - [x] Move `SyntaxHighlighting.t` there
- [x] Update theme when it changes
- [x] Run syntax highlight job continuously in `Oni_Syntax_Server`
- [x] Create simpler client-side state for managing syntax highlighting - just the tokens
- [x] Wire up 'textmate' strategy again
   - [x] Hook up grammar discovery
   - [x] Add `getUpdatedLines` to job / interface
   - [x] Add `clearUpdatedLines` to job / interface
- [x] Wire up 'treesitter' strategy again
   - [x] Add TreeSitterConverter repository (like GrammarRepository but for tree-sitter)
   - [x] Hook up tree-sitter json discovery
   - [x] Add `getUpdatedLines` to job / interface
   - [x] Add `clearUpdatedLines` to job / interface
- [x] Hook up configuration / configuration changes (to decide if we need to pick up tree-sitter)
- [x] Hook up logic to determine which highlight strategy to use
- [x] Add `rei` interfaces + review and consolidate
- [x] Implement 'shift' logic when adding / removing lines
- ~Debug issue where new characters are not highlighted with previous token~ (this was an pre-existing issue, will test further to explore a fix in separate PR)
- [x] Add integration test for syntax highlighting 
   - [x] Windows: Fix crash when spinning up syntax server
- ~Turn on 'C' tree-sitter?~ (will defer to a separate change)
- [x] Keep track of ranges that were updated, send those up, and clear them
   - [x] Store updated ranges as part of job
   - [x] Add 'getPendingRanges'
   - [x] Call 'clearPendingRanges'
- [x] Add health check for syntax server
- [x] Test release builds
  - [x] OSX
  - [x] Windows
  - [x] Linux

... later - keyword extraction would be nice to have!


src/Syntax_Client/Oni_Syntax_Client.re
```diff
@@ -0,0 +1,114 @@
+/*
+ Syntax client
+ */
+
+module Core = Oni_Core;
+
+open Oni_Syntax;
+
+module LogClient = (val Core.Log.withNamespace("Syntax.client"));
+module LogServer = (val Core.Log.withNamespace("Syntax.server"));
```
glennsl: I've so far prefixed every namespace with `Oni2.`, except for the revery log which is `Revery` right now. The idea is that the first part should identify the application or library it belongs to. That's really the only convention I've used so far.

I would also call the log modules `ClientLog` and `ServerLog`, so they're more noun-like than verb-like.

Not set on any of this of course, but that's what I've done elsewhere so far.

src/Syntax_Client/Oni_Syntax_Client.re
```diff
@@ -0,0 +1,114 @@
...
+
+  let shouldClose = ref(false);
+
+  let in_channel = Unix.in_channel_of_descr(stdout);
+  let out_channel = Unix.out_channel_of_descr(stdin);
+
+  let waitThread =
+    Thread.create(
+      () => {
+        let _ = Unix.waitpid([], pid);
```
glennsl: Ignoring values without type annotations produce a warning in 4.08.

src/Syntax_Server/State.re
```diff
@@ -0,0 +1,151 @@
...
+let getActiveHighlighters = (v: t) => {
+  getVisibleHighlighters(v)
+  |> List.filter(hl => NativeSyntaxHighlights.anyPendingWork(hl));
+};
+
+let anyPendingWork = (v: t) => {
+  switch (getActiveHighlighters(v)) {
+  | [] => false
+  | _ => true
+  };
```
glennsl: `getActiveHighlighters(v) != []`

src/Syntax_Server/State.re
```diff
@@ -0,0 +1,151 @@
...
+              Some(NativeSyntaxHighlights.updateVisibleRanges(ranges, v))
+            },
+          prev,
+        );
+      },
+      v.highlightsMap,
+      buffers,
+    );
+
+  {...v, visibleBuffers, highlightsMap};
```
glennsl: There's a lot of `v`s shadowing each other here.

src/Syntax_Server/Oni_Syntax_Server.re
```diff
@@ -0,0 +1,126 @@
...
+      () => {
+        while (isRunning^) {
+          queue(Log("Waiting for message from client..."));
+          let msg: Oni_Syntax.Protocol.ClientToServer.t =
+            Marshal.from_channel(Stdlib.stdin);
+          queue(Log("Got message from client!"));
+          switch (msg) {
+          | exception _ => queue(Exception)
+          | msg => queue(Message(msg))
+          };
```
glennsl: This looks odd. `msg` is just a value. It can't throw an exception. Did you mean to use on `Marshal.from_channel`?

src/Syntax_Client/Oni_Syntax_Client.re
bryphe: Sounds good! I'll stick with the convention you've started 👍

src/Syntax_Client/Oni_Syntax_Client.re
bryphe: Good point - I'm pulling the return code now, and still ignoring the status, but annotated it.

src/Core/Utility.re
```diff
@@ -263,6 +263,12 @@ module Option = {
     | Some(x) => f(x)
     | None => None;
 
+  let flatten =
+    fun
+    | Some(Some(x)) => Some(x)
+    | Some(None) => None
+    | None => None;
```
glennsl: This is equivalent to 
```
    | Some(x) => x
    | None => None;
```

Because the last branch is `None`, `x` has to be an `option(_)`.

src/Syntax_Server/Oni_Syntax_Server.re
bryphe: Ouch, nice catch @glennsl ! My intention was to have a `try...catch` around the Marshal call, since that can potentially fail (I believe) and at least log it.

Fixed by wrapping in a `try` now

src/Core/Utility.re
bryphe: Makes sense - nice!

src/Syntax_Server/State.re
bryphe: Good point... Renamed the top-level `v`'s to `state` and used `fun` to get rid of some extraneous names.

PR #929, opened 2019-11-13
Upgrade to latest revery (870ae23)

glennsl:
With the brisk-reconciler upgrade, and animation/time revamp.

Note that this depends on https://github.com/revery-ui/revery/pull/641


glennsl:
Not sure why CI fails. Test and bench both run and everything has been checked in.

bryphe:
Wow, thanks for updating all the components to use the new API, @glennsl ! 😍 Awesome!

> Not sure why CI fails. Test and bench both run and everything has been checked in.

Ahh, just checked - there was a misleading message - the `integrationtest.esy.lock` was the one not up-to-date, but the job description in CI said `test.esy.lock`... so very confusing. I fixed it, and updated the guidance to run `./update-lockfiles.sh`, since that updates all of them in one go.

Hope it passes now 🤞

glennsl:
`QuickOpenEventuallyCompletesTest` seems a bit flaky now. It sometimes succeeds, and sometimes not. Not entirely sure why upgrading revery would have an effect on this.

bryphe:
> QuickOpenEventuallyCompletesTest seems a bit flaky now. It sometimes succeeds, and sometimes not. Not entirely sure why upgrading revery would have an effect on this.

I wonder if this is due to some change from https://github.com/onivim/oni2/pull/914 ? It seems more likely to have an impact on this test vs the Revery upgrade 🤔  It might be a timing condition (the quick menu completes w/o setting 'loading' to true, for example).

It looks like the latest run failed at the `LanguageReasonMerlinDiagnosticsTest`. I'll pull this flaky test out for now, since I plan on switching to testing the VSCode extension for diagnostics instead.

glennsl:
> I wonder if this is due to some change from #914 ?

That certainly seems more likely, but I ran those tests a lot of times doing the final fixes on that PR yesterday without coming across this problem.

bryphe:
Got a green build! 🎉 

Thank you @glennsl for doing the work to upgrade all the components across Onivim!

PR #930, opened 2019-11-13
Exthost - Language Features: Initial completion integration

bryphe:
This is an initial hook-up to VSCode's language features in the extension host - to pick up the completion / suggest functionality.

A few pieces are needed to make this work:
- The extension host sends us a `$registerSuggestSupport` message, which lets us know which documents the suggestion provider applies to, as well as other context (like trigger characters)
- When we want completion, we call `ExtHostLanguageFeatures:$provideCompletionItems` with the id for the registered suggestion provider (and some other context - URI, position, etc). The extension host will reply with the completions available.

I've added a special file-type for development `oni-dev` which has a completion provider to exercise this.

Work remaining on this PR:
- [x] Store the information about the registered support providers in a LanguageFeatures data structure
- [x] On the completion meet event, figure out which completion providers apply, and call `provideCompletionItems` for them. Add them to the active completion context.


src/Core/Utility.re
```diff
@@ -230,3 +230,27 @@ let ranges = indices =>
...
+  let some = x => Some(x);
+
+  let bind = f =>
+    fun
+    | Some(x) =>
+      switch (f(x)) {
+      | None => None
+      | Some(_) as v => v
+      }
+    | None => None;
```
glennsl: Beware that the argument order here is the wrong way around compared to `Stdlib.Option.bind`. It's weird, but apparently that's the convention for `bind`. A `flat_map` function was initially also included with the arguments in the"right" order, but that was taken out because it traditionally originated on the `List` type or something, and it would apparently be weird to have it in `Option` and not in `List`.

src/Model/Selectors.re
```diff
@@ -58,6 +58,20 @@ let getActiveBuffer = (state: State.t) => {
...
+  switch (buffer) {
+  | None => ()
+  | Some(buf) =>
+    let fileType = Buffer.getFileType(buf);
+    switch (fileType) {
+    | None => ()
+    | Some(ft) => f(buf, ft)
+    };
+  };
+};
```
glennsl: `Option.map` + `Option.iter` would shorten this significantly.

src/Core/Utility.re
bryphe: That's good to know! I didn't realize that. I would've expected this is the 'right way' because it is so amenable for piping.

I guess, once we upgrade to 4.08, and pick up the `Stdlib.Option.bind`, we'd switch these to use the monadic-let anyway?

glennsl: Yeah, that would be nice. I hope I can get used to them quickly though, as they're not exactly intuitive. Maybe Reason will provide a better syntax for it.

src/Model/Selectors.re
bryphe: Good idea! I added `Option.iter` 👍 

One thing that was a bit awkward is we needed to keep the `buf` around along with the filetype `ft` - I'm not sure if I did it in the cleanest way.

bryphe: I have another change that depends on this, so I'll bring it in... but if you happen to have feedback I'll incorporate in a separate PR. Thanks for the suggestion @glennsl !

PR #933, opened 2019-11-14
[WIP] fix(language-support/#782): synchronize vim filetype 

### LINKED ISSUE #782: Hook up Vim filetypes for Syntax
I've discussed this with a few people, but it would be nice if we could hook up the vim filetype commands to Oni2.

That is:
 - User issues a `:set syntax=XXXX` command. Would probably also make sense to include the `ft`/`filetype` based commands too.
 - `libvim` sends a notification of that event.
 - We match that against the filetypes in the `textmate` grammars / `tree-sitter` and update the syntax highlights accordingly.

Part of this is going to be needed / useful in just the command pallete, to just say "Swap the syntax HL of this file to X", so that would be a useful first step. Then we can look at hooking up the notification and lookup.

There is probably something I'm missing here and I'm only assuming that a notification like that is possible, though there are `autocmds` for them all, so I'd assume it would be. (As an aside, do we have / is there any value in adding a generic way of hooking up to `autocmds`? Though I guess that would be a nightmare to add assumedly).

bryphe:
This adds support for `set filetype=<some filetype>`, which Onivim2 listens to and then will force the buffer to be set to the specified filetype. The available filetypes are contingent on the VSCode extensions included at runtime.

__TODO:__
- [ ] Hook up `:set filetype` to show a menu that allows for selection from list of filetypes
- [ ] Hook up syntax highlighting to respond to buffer update and re-highlight the file
- [ ] Hook up exthost to respond to buffer update and send appropriate notification (which notification is needed to adjust filetype of document)?

This fixes #782 , although it doesn't support `set syntax` - we don't have a concept of differentiating between syntax highlighting and filetype for the purpose of icons/language services at this time.


bryphe:
Superseded by #2303

PR #937, opened 2019-11-16
Fix quickmenu retaining focus after being closed

glennsl:
This fixes a bug introduced in #914 where closing the menu in any way other than pressing `Esc` would retain focus, and therefore not forward input to `libvim`. This is a known bug in revery-ui/revery#412, which had a workaround that was only partly removed and not replaced.


glennsl:
Merging this immediately since it's just a bugfix

bryphe:
Thanks for catching this and the quick fix, @glennsl !

PR #944, opened 2019-11-19
Dependency: revery -> fed5405

### LINKED ISSUE #942: Performance: Slow redraw on editor surface
Profiling scrolling through the editor surface on OSX - I saw that rendering was slower than expected (the `draw` call was taking around `24ms` on average - it should be taking less <`8ms` ideally).

The graph showed the following:

image

The trace showed that both the `Revery__Draw__Text_render` and `Revery_Draw__Shapes__drawRect_inner` methods were very expensive, primarily due to the matrix multiplications via `caml_mat4_multiply`.

We are using significantly more matrix multiplications than we strictly need to in Revery, and most of the transformations used in those paths are simple (translation + scaling) that could be done more efficiently.

bryphe:
Pick up `reason-gl-matrix` updates, fixes #942


bryphe:
Superseded by #951

PR #957, opened 2019-11-20
Exthost: Integrate css-language-features plugin

bryphe:
This is a baby step into language features, adding a test case for diagnostics in CSS, and hooking up completion from the CSS-language-features plugin.

__TODO:__
- [x] Fix the DocumentSelector parsing, instead of hardcoding it
- [x] Merge and fix up test case for CSS
- [x] Investigate why completion isn't working in the plugin

(getting errors like this):
```
[INFO] [ExtHost]:   Message: Request textDocument/completion failed with message: Cannot read property 'uri' of undefined
  Code: -32603 

[INFO] [ExtHost]: [Error - 5:07:45 PM] Request textDocument/completion failed
```

This also sets up an installation pipeline to bring in dependencies for the JS extensions - note that, if you are building from source, you'll need to run `node install-node-deps.js` from the root (build instructions have been updated).

I'd like to find a better way to bundle the extensions, but for now, I've included the built sources from the `vscode-exthost` repo for the `css-language-features` extension and vendored it here.


src/Core/Uri.re
```diff
@@ -31,7 +31,8 @@ let fromMemory = (path: string) => {scheme: Scheme.memory, path};
 let fromPath = (path: string) => {scheme: Scheme.file, path};
 
 let toString = (uri: t) => {
-  Scheme.toString(uri.scheme) ++ "://" ++ uri.path;
+  //Scheme.toString(uri.scheme) ++ "://" ++ uri.path;
```
bryphe: This looks like it should be reverted

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -181,6 +196,8 @@ let start = (extensions, setup: Core.Setup.t) => {
     let providers =
       LanguageFeatures.getSuggestProviders(fileType, languageFeatures^);
 
+    print_endline("FOUND " ++ string_of_int(List.length(providers)) ++ " suggest providers for filetype: " ++ fileType);
```
bryphe: Remove

src/Extensions/DocumentSelector.re
```diff
@@ -11,7 +11,15 @@ type item = {language: string};
...
 let matches = (v: t, fileType: string) => {
-  List.exists(item => item.language == fileType, v);
+  List.exists(item => String.equal(item.language, fileType), v);
+};
+
+let toString = (v: t) => {
+  "DocumentSelector("++ 
+  List.fold_left((prev, curr) => {
+    prev ++ "," ++ curr.language
+    }, "", v)
```
glennsl: This would be simpler as `List.map(x => x.language) |> String.concat(",")` I think.

Also, the convention is to use `acc` (short for "accumulator") for `prev` here. `prev` seems to me to imply that it has the same type as `curr`.

src/Extensions/ExtHostClient.re
```diff
@@ -66,14 +67,29 @@ let start =
...
     | (s, m, _a) =>
-      Log.error("Unhandled message - " ++ s ++ ":" ++ m ++ " | ");
+      Log.error(
+        "Unhandled message - "
+        ++ s
+        ++ ":"
+        ++ m
+        ++ " | "
+        ++ Yojson.Safe.to_string(`List(_a)),
+      );
```
glennsl: Sometimes I think `refmt` might be just a little bit silly (but oh my is it better than it was a bit over a year ago). I think this would look much more readable with `Printf.sprintf`.

src/Extensions/ExtHostProtocol.re
```diff
@@ -280,8 +280,19 @@ module DiagnosticsCollection = {
...
-        List.map(Diagnostics.of_yojson, perFileDiagnostics)
-        |> Utility.filterMap(Utility.resultToOption);
+        List.map(Diagnostics.of_yojson, perFileDiagnostics);
+
+      List.iter(
+        v =>
+          switch (v) {
+          | Ok(_) => ()
+          | Error(msg) => print_endline("ERROR: " ++ msg)
+          },
```
glennsl: This could be shortened a bit using `fun | ...`. That's what it's really for, I think. In 4.08 I think you can also use `Result.iter_error`.

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -121,13 +121,28 @@ let start = (extensions, setup: Core.Setup.t) => {
...
+    switch (fileType) {
+    | None => ()
+    | Some(ft) =>
+      switch (Hashtbl.find_opt(activatedFileTypes, ft)) {
+      | None =>
+        ExtHostClient.activateByEvent("onLanguage:" ++ ft, extHostClient);
+        Hashtbl.add(activatedFileTypes, ft, true);
+      | Some(_) => ()
+      }
+    };
```
glennsl: `Option.bind` could flatten this a bit

bryphe:
Found the first issue with completion - we were failing with an error: ```
[ERROR] Unhandled message - MainThreadOutputService:$reveal | [{},true]
APPEND:   Message: Request textDocument/completion failed with message: Cannot read property 'uri' of undefined
  Code: -32603 

[INFO] [ExtHost]:   Message: Request textDocument/completion failed with message: Cannot read property 'uri' of undefined
  Code: -32603 
```

The language server was expecting an object with the completion parameters - like: 
```
{"textDocument":{"uri":"file:///Users/bryphe/oni2/test.css"},"position":{"line":0,"character":1},"context":{"triggerKind":1}}
```
but it was getting an array instead:
```
[{"textDocument":{"uri":"file:///Users/bryphe/oni2/test.css"},"position":{"line":0,"character":1},"context":{"triggerKind":1}},null]
```

And of course, trying to check `textDocument.uri` of the array would fail.

There's logic in `vscode-jsonrpc` to special-case the situation where there is an array passed, but the last parameter is a `CancellationToken`: https://github.com/microsoft/vscode-languageserver-node/blob/dd78cba181136e057cfcc4bb7bc17c7c78877648/jsonrpc/src/main.ts#L1050

in that case - it would extract the parameters out.

The problem was - we weren't sending a cancellation token, we were just sending null. So we'll have to figure out how to serialize this token for the language server.

 it turns out that the vscode-language-client / vscode-language-server / vscode-jsonrpc have some

File: src/Extensions/DocumentSelector.re (hunk 3 above)
Comment (bryphe) [reply src/Extensions/DocumentSelector.re]: Good call! Makes sense about `acc` vs `prev` - I'll switch to that convention.

File: src/Extensions/ExtHostProtocol.re (hunk 5 above)
Comment (bryphe) [reply src/Extensions/ExtHostProtocol.re]: Oh, ya, good catch - I hadn't used the `fun` shorthand much but saw it in a few of your PRs. Indeed, it's much more streamlined:
```
v => switch(v) {
| ...
}
```
to: 
```
fun
| ....
```

I switched to use `fun`, but the `Result.iter_error` would be even more concise - might be worth bringing in a polyfilled `Result` module like we did for `Option`?

src/Extensions/ExtHostClient.re
bryphe: Definitely looks better with `Printf.sprintf` - thank you!

src/Store/ExtensionClientStoreConnector.re
bryphe: Good point! I used `Option.iter` and added an `Option.iter_none`

PR #964, opened 2019-11-20
Dependency: reason libvim -> d9a3812

bryphe:
Pick up improvements in auto-closing pairs from `reason-libvim`:


bryphe:
#965 has a newer reason-libvim

Issue #943, opened 2019-11-18
Workspace Search


glennsl:
We 'bout to get search! But how?

Here's a rough outline of the (current) plan:

- [ ] Start with a basic UI similar to the search panel in vscode
- [ ] Add first-class keyboard navigation
- [ ] Implement all the bells and whistles vscode has
- [ ] Somehow make the the search, or search results open in an ordinary buffer, to be positioned and moved around in the layout as desired.

Copying vscode is relatively straight-forward, so while other feedback is of course also welcome, my only questions right now are:

- How should the keyboard navigation work? Also consider how this will work if there are multiple search buffers open, for example. I understand vim sets some precedence and that @bryphe has some ideas.
- How should the "search buffer" work? Should there just be a Ctrl+P command to open the buffer, and then have the entire search UI in the buffer? Or should the UI basically be in the command palette? Or should it hijack/extend one or more vim commands?

Progress will be tracked in https://github.com/onivim/oni2/projects/9

bryphe:
So excited for this! It's a huge gap today 😄 I'll be very happy when I no longer need to leave Onivim 2 to `git grep` 🎉 Thank you for thinking about this @glennsl !

> How should the keyboard navigation work? Also consider how this will work if there are multiple search buffers open, for example. I understand vim sets some precedence here and that @bryphe has some ideas as well.

I've been thinking a bit about this - there are two pieces we need:

__Piece 1:__ The ability to focus the 'sidebar' or other panes. Ideally, we could use the `<C-w>h` shortcut to move window-left into the sidebar, and give the sidebar focus. Our current window manager only knows about focusing editors - so we'll need to augment it to know about giving focus to the sidebar. 

IE, this `moveCore` method in `WindowManager`: https://github.com/onivim/oni2/blob/b3fb1c61dc82b91be3acfa8347a8c3c01126b57c/src/Model/WindowManager.re#L105 only knows about moving between window splits, not the sidebar / activity bar (the `docks`).

__Piece 2:__ For vim keybindings, I anticipate there are multiple places we will want vim keybindings to integrate into GUI actions - the file explorer being one, the search result being another, maybe output logs / notifications like in #849  . 

I'd like to provide a special hook for this, like:

```
let%hook () = Hooks.vimKeyboardNavigation(
   ~active=true, // Set to `true` if we have focus, otherwise false
   ~onMoveUp=(count) => setFocused(state => state.focusedIndex - count),
[...]

PR #971, opened 2019-11-21
Basic Search

glennsl:
Search is now usable. The UI is a very basic drawer popping up along the bottom. No fancy stuff, it's just to get a feel for the layout and to test the functionality. There are still a few things I'd like to get in before I'd call the basic experience complete, however:

- [x] Match highlights
- [x] Put cursor on the match when selecting a result

There's also a fair bit of duplication here, and plenty of opportunity for refactoring. SearchSubscription is very similar to RipgrepSubscription, and could perhaps be merged or have some common code factored out. The entire Ripgrep implementation has also basically been copied before tweaking it and has large parts in common.

2019-11-22-001109_797x623_scrot


bryphe:
Wow, looks amazing @glennsl ! Great start to test the layout and get initial feedback.

I like the design you chose in terms of layout 👍 

>  Match highlights

Awesome! This will be a nice improvement

> Put cursor on the match when selecting a result (any hints for this @bryphe?)

Do you mean opening the file? The main way that we have this set up, right now, is the dispatching a `Model.Actions.OpenFileByPath` action - that gets picked up here: https://github.com/onivim/oni2/blob/349fdbda43ada69c2de71c701a8b06202a2f936f/src/Store/VimStoreConnector.re#L250

The one thing we're missing though, is picking a line number along with it. It might make sense to add an optional argument / parameter for a cursor position - and then we could use the `Vim.Cursor.setPosition` API to put it in the right spot, along with the file opening - we could add that capability to the `openFileByPathEffect` in `VimStoreConnector`.

The other item I'm thinking about in terms of selecting a result is the Vim keyboard accessibility - once we press 'Enter' on search, it'd be great to bring the 'focus' into the selection result - and then we can use a special hook to manage the keyboard navigation. We might have to think about the Revery-focus-management infrastructure to enable this. This can come as a separate change, though!

> There's also a fair bit of duplication here, and plenty of opportunity for refactoring. SearchSubscription is very similar to RipgrepSubscription, and could perhaps be merged or have some common code factored out. The entire Ripgrep implementation has also basically been copied before tweaking it and has large parts in common.

Cool! Something for us to think about as we add more subscriptions.

src/Core/Ripgrep.re
```diff
@@ -2,6 +2,24 @@ open Rench;
 
 module Time = Revery_Core.Time;
 
+// TODO: remove after 4.08
+module List = {
+  include List;
+
+  let filter_map = f => {
```
bryphe: We have a dup of this that we could replace / remove: https://github.com/onivim/oni2/blob/349fdbda43ada69c2de71c701a8b06202a2f936f/src/Core/Utility.re#L94

src/Core/Ripgrep.re
```diff
@@ -14,7 +32,60 @@ let getCompletedCount = () => _ripGrepCompletedCount^;
...
+  type t = {
+    file: string,
+    text: string,
+    lineNumber: int,
+    charStart: int,
+    charEnd: int,
+  };
+
+  let fromJsonString = str => {
+    Yojson.Basic.Util.(
```
bryphe: Just curious - when do you prefer `Yojson.Basic` to `Yojson.Safe` - I'm not very aware of the differences?

src/Core/Ripgrep.re
```diff
@@ -14,7 +32,60 @@ let getCompletedCount = () => _ripGrepCompletedCount^;
...
+          let data = member("data", json);
+
+          let submatches = data |> member("submatches") |> to_list;
+
+          let matches =
+            submatches
+            |> List.map(submatch =>
+                 {
+                   file:
+                     data |> member("path") |> member("text") |> to_string,
```
bryphe: Cool! I think I've duplicated some of these Util methods like `to_string` before - good to know for the future.

src/Store/StoreThread.re
```diff
@@ -199,6 +210,8 @@ let start =
     // TODO: Wire this up properly
     let quickmenuSubs = quickmenuSubscriptionsUpdater(newState);
```
bryphe: As we add more of these, we might want to add first-class support subscription like in TEA

src/UI/Dock.re
```diff
@@ -28,5 +33,12 @@ let make = (~state: State.t, ()) => {
     <Clickable onClick={toggleExplorer(state)} style=button>
       <FontIcon backgroundColor=bg color=Colors.white icon=FontAwesome.file />
     </Clickable>
+    <Clickable onClick={toggleSearch(state)} style=button>
```
bryphe: 🎉 Nice to have this actually do something, now!

src/UI/SearchPane.re
```diff
@@ -0,0 +1,149 @@
...
+      onClick={() =>
+        GlobalContext.current().dispatch(SearchSelectResult(match))
+      }>
+      <View
+        style={Styles.result(~theme, ~isHovered=hovered == i)}
+        onMouseOver
+        onMouseOut>
+        <Text
+          style={Styles.fileText(~font, ~theme)}
+          text={Printf.sprintf(
```
bryphe: One thing I noticed is that, for very long lines, there was a slowdown - we might want to cap the length of the text (or perhaps Revery could do a better job of not rendering text past the window / current clipping region..) 

I was able to reproduce this by searching for `testing` - it seems like some of the yarn lines are very long and hit this case:
image

src/Store/SearchStoreConnector.re
```diff
@@ -0,0 +1,83 @@
...
+      }
+
+    | SearchStart => {...state, query: state.queryInput,hits: []}
+
+    | SearchUpdate(items) => {...state, hits: state.hits @ items}
+
+    // | SearchComplete
+
+    | SearchSelectResult(match) =>
+      print_endline("!! SELECT: " ++ match.file);
```
bryphe: It seems like here is where we could `dispatch(OpenFileByPath(match.file, ...))` and maybe hook up the line number, too!

glennsl:
> Do you mean opening the file? The main way that we have this set up, right now, is the dispatching a Model.Actions.OpenFileByPath action

I already had that implemented actually, but apparently I forgot to check it in before pushing :neutral_face: 

> It might make sense to add an optional argument / parameter for a cursor position - and then we could use the Vim.Cursor.setPosition API to put it in the right spot, along with the file opening - we could add that capability to the openFileByPathEffect in VimStoreConnector.

Yeah, that's exactly what I was looking for. Thank you :slightly_smiling_face: 

>  once we press 'Enter' on search, it'd be great to bring the 'focus' into the selection result - and then we can use a special hook to manage the keyboard navigation. 

Not sure what you mean here by "selection result" and "special hook". My thinking has been that after a selection has been selected with 'Enter' you should be able to follow up right away with "cw" for example, to replace the matched word.

src/Core/Ripgrep.re
glennsl: I wasn't even aware of it until you mentioned it just now. Or rather, I think I've seen it mentioned before, but had no awareness of the difference between that and anything else. But it seems like `Yojson.Safe` is a superset of `Yojson.Basic` with some "extended" and non-standard primitives. How that makes it "safe" I don't know, but it seems to offer some more convenience if both producer and consumer is using OCaml, and `Yojson.Safe` in particular.

src/Core/Ripgrep.re
glennsl: I'm more fond of the more declarative combinator-style API of Elm's ['Json.Decode'](https://package.elm-lang.org/packages/elm/json/1.1.3/) or [`bs-json`](https://github.com/glennsl/bs-json), but this isn't too far away from that.

It's a nice API and it's usually better to decode explicitly like this than to use 'deriving', I think, since this decodes into exactly what you need rather than into some intermediate structure that just ends up being more boilerplate. Or worse, you end up molding your application to fit external data structures rather than using the data structures that are appropriate for your application needs, which tends to lead to all sorts of trouble later.

src/UI/SearchPane.re
glennsl: Ah, yeah I noticed this during some earlier testing, before using FlatList, where just 500 results slowed it down to ~1fps, which I thought it should have handled better. Good idea to just cap the length!

src/Core/Ripgrep.re
glennsl: I can move this into `Utility` like you've done with `Option` somewhere.

src/UI/SearchPane.re
glennsl: Is there any way we can calculate a reasonable maximum number of characters based on font size and available width? Perhaps we can estimate an average character width, divide the available width by that and multiply by 1.5 or 2 for a good safety margin?

glennsl: Figured out on Discord I think. We can just use `measuredWidth` from `EditorFont.t`.

gitbugr:
That's a lot of empty space
Screenshot 2019-11-22 at 23 39 41

Unless there are plans to make use of it, wouldn't it be better to stack the text box horizontally?
In neovim, I use Ag (fzf.vim) in a floating window which is pretty nice to work with

szbergeron:
I kind of agree with @gitbugr, since some projects I work on are really annoyingly large the paths can actually struggle to fit on a half width fuzzy window. In the use case of having oni as one half of a screen then the extra space used by the search bar becomes somewhat consequential. I'll admit it does look nicer having it off to the side IMO. Would some responsive layout be feasible or would that take a while to implement?

glennsl:
> Unless there are plans to make use of it...

There are. We need to fit in three more inputs, one for the text to replace and two for globs to include and exclude files, a bunch of buttons for search options and a "replace all" button.

>  In neovim, I use Ag (fzf.vim) in a floating window which is pretty nice to work with

Can you illustrate? I'm not familiar with it.

> ...some projects I work on are really annoyingly large the paths can actually struggle to fit on a half width fuzzy window.

What are the relevant parts of those paths? We could do like vscode and put the filename first, followed by the path, for example:

vscode search

> Would some responsive layout be feasible or would that take a while to implement?

We could, and I don't think it would take all that much effort to implement. Over in https://github.com/onivim/oni2/issues/943#issuecomment-557314995 I've also mused about having the ability to pop the search results out into a buffer that can be arranged as you please, which would also leave plenty of room at the bottom for the query UI.

gitbugr:
@glennsl 

> > In neovim, I use Ag (fzf.vim) in a floating window which is pretty nice to work with
> 
> Can you illustrate? I'm not familiar with it.

sure:
Screenshot 2019-11-23 at 11 31 45

also a note about the tab-selection aspect of this - when it opens the quickfix menu, this suffices as my search & replace tool, in fact, it's more powerful. I can create a macro which will jump between the files and do more than just replacing strings.

output

The only thing I don't have a solution for is explicitly excluding types of files.

glennsl:
Thank you @gitbugr! Also sorry, I should have asked this of you earlier: Could you post this in #943? This might be merged as-is, but that certainly doesn't mean the UI would be set in stone and it would be a shame to lose this part of the discussion.

I like the idea of selecting a "work set" of results, but also have some concerns: It overlaps quite a bit with the quick open UI, it would seem to require an indexer to be sufficiently responsive in larger projects, and the UI would need to be beefed up quite a bit to be accessible. But these all seem solvable. For example by adding an indexer, replacing Quick Open with this and having a separate and more elaborate UI for accessibility and advanced use cases.

glennsl:
I've added highlighting and a smart-ish snippet extraction algorithm (that could definitely be smarter) that tries to cut the text such that the highlight remains visible. There are still some issues with estimation of the max number of characters, because the editor font is too big and remeasuring it after adjusting is a massive pain, so it currently just estimates based on proportional difference in font sizes. The snippet algorithm could also definitely be smarter.

search-snippets

Known issues:
- [ ] Font measure estimation is off
- [ ] Snippet extraction algorithm is dumb
- [ ] Cursor moves to the match only if the editor is already open and selected
- [ ] Editor does not get focus when a match is selected

There's also a still bit of refactoring that could be done

src/Core/Utility.re
```diff
@@ -332,14 +332,23 @@ module StringUtil = {
 
       let (offset, length) =
         if (matchLength >= maxLength) {
-          // match is lonegr than allowed
-          (charStart, maxLength);
+          (
+            // match is lonegr than allowed
+            charStart,
+            maxLength,
+          );
```
glennsl: refmt :boom: :gun:

src/Core/Utility.re
```diff
@@ -332,14 +332,23 @@ module StringUtil = {
...
         } else if (charEnd - indentation > maxLength) {
-          // match ends out of bounds
-          (charEnd - maxLength, maxLength);
+          (
+            // match ends out of bounds
+            charEnd - maxLength,
+            maxLength,
+          );
         } else {
```
glennsl: refmt :bomb: :boom: :gun:

glennsl:
search-snippets

I've fixed font measurement and the snippet extraction algorithm. There's still some overflow however, which is probably from margins and the scrollbar not being taken into account. But even if they were, I think we'd want it to overflow a little since the last character is likely to only be partially visible. I think we need to add some "padding" to the matched text. Which is great, because I _love_ fiddling with the snippet algorithm (/s).

The two other remaining issues I'm not sure how to go about fixing and could use some help. There's also a weird rendering issue where the ellipsis suddenly turns into four dots with wider spacing. It can be seen at the end of the gif above, on the first yarn result. I consider that an unrelated issue that just happens to show up here, however.

gitbugr:
> There's also a weird rendering issue where the ellipsis suddenly turns into four dots with wider spacing.

That appears to be the `.` from the end of the sentence `consumes fewer resources.` and it will render with wider spacing as it breaks the ligature in the font you're using (Fira Code?)

glennsl:
Ah, right I forgot there were ligatures. Good catch @gitbugr

bryphe:
Just sync'd up to the latest - looking great @glennsl ! Love the highlights in place now. Amazing progress! 

And thanks everyone for the feedback in #943 .

>  [ ] Cursor moves to the match only if the editor is already open and selected
>  [ ] Editor does not get focus when a match is selected

I'll take a look at these blockers shortly and give some ideas-

bryphe:
>  [ ] Editor does not get focus when a match is selected

Regarding this blocker - the issue is that our input strategy checks if a Revery UI element has focus. 

If it does, it doesn't 'pass-through' the input to the Vim layer:
https://github.com/onivim/oni2/blob/092487ecdf58541a3f96382f94599accf88fd43f/src/Store/InputStoreConnector.re#L125

I fixed this by removing the focus on the click event, updating the `onClick` in the `LocationListView`: 
```
  let onClick = () => {
    GlobalContext.current().dispatch(
      OpenFileByPath(item.file, None, Some(item.location)),
    );

    Revery.UI.Focus.loseFocus();
  };
```

There may be a similar issue in the file explorer, too.

bryphe:
> [ ] Cursor moves to the match only if the editor is already open and selected

Ah, if we replace this:
```
      switch (location) {
      | Some(Position.{line, character}) =>
        Vim.Cursor.setPosition(Index.toInt1(line), Index.toInt1(character))
      | None => ()
      };
```

with this:
```
      switch (location) {
      | Some(pos) =>
        dispatch(Model.Actions.CursorMove(pos));
        dispatch(Model.Actions.EditorScrollToLine(Index.toInt0(pos.line)));
      | None => ()
      };
```

This is another case where we have a complicated interplay between actions <-> side effects and ordering. What happens is our `Editor` model has a notion of `cursorPosition` - when the `BufferEnter` effect occurs, there is a `vim.synchronizeEditor` effect that tries to apply the current editor state (cursor, scrolls, etc) to vim -> this overrides our setting via `Vim.Cursor.setPosition`. Using the `CursorMove` and `EditorScrollToLine` set the editor state, too, so that when it gets 'synchronized' , it pushes those values to Vim, _and_ keeps them in our store.

I'd like to look at simplifying this - the main reason we have this synchronization logic is to synchronize a few items:
- Cursor position
- 'Top line'
- 'Left line'
- window width / height

I'd really just like for Onivim to own these values though. The window width / height / top line / left line are needed for some Vim commands (like `zz`, `zt`, `zb`, `H`, `L`, `M`, etc) - but these could be a callback to Onivim (a pull instead of a push). Cursor position could be incorporated into the Vim input/command model - we give it a cursor position (or, maybe a list of cursor positions for multi-cursor) along with an input, and it gives us a new list of cursor positions.

...anyway, just a tangent, not needed for this PR. Hope that helps with the last two issues though!

glennsl:
Thanks for the help and detailed explanation @bryphe!

> the issue is that our input strategy checks if a Revery UI element has focus.

Ah, yes of course. But wouldn't it be better to have the editor "steal" focus by setting it to 'None' when selecting an editor so we don't have to manually unfocus everywhere? Or would that mess with something else, like mouse capture?

> if we replace this: [...] with this: [...]

Unfortunately the `CursorMove` action doesn't seem to really work either. It _appears_ to work, by putting the cursor indicator on the match (well, actually on the character just before), but as soon as you do anything it jumps back to where it was. So it seems to not get synced up properly.

`EditorScrollToLine` also aligns the match to the top of the editor, but I think we'd want it in the middle, like "42gg" or ":42", or to scroll the least amount necessary, like "n". Shouldn't be too hard to fix, but I'm not sure where the fix should be applied, or if this is behaviour that can be altered through vim configuration.

Lastly, do we also want to select the match? It doesn't seem necessary if it's just a single word, since there are commands to easily manipulate those, but it might match something that's much harder to select and manipulate with vim commands.

bryphe:
Added #977 to pick up some `reason-libvim` changes to remove the cursor synchronization and keep the `Editor` as the source of truth for cursor positioning... this means that dispatching (`CursorMove`, or `EditorCursorMove` after #977 ) should be good enough.

I've ported this logic to #979 which seems to work - testing with a hardcoded position from the file explorer.

src/UI/OniInput.re
```diff
@@ -260,14 +260,12 @@ let%component make =
...
     resetCursor();
     onFocus();
-    Sdl2.TextInput.start();
     dispatch(Focus);
   };
 
   let handleBlur = () => {
     resetCursor();
     onBlur();
-    Sdl2.TextInput.stop();
```
bryphe: Nice catch!

src/Core/Ripgrep.re
```diff
@@ -148,30 +192,60 @@ let process = (rgPath, args, callback, completedCallback) => {
...
-    callback,
+    items => items |> dedup |> callback,
+    completedCallback,
+  );
+};
+
+let findInFiles = (path, workingDirectory, query, callback, completedCallback) => {
+  process(
+    path,
+    [|"--smart-case", "--json", "--", query, workingDirectory|],
```
CrossR: I'm assuming this is duplicating the flags from the file opening search, but may make sense to have `--hidden` in both searches. Otherwise, we aren't searching any hidden directories.

This is especially noticable in repos that contain only/lots of hidden files like a dotfiles repo.

(Not exactly urgent or hard, but just something I bump up against occasionally in the existing quick open.)

glennsl: I have no strong opinion on this, but we might ask "What would vscode do?"

CrossR: VSCode by deafult does seem to search in hidden folders. Testing in the Oni2 dir with `esy__cache_install` returns 3 files in VSCode vs the 1 file in Oni2. The results are the same for `azure-pipelines.yml` but the Oni2 searches misses the two `.ci/` files.

glennsl: Then I guess we're doing that :)

src/Store/SearchStoreConnector.re
```diff
@@ -29,20 +29,25 @@ let start = () => {
...
-      | _ =>
-        switch (state.searchPane) {
-        | Some(searchPane) => Some(searchUpdater(searchPane, action))
-        | None => None
-        }
-      };
-
-    ({...state, searchPane}, Isolinear.Effect.none);
+    switch (action) {
+    | Tick(_) => (state, Isolinear.Effect.none)
```
glennsl: I just spent a few hours learning that I need this "hack" to prevent `TickTest` from failing. I understand it's needed because there's a physical equality check somewhere that prevents re-rendering if the state hasn't actually changed, and without this it would re-render on every tick. Which is bad. But is the `Tick` really needed anymore? Can these use cases be solved using the new animation hooks and subscriptions instead?

bryphe: > I just spent a few hours learning that I need this "hack" to prevent TickTest from failing.

Ouch, sorry @glennsl ! 

>  I understand it's needed because there's a physical equality check somewhere that prevents re-rendering if the state hasn't actually changed, and without this it would re-render on every tick. Which is bad.

This is exactly right - don't want to re-render on every `Tick`.

> But is the Tick really needed anymore? Can these use cases be solved using the new animation hooks and subscriptions instead?

I think we could remove it (and it would be ideal, so we don't need the complexity of having our reducers by resilient about ticks). 

The main places that we still use it are:
- `SyntaxHighlightingStoreConnector` - used to execute periodic syntax highlighting jobs w/ time slicing. This could go away with #922  - since we'll be deferring that work to a separate process running full-tilt (we'd get responses from the new syntax highlight process).
- `HoverStoreConnector` - this was used for some animation, but could go away with a refactoring to use the new animation hooks
- `KeyDisplayerConnector` - this uses the `Tick` to expire some events when active. But we could refactor the API such that this isn't needed.
- `ExtensionClientStoreConnector` - it uses the `Tick` to `pump` the Node RPC client. This could be refactored to use a 'scheduler' instead.

It would be great to remove that and not have that 'gotcha' around.

test/Core/UtilityTests.re
```diff
@@ -30,3 +30,190 @@ describe("dropLast", ({test, _}) => {
     expect.list(dropLast([1, 2, 3])).toEqual([1, 2]);
   });
 });
+
+describe("StringUtil", ({describe, _}) => {
+  open StringUtil;
+
+  describe("trimLeft", ({test, _}) => {
```
bryphe: Nice!

src/Store/SearchStoreConnector.re
glennsl: Awesome! I'll add that to the "Technical" project.

glennsl:
> Only question I have - does it make sense to have the search pane enabled by default?

Uhmm, no. That wasn't supposed to have been checked in. I've fixed that and rebased, so if CI passes now I think this is ready to be merged.

bryphe:
> I've fixed that and rebased, so if CI passes now I think this is ready to be merged.

Awesome @glennsl ! Sounds great to me! 🎉

bryphe:
CI failure is not related - so I'll bring this in now! Thanks for all the work on this, @glennsl - so nice to finally have a search experience.

PR #973, opened 2019-11-25
Perf: Don't reconcile on every state change

bryphe:
When the are a flurry of `Actions` dispatched against the `Store`, we were 'reconciling' our UI after each one.

This 'debounces' - when there is a new state, we'll make note of it, but only reconcile once per frame.


src/bin_editor/Oni2_editor.re
```diff
@@ -56,13 +56,26 @@ let init = app => {
 
   let update = UI.start(w, <Root state=currentState^ />);
 
+  let isDirty = ref(false);
   let onStateChanged = state => {
     currentState := state;
+    isDirty := true;
     GlobalContext.set({...GlobalContext.current(), state});
-
     update(<Root state />);
```
izik1: Was this line supposed to poof?

bryphe: Indeed it was! Nice catch @izik1, thank you 👍

PR #983, opened 2019-11-27
Completion: Fuzzy match + highlight completions

bryphe:


src/Model/Completions.re
```diff
@@ -6,11 +6,15 @@
 
 open Oni_Core;
 
+module Zed_utf8 = Oni_Core.ZedBundled;
+
+type filteredCompletions = Filter.result(Actions.completionItem);
```
glennsl: Is this supposed to be plural?

bryphe: Ah, no - good catch @glennsl !

PR #985, opened 2019-11-28
Refactoring: Editor model cleanup

glennsl:
I might have gotten a bit carried away with this, but to my defense I _wanted_ to go a lot further! Initially I just wanted to add a few interfaces to get a better overview of what's in each module, but then it snowballed a bit.

The changes are roughly as follows:
- Add .rei
- Hide internal implementation details as much as possible, to have better control of potential invariants and to have the flexibility to change it without breaking external code.
- Mirror the type definition from `Actions`  so it's easier to look up, so `open Actions` isn't needed to bring the fields into scope and so unnecessary type annotations can be removed in the implementation. That the type is initially defined elsewhere is now a minor detail that doesn't otherwise affect how it's used.
- Replace non-descriptive abbreviations with more descriptive names. Especially every occurrence of `v`, which sometimes had multiple different meanings in a single function.
- Remove unnecessary `open`s that pollute the namespace and obscure the origin of functions. At the top level, generally only modules that are explicitly designed to be opened should be.
- Pass the model down through the view, instead of an id that needs to be looked up later.
- Some unused functions have been removed, but it might be that they ought to be commented out or exposed for debugging purposes, for example, but then they ought to have names that reflect their purpose.

Longer term I'd like to remove all type definitions (except `t` of course) from `Actions`, as having them there prevents proper encapsulation and hiding of internal state. As far as I can see, the only reason they're there is because some modules contain a `reducer` function which depends on `Actions`, causing a dependency cycle. I don't think `reducer ` functions should be associated with specific models, however, and might be better to move them to the `Store` project. If there are dependency cycles between other types, it might be better to use recursive modules defined together, and then include their implementations in separate "module files" which hides the recursion behind its interface.

### DIFF
--- a/bench/lib/EditorSurfaceBench.re
+++ b/bench/lib/EditorSurfaceBench.re
@@ -21,7 +21,7 @@ let editorSurfaceMinimalState = () => {
       rootNode,
       <EditorSurface
         isActiveSplit=true
-        editorGroupId
+        editorGroup
         editor=simpleEditor
         state=simpleState
         metrics
@@ -36,7 +36,7 @@ let editorSurfaceThousandLineState = () => {
       rootNode,
       <EditorSurface
         isActiveSplit=true
-        editorGroupId
+        editorGroup
         editor=simpleEditor
         state=thousandLineState
         metrics
@@ -51,7 +51,7 @@ let editorSurfaceThousandLineStateWithIndents = () => {
       rootNode,
       <EditorSurface
         isActiveSplit=true
-        editorGroupId
+        editorGroup
         editor=simpleEditor
         state=thousandLineStateWithIndents
         metrics
@@ -66,7 +66,7 @@ let editorSurfaceHundredThousandLineState = () => {
       rootNode,
       <EditorSurface
         isActiveSplit=true
-        editorGroupId
+        editorGroup
         editor=simpleEditor
         state=hundredThousandLineState
         metrics
@@ -87,7 +87,7 @@ let setupSurfaceThousandLineLayout = () => {
     container,
     <EditorSurface
       isActiveSplit=true
-      editorGroupId
+      editorGroup
       editor=simpleEditor
       state=thousandLineState
       metrics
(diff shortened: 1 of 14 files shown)


bryphe:
> I might have gotten a bit carried away with this, but to my defense I wanted to go a lot further! Initially I just wanted to add a few interfaces to get a better overview of what's in each module, but then it snowballed a bit.

This is a great 'chunk' of refactoring - thank you @glennsl !

> Add .rei

> Hide internal implementation details as much as possible, to have better control of potential invariants and to have the flexibility to change it without breaking external code.

Awesome. Very helpful - I need to get into the habit of doing this by default. 💯 

> Replace non-descriptive abbreviations with more descriptive names. Especially every occurrence of v, which sometimes had multiple different meanings in a single function.
> Remove unnecessary opens that pollute the namespace and obscure the origin of functions. At the top level, generally only modules that are explicitly designed to be opened should be.
 
I need to break both of these habits...

> Mirror the type definition from Actions so it's easier to look up, so open Actions isn't needed to bring the fields into scope and so unnecessary type annotations can be removed in the implementation. That the type is initially defined elsewhere is now a minor detail that doesn't otherwise affect how it's used.

Good idea! Jumping back is difficult

> Longer term I'd like to remove all type definitions (except t of course) from Actions, as having them there prevents proper encapsulation and hiding of internal state. A

Ya, totally - this is very awkward at the moment. Aside from the duplication in the model type definitions, it also forces us to have overly verbose field names (since you can't have field names with the same name in the `and` type definitions).

> As far as I can see, the only reason they're there is because some modules contain a reducer function which depends on Actions, causing a dependency cycle.

This is correct - was looking for a better solution.

>  I don't think reducer functions should be associated with specific models, however, and might be better to move them to the Store project. 

This seems like it would work, be a better design, and break the dependency cycle 👍



Thanks for this refactoring work @glennsl - I always learn a lot from your changes!

glennsl:
/azp run

Doesn't seem like these all work quite as advertised, but this one does at least.

glennsl:
Oh, so that's how it works. Well that didn't seem all that useful. Not for my needs at least.

PR #988, opened 2019-11-30
Fix #895 - OSX: Themable titlebar

### LINKED ISSUE #895: make titlebar themeable
As of now under macOS the titlebar is the default titlebar of the OS. Especially on macOS darkmode the light titlebar is really irritating and I think the whole editor look would benifit from a themable titlebar.

VSCode has the titlebar in the same style as the editor, see this screenshot for comparison:

titlebar

bryphe:
Fixes #895 - adds a custom, themable titlebar for macOS - picking up the new `titlebarStyle` API in Revery from @zbaylin.

One downside is that the foreground text color isn't themable - macOS picks a color based on the background color. So we'll need to sync our background color with the macOS background color to get good results for the foreground (it'd be even better if we had a way to draw the text ourselves!)

__Before:__

image

__After:__

image

__Remaining work:__
- [x] Merge up https://github.com/revery-ui/reason-sdl2/pull/39
- [x] Reset titlebar in Revery after resize / restore
- [x] Get proper theme colors for title background / foreground (https://code.visualstudio.com/api/references/theme-color#title-bar-colors)
- [x] Track fullscreen / maximized state, and don't render the titlebar spacer when maximized
- [x] Track active / inactive state
- [x] Test on Windows (should still render the same as before, as this fix is specific to OSX)
- [x] Test on Linux (should still render the same as before, as this fix is specific to OSX)


zbaylin:
You can draw the text yourself if you set the title to be hidden. It wouldnt feel very native though. However, it is an option.

PR #991, opened 2019-12-01
Fix a typo in osx notarize script

necabo:


bryphe:
Nice catch @necabo - thank you!

PR #993, opened 2019-12-02
File explorer improvements

### LINKED ISSUE #376: FileExplorer doesn't expand when opening a directory in the bottom
When opening a directory at the bottom, the scrollbar doesn't expand automatically. As soon as you scroll up, the scrollbar expands and you can scroll down.

I guess this is related to: https://github.com/revery-ui/revery/issues/514

### LINKED ISSUE #602: Slow performance on directories with many files/folders
Steps to reproduce:
1. Create a directory with many files/folders (e.g. in Bash: `for i in {1..10000}; do mkdir "$i"; done;`)
2. Open Onivim in the folder

### LINKED ISSUE #649: First launch on large directory, the tree will not scroll down
# Bug description
When Onivim2 launches in a directory that is very large, It will only scroll upwards and not downwards until I select another directory in it. Then is scrolls fine.

# Steps to reproduce
Open Onivim2 in a large directory on macOS. Before clicking any other directory or file, try to scroll down the list.

# What happens
It will not scroll downwards. Scrolling upwards will bring the topmost directory down the screen until it is at the bottom.

# Expected results
Being about to scroll up and down freely.

### LINKED ISSUE #717: Explorer wraps lines
image

Usual behaviour in most editors is to overspill and have a horizontal scrollbar.

### LINKED ISSUE #851: Crash: Clicking on the Directory Slider would not release the mouse...Lock up.
2019-10-15

# What I did:

1) I just did a `git pull` on the oni2 project and recompiled. 
2) Opened oni2 in my project directory.
3) Click and slide the slider on the right for the directory tree.
4) Opened a file in the directory tree
5) Tried to place my cursor in the file, but it is still hand with index finger extended. I clicked several times to try to get it unstuck when it went totally unresponsive with the spinning ball.

# Repeatable?

Yes.

# OS

macOS Mojave 10.14.6

glennsl:
* Design tweaks
* Virtualized rendering, Should fix #602 (though there's still room for improvement, see below)
* Fixes #376 
* Fixed #649 
* Fixes #717 
* Might fix #851 (or it might make it worse, it's a different scrollbar implementation at least)

Re: additional potential performance improvement. Currently items out of view are replaced individually by empty placeholders that take up the same amount of space. These could be collapsed into a single placeholder on each level, which I've tried to do in https://github.com/glennsl/oni2/commit/800b5ceb61f666ded9adf336a4dfe07f06c957fc, but this implementation is a bit off and when trying to fix it I encountered what seemed like reconciliation bugs (elements being reordered). I'm also not sure it's worth the added complexity anyway. Is the cost of rendering empty fixed size elements significant?

TODO:
- [x] Cleanup
- [x] Consider changing loading strategy
- [x] Reorganize actions
- [x] Items being clipped after closing opened folders (probably precomputation issue)
- [x] FontSizeChangeTest fails (?!)


src/UI/TreeView.re
```diff
@@ -1,160 +1,233 @@
...
-      switch (icon, secondaryIcon, isOpen) {
-      | (Some(_), Some(secondary), true) => secondary
-      | (Some(primary), Some(_), false) => primary
-      | (Some(primary), None, _) => primary
-      | (None, Some(secondary), _) => secondary
-      | (None, None, false) => makeIcon(~character=primaryRootIcon)
-      | (None, None, true) => makeIcon(~character=secondaryRootIcon)
+    if (subtreeSize < clipStart || clipEnd < 0) {
+      <placeholder
+        // If the entire node is out of view, render a placeholder with the appropriate height
```
glennsl: refmt...

src/UI/TreeView.re
```diff
@@ -1,160 +1,233 @@
...
+        size={subtreeSize * itemHeight}
+      />;
```
glennsl: refmt...

src/Store/TitleStoreConnector.re
```diff
@@ -9,65 +9,78 @@ module Model = Oni_Model;
...
+    | Init => (state, Effects.updateTitle(state))
+    | BufferEnter(_) => (state, Effects.updateTitle(state))
+    | BufferSetModified(_) => (state, Effects.updateTitle(state))
+
+    // TODO: This shouldn't exist, but needs to  be here because it deoends on
+    // `setTitle` being passed in. It would however be ebtter to have a more
+    // general mechanism, like "Effect actions" handled by an injected dependency
+    // or have effects parameterized by an "environment" passed in along with
+    // `dispatch`
+    | SetTitle(title) => (state, internalSetTitleEffect(title))
```
glennsl: :eyes: 

Maybe we should make our own algebraic effect system!

src/Store/FileExplorerStore.re
```diff
@@ -6,102 +6,154 @@
...
-        FileExplorer.getDirectoryTree(cwd, languageInfo, iconTheme, ignored);
-      dispatch(Actions.SetExplorerTree(newTree));
+        FileExplorer.getDirectoryTree(
+          directory,
+          languageInfo,
+          iconTheme,
+          ignored,
+        );
+
+      dispatch(Actions.FileExplorer(TreeUpdated(newTree)));
```
glennsl: I might moderate my strong stance on making all actions in effects explicit. In cases like this were the action is internal it seems to make sense that it's implicit. Might still be useful to have an optional `onComplete` action though. I'm sure we'll find out eventually!

glennsl:
I've tried to reorganize the actions and effects of the File Explorer according to out discussions in https://github.com/onivim/oni2/pull/993/commits/926c327699a72fe1a3aadbad9aa3cad036cd6cb5, @bryphe. But once you start tugging on a thread things start unravelling fast. So in order to keep this relatively contained, some things are still a bit awkward. I see paths forward for everything though.

Also, the integration test fails for mysterious reasons but likely has something to do with these changes.

src/Store/TitleStoreConnector.re
bryphe: @jordwalke had some ideas around user-space algebraic effects!

src/Store/VimStoreConnector.re
```diff
@@ -79,7 +79,7 @@ let start =
 
   let _ =
     Vim.onDirectoryChanged(newDir =>
-      dispatch(Model.Actions.OpenExplorer(newDir))
+      dispatch(Model.Actions.VimDirectoryChanged(newDir))
```
bryphe: This looks like a good example of renaming from the 'expected action' to a describing an event that has occurred 👍

src/Store/FileExplorerStore.re
```diff
@@ -6,102 +6,154 @@
 open Oni_Core;
 open Oni_Model;
 
-let start = () => {
-  let (stream, _) = Isolinear.Stream.create();
-
-  let getExplorerFilesEffect = (cwd, languageInfo, iconTheme, ignored) => {
-    Isolinear.Effect.createWithDispatch(~name="explorer.open", dispatch => {
+module Effects = {
```
bryphe: Nice 👍

src/Store/TitleStoreConnector.re
bryphe: Would be nice to have a better solution for this for sure. I've seen a few articles on Free monads which look interesting: http://degoes.net/articles/modern-fp

It would describe the semantics of effects - like `SetTitle(title)` - but leave it up to an 'interpreter' to decide _what_ to do.

src/Store/FileExplorerStore.re
```diff
@@ -6,102 +6,154 @@
...
 
-  (updater, stream);
-};
+  (
+    (state: State.t) =>
+      fun
+      // TODO: Should be handle by a more general init mechanism
+      | Actions.Init => {
+          let cwd = Rench.Environment.getWorkingDirectory();
+          let newState = {
```
bryphe: This nested update always feels awkward - I wonder if lenses could streamline this? Like [`lens`](https://opam.ocaml.org/packages/lens/).

src/Store/VimStoreConnector.re
glennsl: Yeah, this is a good example of something that has a high likelihood of causing a bug in the future. If we want something more to happen when we click the explorer button in the sidebar for example, it seems natural to put that in the `OpenExplorer` action. But you probably do not want that same thing to happen when you `:cd`.

src/Store/FileExplorerStore.re
glennsl: I don't think we can put most effects in `Effects` submodules in the stores though, because it'll cause dependency cycles. But then I also think very few updaters should be in "stores". They should instead be associated more with UI components since that's what the "actions"/"events" will refer to. But it still makes sense to have a "VimStore" for example.

src/Store/FileExplorerStore.re
glennsl: I'm not sure lenses really solve the actual problem. Having deeply nested updates is usually an indication of architectural flaws, and using lenses is a bit like just brushing it under the carpet.

This isn't really a nested update though, just a value being set on a record that also happens to contain a record. We might make it a bit smoother by adding a `setWorkspace(cwd, state)` or `Workspace.create(cwd)` function for example.

glennsl:
`FontSizeChangeTest` failed for me locally because I had configured a different font. I guess I can live with FiraCode for now, but it might be an idea to not load user configuration during integration testing. I've created a task for this in the "Technical" project.

src/Core/Constants.re
```diff
@@ -24,11 +24,6 @@ type t = {
   minimapLineSpacing: int,
   scrollBarThickness: int,
   minimapMaxColumn: int,
-  /*
-   * Maximum levels of the file system to traverse
-   * when initially populating the file explorer
-   */
-  maximumExplorerDepth: int,
```
bryphe: Thanks for simplifying this 👍

src/Store/TitleStoreConnector.re
glennsl: I think tagless final might work better with Reason's module system, but yeah something along those lines might be nice. Or overcomplicated. Either way it would be fun to try!

src/Model/Actions.re
```diff
@@ -62,11 +62,9 @@ type t =
   | EditorScroll(EditorId.t, float)
   | EditorScrollToLine(EditorId.t, int)
   | EditorScrollToColumn(EditorId.t, int)
-  | OpenExplorer(string)
   | ShowNotification(notification)
   | HideNotification(int)
-  | SetExplorerTree([@opaque] UiTree.t)
-  | UpdateExplorerNode([@opaque] UiTree.t, [@opaque] UiTree.t)
+  | FileExplorer(FileExplorer.action)
```
bryphe: This is a good precedent for us to follow as we add more actions. I'll work on moving to a similar model for the language features I'm integrating.

src/UI/FileTreeView.re
```diff
@@ -0,0 +1,106 @@
...
+    | _ => <icon />
+    };
+
+  <View style=Styles.item>
+    <icon />
+    <Text text={node.displayName} style={Styles.text(~fg, ~bg, ~font)} />
+  </View>;
+};
+
+module TreeView = TreeView.Make(FsTreeNode.Model);
```
bryphe: Nice use of functors here... this seems like a great way to generalize and keep the data model independent of the actual TreeView.

glennsl: I first tried using a first-class module, but the brisk-reconciler ppx was giving me issues when using a locally abstract type. Using a functor turned out simpler anyway though, so win-win!

bryphe:
> FontSizeChangeTest failed for me locally because I had configured a different font. I guess I can live with FiraCode for now, but it might be an idea to not load user configuration during integration testing. I've created a task for this in the "Technical" project.

Sorry you had to spend time investigating this! Indeed, we should not be loading user config for tests, just a recipe for problems... I opened up #997 to address that task.

> I've tried to reorganize the actions and effects of the File Explorer according to out discussions in 926c327, @bryphe. But once you start tugging on a thread things start unravelling fast. So in order to keep this relatively contained, some things are still a bit awkward. I see paths forward for everything though.

For sure... I think you've made some great incremental improvements, though. This seems like the right balance to me for this PR - and then we can continue moving towards our architectural goals in subsequent PRs.

I just tested out the latest and the change looks awesome - big improvement in both performance and look and feel. Thank you!

PR #995, opened 2019-12-03
ExtHost - Workspace: Populate workspace rootPath

bryphe:
In testing reason-language-server, one of the initial blockers was that we were not populating `vscode.workspace.rootPath` - which is used to find the pre-built RLS binaries: https://github.com/jaredly/reason-language-server/blob/a04d4feda53ddb1a5df9534d60b805711418a23c/editor-extensions/vscode/src/index.js#L24

This change sets up our `ExtHostClient` to take an initial workspace when initializing the extension host, and the plumbing to update the workspace path when it changes. 

It also adds some commands in our developer VSCode extension to show / test the rootPath.


development_extensions/oni-dev-extension/extension.js
```diff
@@ -97,17 +97,23 @@ function activate(context) {
...
+		// Display a message box to the user
+		vscode.window.showInformationMessage("rootPath: " + vscode.workspace.rootPath);
+	});
 
     // Helper command to show buffer text
     // This helps us create a test case to validate buffer manipulations
-	let disposable5 = vscode.commands.registerCommand('developer.oni.getBufferText', () => {
+	let disposable6 = vscode.commands.registerCommand('developer.oni.getBufferText', () => {
 		vscode.window.showInformationMessage("fulltext:" + latestText);
 	});
```
glennsl: Instead of using non-descriptive numbered identifiers, and having to renumber them when adding a new one, it might be better to write a local helper function that both does the registration  and adds the dispose function to `context.subscriptions`.

bryphe: Good suggestion, thanks!

PR #996, opened 2019-12-03
Extensions - CLI: Add --extensions-dir argument

bryphe:


docs/docs/using-onivim/command-line.md
```diff
@@ -9,3 +9,34 @@ sidebar_label: Command Line
...
+This can be overridden via the `--extensions-dir`, like:
+
+```
+oni2 --extensions-dir /some/path/with/extensions
+```
+
+### Logging
+
+The `-f` argument can be used to keep Onivim 2 attached to the terminal,
+and provide logging output.
```
glennsl: Would it be a good idea to mention `ONI2_DEBUG` and `ONI2_LOG_FILE` here as well?

bryphe: Would be a good place for it!

PR #998, opened 2019-12-03
ExtHost - Documents: Hook up initial document load

bryphe:
__Issue:__ When opening an existing file, the contents from the open wouldn't get synchronized to the extension host.

We would send an update, but we would send with an endline that didn't exist yet - which the extension wouldn't handle. For an initial update, we should send our lines but at a start / end one-based position of `(line 1, column 1)` so that the insertion of the initial update is handled correctly.

Along the way, I stumbled across a couple of issues with our buffer updates - we should push some fixes (and improve the granularity / typing of updates we get from `reason-libvim`), so I also added more test coverage in our integration tests around buffer synchronization. This is something that must _always_ work.


src/Core/Types.re
```diff
@@ -102,6 +102,8 @@ module Position = {
 module BufferUpdate = {
```
bryphe: Move this out of Core/Types

src/Store/VimStoreConnector.re
```diff
@@ -317,17 +319,51 @@ let start =
...
           (),
         );
 
-      dispatch(Model.Actions.BufferUpdate(bu));
+      let shouldApply =
+        bufferMaybe
+        |> Option.map(Model.Buffer.shouldApplyUpdate(bu))
+        |> Option.value(~default=true);
+
+      if (shouldApply) {
```
bryphe: Add comment that we should be handling the buffer updates better from `reason-libvim` - in other words, we shouldn't _need_ this logic at this layer

src/Core/BufferUpdate.re
```diff
@@ -0,0 +1,60 @@
...
+  let ret: t = {
+    isFull,
+    id,
+    startLine: Index.OneBasedIndex(startLine),
+    endLine: Index.OneBasedIndex(endLine),
+    lines,
+    version,
+  };
+  ret;
+};
```
glennsl: Why not just use `Index.t` here?

Also, `{ let ret: {isFull, ...}; ret; }` is equivalent to `{isFull, ...}`

src/Store/VimStoreConnector.re
```diff
@@ -317,17 +319,59 @@ let start =
...
       Log.info("Vim - Buffer update: " ++ string_of_int(update.id));
       open Core.Types;
+      open Model.State;
+
+      let isFull = update.endLine == (-1);
+
+      let bufferMaybe =
+        Some(getState())
+        |> Option.map(state => state.buffers)
+        |> Option.bind(Model.Buffers.getBuffer(update.id));
```
glennsl: Isn't this equivalent to

```reason
getState().buffers |> Model.Buffers.getBuffer(update.id);
```
?

src/Store/VimStoreConnector.re
```diff
@@ -317,17 +319,59 @@ let start =
       open Vim.BufferUpdate;
       Log.info("Vim - Buffer update: " ++ string_of_int(update.id));
       open Core.Types;
+      open Model.State;
+
+      let isFull = update.endLine == (-1);
+
+      let bufferMaybe =
```
glennsl: The convention is usually `maybeBuffer` I think. At least from what I've seen. It also reads better to me.

Good descriptive naming either way though :+1:

src/Store/VimStoreConnector.re
```diff
@@ -317,17 +319,59 @@ let start =
...
+      // In particular, with undo / redo, we get multiple updates - a full update (version+2) and
+      // an incremental update (version+1) _after_ the full update. If we apply the incremental update
+      // after the full update, ignoring the version, we'll get incorrect results.
+      //
+      // The fix really belongs in reason-libvim - we should always be trust the order we get from the updates,
+      // and any of this filtering or manipulation of updates should be handled and tested there.
+      let shouldApply =
+        bufferMaybe
+        |> Option.map(Model.Buffer.shouldApplyUpdate(bu))
+        |> Option.value(~default=true);
```
glennsl: This feels a bit convoluted. How about `Option.map(Model.Buffer.shouldApplyUpdate(bu), bufferMaybe) != Some(false)`. Isn't that at least a little bit clearer?

Also, just opening `Oni_Model` at the top would remove quite a bit of noise. opening modules that only contain module aliases is usually fine.

src/Store/VimStoreConnector.re
bryphe: Sure is... don't know what I was thinking 😄

src/Store/VimStoreConnector.re
bryphe: Sounds good - I think the confusing aspect of the logic is that you still have to think about the `None` case. But it is cleaner. And hopefully the confusing part we can go away by fixing the root-cause bug in reason-libvim.

src/Core/BufferUpdate.re
bryphe: Sounds good - I switched to just `create` (and cleaned up the unnecessary `ret`)

Issue #523, opened 2019-07-31
Startup is slow on Windows 10


Raikiri:
Unfortunately I don't have enough profiling tools to measure start time exactly, but it definitely feels 2x-4x slower than that of Sublime Text 3. I'm running Oni2.exe directly from file explorer, it's installed on an SSD and takes around 1-2 seconds to launch, when sublime takes around 200ms or so.

Raikiri:
Also tested it on another Windows 10 machine (core i7, SSD) and startup time feels basically the same, so it's not just my machine. I'd love to provide more info if there were some profiling tools included.

satvikpendem:
I can replicate as well.

Azumgi:
I'm on Windows 10 machine with oni2 installed on SSD and startup time for me is ~500ms.

bryphe:
Looking into this - the current startup flow isn't optimal on Windows. We actually start up two processes - `Oni2` spins up `Oni2_editor`.

Both processes are also _console_ process which mean they actually spin up `conhost` and allocate console instances. In the case we are 'attached' to a console (via the `-f` argument), they just re-use the existing console and stay attached. When we aren't 'attached', the `Oni2_editor` process frees its console.

This isn't great for performance - as each process spawn is a non-trivial cost on Windows. In addition, we should be, by default, a _win32 gui_ process (not a _console_ process) - and only attach/allocate a console with the `-f` flag.

So some fixes for this would be:
- Verify the compilation flags we are using for the binaries, and set them up to run as _win32 gui_ processes instead of _console_ processes
- In the case of `-f` flag, attach to an existing console, or allocate a new one (instead of the reverse behavior we have today - `f` frees the attached console).

And we'd want to validate:
- Running `Oni2` from the GUI (there should be no console-window-flash)
- Running `Oni2` from the terminal (should detach immediately)
- Running `Oni2 -f` from the terminal (should attach to the current console instance)
- Running `Oni2 -f` from the GUI (should allocate a new console instance)

bryphe:
An interesting idea here: https://blogs.msdn.microsoft.com/junfeng/2004/02/06/how-to-make-an-application-as-both-gui-and-console-application/

[...]

PR #999, opened 2019-12-03
Switch to using "logs" as the logging framework

glennsl:
This switches over to using dbuenzli's `logs` library. It's at feature parity and backwards compatible with the old `Log` module, but now also prints in shiny colors!

I've also added some preliminary namespace support using tags as a proof of concept, and I think we should be able to have a namespace hierarchy with both fixed and ad hoc namespaces., filtering and hash-based coloring. I'd like to add that too before merging this, but have to do a bit of research to find a good way to do the filtering first.

- [x] Filtering
- [x] CLI options
- [x] Hash-based coloring of namespaces
- [x] Timing
- [ ] Ad-hoc namespaces?
- [ ] Remove the old non-namespaced logging functions?

### DIFF
--- a/package.json
+++ b/package.json
@@ -210,7 +210,8 @@
     "reason-textmate": "^2.1.0",
     "esy-sdl2": "*",
     "@opam/fmt": "^0.8.8",
-    "@opam/logs": "^0.7.0"
+    "@opam/logs": "^0.7.0",
+    "@opam/re": "*"
   },
   "resolutions": {
     "reason-jsonrpc": "github:bryphe/reason-jsonrpc#95f2427",
--- a/package.json
+++ b/package.json
@@ -208,7 +208,9 @@
     "revery": "0.27.0",
     "reason-tree-sitter": "*",
     "reason-textmate": "^2.1.0",
-    "esy-sdl2": "*"
+    "esy-sdl2": "*",
+    "@opam/fmt": "^0.8.8",
+    "@opam/logs": "^0.7.0"
   },
   "resolutions": {
     "reason-jsonrpc": "github:bryphe/reason-jsonrpc#95f2427",
--- a/src/Core/Log.re
+++ b/src/Core/Log.re
@@ -96,7 +96,10 @@ module Namespace = {
 };
 
 module DeltaTime = {
-  let pp = (ppf, dt) => Fmt.pf(ppf, "%+5.0fms", dt *. 1000.);
+  let pp = (ppf, dt) =>
+    dt > 10.
+      ? Fmt.pf(ppf, "%+6.2fs", dt) : Fmt.pf(ppf, "%+5.0fms", dt *. 1000.);
+
   let tag = Logs.Tag.def("time", pp);
 };
 
--- a/src/Core/Log.re
+++ b/src/Core/Log.re
@@ -19,25 +19,24 @@ module Option = {
 };
 
 module Constants = {
-  let colors =
-      [|
-        // `Black
-        `Blue,
-        `Cyan,
-        `Green,
-        `Magenta,
-        `Red,
-        //`White,
-        `Yellow,
-        `Hi(`Black),
-        `Hi(`Blue),
-        `Hi(`Cyan),
-        `Hi(`Green),
-        `Hi(`Magenta),
-        `Hi(`Red),
-        `Hi(`White),
-        `Hi(`Yellow),
-      |]
+  let colors = [|
+    // `Black
+    `Blue,
+    `Cyan,
+    `Green,
+    `Magenta,
+    `Red,
+    //`White,
+    `Yellow,
+    `Hi(`Black),
+    `Hi(`Blue),
+    `Hi(`Cyan),
+    `Hi(`Green),
+    `Hi(`Magenta),
+    `Hi(`Red),
+    `Hi(`White),
+    `Hi(`Yellow),
+  |];
 };
 
 module Env = {
@@ -72,32 +71,28 @@ module Namespace = {
 
   let setFilter = filter => {
     let filters =
-      filter
-      |> String.split_on_char(',')
-      |> List.map(String.trim);
+      filter |> String.split_on_char(',') |> List.map(String.trim);
 
     let (incs, excs) =
       filters
-      |> List.fold_left(((includes, excludes), filter) =>
-        if (filter == "") {
-          (includes, excludes)
-        } else if (filter.[0] == '-') {
-          let filter =
-            String.sub(filter, 1, String.length(filter) - 2)
-            |> Re.Glob.glob
-            |> Re.compile;
+      |> List.fold_left(
+           ((includes, excludes), filter) =>
+             if (filter == "") {
+               (includes, excludes);
+             } else if (filter.[0] == '-') {
+               let filter =
+                 String.sub(filter, 1, String.length(filter) - 2)
+                 |> Re.Glob.glob
+                 |> Re.compile;
 
-          (includes, [filter, ...excludes])
-        } else {
-          let filter =
-            filter
-            |> Re.Glob.glob
-            |> Re.compile;
+               (includes, [filter, ...excludes]);
+             } else {
+               let filter = filter |> Re.Glob.glob |> Re.compile;
 
-          ([filter, ...includes], excludes)
-        },
-        ([], [])
-      );
+               ([filter, ...includes], excludes);
+             },
+           ([], []),
+         );
 
     includes := incs;
     excludes := excs;
@@ -106,19 +101,26 @@ module Namespace = {
 
 type msgf('a, 'b) = (format4('a, Format.formatter, unit, 'b) => 'a) => 'b;
 
-let fileReporter =
-  switch (Env.logFile) {
-  | Some(path) =>
-    let channel = open_out(path);
-    Printf.fprintf(channel, "Starting log file.\n%!");
-    let ppf = Format.formatter_of_out_channel(channel);
+let logFileChannel = ref(None);
 
-    Logs.{
-      report: (_src, level, ~over, k, msgf) => {
-        let k = _ => {
-          over();
-          k();
-        };
+let setLogFile = path => {
+  Option.iter(close_out, logFileChannel^);
+  let channel = open_out(path);
+  Printf.fprintf(channel, "Starting log file.\n%!");
+  logFileChannel := Some(channel);
+};
+
+let fileReporter =
+  Logs.{
+    report: (_src, level, ~over, k, msgf) => {
+      let k = _ => {
+        over();
+        k();
+      };
+
+      switch (logFileChannel^) {
+      | Some(channel) =>
+        let ppf = Format.formatter_of_out_channel(channel);
 
         msgf((~header=?, ~tags as _=?, fmt) => {
           Format.kfprintf(
@@ -129,10 +131,10 @@ let fileReporter =
             (level, header),
           )
         });
-      },
-    };
 
-  | None => Logs.nop_reporter
+      | None => k()
+      };
+    },
   };
 
 let consoleReporter =
@@ -156,7 +158,7 @@ let consoleReporter =
           namespace,
         );
       });
-    }
+    },
   };
 
 let reporter =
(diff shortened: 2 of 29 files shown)


src/Store/StoreThread.re
```diff
@@ -15,6 +15,8 @@ module Model = Oni_Model;
 
 open Oni_Extensions;
 
+module Log = (val Core.Log.withNamespace("StoreTread"));
```
glennsl: We could also use a functor for this. In that case it would look like:

```reason
module Log = Core.Log.WithNamespace({ let namespace = "StoreThread" });
```

None of them are particularly intuitive, but using `val` is at least less noisy.

bryphe: Thanks for calling out the possibilities! I agree, the `val` first-class module is at least more concise

src/Core/Log.re
```diff
@@ -4,77 +4,143 @@
...
-    Printf.fprintf(oc, "%s\n", msg);
-    flush(oc);
-  | None =>
-    switch (error) {
-    | false => print(msg)
-    | true => prerr(msg)
-    }
+// defaults
+let () = {
+  Fmt_tty.setup_std_outputs(~style_renderer=`Ansi_tty, ());
```
bryphe: Does this send anything to `stdout`? We have to be careful on Windows - we can actually get a crash by emitting to OCaml's stdout if the console isn't attached (it seems like a bug... `printf` from C is fine, but eventually OCaml `print_endline` hits a limit where it crashes if we log too much w/o an attached console)

bryphe:
> It's at feature parity and backwards compatible with the old Log module, but now also prints in shiny colors!

> I've also added some preliminary namespace support using tags as a proof of concept, and I think we should be able to have a namespace hierarchy with both fixed and ad hoc namespaces., filtering and hash-based coloring.

Wow, that's excellent!

For filtering - we could potentially use the [`Re.Glob`](https://github.com/ocaml/ocaml-re/blob/master/lib/glob.mli) module or another one... or the [`reason-oniguruma`](https://github.com/onivim/reason-oniguruma) regex library. Globs might be the simplest to use, though. 

Some possibilities for specifying it:
- An environment variable
- A CLI arg
- A configuration setting (might be nice for runtime tweaking of the filtering granularity, if doable)

src/Core/Log.re
glennsl: I don't think it does. From what I can see it just sets some attributes used later to determine how to print.

glennsl:
> For filtering - we could potentially use the Re.Glob module or another one... or the reason-oniguruma regex library. Globs might be the simplest to use, though.

Ah, yeah I was wondering if we had access to anything that could do globbing already.

I think the simplest approach is to just use a dotted namespace string (e.g. "Oni2.StoreThread") and then use globbing for filtering with an added `-` modifier to specify exclusion rather than inclusion. Shamelessly stolen from [debug](https://github.com/visionmedia/debug#wildcards).

> Some possibilities for specifying it:

Good suggestions! I think we can do all these relatively easily.

package.json
```diff
@@ -210,7 +210,8 @@
     "reason-textmate": "^2.1.0",
     "esy-sdl2": "*",
     "@opam/fmt": "^0.8.8",
-    "@opam/logs": "^0.7.0"
+    "@opam/logs": "^0.7.0",
+    "@opam/re": "*"
```
glennsl: Apparently `re` wasn't a dependency already, though it is an indirect dependency of something. Is it okay to add as direct dependency, or should we try something else?

src/bin_launcher/Oni2.re
```diff
@@ -12,10 +12,14 @@ let version = () => {
...
 let passthrough = Arg.Unit(() => ());
 let passthroughFloat = Arg.Float(_ => ());
+let passthroughString = Arg.String(_ => ());
 
 let spec = [
   ("-f", Arg.Set(stayAttached), ""),
   ("--nofork", Arg.Set(stayAttached), ""),
+  ("--debug", passthrough, ""),
+  ("--log-file", passthroughString, ""),
+  ("--log-filter", passthroughString, ""),
```
glennsl: Why is this needed? Can it not just use the flags specified in `Cli'?

src/Core/Log.re
```diff
@@ -106,19 +101,26 @@ module Namespace = {
...
 type msgf('a, 'b) = (format4('a, Format.formatter, unit, 'b) => 'a) => 'b;
 
+let logFileChannel = ref(None);
+
+let setLogFile = path => {
+  Option.iter(close_out, logFileChannel^);
+  let channel = open_out(path);
+  Printf.fprintf(channel, "Starting log file.\n%!");
+  logFileChannel := Some(channel);
+};
```
glennsl: Should this also always set the log level to debug, as it does when the env variable is set?

src/Core/Log.re
```diff
@@ -143,17 +174,17 @@ let enableDebugLogging = () =>
...
-let info = msg => infof(m => m("%s", msg));
-
-let debugf = msgf => Logs.debug(m => msgf(m(~header=?None, ~tags=?None)));
-let debug = msgf => debugf(m => m("%s", msgf()));
+let log = (~namespace="Global", level, msgf) =>
+  Logs.msg(level, m => {
+    if (Namespace.isEnabled(namespace)) {
+      let tags = Logs.Tag.(empty |> add(Namespace.tag, namespace));
+      msgf(m(~header=?None, ~tags))
+    }
```
glennsl: This filters on all reporters. We can also set up filtering for just the console reporter, but that will add a bit of overhead because at that point it's wrapped in a tag. Probably a negligible overhead, but still.

src/Extensions/ExtHostClient.re
```diff
@@ -85,12 +87,12 @@ let start =
     | ("MainThreadStatusBar", "$setEntry", args) =>
       In.StatusBar.parseSetEntry(args) |> apply(onStatusBarSetEntry);
       Ok(None);
-    | (s, m, _a) =>
-      Log.error(
-        Printf.sprintf(
+    | (system, method, _a) =>
+      Log.errorf(m => 
+        m(
```
glennsl: I have no idea what `s` and `m` is here, so just guessing. `m` is the convention used for the logger though. No idea why that is either, but one of them will have to change.

package.json
bryphe: Should be fine to add as a direct dependency. If I recall correctly, we use it in revery anyway (so it gets compiled in regardless) 👍

src/bin_launcher/Oni2.re
bryphe: Oh, good question... This isn't really documented anywhere, and should be.

Because of the way we package - `Oni2.exe` (the launcher) needs to have an extreme minimum set of dependencies. On Linux and Mac we go through some special gymnastics to bake in the dynamic libraries used:
- `OSX` - we run the `dylibbundler` tool which remaps all the dynamic-lib paths (here: https://github.com/onivim/oni2/blob/69c05a9050cca840c2fe5d082bae7233d3447566/scripts/release.js#L157)
- `Linux` - we run the `linuxdeploy` tool, which does a similar thing - moves the dynamic libs and remaps the references in `Oni2_editor` to point to them: https://github.com/onivim/oni2/blob/69c05a9050cca840c2fe5d082bae7233d3447566/scripts/linux/package-linux.sh#L26

We only run it for `Oni2_editor`, though, and not `Oni2`. `Oni_Core` has some dependencies that need that dynamic-library-packaging to work, like `textmate`, `Revery`, etc. So if the `Oni2.exe` launcher brought in `Oni_Core`, we'd get linker errors when the user tries to run it from a packaged build...

A way we could re-use modules like the `Cli` one, though, is to create another module that has no external dependencies - those would be safe to use between binaries. Something like `Oni_Common` or `Oni_Bootstrap`, with a `dune` file that has nothing but pure-ocaml dependencies.

Alternatively... we could reuse the same binary as a 'launcher' (remove building two binaries: `Oni2_editor` and `Oni2.exe`. The only reason we need the launcher, at the moment, is to detach from the Windows console... When we fix https://github.com/onivim/oni2/issues/523, the need for this split might go away completely.

src/Core/Log.re
bryphe: I think keeping them as separate options is OK 👍

src/Core/Log.re
bryphe: Sounds reasonable to me!

src/Extensions/ExtHostClient.re
bryphe: Oops, bad variables from me...

It should be `scope`, `method`, and `args` - the extension host sends us requests of the form:
`[scope, method, argsAsJson]`, like: `["MainThreadDiagnostics", "$changeMany", diagnosticsJson]`

bryphe:
Change looks great overall!

I'm testing on Windows though and seeing a crash on startup - trying to figure out the cause. Seems like it's crashing somewhere in `Core.Cli.parse`... trying to narrow down exactly where

src/bin_launcher/Oni2.re
glennsl: Ah, thanks for explaining! Could we not just ignore omit the ones we don't need to handle and ignore the `Arg.Bad` exceptions? I assume it'll produce an error when parsed in `Oni2_editor` anyway?

src/Extensions/ExtHostClient.re
glennsl: Fixed in https://github.com/onivim/oni2/pull/999/commits/81033d07085beb2ebf9d234d927f42fc0f8ee0cd

glennsl:
Hmm, are you passing any options? [`setLogFIle`](https://github.com/onivim/oni2/pull/999/files#diff-ef78154a4741a749d1eb8942e9e5e6baR107-R112) seems the most likely culprit to be called from `Cli` as it closes, opens and prints to files. [`setFilter`](https://github.com/onivim/oni2/pull/999/files#diff-ef78154a4741a749d1eb8942e9e5e6baR68-R96) uses `re`, but that should be pure OCaml and seems unlikely to crash, but regexes are pretty complicated stuff.

bryphe:
> Hmm, are you passing any options?

Just `esy run -f`! 

I believe the root cause is that, with the `-f` flag, `Log.isPrintingEnabled()` is still reporting `false` - I'd expect it to return `true` with the `-f` flag.

bryphe:
If I hardcode `Log.isPrintingEnabled` to return `true`, it works great!

image

glennsl:
Oops,

```reason
let isPrintingEnabled = () => Logs.reporter() === Logs.nop_reporter;
```

should be

```reason
let isPrintingEnabled = () => Logs.reporter() !== Logs.nop_reporter;
```

glennsl:
Do you have any opinion on whether we need ad hoc namespaces @bryphe? Ie. the possibility of adding to the namespace when calling the logging functions.

I'm leaning towards not needing it. Adding to the namespace is only necessary if we want to be able to filter on it and have it colored. Otherwise you can just add a prefix to the message, like we've done until now. It's even possible to add color using `fmt`, though it's not very ergonomic.

If a namespace  is really needed for filtering, it's also still possible to create multiple loggers with different namespaces at the top level. I've done so with `ReveryLog` in `Oni2_editor` and `DispatchLog` in `StoreThread` for example. I think that'll do fine for now, and that we can hold off and see over time what we actually need.

bryphe:
> I'm leaning towards not needing it. Adding to the namespace is only necessary if we want to be able to filter on it and have it colored.

Agreed, I'm okay deferring it, until we have a demonstrated need. This is already such a big improvement over what we have today!

Confirmed it works on Windows w/ the latest changes too 👍

src/bin_launcher/Oni2.re
glennsl: Still a bit curious about this @bryphe

CrossR: Unless I'm mistaken, we'd still want the ones we don't want to handle here, such that they show up when you call `oni2 --help`. Unless there is a nicer way of achieveing that here?

(That is what I would say if we had some help text rather than just `""`, but I think currently none of them show up on `oni2 --help` as we haven't populated that string yet. If that is the case, we should probably add something to the strings to help.)

glennsl: There's an exception that happens on `--help` as well, so I think we can just forward that too.

Otherwise, yeah, `--help` doesn't show anything useful right now and it would be nice if it did. We do have a few options and it's starting to get hard to remember them all.

Issue #880, opened 2019-10-21
Feature: Additional variables for titlebar


bryphe:
PR #879 adds a custom titlebar, that can be specified via the `window.title` configuration setting.

VSCode supports a rich set of configuration settings:
```
  // Controls the window title based on the active editor. Variables are substituted based on the context:
  // - `${activeEditorShort}`: the file name (e.g. myFile.txt).
  // - `${activeEditorMedium}`: the path of the file relative to the workspace folder (e.g. myFolder/myFileFolder/myFile.txt).
  // - `${activeEditorLong}`: the full path of the file (e.g. /Users/Development/myFolder/myFileFolder/myFile.txt).
  // - `${activeFolderShort}`: the name of the folder the file is contained in (e.g. myFileFolder).
  // - `${activeFolderMedium}`: the path of the folder the file is contained in, relative to the workspace folder (e.g. myFolder/myFileFolder).
  // - `${activeFolderLong}`: the full path of the folder the file is contained in (e.g. /Users/Development/myFolder/myFileFolder).
  // - `${folderName}`: name of the workspace folder the file is contained in (e.g. myFolder).
  // - `${folderPath}`: file path of the workspace folder the file is contained in (e.g. /Users/Development/myFolder).
  // - `${rootName}`: name of the workspace (e.g. myFolder or myWorkspace).
  // - `${rootPath}`: file path of the workspace (e.g. /Users/Development/myWorkspace).
  // - `${appName}`: e.g. VS Code.
  // - `${remoteName}`: e.g. SSH
  // - `${dirty}`: a dirty indicator if the active editor is dirty.
  // - `${separator}`: a conditional separator (" - ") that only shows when surrounded by variables with values or static text.
  "window.title": "${dirty}${activeEditorShort}${separator}${rootName}${separator}${appName}",
```

However, we only support the following right now:
- `${activeEditorShort}`
- `${separator}`
- `${rootName}`
- `${dirty}`

This is tracking adding the additional variables. The place these would be added is here: https://github.com/onivim/oni2/blob/d99801e4c3c7a8e00c774d7befd2e0449c1d26b6/src/Store/TitleStoreConnector.re#L15

The `getTemplateVariables` takes our application state as a function, and returns a list of template variables -> values.  We have both the workspace path as `state.workspace.rootName` and the active buffer file path available in that method, which should allow us to derive most of those.

shulhi:
[...]

PR #1001, opened 2019-12-04
Add activeEditorShort and activeFolderShort

shulhi:
This is related to https://github.com/onivim/oni2/issues/880.

I added two additional variables `activeEditorShort` and `activeFolderShort`. I think I can add more of the variables but I'm not sure how workspaces work in oni2 at the moment (I'll probably create another PR for this when I have the time).


bryphe:
Looks great @shulhi ! Thank you for the contribution 👍

PR #1002, opened 2019-12-04
Exthost: Configuration - Initial configuration modelling

bryphe:
This picks up the latest `vscode-exthost`, which now requires a configuration be specified. And this creates type info to model configuration for the vscode extension API.

This will be used to specify a few configurations for incoming extensions.

### DIFF
--- a/src/Extensions/Configuration.re
+++ b/src/Extensions/Configuration.re
@@ -1,0 +1,87 @@
+/*
+ * Configuration.re
+ * 
+ * Types related to creating configuration
+ *
+ */
+
+let emptyJsonObject = `Assoc([]);
+let emptyJsonArray = `List([]);
+
+// Type relating to 'ConfigurationModel' in VSCode
+// This is an 'instance' of configuration - modelling user, workspace, or default configuration.
+// The full configuration is set up by combining the various configuration 'instances'.
+module Model = {
+
+  type t = {
+    contents: Yojson.Safe.t, 
+    keys: list(string),
+    // TODO: Investigate hooking up the 'overrides' setting
+    // overrides: ?
+  };
+
+  let empty = {
+    contents: emptyJsonObject,
+    keys: [], 
+  }
+
+  let create = (~keys, contents) => {
+    keys,
+    contents,
+  }
+
+  let to_yojson = (model) => {
+   let { contents, keys } = model;
+ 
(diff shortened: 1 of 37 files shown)


src/Extensions/Configuration.re
```diff
@@ -0,0 +1,76 @@
+/*
+ * Configuration.re
```
bryphe: @glennsl - do you have any opinions on naming?

With this change - we'll have `src/Core/Configuration.re` and `src/Extensions/Configuration.re`. With the extensions folder, we have some prefixed with `Ext` - but that gets verbose if you alias the module, like:
```
module Ext = Oni_Extensions;

let transport = Ext.ExtHostTransport.create();
```

We could either start removing the prefixes for the `ExtHost*` files in `src/Extensions`, or consistently use them - wanted to see if you have a preference.

glennsl: I definitely prefer this style, where module names do not have prefixes, but are contextual and relatively short internally where they're used the most.

If there's a naming collision you can usually disambiguate easily by just prefixing the project, even with opened modules, like e.g. `Oni_Core.Configuration`.

I also prefer aliasing project modules rather than opening them because it pollutes the namespace and makes it harder to identify where modules come from, especially when you're unfamiliar with the code base. Although with some exceptions like `Oni_Core`, `Oni_Model` and `Revery` when they're heavily used. I also prefer to avoid abbreviations, because they can easily become non-descriptive and unpredictable, but `Ext` might be an exception because it seems like a reasonably predictable and descriptive convention. Abbreviations like `Mod` or `M` for `Model` is a no-go though.

I don't have a very strong opinion on most of this, except for general use of `open` and abbreviations, but this is my preference right now at least. Thanks for asking!

bryphe: Cool, thanks for the feedback! 

> I definitely prefer this style, where module names do not have prefixes, but are contextual and relatively short internally where they're used the most.

This is what I'm leaning towards now too - after experimenting w/ the baking-of-prefixes in the `Extensions` model. I added a note in the 'Technical' project to track removing the `Ext` extensions.

> I also prefer aliasing project modules rather than opening them because it pollutes the namespace and makes it harder to identify where modules come from, especially when you're unfamiliar with the code base.

Makes sense!

PR #1007, opened 2019-12-05
Add copyright / license for the logs library

bryphe:


bryphe:
> Should licenses for re and fmt be added as well?

@glennsl - oh, good catch! Yes, we need to include all the top-level dependencies licenses. I'll bring them over as part of this PR.

PR #1021, opened 2019-12-08
Only move the selected line if there is actually a line to move to.

CrossR:
This just helps sort out issues I was seeing where you could click outside the buffer and the selected line would seem to move. (I.e. open Oni2 and just click in the buffer that isn't line 1).

For now I've gone with a "don't move at all" but I could see that "move to the last line" would also be nice, so any opinions?


bryphe:
Ah, nice catch @CrossR , thank you for the fix!

> For now I've gone with a "don't move at all" but I could see that "move to the last line" would also be nice, so any opinions?

I don't have a strong opinion here, the "don't move at all" behavior seems reasonable to me - we can change it down the road if there is a compelling reason 👍

PR #1022, opened 2019-12-09
File explorer focus

glennsl:
This highlights the current buffer in the file explorer, opens all directories in the path to the current buffer and scrolls the view so it's visible. You'd think this would be pretty straight forward, but unfortunately it is not.

One issues is that the path to the current buffer might only be partially loaded. We therefore need to traverse the tree until encountering an unloaded directory, load that asynchronously, wait for the result and then try again. This works fine as it is now, but is a bit complicated.

Another issue is that it should only scroll to the focused item on `BufferEnter`, but currently does so whenever the focus changes AND even just when opening a directory since that updates the tree including node ids and hence causes comparison of the node path to fail. The root issue here is that the scroll position is kept in the `TreeView` component, but the `BufferEnter` event happens outside it. The only solution I've found so far is to move the scroll position out of the component and into the `FileExplorer` model, but since we need pixel precision and only the component knows about that, we need to store the position as a float with whole numbers representing the start of an item, and compute the pixel position from that. We might also want to reuse node ids so tree comparison doesn't break down, although if the scroll position is moved out I don't think we need this here at least.

A third issue is that there seems to be some blocking on `BufferEnter`, causing a delay in updating the explorer until the editor has finished loading and rendering.

There's also room for performance improvement here if needed by using hashing instead of direct string comparison, for example, but unless it's really needed I think this is complicated enough as it is.

### DIFF
--- a/src/Core/BufferUpdate.re
+++ b/src/Core/BufferUpdate.re
@@ -7,7 +7,7 @@ type t = {
   id: int,
   startLine: Index.t,
   endLine: Index.t,
-  lines: array(string),
+  lines: [@opaque] array(string),
   version: int,
 };
 
--- a/src/Model/FileExplorer.re
+++ b/src/Model/FileExplorer.re
@@ -9,8 +9,9 @@ type t = {
 
 [@deriving show({with_path: false})]
 type action =
-  | TreeUpdated([@opaque] FsTreeNode.t)
-  | NodeUpdated(int, [@opaque] FsTreeNode.t)
+  | TreeLoaded([@opaque] FsTreeNode.t)
+  | NodeLoaded(int, [@opaque] FsTreeNode.t)
+  | FocusNodeLoaded(int, [@opaque] FsTreeNode.t)
   | NodeClicked([@opaque] FsTreeNode.t);
 
 module ExplorerId =
--- a/src/Model/FileExplorer.re
+++ b/src/Model/FileExplorer.re
@@ -134,6 +134,6 @@ let getDirectoryTree = (cwd, languageInfo, iconTheme, ignored) => {
 let initial = {
   tree: None,
   isOpen: true,
-  focus: None,
   scrollOffset: `Start(0.),
+  focus: None,
 };
--- a/src/Model/FileExplorer.re
+++ b/src/Model/FileExplorer.re
@@ -4,6 +4,7 @@ open Oni_Core;
 type t = {
   tree: option(FsTreeNode.t),
   isOpen: bool,
+  scrollOffset: float,
   focus: option(string) // path
 };
 
@@ -12,7 +13,8 @@ type action =
   | TreeLoaded([@opaque] FsTreeNode.t)
   | NodeLoaded(int, [@opaque] FsTreeNode.t)
   | FocusNodeLoaded(int, [@opaque] FsTreeNode.t)
-  | NodeClicked([@opaque] FsTreeNode.t);
+  | NodeClicked([@opaque] FsTreeNode.t)
+  | ScrollOffsetChanged(float);
 
 module ExplorerId =
   UniqueId.Make({});
@@ -129,4 +131,4 @@ let getDirectoryTree = (cwd, languageInfo, iconTheme, ignored) => {
   );
 };
 
-let initial = {tree: None, isOpen: true, focus: None};
+let initial = {tree: None, isOpen: true, focus: None, scrollOffset: 0.};
--- a/src/Model/FileExplorer.re
+++ b/src/Model/FileExplorer.re
@@ -4,6 +4,7 @@ open Oni_Core;
 type t = {
   tree: option(FsTreeNode.t),
   isOpen: bool,
+  focus: option(string) // path
 };
 
 [@deriving show({with_path: false})]
@@ -127,4 +128,4 @@ let getDirectoryTree = (cwd, languageInfo, iconTheme, ignored) => {
   );
 };
 
-let initial = {tree: None, isOpen: true};
+let initial = {tree: None, isOpen: true, focus: None};
--- a/src/Model/FileExplorer.re
+++ b/src/Model/FileExplorer.re
@@ -4,7 +4,7 @@ open Oni_Core;
 type t = {
   tree: option(FsTreeNode.t),
   isOpen: bool,
-  scrollOffset: float,
+  scrollOffset: [ | `Start(float) | `Middle(float)],
   focus: option(string) // path
 };
 
@@ -14,7 +14,7 @@ type action =
   | NodeLoaded(int, [@opaque] FsTreeNode.t)
   | FocusNodeLoaded(int, [@opaque] FsTreeNode.t)
   | NodeClicked([@opaque] FsTreeNode.t)
-  | ScrollOffsetChanged(float);
+  | ScrollOffsetChanged([ | `Start(float) | `Middle(float)]);
 
 module ExplorerId =
   UniqueId.Make({});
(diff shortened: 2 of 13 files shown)


src/Model/FileExplorer.re
```diff
@@ -4,12 +4,14 @@ open Oni_Core;
 type t = {
   tree: option(FsTreeNode.t),
   isOpen: bool,
+  focus: option(string) // path
```
glennsl: The focus is stored as absolute filesystem path, since the node might not be loaded, or even in the workspace at all, and we probably want to preserve it if we change the workspace root too. It also makes it very easy to set in `BufferEnter` and elsewhere.

src/UI/FileTreeView.re
```diff
@@ -75,31 +72,55 @@ let nodeView = (~font: Core.Types.UiFont.t, ~fg, ~bg, ~node: FsTreeNode.t, ()) =
...
+      let localPath = path |> Str.replace_first(re, "");
+
+      switch (FsTreeNode.findNodesByLocalPath(localPath, tree)) {
+      | `Success(nodes) => Some(nodes)
+      | `Partial(_)
+      | `Failed => None
+      };
+
+    | _ => None
+    };
```
glennsl: Note to self: Extract to a function with a descriptive name (`fsPathToNodePath`?)

src/UI/TreeView.re
```diff
@@ -81,23 +89,69 @@ module Make = (Model: TreeModel) => {
       style=Style.[width(Constants.arrowSize), height(Constants.arrowSize)]
     />;
 
+  // COunts the number of axpanded nodes before the node specified by the given path
```
glennsl: Note to self: typo

src/UI/TreeView.re
```diff
@@ -81,23 +89,69 @@ module Make = (Model: TreeModel) => {
...
+            loopChildren(1, Model.children(node));
+          };
+        }
+      };
+
+    switch (loop(tree, path)) {
+    | `Found(count) => Some(count)
+    | `NotFound(_) => None
+    };
+  };
```
glennsl: note to self: Try to factor out a more general tree traversal function from this and `findNodesByLocalPath`

src/UI/TreeView.re
```diff
@@ -177,6 +235,22 @@ module Make = (Model: TreeModel) => {
...
+          |> Option.bind(nodeOffsetByPath(tree))
+          |> Option.iter(offset => {
+               // center focused item
+               setScrollTop(_ =>
+                 offset * itemHeight - menuHeight / 2
+               )
+             });
+          None;
+        },
+      );
```
glennsl: This doesn't really do its job, because `focus` changes structurally whenever part of the tree is loaded, because we don't reuse ids and of course because the tree actually changes structurally.

Structural comparison on `FsTreeNode`'s is also amazingly inefficient (exponential?). If we preserve node ids we can just compare those, but I don't think the effect hook approach is viable here anyway since we need to distinguish the origin of the change (`BufferEnter` vs anything else), which we can't do here.

glennsl: Fixed by allowing scrollTop to be controlled externally

src/Store/QuickmenuStoreConnector.re
```diff
@@ -49,12 +49,7 @@ let start = (themeInfo: Model.ThemeInfo.t) => {
     });
 
   let makeBufferCommands = (languageInfo, iconTheme, buffers) => {
-    let currentDirectory = Rench.Environment.getWorkingDirectory();
-
-    let getDisplayPath = (fullPath, dir) => {
-      let re = Str.regexp_string(dir ++ Filename.dir_sep);
-      Str.replace_first(re, "", fullPath);
-    };
+    let currentDirectory = Rench.Environment.getWorkingDirectory(); // TODO: This should be workspace-relative
```
glennsl: I found a few of these, which are going to break functionality when you `:cd`

src/Core/BufferUpdate.re
```diff
@@ -7,7 +7,7 @@ type t = {
   id: int,
   startLine: Index.t,
   endLine: Index.t,
-  lines: array(string),
+  lines: [@opaque] array(string),
```
bryphe: Thanks for adding this! The buffer updates could be noisy in the logs, especially for large files

src/Store/QuickmenuStoreConnector.re
bryphe: Ah, good catch

src/UI/FileTreeView.re
```diff
@@ -75,32 +82,59 @@ let nodeView = (~font: Core.UiFont.t, ~fg, ~bg, ~node: FsTreeNode.t, ()) => {
...
   let State.{theme, uiFont as font, _} = state;
 
   let bg = state.theme.sideBarBackground;
   let fg = state.theme.sideBarForeground;
 
+  let focus =
+    switch (focus, state.workspace) {
+    | (Some(path), Some(workspace)) => toNodePath(workspace, tree, path)
+    | _ => None
+    };
```
bryphe: Is this expensive to run every time we render render?

bryphe:
Really cool @glennsl ! This is a huge quality of life improvement for visualizing the source tree while working.

I noticed one bug when I tested it:
scroll-to-file

In this case when I opened `Node.re` - it expanded the folders correctly, but didn't scroll _all the way_ to it.

src/UI/FileTreeView.re
glennsl: Depends how many nodes need to be traversed. In practice I don't think it'll be a problem because the file explorer will likely become impractical to use before there's a noticeable performance impact. For example, expanding `node_modules` isn't going to have an impact since its contents will be skipped if it's not in the path. You have to select a file deep in the hierarchy for it to be traversed, and then it'll still only traverse the nodes that are before it, expanded and whose parent is in the path.

There are still way we can improve it though. The simplest would be to store a hash of the name and compare those instead. We should probably do that anyway since using `displayName` is based on an assumption that might not hold. We could also precompute it, but that's an invariant that might be tricky to maintain. It goes against "single source of truth" so we have to remember to recompute it every time the `foxus` OR `tree` changes.

glennsl:
> In this case when I opened Node.re - it expanded the folders correctly, but didn't scroll all the way to it.

Fixed in https://github.com/onivim/oni2/pull/1022/commits/c2190b9fcb66cf4970d02ef997986d131c765117. Caused by comparing nodes structurally instead of by `id` which would fail after opening the nodes since that of course changes the structure.

src/UI/FileTreeView.re
bryphe: Thanks for the details @glennsl ! Do we have a way to do 'perf' logging today with our new framework? It might be worth wrapping this in a 'perf' block so we can track it. 

> We could also precompute it, but that's an invariant that might be tricky to maintain. It goes against "single source of truth" so we have to remember to recompute it every time the foxus OR tree changes.

Right, I've been thinking about this in the context of a few other places (like the `EditorSurface`) - sometimes there are things we'd really like to cache, but it creates new sources of truth. I wonder if we could have some sort of `useMemo` hook that takes an object, a comparison function, and an 'expensive' calculation, and caches the result of that calculation (or recomputes) based on the comparison result.

bryphe: No action needed for this PR - just thinking out loud)

glennsl: `Log.perf` is still there. I'll check it out and probably add a hashcode to `FsTreeNode` for comparison.

> I wonder if we could have some sort of useMemo hook that takes an object, a comparison function, and an 'expensive' calculation, and caches the result of that calculation (or recomputes) based on the comparison result.

That's a good idea!. `Memo` seems a bit misleading, but it sounds basically like `Lazy.t` with an invalidation condition wrapped up in a `ref` hook.

Issue #1017, opened 2019-12-07
Various UI Bugs


jordwalke:
I've noticed a few menu bugs and since they might all have the same root cause, I thought I'd file one issue. Please let me know if these specific issues are already being tracked:

- When hitting `command+p` for file search, the list updates and reorders as it finds new
  files, but I think it should be append only.
  - You can repro by `:cd ~/` (a large directory) then hitting `command+p`.
- `command+shift+p` also has a flickering reordering upon open.
- When toggling Explorer, the right hand scroll bar flashes in the middle of
  the editor for one frame.
- `Control+tab` switcher behavior:
  - Reorders once upon opening after a slight delay. (Can create screencast if necessary).
  - Stays open upon key-up. I think it should close.
  - When pressing `control+tab,` and keeping `control` down, you can't hit tab
    repeatedly to cycle through (or `shift+control+tab` to cycle back).
  - Question:
    - When pressing `control+tab` should it be in MRU order?
    - If mru ordering, when pressing `control+tab:`
      - The first entry should always be the current file.
      - The second entry should be the second most recently opened file.
      - When pressing `control+tab,` it should probably initialized with the
        second entry highlighted.
      - When pressing `control+tab` and then releasing `control+tab,` it would
        select the second item in the menu which was highlighted when the menu
        opened.
- For the `:` menu:
  - When I press `:cd ~/D` and press tab I would expect it to fill in to `:cd
    ~/Desktop/`
    - Currently, I have to press `Tab` twice to get it to complete a path segment.
    - It appears to highlight the wrong row. (Screenshot attached)
    - I cannot continue pressing tab in order to get back to my "uncompleted"
      path segment. `:cd ~/D`
  - I think it is supposed to behave more like `wildmode=full`. In this mode,
    `<Tab>`(ctrl-n/ctrl-p) acts like a kind of "temporary" completion, until
    you hit enter or type any character after the temporary completion - at
    which point, that becomes the new content of the input. But if you don't hit enter, or type any more characters, the original contents of the input are secretly retained and you can cycle back to "uncompletedness" if you press `<Tab>/ctrl-n/ctrl-p` enough times.

Here is how `wildmode=full` behaves (I think it would be a good workflow to model the `:` completion off of. This would also be totally compatible with Sublime/VSCode style popup interactions because neither of those make use of the tab / shift+tab keys in their dialogs. That way both can be supported (control+p/n to move up or down, or tab/shift+tab if you are more familiar with wildmode vim defaults).

```
[...]

PR #1025, opened 2019-12-09
Sort Ctrl+Tab menu by MRU

glennsl:
This addresses a few issues from #1017:

- Sorting Ctrl+Tab menu by MRU
- Initial "flickering" in Ctrl+Tab and Ctrl/Cmd+Shift+P menus

The flickering issue was caused by FilterProcessingJob reversing the items while processing them because it uses lists instead of queues when processing. The fix here is just to bypass filtering entirely if there's no filter. Note however when a filter is applied, this will still cause similarly ranked items to appear in reverse order. Probably not an issue in practice, but might be worth considering using a queue here anyway to alleviate it.

Note also that this doesn't fix the non-determinism of the Ctrl/Cmd+P menu. That definitely does need to use a queue during processing.

### DIFF
--- a/src/Model/Buffer.re
+++ b/src/Model/Buffer.re
@@ -15,6 +15,7 @@ type t = {
   lines: array(string),
   indentation: option(IndentationSettings.t),
   syntaxHighlightingEnabled: bool,
+  lastUsed: float,
 };
 
 let show = _ => "TODO";
@@ -28,6 +29,7 @@ let ofLines = (lines: array(string)) => {
   lines,
   indentation: None,
   syntaxHighlightingEnabled: true,
+  lastUsed: 0.,
 };
 
 let empty = ofLines([||]);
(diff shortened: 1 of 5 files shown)


src/Model/Buffer.re
```diff
@@ -15,6 +15,7 @@ type t = {
   lines: array(string),
   indentation: option(IndentationSettings.t),
   syntaxHighlightingEnabled: bool,
+  lastUsed: float,
```
glennsl: This ought to be modeled as an `option(float)`, but I think that's more bother than it's worth. If at some point we do need to know that a buffer has never been used, it'll be most likely be just as simple to change it then.

bryphe:
> The flickering issue was caused by FilterProcessingJob reversing the items while processing them because it uses lists instead of queues when processing.

Nice catch!

PR #1027, opened 2019-12-10
Fix #818 - Non-deterministic Fuzzy search results

### LINKED ISSUE #818: Initial results in ´Ctrl+P´ menu not deterministic.

glennsl:
Fixes #818 
Addresses #1017 

This gist of this is just a switch from using `list` to `Queue.t` for the processing in `FilterJob` and `Ripgrep` to prevent the reversion that naturally occurs when working recursively with `list`s. But since `Queue.t` is a mutable data structure this wasn't just a slot-in replacement. So while I was at it I also took the opportunity to clean these modules up a bit and remove a lot of the imperativeness from them.

### DIFF
--- a/src/Core/Ripgrep.re
+++ b/src/Core/Ripgrep.re
@@ -83,33 +83,26 @@ and dispose = unit => unit;
 module RipgrepProcessingJob = {
   type pendingWork = {
     callback: list(string) => unit,
-    bytes: list(Bytes.t),
+    queue: Queue.t(Bytes.t) // WARNING: mutable data structure
   };
 
-  let pendingWorkPrinter = (p: pendingWork) => {
-    "Byte chunks left: " ++ string_of_int(List.length(p.bytes));
-  };
+  let pendingWorkPrinter = pending =>
+    Printf.sprintf("Byte chunks left: %n", Queue.length(pending.queue));
 
   type t = Job.t(pendingWork, unit);
 
-  let doWork = (pendingWork, c) => {
-    let newBytes =
-      switch (pendingWork.bytes) {
-      | [] => []
-      | [hd, ...tail] =>
-        let items =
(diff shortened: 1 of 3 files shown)


src/Core/Ripgrep.re
```diff
@@ -83,33 +83,26 @@ and dispose = unit => unit;
 module RipgrepProcessingJob = {
   type pendingWork = {
     callback: list(string) => unit,
-    bytes: list(Bytes.t),
+    queue: Queue.t(Bytes.t) // WARNING: mutable data structure
```
bryphe: Cool! I hadn't used `Queue.t` from the stdlib much... interesting to see it in action.

src/Model/FilterJob.re
```diff
@@ -21,215 +21,194 @@ module Make = (Config: Config) => {
...
-    // This never gets filtered - it's persisted in case we need
-    // the full set again
-    explodedFilter: list(UChar.t),
-    shouldLower: bool,
-    totalItemCount: int,
-    allItems: list(list(Config.item)),
-    // Commands to filter are commands we haven't looked at yet.
-    itemsToFilter: list(list(Config.item)),
-  };
+  module PendingWork = {
```
bryphe: Good idea to factor these to modules

src/Core/Ripgrep.re
glennsl: First time I've used it too!

PR #1030, opened 2019-12-10
Various quickmenu bugfixes

glennsl:
Bugs fixed:

- Input prefix multiplied like rabbits. Caused by not properly distinguishing between the actual value and the value to display, which should include the prefix.

- Filtered items where duplicated. Caused by the FilterJob revamp earlier today, although I still don't understand how it could have worked before without this fix either.

- Ripgrep seemed to never complete, causing the loading animation to animate forever and ever. Root cause was that `onComplete` was called when the ripgrep process had completed, but the processing still hadn't, triggering further `onUpdate` calls that set the state to `Loading`.

- FlatList was overflowing its container when displaying many items initially. Caused by the initial height estimation overriding the later measurement. It now uses the estimation only to calculate which items to render, not to set the height of the container.

- Input lost focus when a submenu, such as the theme picker, was selected. Caused by `loseFocus` being called on `ListSelect`, but since the menu, due to reconciliation,  is never removed and remounted `focus` wasn't being called when opening the submenu. Fixed by adding a very hacky `QuickmenuMaybeLoseFocus` action that checks whether `state.quickmenu` is `None` and calls `loseFocus` only after the selected item's command has been dispatched and the state updated.

- Wildmenu completion out of sync. This was caused by an ordering issue between updaters where VimStore applied the completion before QuickmenuStore had updated the focus. This was originally fixed in https://github.com/onivim/oni2/pull/858/commits/44a3f453440254bc649405199ea094a93c12536b, but seems to have been lost in some rebase.

Addresses parts of #1017

### DIFF
--- a/src/Store/QuickmenuStoreConnector.re
+++ b/src/Store/QuickmenuStoreConnector.re
@@ -28,6 +28,7 @@ let start = (themeInfo: Model.ThemeInfo.t) => {
     Isolinear.Effect.createWithDispatch(~name="quickmenu.selectItem", dispatch => {
       let action = item.command();
       dispatch(action);
+      dispatch(Actions.QuickmenuMaybeLoseFocus);
     });
 
   let executeVimCommandEffect =
@@ -260,7 +261,6 @@ let start = (themeInfo: Model.ThemeInfo.t) => {
       )
 
     | ListSelect =>
-      Revery_UI.Focus.loseFocus(); // TODO: Remove once revery-ui/revery#412 has been fixed
       switch (state) {
       | Some({variant: Wildmenu(_), _}) => (None, executeVimCommandEffect)
 
@@ -271,7 +271,7 @@ let start = (themeInfo: Model.ThemeInfo.t) => {
         }
 
       | _ => (state, Isolinear.Effect.none)
-      };
+      }
 
     | ListSelectBackground =>
       switch (state) {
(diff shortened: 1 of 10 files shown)


glennsl:
This CI failure is extremely weird. Does it do an automatic rebase or merge if there's no conflicts?

bryphe:
> This CI failure is extremely weird. Does it do an automatic rebase or merge if there's no conflicts?

Ah, ya, looks like this was a silent merge conflict with #1029 - it does merge with `master` for the PR builds if there are no conflicts.

Great set of fixes!

> Ripgrep seemed to never complete, causing the loading animation to animate forever and ever.

I was wondering about this, glad it's fixed 👍 

> Wildmenu completion out of sync. This was caused by an ordering issue between updaters where VimStore applied the completion before QuickmenuStore had updated the focus.

This synchronization is challenging (the update-ordering issues are really rough) - these sorts of issues make me wonder if it's better for us to manage the state outside of Vim. It's a tradeoff - getting the Vim aspects 'for free' but dealing with the synchronization... or avoiding the synchronization and having more logic in our UX.

src/Store/QuickmenuStoreConnector.re
```diff
@@ -285,6 +286,14 @@ let start = (themeInfo: Model.ThemeInfo.t) => {
       | _ => (state, Isolinear.Effect.none)
       }
 
+    // Triggered by selectItemEffect in order to lose focus iff the item command
```
bryphe: Thanks for the comment  explaining it 👍

glennsl:
> these sorts of issues make me wonder if it's better for us to manage the state outside of Vim.

Yeah, I was originally thinking it could be better handled with subscriptions, which would ensure the right ordering, but the vim interaction here with a lot of `Vim.input` calls make that difficult and prone to infinite loops.

bryphe:
> but the vim interaction here with a lot of Vim.input calls make that difficult and prone to infinite loops.

Ah, ya... We might be able to provide a different level of abstraction for this. If we control more of the UX - and manage the state - we could potentially use less `Vim.input` (at the cost of managing more of the state ourselves). 

An alternate 'boundary' for this, that I could see, is using the menu essentially as-is, but then have a subscription that subscribes to a 
```
let completions = Vim.CommandLine.getCompletions(commandLineString, cursorPosition);
```

And then, only when we commit, do we send a `Vim.command(..)`.

bryphe:
...just need some work to tweak the APIs on the libvim / reason-libvim side

PR #1031, opened 2019-12-11
Refactoring - Extensions: Move Language Features to Model

bryphe:
As I was working on the completions and the definition provider, I realized that some of our language integrations were 'tightly coupled' to the VSCode extension implementation. For example,  our `CompletionProvider` was defined by a VSCode extension id and metadata (a `DocumentSelector`).

However, it's desirable to have these language features decoupled from the VSCode extension host implementation, for a few reasons:
- Less complicated, easier to reason about in isolation
- We can create features like completion providers external to the VSCode extension host (a great use case would be _keyword completion_!)... or even expose Reason/OCaml APIs for them, independent of the extension host.
- Easier to test

I also tightened up the nomenclature - most of the code we were using `Completion` vs `Suggest` - I left Suggest in the VSCode specific parts (although, it's inconsistent even there - the extension API uses completion, while other places, like the configuration or input, use suggest).

I started this for the definition provider in #1026, but it was awkward, in that now the `CompletionProvider` and `DefinitionProvider` operated at different levels of abstraction - the `CompletionProvider` was coupled to the extension host, but the `DefinitionProvider` was not.

This change pivots the idea of a `CompletionProvider` - from being something defined in `Oni_Extensions` that is coupled to the VSCode extension host model, to a function: 
```
type t = (Buffer.t, CompletionMeet.t, Position.t) => option(Lwt.t(list(CompletionItem.t)))
```

It might look a bit strange to have this return an `option(Lwt.t(...))`, but this gives completion providers a way to filter - if they don't want to handle a particular buffer (based on filetype, etc), they can just return `None`.

Then, we have a separate `ExtensionCompletionProvider` that, given an extension host and some registration info, can create a `CompletionProvider.t`

As I moved the `LanguageFeatures.t` to `Model` from `Extensions`, a few other changes had to be made to break circular dependencies - I pulled out the `Actions.completionMeet` (seems like there was duplication between that and `CompletionMeet.t`), as well as `Actions.completionItem` and factored that to a `CompletionItem.t`

__TODO:__ 
- [x] Get tests green again - need to bring back a deterministic way to see a completion provider registered.


src/Model/CompletionMeet.re
```diff
@@ -28,6 +28,20 @@ let show = (v: t) =>
 
 let defaultTriggerCharacters = [UChar.of_char('.')];
 
+let getLine = maybeMeet => {
```
bryphe: Add REI file

src/Model/LanguageFeatures.re
```diff
@@ -0,0 +1,77 @@
+/*
+ * LanguageFeatures.re
+ *
+ * This module documents & types the protocol for communicating with the VSCode Extension Host.
+ */
+
+open Oni_Core;
```
bryphe: Add REI file

src/Model/LanguageFeatures.re
```diff
@@ -0,0 +1,77 @@
...
+let _identity = v => v;
+
+let requestCompletions =
+    (~buffer: Buffer.t, ~meet: CompletionMeet.t, ~position: Position.t, lf: t) => {
+  let promises =
+    lf.completionProviders
+    |> List.map(suggestor => suggestor(buffer, meet, position))
+    |> Utility.List.filter_map(_identity);
+
+  let completorCount = ref(List.length(promises));
```
bryphe: This is a little messy... I wasn't sure if there was a built-in method in Lwt that does this (I couldn't find it, searching around the docs).

I wanted to return a promise that completes when all promises are resolved, joining the successful results with a function `f` (and in this case, `f` is just list concatenation).

src/Store/CompletionStoreConnector.re
```diff
@@ -14,7 +14,11 @@ module Quickmenu = Model.Quickmenu;
 
 open Actions;
 
-type lastCompletionMeet = Actions.completionMeet;
+type lastCompletionMeet = {
```
bryphe: There is still further refactoring that could be done in this file (still some duplication left between `Model.CompletionMeet.t` and `CompletionStoreConnector.lastCompletionMeet` types)

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -181,100 +240,95 @@ let start = (extensions, setup: Core.Setup.t) => {
...
-               Lwt.return();
-             },
+    Lwt.bind(
+      completionPromise,
+      completions => {
+        dispatch(Model.Actions.CompletionAddItems(meet, completions));
+        Lwt.return();
+      },
+    );
+    /*providers
```
bryphe: Remove now unused code

src/Model/LanguageFeatures.re
```diff
@@ -0,0 +1,77 @@
...
+
+module CompletionProvider = {
+  type t =
+    (Buffer.t, CompletionMeet.t, Position.t) =>
+    option(Lwt.t(list(CompletionItem.t)));
+};
+
+type t = {completionProviders: list(CompletionProvider.t)};
+
+let empty = {completionProviders: []};
```
bryphe: This type gets stored in the `State` - and there are some functional values here (as the `CompletionProvider.t` is a function). @glennsl - I believe you had some concerns about storing functions on the `State` (if I recall correctly), so wanted to run this by you.

src/Model/LanguageFeatures.re
glennsl: Like `Promise.all`? You should be able to do it using `List.fold_left` and `Lwt.bind` I would think. Something like:

```reason
List.fold_left(
  (acc, promise) => {
    let%lwt completions = acc;
    let%lwt completion = promise;
    Lwt.return([completion, ...completions]);
  }),
  Lwt.return([]),
  promises
);
```

FYI: You can drag the "add comment button" to apply a comment to a block of text. It's easier to see what specifically you're referring to then.

src/Store/CompletionStoreConnector.re
```diff
@@ -14,7 +14,11 @@ module Quickmenu = Model.Quickmenu;
...
+  completionMeetBufferId: int,
+  completionMeetLine: Index.t,
+  completionMeetColumn: Index.t,
```
glennsl: Why not `Position.t`?

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -21,6 +21,60 @@ module Protocol = Extensions.ExtHostProtocol;
...
+           Some(
+             ExtHostClient.getCompletions(
+               suggestProvider.id,
+               uri,
+               position,
+               client,
+             )
+             |> Lwt.map(suggestionsToCompletionItems),
+           )
+         };
```
glennsl: `if` expressions still work fine in Reason (and isn't any less functional) ;) I've noticed a pattern with beginners especially tending to prefer pattern matching over `if` expressions. Probably just to feel it out, but `if` expressions tend to be better suited for boolean conditions because it's a less general construct, it puts the condition and "consequence" closer together, and it evolves more easily when more conditions are required.

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -70,9 +124,14 @@ let start = (extensions, setup: Core.Setup.t) => {
     );
   };
 
-  let onRegisterSuggestProvider = (sp: LanguageFeatures.SuggestProvider.t) => {
+  let onRegisterSuggestProvider = (client, sp) => {
```
glennsl: `sp` is too abbreviated I think. I would just use `provider` here, since the context strongly suggests (^^) that a `SuggestProvider` is the primary kind of `provider` here.

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -181,100 +240,95 @@ let start = (extensions, setup: Core.Setup.t) => {
...
-               };
-               Lwt.return();
-             },
+    Lwt.bind(
+      completionPromise,
+      completions => {
+        dispatch(Model.Actions.CompletionAddItems(meet, completions));
+        Lwt.return();
+      },
+    );
```
glennsl: Does this need to return a promise? I see it's just discarded later anyway, so seems like this could use `Lwt.on_success` instead.

src/Model/LanguageFeatures.re
glennsl: Why is this a module? Is there an actual (future) need for it, or is it just because VSCode/LSP organizes it this way? Could it not just be a `Completions` module and have other "language features" as separate modules as well?

src/Model/LanguageFeatures.re
glennsl: Yeah, functions don't have a "structure" like other values. They can't be printed, they can't be compared and as closures can hold state that you cannot see, which can cause bugs that are very hard to track down. If you try to compare functions structurally it will just crash, which isn't great, and might also remove opportunities for optimizations like only rendering the parts of the UI whose state has changed for example.

I also think it's largely unnecessary. Unless you need dynamic linking you'll have all the functions available at compile-time. If you need "Inversion of Control" or "dependency injection" you can do so using partial application, a closure or a functor along with some associated value, a string, generated unique id or proper variant, that can be put in the model and actions.

Or if the limitations of interacting with an imperative API makes that difficult, it might be better to treat the entire thing as external. Abstract it so you have a single completion provider that can register other providers or be swapped out with a dummy. Request completions through an effect instead of an ordinary function and deliver the results via an action.

Or isolate the imperativeness to the extension host API, if that's likely the only "culprit", by having a single completion provider for the extension host which handles registering and such.

src/Extensions/ExtHostClient.re
```diff
@@ -20,6 +20,7 @@ type t = {transport: ExtHostTransport.t};
 type simpleCallback = unit => unit;
 let defaultCallback: simpleCallback = () => ();
 let defaultOneArgCallback = _ => ();
+let defaultTwoArgCallback = (_, _) => ();
```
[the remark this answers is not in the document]
bryphe: I replaced them with the shorter `noop`/`noop0`/`noop1` - it's shorter at least, hopefully clearer too.

src/Extensions/ExtHostClient.re
```diff
@@ -41,16 +42,28 @@ let start =
...
     ) => {
+  // Hold onto a reference of the client, so that we can pass it along with
+  // '$register' actions.
+  let client: ref(option(t)) = ref(None);
+
+  let withClient = f =>
+    switch (client^) {
+    | None => ()
+    | Some(client) => f(client)
+    };
```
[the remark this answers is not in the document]
bryphe: Good point... we had an `apply` function here that was basically `Option.iter` as well. Replaced both.

src/Extensions/ExtHostClient.re
```diff
@@ -112,6 +125,7 @@ let start =
       setup,
     );
   let ret: t = {transport: transport};
```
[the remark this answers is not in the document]
bryphe: I think it had more originally, and then got trimmed, but still left as a record - I streamlined this now

src/Extensions/ExtHostClient.re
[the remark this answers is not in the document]
glennsl: They could be nice to have in `Utility` perhaps, along with an `identity` or `id` function.

src/Model/LanguageFeatures.re
bryphe: > Why is this a module? (Edit: Or rather, why is it called LanguageFeatures?) Is there an actual (future) need for it, or is it just because VSCode/LSP organizes it this way? 

Just seemed like a natural grouping for now - I could see splitting it up though (maybe in the future as we have more in here, and can see the boundaries clearly). I was going to add some other 'language features' here:
- DefinitionProvider
- FindAllReferencesProvider
- DocumentHighlightsProvider
- DocumentSymbolProvider
- WorkspaceSymbolProvider
etc

> Could it not just be a `Completions` module and have other "language features" as separate modules as well?
We actually have a `Completions` module right now, that tracks the state of completion _items_. So we're differentiating between the 'provider' (the strategy for obtaining completions), vs the actual completions available to the user. 

It might be though we could split it out into a `CompletionsProvider` module - but I'd prefer to defer this until we have more examples here and a better overview of the constraints for refactoring.

src/Model/LanguageFeatures.re
```diff
@@ -0,0 +1,77 @@
...
+
+let _identity = v => v;
+
+let requestCompletions =
+    (~buffer: Buffer.t, ~meet: CompletionMeet.t, ~position: Position.t, lf: t) => {
+  let promises =
+    lf.completionProviders
+    |> List.map(suggestor => suggestor(buffer, meet, position))
+    |> Utility.List.filter_map(_identity);
```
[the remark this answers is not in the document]
bryphe: Cool, this would be handy! Added it

src/Model/LanguageFeatures.re
bryphe: Cool, that's much nicer, thanks @glennsl !

> FYI: You can drag the "add comment button" to apply a comment to a block of text. It's easier to see what specifically you're referring to then.

Oh.. nice.

src/Store/ExtensionClientStoreConnector.re
bryphe: Good catch

src/Extensions/ExtHostClient.re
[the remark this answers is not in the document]
bryphe: Moved to `Utility` 👍

src/Model/LanguageFeatures.re
bryphe: Thanks for the details @glennsl ! Really appreciate your thoughts.

These concerns make sense to me:
1) Can't be printed
2) Can't be compared
3) Hold invisible state / hard-to-track bugs
4) Can't compare state
5) Can't persist state
6) Breaks incremental rendering

It seems like having an 'id' could address 1 & 2, and having logging around / in could help 3.

The comparing the state / can't persist state 4 & 5 is a price that we pay - but the comparison today we only are using a reference equality (shallow equality) check at the top-level, so I'm not as concerned about that piece (persistence would need to be custom - we won't persist the whole state anyway, but maybe subsets / pieces of it).

The incremental rendering is a good point, but not something we support yet.

> If you need "Inversion of Control" or "dependency injection" you can do so using partial application, a closure or a functor along with some associated value, a string, generated unique id or proper variant, that can be put in the model and actions.

Wouldn't a closure or a functor have the same issues? Trying to understand the difference - is it having the associated value / identifier / some proper variant that can at least help with logging?

> Abstract it so you have a single completion provider that can register other providers or be swapped out with a dummy.

This is (sort of) what we have with this refactoring - the `LanguageFeatures.requestCompletions` is the 'aggregate provider' that registers other providers. The key idea is what constitutes a completion provider across various strategies? Like the extension host has one way of getting completions (call `provideSuggestions` API), and keyword completion has another (consulting a per-buffer, per-filetype map to see what keywords are available) - the simplest common way that I can see, right now, to express these is a function. Pretty easy for both those places to implement that completion provider function and register it.

> Or isolate the imperativeness to the extension host API, if that's likely the only "culprit", by having a single completion provider for the extension host which handles registering and such.

This was essentially the pre-refactoring design, if I understand correctly - but it doesn't extend well to things like keyword completion (registering a separate provider that can get completions from syntax highlighting)

src/Model/LanguageFeatures.re
```diff
@@ -0,0 +1,77 @@
...
+ * This module documents & types the protocol for communicating with the VSCode Extension Host.
+ */
+
+open Oni_Core;
+
+module CompletionProvider = {
+  type t =
+    (Buffer.t, CompletionMeet.t, Position.t) =>
+    option(Lwt.t(list(CompletionItem.t)));
+};
```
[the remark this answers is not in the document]
bryphe: I'll add an `id` at the minimum to it - I need some better way to track it anyway to get the integration test green (to deterministically wait for the completion providers to be registered)

src/Model/LanguageFeatures.re
glennsl: Cool, it's definitely not a big deal, but if there' such a thing as "premature organization" I think this might be it :wink: It just seems to me to add indirection for no real purpose, and I guess I tend to start in the other end, creating individual pieces that I might later compose if needed.

bryphe:
I added a card in the Technical project (https://github.com/onivim/oni2/projects/10#card-30447137) to track some remaining issues:
- Splitting up `LanguageFeatures` to separate modules
- Moving from the 'function' implementations stored in State to an alternative design

I think as we get a few more of the integrations hooked up from the Extension Host, as well as some non-Extension Host integrations (like keyword completion), the constraints will be clearer and we can look at further improvements in the design.

src/Model/LanguageFeatures.re
bryphe: Cool, thanks for the feedback! I'll track revisiting it here: https://github.com/onivim/oni2/projects/10#card-30449023

src/Model/LanguageFeatures.re
glennsl: > ... the comparison today we only are using a reference equality (shallow equality) check at the top-level, so I'm not as concerned about that piece

I'd be more concerned about a rogue unintentional structural equality check causing a crash. That can very easily happen. Though it might be an effective way of revealing what might otherwise have been a performance problem. But I think users will prefer a minor performance hit over a crash.

> Wouldn't a closure or a functor have the same issues?

If you put them in the state they would. But that's not what I'm suggesting.

> is it having the associated value / identifier / some proper variant that can at least help with logging?

No, I mean as a way of doing dependency injection without putting the dependencies in the state. From what I can see there are two different dependency relations though, and they seem pretty entangled here, so let me try to untangle and explain them separately:

One is between the model/store and the "first level" of completion providers such as `ExtensionCompletionProvider` and `KeywordCompletionProvider`. You probably don't want the model and store to depend on them, and you want to be able to switch them out for testing. For this you can use a functor, partial application or closure to inject the dependencies in `StoreThread` when you wire everything up. This is done elsewhere where `*StoreConnector.start` takes dependencies as arguments and the updater closes over them.

The other dependency relation is really just between `ExtensionCompletionProvider` and the extensions themselves, which, as I understand, uses the `onSuggestionProvider` to register themselves. And it does so by providing an `id` and a set of capabilities, not a callback function. So I'm not sure why you're turning it into a function and storing that, when the id and capabilities could be stored instead.

However, if for some reason you _do_ need something like a registry of functions, what I'm suggesting is to keep the complexities of that external to the general CompletionProvider mechanism. Keep the dirt in the corner instead of spreading it around the room. It ought to be a concern solely for the specific `CompletionProvider`, which then provides a single interface for all extension completion providers. Treat it as a side-effect that just dispatches an action with results when it's done.

> the simplest common way that I can see, right now, to express these is a function

Functions may be easy, but I don't think they're simple. It's probably the most complex construct in the language, which is why they're so hard to serialize and compare, and because they can do pretty much anything they're hard to debug. Data is simple, and you can easily map data to code. If polymorphism is needed, a GADT with a constructor for each completion provider with an existential to store provider-specific state is much simpler than a function, but definitely not as easy. From what I can tell a simple variant should be sufficient though. Then you select the code to run by using a simple switch:

```rust
switch (completionProvider) {
| Extension(id) => ExtensionCompletionProvider.getCompletions(id)
| Keyword(state) => KeywordCompletionProvider.getCompletions(state)
};
```

I don't see why it should need to be any more complicated than this, though of course that certainly doesn't mean there isn't good reason.

bryphe: Thanks for the thoughtful & detailed response, @glennsl !

I think this is the crux of our different perspectives:

> Functions may be easy, but I don't think they're simple. It's probably the most complex construct in the language, which is why they're so hard to serialize and compare, and because they can do pretty much anything they're hard to debug. Data is simple, and you can easily map data to code. If polymorphism is needed, a GADT with a constructor for each completion provider with an existential to store provider-specific state is much simpler than a function, but definitely not as easy.

In the current iteration of design w/ this feature - at this time, I do think that the functions are a simpler expression of these integrations than the proposed alternatives (GADT or variants), even with the trade-offs we've discussed, and will be the easiest path forward (at least for me!) to implement the ExtHost integration. 
 
But it certainly is very helpful to review the trade-offs and examine alternate designs, and I am open to refactoring to an alternate design, especially as we have more examples in place.

So I'd prefer to move forward with the function approach for now - but I'm open to revisiting in the future - I put more details in a card specific to the 'functional' parts of this module: https://github.com/onivim/oni2/projects/10#card-30447137

bryphe:
I'll bring this in, so that I can unblock some of the additional extension-host integration work - but will track revisiting the design in:
https://github.com/onivim/oni2/projects/10#card-30447137
https://github.com/onivim/oni2/projects/10#card-30449023

PR #1032, opened 2019-12-11
[docs] Fix key-bindings example [noci]

glennsl:


bryphe:
Ouch, thanks for fixing this @glennsl !

bryphe:
CI failure unrelated; will bring this in now

PR #1033, opened 2019-12-12
Refactor to use editor-core-types

glennsl:
There is a non-zero probability that something may cause a crash due to the extra check when creating one-based indexes. I've fixed a few already, caused by mistakes in translation. They're easy to find and fix though.


bryphe:
I'll bring this in now, just so you don't have to deal with any merge / rebases with other PRs coming in 👍

Issue #849, opened 2019-10-14
UI: Notification experience improvements


bryphe:
@jordwalke had some great ideas around improving our notifications story in the Discord.

A persistent number of 'new messages' that are 'unread' in the statusbar. 
![image](https://user-images.githubusercontent.com/13532591/66783220-53c8a000-ee8c-11e9-8c94-21ef893027ed.png)

Message animates in when a new notification comes in:
![image](https://user-images.githubusercontent.com/13532591/66783234-59be8100-ee8c-11e9-99ae-b5c8dc0cae40.png)

Messages are shown in the bottom:
![image](https://user-images.githubusercontent.com/13532591/66783313-8377a800-ee8c-11e9-849b-6d0285a938d0.png)

And can expand the pane:
![image](https://user-images.githubusercontent.com/13532591/66783342-912d2d80-ee8c-11e9-901a-5b2c7b00da12.png)

bryphe:
The UX for notifications is very spammy and not very friendly right now. I really like the proposal here - maybe something we can explore @glennsl ?

A few things we'd have to consider:
- How long do messages 'stay' in the status bar?
- Do they cover up the items in the status bar? Or fit in between them?

I think the 'search pane' + the diagnostics pane in #1034 would be a natural place we could surface these.

jordwalke:
Only problem with the search pane position as the sole locations that I can see, is that many of these messages are intended to be very lightweight, and should be visible even in a minimalist windowing configuration. I think the search pane would be a great place to see the entire history of logs though. You definitely want a place to go explore old logs and that pane seems like the right place.

bryphe:
> Only problem with the search pane position as the sole locations that I can see, is that many of these messages are intended to be very lightweight, and should be visible even in a minimalist windowing configuration.

Definitely! I like the idea of using the status bar for the showing of the notifications - the experience seems much better than what we have now. 

I think my comment above made it sound like I only wanted them in the 'search pane', so I edited it a bit to clarify-

PR #1034, opened 2019-12-12
Feature - UX: Initial Diagnostics Pane

bryphe:
This extends the `SearchPane` to a `Pane` that can contain a few different kinds of sub-panes:
image

Right now, this adds a placeholder for a 'Problems' pane, which can show all the errors from plugins / language services.

There are a few other 'panes' I'd like to expose here:
- Find all references, when available - this could be a very similar UX to the search, reusing the LocationListVIew
- Notifications - described in https://github.com/onivim/oni2/issues/849

__TODO:__
- [x] Merge in focus change for search
- [x] Add a `LocationListView` for errors
- [x] Hook up status bar click to pop open errors
- [x] Add keyboard shortcut for opening diagnostics pane

Would also be a convenient place to bubble up QuickFix and Location Lists from Vim.

Next steps:
- Populate the 'problems' pane - we have all the data available via the diagnostics in the model, just need to distill it into any appropriate tree view.
- Add an error count in the status bar, and make the 'problems' pane show when clicked
- Add keyboard shortcuts for next error / previous error


src/Model/Pane.re
```diff
@@ -0,0 +1,28 @@
...
+let getType = pane => !pane.isOpen ? None : Some(pane.activePane);
+
+let isOpen = pane => pane.isOpen;
+
+let isTypeOpen = (paneType, pane) =>
+  pane.isOpen && pane.activePane == paneType;
+
+let hide = pane => {...pane, isOpen: false};
+
+let show = paneType => {activePane: paneType, isOpen: true};
```
glennsl: This confused me a bit because I expected `show` to return a string as it usually does. It might be an idea to call these something like `setOpen` and `close` until we're weaned off the `show` functions.

src/Model/Actions.re
```diff
@@ -111,8 +111,8 @@ type t =
   | EnableZenMode
   | DisableZenMode
   | CopyActiveFilepathToClipboard
-  | SearchShow
-  | SearchHide
+  | PaneShow(Pane.paneType)
+  | PaneHide
```
glennsl: This seems like a good opportunity to make these non-function-like.

src/Model/Pane.re
bryphe: Oh, good point, that is confusing. I've switched it to `setOpen` and `setClosed` :

src/Model/Actions.re
bryphe: Renamed to more event-like `PaneOpen(paneType)` and `PaneHidden` - does that sound better?

glennsl: This seems to describe the event that _will_ happen after it has been processed. Or the _intent_ perhaps, which is really just a different naming scheme for a function. What you want to describe is the event that _caused_ it, which requires a more fundamental change than just a rename. So in this case what we might (eventually) want is `ActivityBar(SearchButtonClicked)` and `Pane(SearchTabSelected | ProblemsTabSelected)`.

src/Store/PaneReducer.re
```diff
@@ -0,0 +1,48 @@
...
+      ...state,
+      pane: Pane.setOpen(Pane.Diagnostics),
+    }
+  | DiagnosticsHotKey when !isDiagnosticsOpen => {
+      ...state,
+      pane: Pane.setOpen(Pane.Diagnostics),
+    }
+  | _ => state
+  };
+};
```
glennsl: The control flow here is very confusing. Seems like it might be better to switch on `(action, pt)`, or use nested switches, instead of using guards, as that would give you exhaustiveness checking as well.

I'm also pretty sure that you're able to come up with a better name than `pt`...

src/Model/Actions.re
bryphe: These actions are removed now 👍

src/Store/PaneReducer.re
bryphe: Made an attempt to clean it up! It's still a bit complex (it's essentially a state machine with 3 states - not focused/open, open but not focused, and then open but focused - and handling the transitions based on each action). Probably still a nicer way to express it! But I think it's better than it was.

PR #1040, opened 2019-12-13
Fix #1037 - Quickopen not showing full set of results

### LINKED ISSUE #1037: Bug: Quickopen not showing full set of results
I just tried out the latest `master`, and am not seeing the full set of results in Quickopen:

Typing in `Diag`, I'd expect to see 4 files - but I only see two on `master`:

__Actual:__
image

__Expected:__
image

I bisected and found:
- [`02e19ff`] (#1030) - not working
- [`9cfd7642`] (#1029) - working

So it looks like there was potentially a regression in #1030 - could you please take a look at this @glennsl ?

glennsl:
Fixes #1037

### DIFF
--- a/bench/lib/FilterJobBench.re
+++ b/bench/lib/FilterJobBench.re
@@ -38,13 +38,24 @@ let job =
   |> Job.map(MenuFilterJob.addItems(largeAmountOfItems))
   |> Job.map(MenuFilterJob.addItems(largeAmountOfItems));
 
-let runJob = () => {
-  let _ = Job.doWork(job);
-  ();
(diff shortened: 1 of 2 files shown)


bryphe:
There are a few problems with the CI right now:
- Something is busted on Azure CI with the Ubuntu 16.04 image (getting bad hash for some deps)
- The `linuxdeploy.appimage` isn't published, so failing when packaging linux on the CentOS Docker run
- ...seems like some bugs on Windows, too.

None of these are due to this change, so will override the checks and bring it in

PR #1044, opened 2019-12-14
Fix #81 - Initial Welcome Experience

### LINKED ISSUE #81: Handling initial [No Name] Buffer

bryphe:
This adds support for custom 'renderers' for buffers - the idea is that we can skip out on rendering the `<EditorSurface />` in some cases.

This will be useful for:
- Welcome experience
- Terminal
- Image rendering

And potentially future displays (markdown, pdf).

For now, though, this is about adding a welcome experience - @glennsl and I discussed this a bit:
image

There's a few goals I have for this immediate PR:
- [x] Animate / polish the logo
- [x] Show some of the basic keybindings for getting around:
  - [x] QuickOpen (Control/Command+P)
  - [x] Command Palette (Control/Command + Shift + P)
  - [x] Vim Ex mode (`:`)
- [x] Add support for overriding tab name w/ custom renderer 

Then, as next steps, I'd like to add to the screen:
- Build info
- Link to our documentation 
- 'Sneak mode' entry point (Control+G)
- Recent files 
- Recent workspaces

One other idea for the welcome screen was to add some very small "mini-tutorials" to learn the basic keybindings - like `h`/`j`/`k`/`l` - or some quick vim tips

Fixes #81


src/UI/WelcomeView.re
```diff
@@ -13,84 +13,120 @@ open Oni_Core;
...
+        width(300),
+      ];
+
+    let commandText = (~theme: Theme.t, ~fontFile, ~fontSize as fs) =>
+      Style.[
+        fontFamily(fontFile),
+        fontSize(fs),
+        backgroundColor(theme.background),
+        color(theme.editorForeground),
+      ];
```
glennsl: If you remove `Style.` here in favor of `open Style;` at the beginning of the `Styles` module, you don't have to rename `~fontSize`. Instead you can qualify the style property: `Style.fontSize(fontSize)`. Much less awkward IMO.

bryphe: Nice!

bryphe:
I'll bring this in for now, but still needs some refinement + polish.

glennsl:
Looks very nice! Neat fade! The buffer is still called `untitled` in the Ctrl+Tab menu though.

bryphe:
Thanks @glennsl ! Animation API is great 😉 

> The buffer is still called untitled in the Ctrl+Tab menu though.

Oh, good catch, I guess we should have a more central location to override the buffer name. Logged #1110 to track

Issue #568, opened 2019-08-02
Documentation: Add documentation for keybindings.json format


bryphe:
Following for #564 and #565 , we should start documentation on the `keybindings.json` format:

Things to be covered:
- The `when` clause and appropriate values
- The available commands

bryphe:
This is added now, but we need to continue to add to it as we add features / commands / when clauses.

PR #1047, opened 2019-12-14
Documentation: Add default key bindings to docs

bryphe:


bryphe:
Fixes #568 , but we'll need to continue to maintain it as we add more.

PR #1050, opened 2019-12-14
Implement and use immutable queue for Ripgrep and FilterJob

glennsl:
Performance-wise the only significant difference is that adding items takes twice as long because the length of each chunk is computed twice, inside each of the queues it is added too. Previously it could be computed once since the total count was dealt with separately.

TODO:
- [x] add tests

### DIFF
--- a/src/Core/Utility.re
+++ b/src/Core/Utility.re
@@ -526,7 +526,7 @@ module StringUtil = {
   };
 };
 
-module Queue : {
+module Queue: {
   type t('a);
 
   let empty: t(_);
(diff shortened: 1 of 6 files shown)


src/Core/Utility.re
```diff
@@ -525,3 +525,152 @@ module StringUtil = {
...
+        ([item, ...items], queue);
+
+      | (None, queue) => ([], queue)
+      };
+    };
+
+  let toList = ({front, rear, _}) => front @ List.rev(rear);
+};
+
+module ChunkyQueue: {
```
bryphe: Very nice @glennsl  - neat abstraction to make the 'chunking' problem simpler

test/Core/UtilityTests.re
```diff
@@ -217,3 +217,256 @@ describe("StringUtil", ({describe, _}) => {
...
+      expect.list(Queue.toList(q)).toEqual([3, 4]);
+    });
+  });
+};
+
+describe("Queue", ({describe, _}) => {
+  testQueue(describe, (module Queue))
+});
+
+describe("ChunkyQueue", ({describe, _}) => {
```
bryphe: Thanks for the solid set of test cases 💪

glennsl:
I'm pretty sure all CI jobs have succeeded at some point, just not simultaneously, so merging this.

PR #1051, opened 2019-12-15
File Explorer performance tweaks

glennsl:
So far I've just added logging. I'm not sure how to find good test cases for this. A good test case would be a file that's both deep in the hierarchy, is nested in directories with many items and also listed late in those directories.

The worst case I found had `toNodePath` use 2e-05s.

I also found that focus does not work when the workspace is set to root (`/`) though. I need to look into that before I implement the hashing.

### DIFF
--- a/src/UI/FileTreeView.re
+++ b/src/UI/FileTreeView.re
@@ -5,15 +5,17 @@ open Revery.UI;
 
 module Core = Oni_Core;
 
-let toNodePath = (workspace: Workspace.workspace, tree, path) => {
-  let localPath = Workspace.toRelativePath(workspace.workingDirectory, path);
+let toNodePath = (workspace: Workspace.workspace, tree, path) =>
+  Core.Log.perf("FileTreeview.toNodePath", () => {
+    let localPath =
+      Workspace.toRelativePath(workspace.workingDirectory, path);
 
-  switch (FsTreeNode.findNodesByLocalPath(localPath, tree)) {
-  | `Success(nodes) => Some(nodes)
-  | `Partial(_)
-  | `Failed => None
-  };
-};
+    switch (FsTreeNode.findNodesByLocalPath(localPath, tree)) {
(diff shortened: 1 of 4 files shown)


glennsl:
The root workspace issue was in `Workspace.toRelativePath`. Good thing I just created that from the same code duplicated across the code base.

Hash code comparison seems to improve the worst by an order of magnitude. Miniscule eitehr way, but now even minisculer. As such a simple tweak I think it's nice to have.

glennsl:
I'm beginning to think these CIs aren't every going to succeed. If it isn't the Windows job getting stuck it almost certainly is the syntax highlight integration test failing. Or both.

src/UI/FileTreeView.re
```diff
@@ -5,15 +5,17 @@ open Revery.UI;
...
-  let localPath = Workspace.toRelativePath(workspace.workingDirectory, path);
-
-  switch (FsTreeNode.findNodesByLocalPath(localPath, tree)) {
-  | `Success(nodes) => Some(nodes)
-  | `Partial(_)
-  | `Failed => None
-  };
-};
+let toNodePath = (workspace: Workspace.workspace, tree, path) =>
+  Core.Log.perf("FileTreeview.toNodePath", () => {
```
bryphe: Cool, thanks for adding this! Forgot we have this...

bryphe: (And appreciate you measuring / following up on the perf 👍 )

bryphe:
> I'm beginning to think these CIs aren't ever going to succeed. If it isn't the Windows job getting stuck it almost certainly is the syntax highlight integration test failing. Or both.

Sorry about this, it does seem like they are pretty flaky right now. I'll take a look at these tests tomorrow and see if we can stabilize them

glennsl:
Pretty sure all these CI jobs have succeeded at some pint as well, just that the stars haven't all lined up perfectly at exactly the right time.

PR #1065, opened 2019-12-17
Extensions - CLI: --list-extensions command

bryphe:
Implement the `--list-extensions` command, by hooking up into the `ExtensionScanner` and reporting the available extensions


src/Store/ExtensionUtility.re
```diff
@@ -0,0 +1,347 @@
...
+                m("Error discovering user extensions: %s", msg)
+              );
+              None;
+            }
+          }
+        )
+        |> Option.map(p => {
+             Log.infof(m => m("Searching for user extensions in: %s", p));
+             p;
+           })
```
glennsl: I often make a `tap` function that's useful for logging in the middle of expressions:

```reason
let tap = (f, x) => {
  f(x);
  x
};
```
Haven't gotten around to it here yet, but maybe you'd want to add it.

bryphe: Cool idea! Added it in our utilities

PR #1069, opened 2019-12-17
Fix #1068 - High-DPI layout bugs

### LINKED ISSUE #1068: Completion UI broken on high-DPI / retina displays
image

Looks like an issue with text measurement / layout in Revery, for high-DPI.

bryphe:
This fixes up the incorrect layout / measurement in the completion experience when using a retina high-DPI display.

This picks up the root Revery fix: https://github.com/revery-ui/revery/pull/669 - and it requires now that we always pass the `window` to the measure APIs, because the measurements need the scaling information.

This also fixes up a couple of minor issues in the `LocationListView` that I happened to run into when testing this fix on high-DPI:
- We're rendering the location text with `editorFont`, but measuring with `uiFont` - this caused an inconsistency
- The `getFontAdvance` looks like it does the same thing as `measure` in Revery - so we might be able to remove that duplication already.

Fixes #1068

### DIFF
--- a/src/UI/LocationListView.re
+++ b/src/UI/LocationListView.re
@@ -74,10 +74,16 @@ let item =
       Index.toOneBased(item.location.line),
     );
 
-  let locationWidth =
-    Oni_Core.EditorFont.(
-      int_of_float(editorFont.measuredWidth) * String.length(locationText)
-    );
+  let locationWidth = {
+    let window = Revery.UI.getActiveWindow();
+    Revery.Draw.Text.measure(
+      ~window,
+      ~fontSize=uiFont.fontSize,
+      ~fontFamily=uiFont.fontFile,
+      locationText,
+    ).
+      width;
+  };
 
   let location = () =>
     <Text
--- a/src/UI/LocationListView.re
+++ b/src/UI/LocationListView.re
@@ -7,10 +7,24 @@ open Oni_Model;
 
 module Option = Utility.Option;
 
-let getFontAdvance = (fontFamily, fontSize) => {
-  let window = Revery.UI.getActiveWindow();
-  Revery.Draw.Text.measure(~window, ~fontSize, ~fontFamily, "x").width
-  |> float_of_int;
+// TODO: move to Revery
+let getFontAdvance = (fontFile, fontSize) => {
+  open Revery.Draw;
+
+  let maybeWindow = Revery.UI.getActiveWindow();
+  let scaledFontSize =
+    Text._getScaledFontSizeFromWindow(maybeWindow, fontSize);
+  let font = FontCache.load(fontFile, scaledFontSize);
+  let shapedText = FontRenderer.shape(font, "x");
+  let Fontkit.{advance, _} =
+    FontRenderer.getGlyph(font, shapedText[0].glyphId);
+
+  let multiplier =
+    switch (maybeWindow) {
+    | None => 1.0
+    | Some(w) => Window.getScaleAndZoom(w) *. Window.getDevicePixelRatio(w)
+    };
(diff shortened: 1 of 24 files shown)


src/UI/LocationListView.re
```diff
@@ -7,18 +7,10 @@ open Oni_Model;
...
-  let shapedText = FontRenderer.shape(font, "x");
-  let Fontkit.{advance, _} =
-    FontRenderer.getGlyph(font, shapedText[0].glyphId);
-
-  float(advance) /. 64.;
+let getFontAdvance = (fontFamily, fontSize) => {
+  let window = Revery.UI.getActiveWindow();
+  Revery.Draw.Text.measure(~window, ~fontSize, ~fontFamily, "x").width
+  |> float_of_int;
 };
```
glennsl: The reason it is duplicated is to get more precision. The conversions back and forth between int and float in `measure` and its helper functions causes a loss of precision. Although with the text rendering we have now I understand this will be a whole integer anyway, I'm not convinced that will be the case if we do sub-pixel rendering. It also seems odd that advance has a precision of 64 units per pixel internally if there is no use for it.

src/UI/LocationListView.re
```diff
@@ -83,17 +75,9 @@ let item =
...
-        ~fontSize=uiFont.fontSize,
-        ~fontFamily=uiFont.fontFile,
-        locationText,
-      ).
-        width
-    | None => String.length(locationText) * uiFont.fontSize
-    };
+    Oni_Core.EditorFont.(
+      int_of_float(editorFont.measuredWidth) * String.length(locationText)
+    );
```
glennsl: This is wrong. `locationText` uses the UI font, not the editorFont.see https://github.com/onivim/oni2/blob/33f1ab892a0720f84830af7178a6c33b0f146db1/src/UI/LocationListView.re#L82-L86

glennsl:
> We're rendering the location text with editorFont, but measuring with uiFont - this caused an inconsistency

From what I can see in the code it uses `uiFont`. What inconsistency did you experience?

src/UI/LocationListView.re
bryphe: Derp, I don't know what i was thinking... tricked by the font I had on my Mac, I guess 😬 

You're right, thanks for catching that!

src/UI/LocationListView.re
bryphe: Right - unfortunately the text rendering in Revery isn't perfect. In particular, the kerning is clamped to integer pixel coordinates (so we don't take full advantage of the [font metrics / precision from FreeType2](https://www.freetype.org/freetype2/docs/tutorial/step2.html)).

>  Although with the text rendering we have now I understand this will be a whole integer anyway, I'm not convinced that will be the case if we do sub-pixel rendering for example.

True! We should probably push this change to Revery - have the `measure` and similar APIs return floating-point to prepare for that (even if it is only integer-precision today). Once Skia is in place, we'll leverage the APIs it has for measuring- like this one: https://github.com/manuhornung/reason-skia/blob/d26e3ddfb7e8cee462d4290f3c29d5298baf72b0/src/Skia.rei#L77 - which would give us floating-point precision. The text rendering provided by Skia is significantly better and higher fidelity than what we have today...

> It also seems odd that advance has a precision of 64 units per pixel internally if there is no use for it.

For sure... I'm looking forward to being out of the text-rendering business and letting Skia handle it...

In the meantime - I'll bring back the previous logic, and a note that we should push the `float` to the upstream measure APIs, and just port over the high-DPI fix here too.

bryphe: Tracking the revery API change here: https://github.com/revery-ui/revery/issues/670

PR #1073, opened 2019-12-17
Fix #1070: Crash when running ':e ~/some/directory' in OSX build

### LINKED ISSUE #1070: running :e command on a directory crashes program
## Problem

When I run `:edit ~/my/project` the whole program crashes.

According to the docs (https://onivim.github.io/docs/using-onivim/working-with-files), this is because `:e` is meant for opening files only, and `:cd` is intended for directories. Nonetheless, since this is the habit we've developed from using vim, this common error should be handled.

## Expected Behaviour

Running the edit command on a directory should fail gracefully, by either opening the directory in the Explorer section (like vim's netrw directory listing), or give a helpful message like: "You cannot use :edit on a directory. Did you mean :cd ~/my/project?"

## Reproduction

1. type `:edit ~/path/to/a/directory`

## My setup

**Onivim version:** MacOSX v2.0.2.0

bryphe:
Fixes #1070 - a crash when trying to edit a directory.

It's a bit of a mystery why we are hitting this case in packaged builds, but _not_ in development builds / running the executable directly.

This prevents the crash, but the focus behavior is not correct (it doesn't actually focus the directory).

### DIFF
--- a/src/Store/FileExplorerStore.re
+++ b/src/Store/FileExplorerStore.re
@@ -30,7 +30,7 @@ module Effects = {
 let nodeOffsetByPath = (tree, path) => {
   let rec loop = (node: FsTreeNode.t, path) =>
     switch (path) {
-    | [] => failwith("Well, this is awkward (ie. unreachable)")
+    | [] => `NotFound(0)
     | [(focus: FsTreeNode.t), ...focusTail] =>
       if (focus.id != node.id) {
         `NotFound(node.expandedSubtreeSize);
--- a/src/Store/FileExplorerStore.re
+++ b/src/Store/FileExplorerStore.re
@@ -30,7 +30,7 @@ module Effects = {
 let nodeOffsetByPath = (tree, path) => {
   let rec loop = (node: FsTreeNode.t, path) =>
     switch (path) {
-    | [] => `NotFound(0)
+    | [] => `Found(0)
     | [(focus: FsTreeNode.t), ...focusTail] =>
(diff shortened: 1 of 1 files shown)


src/Store/FileExplorerStore.re
```diff
@@ -30,7 +30,7 @@ module Effects = {
 let nodeOffsetByPath = (tree, path) => {
   let rec loop = (node: FsTreeNode.t, path) =>
     switch (path) {
-    | [] => failwith("Well, this is awkward (ie. unreachable)")
+    | [] => `NotFound(0)
```
glennsl: I think this should be `` `Found(0)``. If the directory you open is expanded in the file explorer it looks like it'll hit this, because it'll try to keep looking. And it'll probably open it in `revealPath`, so it might always hit this when opening a directory.

bryphe: Ah, you're right! Nice catch.

I'm able to reproduce this on developer builds by running `:e a-directory-in-file-explorer`. So it wasn't actually a packaged-vs-developer-builds issue... it just happened that the packaged build started it in a different directory.

bryphe: And the focus works correctly now with `Found(0)` too

glennsl: Awesome! That was an easy fix. Clearly I just assumed that opening directories is a silly thing to do that's not worth bothering with. And you know what they say about assumptions... 🤦‍♂

PR #1082, opened 2019-12-18
Dependency: Sync up JSOO dependencies with Revery

bryphe:


bryphe:
Bring in #1084 instead 👍

Issue #528, opened 2019-08-01
Feature: Vim navigation for the file explorer


bryphe:
In Onivim 1, we had vim-style keybindings for the file explorer. In this model, the file explorer behaved as another window split, in that you could `<C-w>h` over to it.

Once in the explorer split, you could use `h`/`j`/`k`/`l` to navigate along with various other vim bindings. In addition, we also were able to delete (with `d`) and yank (with `y`) - and even `u` undo.

This was actually really nice! But there were two problems:

- The `d` key was too easy to hit. It as easy to start typing, not realize you're focused in the file explorer, and accidentally delete a file. I think, if we bring this behavior forward, we need to find a UX that is more deliberate for deletion, yanking, pasting, and undoing.

- We never implemented search for it (`/`). I think a useful model would be search-as-a-filter in this view, but open to ideas.

ImCesar:
Nerdtree has the menu you have to open before you can delete files and then prompts you if you are sure. So if your cursor was over a file you would hit m to open the menu, d to delete, and then it would prompt you y/N to verify.

For me the deal breaker on switching from vim to VS Code was how slow I felt not being able to use Nerdtree style controls. Being able to open splits/tabs with a key press is pretty handy (same with ctrl+p). Also having the option to rebind <C-w>h to ctrl + h (or any of the directions) would be awesome.

bryphe:
Cool, thanks for the feedback @ImCesar !

> Nerdtree has the menu you have to open before you can delete files and then prompts you if you are sure. So if your cursor was over a file you would hit m to open the menu, d to delete, and then it would prompt you y/N to verify.

I was mostly an `Ex` user so I wasn't familiar with this - makes sense to check it out. Lots of thought went into though. Sounds better than our paradigm in Onivim 1, since there is less chance for an accidental deletion of a file.

> For me the deal breaker on switching from vim to VS Code was how slow I felt not being able to use Nerdtree style controls.

I can relate to this! 😎 I missed being able to access everything from the keyboard.

jordanjones98:
> Also having the option to rebind h to ctrl + h (or any of the directions) would be awesome.

I agree with this 100%, maybe this is another issue/feature request but it is my #1 reason for not switching over to some non-vim editor.

repromancer:
These days I do most of my file browsing with FZF (fuzzy finding by either file name or file content), and whenever I need directory tree navigation, I just open up `ranger`, which takes over the entire pane and acts like a "open file in current tab" dialogue.

[...]

PR #1086, opened 2019-12-19
File Explorer Keyboard Navigation - Part 1

glennsl:
This implements basic keyboard navigation (up arrow/down arrow) and selection (enter) _within_ the file explorer. There's no way to use the keyboard to focus the file explorer, however, so it's not very useful for end-users yet. And because of the mechanism used to capture keyboard input it also blocks all keybindings. But it's useful for testing keyboard input and relatively self-contained, so I'm submitting this now to try to keep PRs manageable.

Addresses #528

### DIFF
--- a/src/Core/Utility.re
+++ b/src/Core/Utility.re
@@ -254,6 +254,24 @@ module RangeUtil = {
     };
 };
 
+module ArrayEx = {
+  exception Found(int);
+
+  let findIndex = (predicate, array) =>
+    try(
+      {
+        for (i in 0 to Array.length(array) - 1) {
+          if (predicate(array[i])) {
+            raise(Found(i));
+          };
+        };
+        None;
+      }
+    ) {
+    | Found(i) => Some(i)
+    };
+};
+
 // TODO: Remove after 4.08 upgrade
 module List = {
   include List;
--- a/src/Core/Utility.re
+++ b/src/Core/Utility.re
@@ -701,3 +701,15 @@ module ChunkyQueue: {
   let toList = ({front, rear, _}) =>
     front @ (Queue.toList(rear) |> List.concat);
 };
+
+module Path = {
+  // Not very robust path-handling utilities.
+  // TODO: Make good
+
+  let toRelative = (~base, path) => {
+    let base = base == "/" ? base : base ++ Filename.dir_sep;
+    Str.replace_first(Str.regexp_string(base), "", path);
+  };
+
+  let explode = String.split_on_char(Filename.dir_sep.[0]);
+};
--- a/src/Model/FileExplorer.rei
+++ b/src/Model/FileExplorer.rei
@@ -4,7 +4,8 @@ type t = {
   tree: option(FsTreeNode.t),
   isOpen: bool,
   scrollOffset: [ | `Start(float) | `Middle(float)],
-  active: option(string) // path
+  active: option(string), // path
+  focus: option(int), // node id
 };
 
 [@deriving show]
--- a/src/Model/FileExplorer.rei
+++ b/src/Model/FileExplorer.rei
@@ -15,6 +15,7 @@ type action =
   | FocusNodeLoaded(string, [@opaque] FsTreeNode.t)
   | NodeClicked([@opaque] FsTreeNode.t)
   | ScrollOffsetChanged([ | `Start(float) | `Middle(float)])
+  | Select
   | FocusPrev
   | FocusNext;
 
--- a/src/Model/FileExplorer.rei
+++ b/src/Model/FileExplorer.rei
@@ -11,8 +11,8 @@ type t = {
 [@deriving show]
 type action =
   | TreeLoaded([@opaque] FsTreeNode.t)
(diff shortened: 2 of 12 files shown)


src/Model/FileExplorer.rei
```diff
@@ -4,16 +4,20 @@ type t = {
   tree: option(FsTreeNode.t),
   isOpen: bool,
   scrollOffset: [ | `Start(float) | `Middle(float)],
-  focus: option(string) // path
+  active: option(string), // path
+  focus: option(string) // node id
```
glennsl: Note to self: path, not id

src/Model/FsTreeNode.re
```diff
@@ -1,5 +1,9 @@
+open Oni_Core;
+
+module ArrayEx = Utility.ArrayEx;
+module Path = Utility.Path;
+
 type t = {
-  id: int,
```
glennsl: I've removed `id` entirely in favor of using paths as identifiers instead. For a few reasons:

- it's simpler.
- Paths don't change (and if they do it's because their identity changes). No more stale ids, they're valid across workspace changes and can even be persisted.
- Searching by path is most likely faster since nodes that aren't in the path can be eliminated from the search. Hashing also makes individual comparisons as fast as on ids.

src/Model/FsTreeNode.re
```diff
@@ -26,49 +34,48 @@ let rec countExpandedSubtree =
...
     expandedSubtreeSize: countExpandedSubtree(kind),
   };
 };
 
-let findNodesByLocalPath = (path, tree) => {
-  let pathHashes =
-    path
-    |> String.split_on_char(Filename.dir_sep.[0])
-    |> List.map(Hashtbl.hash);
+let equals = (a, b) => a.hash == b.hash && a.path == b.path;
```
glennsl: Hashes aren't guaranteed to be unique, so use this instead of comparing hashes directly, and especially instead of comparing nodes structurally.

src/UI/FileTreeView.re
```diff
@@ -67,56 +86,80 @@ let nodeView = (~font: Core.UiFont.t, ~fg, ~node: FsTreeNode.t, ()) => {
...
+    // arrow down
+    | v when v == 1073741905 =>
+      GlobalContext.current().dispatch(Actions.FileExplorer(FocusNext))
+
+    | _ => ()
+    };
+  };
+
+  <View
+    onKeyDown style=Styles.container ref={ref => setContainerRef(Some(ref))}>
```
glennsl: This should all be removed in favor of handling input "bottom-up".

glennsl:
Woo! What is this new and amazing CI we got? Didn't even take 30 mins to finish! :tada:

src/Core/Utility.re
```diff
@@ -683,3 +701,15 @@ module ChunkyQueue: {
...
+  // Not very robust path-handling utilities.
+  // TODO: Make good
+
+  let toRelative = (~base, path) => {
+    let base = base == "/" ? base : base ++ Filename.dir_sep;
+    Str.replace_first(Str.regexp_string(base), "", path);
+  };
+
+  let explode = String.split_on_char(Filename.dir_sep.[0]);
+};
```
bryphe: We could consider using the `Fp`/`Path` library from reason-native: https://github.com/facebookexperimental/reason-native/tree/master/src/fp
(if it makes sense)

src/Model/FileExplorer.rei
bryphe: That would be one nice thing about using a path library like `Path`/`fp` - we could have a type for it

src/Model/FsTreeNode.re
bryphe: Makes sense to me!

bryphe:
> There's no way to use the keyboard to focus the file explorer, however, so it's not very useful for end-users yet. And because of the mechanism used to capture keyboard input it also blocks all keybindings. But it's useful for testing keyboard input and relatively self-contained, so I'm submitting this now to try to keep PRs manageable.

I like the way you divided the work - seems like a good chunk to bring in. I tried it out and it's so nice to have some keyboard input in the file explorer!

src/Core/Utility.re
glennsl: Yeah, that looks pretty good. It doesn't seem to have any way of destructuring it like `explode` though. But we could contribute that.

bryphe:
> Woo! What is this new and amazing CI we got? Didn't even take 30 mins to finish! 🎉

I tried to go through and fix most of the hangs / issues... hopefully it works better now 🤞

JonDum:
This is awesome. Looking forward to part 2 and 3! Once it's "nerdtree-like" I will be able to use oni2 for real I think.

gordonrust:
Any update on the 'nerdtree-like' behaviour. I am also eagerly waiting for this behaviour to switch to onivim2.
Also, if possible, could the panel in the bottom of the UI also behave like another hsplit window as is the behavoiur in vim. These two are the most important things that have halted my adoption of vscode. I have my hopes on oni2.

PR #1087, opened 2019-12-19
Search hotkey

### LINKED ISSUE #1054: Feature - Search: Add bindable command + default key binding to focus search
We should have an easy, keyboard-only entry point for the shiny new search functionality - a command to open the pane, and focus the search text box (`workbench.action.findInFiles`, for VSCode compatibility), as well as some default key bindings:

- Mac: `Command+Shift+F`
- Linux / Windows: `Control+Shift+F`

glennsl:
Fixes #1054 (At least partially)

Adds a new keybinding to toggle Search, and gives focus to the input when pane is opened. Unfortunately it's not very practical to focus the input outside the view as it would require storing the input ref in the state or a global. So instead the hotkey just toggles the pane. We can revisit this once I've changed the focus mechanism so it's controlled bottom-up.

### DIFF
--- a/src/Store/KeyBindingsStoreConnector.re
+++ b/src/Store/KeyBindingsStoreConnector.re
@@ -34,6 +34,16 @@ let start = () => {
         condition: "quickmenuCursorEnd" |> parseExp,
       },
       {
+        key: "<S-C-F>",
+        command: "workbench.action.findInFiles",
+        condition: "editorTextFocus" |> parseExp,
+      },
+      {
+        key: "<D-S-F>",
+        command: "workbench.action.findInFiles",
+        condition: "editorTextFocus" |> parseExp,
+      },
+      {
         key: "<C-TAB>",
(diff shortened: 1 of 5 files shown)


src/Store/KeyBindingsStoreConnector.re
```diff
@@ -33,6 +33,16 @@ let start = () => {
         command: "list.selectBackground",
         condition: "quickmenuCursorEnd" |> parseExp,
       },
+      {
+        key: "<S-C-F>",
+        command: "workbench.action.findInFiles",
+        condition: "editorTextFocus" |> parseExp,
+      },
+      {
+        key: "<D-S-F>",
```
glennsl: I think this is how you specify `Cmd-Shift-F`?

bryphe: Correct! We need to have a nicer parser for keybindings though (something that can support both the VSCode style: Cmd+Shift+F and the Vim style: <D-S-F>)

PR #1090, opened 2019-12-19
UX - Extensions Pane - Show bundled extensions

bryphe:
Right now, our 'extensions' pane is totally empty. 

This change shows the set of bundled extensions we bring in:
image

__TODO:__
- [x]  Filter to set of extensions with language features (typescript, css, html, js, reason) and third-party themes (nord, gruvbox, laserwave).
- [x] Fix `%displayName%` in extension titles
- [x] Show author and version

__Next steps:__
- Add an 'accordion' control to filter between bundled, user installed, and from workspace


src/Extensions/ExtensionManifest.re
```diff
@@ -34,8 +36,16 @@ type t = {
...
+let getDisplayName = (manifest: t) => {
+  manifest.displayName |> Option.value(~default="Unknown Extension");
+};
+
+let getIcon = (manifest: t) => manifest.icon;
+
 let remapPaths = (rootPath: string, manifest: t) => {
   ...manifest,
+  main: Option.map(mainPath => Path.join(rootPath, mainPath), manifest.main),
+  icon: Option.map(iconPath => Path.join(rootPath, iconPath), manifest.icon),
```
glennsl: Partial application opportunities:

```reason
Option.map(mainPath => Path.join(rootPath, mainPath), manifest.main)
// is equivalent to
Option.map(Path.join(rootPath), manifest.main)
```

src/Store/ExtensionsReducer.re
```diff
@@ -0,0 +1,13 @@
+/*
+ * ExtensionsReducer.re
+ */
+open Oni_Model;
+
+let reduce = (action: Actions.t, v: Extensions.t) => {
+  switch (action) {
+  | Extension(Activated(id)) => Extensions.activateExtension(id, v)
```
glennsl: The cause and effect is very unclear here. Either `Extensions(Activated(id))` does not actually mean an extension has been activated, but rather a request for activation, or `Extensions.activateExtension` doesn't actually activate the extension but just marks it as activated. At least one of these must be wrong.

bryphe: Good point - `Extensions.activeExtension` sounds like it is actually doing the activation - but in actuality, it's just recording that the activation happened.

I changed the naming of both these methods to `notifyActivated` and `notifyDiscovered`, but let me know if there was some other nomenclature you had in mind.

glennsl: Hmm, `notify` seems to me imply an active form of messaging. Perhaps `setActivated`, `markActivated` or `addToActivated` depending on the level of abstraction here?

bryphe: Thanks @glennsl ! Sounds good - I used `markActivated` instead of `notifyActivated`, and `add` instead of `notifyDiscovered`

PR #1094, opened 2019-12-21
File Explorer Keyboard Navigation - Part 2: Input Focus Revamp

glennsl:
This changes the input handling to manage focus through the state and routes input "bottom-up" instead of "top-down" (or rather sometimes "top-down" as it was a bit of a hybrid) where focus is handled imperatively and focused nodes get input events on the nodes themselves.

This introduces a focus stack kept int he state to keep track of which parts of the UI has requested focus, but also a `FocusManager` that computes the actual focus based on the entire state in order to maintain other invariants, such as focus always being with he Quickmenu input if it is open, and (potentially) not being able to focus an element that is not actually visible.

A few other benefits:
- A number of hacks have been removed because they're no longer necessary.
- Input handling has been removed from `OniInput` to be put in `InputModel`, which provides for a much cleaner separation of concerns.


bryphe:
> This changes the input handling to manage focus through the state and routes input "bottom-up" instead of "top-down" (or rather sometimes "top-down" as it was a bit of a hybrid) where focus is handled imperatively and focused nodes get input events on the nodes themselves.

Cool! Thanks for thinking about this, @glennsl . This bottom-up modeling seems much cleaner than what we had before, because there is a consistent source of the truth for what is focused (whereas what wasn't necessarily the case with the 'hybrid' approach before..)

> A number of hacks have been removed because they're no longer necessary.

Nice!

src/Model/FocusManager.re
```diff
@@ -0,0 +1,26 @@
+open Oni_Core;
+
+module Option = Utility.Option;
+
+let push = (focusable: Focus.focusable, state: State.t) =>
+  switch (focusable) {
+  | Quickmenu
```
bryphe: At first I wasn't sure why `Quickmenu` and `Wildmenu` were special cased here - but then I re-read the PR description and the `current` method and see - those receive implicit focus based on the `state`, and thus shouldn't explicitly need to be pushed / popped.

src/Model/InputModel.re
```diff
@@ -0,0 +1,37 @@
+open Oni_Core;
```
bryphe: Cool! It's nice to have this logic decoupled from the rendering.

src/Model/FocusManager.re
glennsl: Exactly, although the more important invariant to maintain here is that it shouldn't be possible for _anything else_ to steal focus when the menus are visible, which would be much harder to ensure without this.

PR #1095, opened 2019-12-22
Docs - Update Docs/CLI Help Text.

CrossR:
Some of the docs are a bit duplicated between CLI/Extensions. Arguably for a flagship feature, I think we should have a distinct page for extensions, so this can hopefully be the start of that.

I also updated the CLI flag text and added it for every flag, such that they show up, and aligned them all.

### DIFF
--- a/docs/docs/configuration/extensions.md
+++ b/docs/docs/configuration/extensions.md
@@ -1,0 +1,45 @@
+---
+id: extensions
+title: extensions
+sidebar_label: VSCode Extensions
+---
+
+Onivim is capable of loading VSCode extensions. The current capabilities of the
+extension host can be found [here](https://github.com/onivim/oni2/issues/1058).
+
+## Installing extensions
+
+Onivim has two ways of installing VSCode extensions currently, though in the future this
+will be expanded to include an in-editor experience.
+
+Extensions by default are installed in the following folders:
+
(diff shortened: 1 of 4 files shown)


CrossR:
Looks like this failed / stalled on this bit:

```
2019-12-22T15:09:48.9301932Z [INFO]     +0ms Global : RUNNING CHECK: Verify node dependencies
```

Not sure why that would fail when it was fine for the zip and exe. (I know its unrelated to this PR, was just interested why/where the CI was being flaky).

bryphe:
Awesome, thanks @CrossR !

Having a separate page for CLI and extensions sounds good 👍

bryphe:
We'll need to add extensions to our `sidebars.json` here:
https://github.com/onivim/oni2/blob/master/docs/website/sidebars.json

You can actually test the documentation here too (from `~/oni2/docs/website`), with:
`yarn install`
`yarn start`

bryphe:
> Not sure why that would fail when it was fine for the zip and exe. (I know its unrelated to this PR, was just interested why/where the CI was being flaky).

I think there is still one intermittent failure left here - it seems, on Windows, sometimes this health check fails (I'm not sure why). I'm going to try replacing with with our custom `NodeTask` instead of the rench child process to see if that is more reliable.

PR #1099, opened 2019-12-23
Include hidden files in quick open

### LINKED ISSUE #860: Add hidden files to quick open
Currently the quick open (`Ctrl/Cmd - P`) doesn't show hidden files or folders (ones starting with a `.`).

To add this we need to do a few things:

 1. Add `--hidden` to the `ripgrep` flags over in `Ripgrep.re`.
 2. Hook up the config option to be passed over.
 3. Build up a bunch of `--glob=!` arguments based on the `files.exclude` configuration setting. (This might also have to include the `.git` folder, I can't remember if that is ignored by default/is included in the `files.exclude`.)

At that point we should have hidden files and excluded files being ignored from the search.

amiralies:
Closes #860

added `--hidden` flag to rg cmd
added globs to rg command using filesExclude config, this makes
filter results like .git node_modules etc.

### DIFF
--- a/src/Core/Ripgrep.re
+++ b/src/Core/Ripgrep.re
@@ -60,6 +60,7 @@ module Log = (val Log.withNamespace("Oni2.Ripgrep"));
 type t = {
   search:
     (
+      ~filesExclude: list(string),
       ~directory: string,
       ~onUpdate: list(string) => unit,
       ~onComplete: unit => unit
@@ -200,7 +201,8 @@ let process = (rgPath, args, onUpdate, onComplete) => {
    order of the last time they were accessed, alternative sort order includes
    path, modified, created
  */
-let search = (~executablePath, ~directory, ~onUpdate, ~onComplete) => {
+let search =
+    (~executablePath, ~filesExclude, ~directory, ~onUpdate, ~onComplete) => {
   let dedup = {
     let seen = Hashtbl.create(1000);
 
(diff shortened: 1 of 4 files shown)


src/Core/Ripgrep.re
```diff
@@ -214,9 +216,20 @@ let search = (~executablePath, ~directory, ~onUpdate, ~onComplete) => {
...
+    filesExclude
+    |> List.map(x => "!" ++ x)
+    |> List.map(x => ["-g", x])
+    |> List.concat;
+
+  let args =
+    globs
+    @ ["--smart-case", "--hidden", "--files", "--", directory]
+    |> Array.of_list;
+
```
glennsl: This is the first time we apply user input as arguments to flags. Do we need to sanitize or escape this to prevent "flag injection", @bryphe? What happens if `x` starts with whitespace? Should it be trimmed?

bryphe: Good question @glennsl . I believe this should be safe from a 'command injection' attack (where some rogue config would set an option that would allow us to escape out and execute arbitrary shell commands). 

The arguments we're passing are split out and separate from the executable command: https://github.com/revery-ui/rench/blob/a976fe51b89b4943e8650f8b9d6ebc49331678b0/src/ChildProcess.re#L126

(It would be more of a concern if we were passing the entire command-line to the shell).

> What happens if x starts with whitespace? Should it be trimmed?

I don't believe this is needed, but it would be worth testing (it could behave differently on different platforms). Worst case would be that the argument would not be processed as expected.

Thanks for reviewing and thinking about this!

bryphe:
Thank you @amiralies ! Looks great! 💯

src/Core/Ripgrep.re
glennsl: > I believe this should be safe from a 'command injection' attack (where some rogue config would set an option that would allow us to escape out and execute arbitrary shell commands).

Yeah I wasn't so much worried about that as being able to pass arbitrary flags to `rg`, accidentally or not.

bryphe: > Yeah I wasn't so much worried about that as being able to pass arbitrary flags to rg, accidentally or not.

Ah, got it. I believe when the arguments are passed as an array with `create_process_env`, it's like having single quotes around them (`rg '-gtest'`), so I'm pretty sure we're safe from arbitrary flags - the glob pattern may break though (ie, `rg '-g! -test'`)

Issue #902, opened 2019-10-24
OSX Sierra (10.12) support


bryphe:
Unfortunately today we don't support OSX Sierra (10.12) - this should be doable, we just need to set up our CI to work with 10.12. 

The earliest build available for the Microsoft-hosted CI is 10.13. We have a couple options:
- Use an alternate CI provider that can use 10.12
- Stand up our own CI agent for 10.12

bryphe:
This has been asked for a few times - something we need to find a solution for-

CrossR:
Travis seems to support 10.12 (Well, it supports back to 10.10 apparently): https://docs.travis-ci.com/user/reference/osx/#macos-version

PR #1102, opened 2019-12-23
CI/Builds - Support macOS 10.12

CrossR:
This is just a test to change the macOS CI to use Xcode 8, which is compatible with macOS 10.12 (to my knowledge). I think this means we'd be compatible with 10.12, though I could be very wrong.

A quick look in my CI looks like there is some build failures over in `esy-sdl2`, so opening just so some people with more understand than me can see.

### DIFF
--- a/.ci/swap-xcode.yml
+++ b/.ci/swap-xcode.yml
@@ -1,0 +1,11 @@
+# Swap XCode version to 8 for 10.12 support.
+
+steps:
+  - script: env | grep XCODE
+    displayName: 'List Xcode versions'
+  - script: xcodebuild -version
+    displayName: 'List current Xcode version'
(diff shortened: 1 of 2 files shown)


CrossR:
I've swapped to Xcode 9, not 8 as this is the minimum for sdl2 as far as I can see (the `@available` syntax is in the code, and was only added in 9). This should let us work with 10.12.6 which is hopefully what most people are using if they are still on 10.12, as that is the only one still getting security updates.

CrossR:
This looks like it builds now, so I think we should be able to support 10.12, which would fix #902.

bryphe:
Wow, great find @CrossR ! 

That's awesome that we can support OSX 10.12 w/o needing a 10.12 build machine.

bryphe:
Merging since the windows validate release failure was unrelated-

Issue #1097, opened 2019-12-23
Crash on opening files


CrossR:
Currently, when you open a file with syntax highlighting (or it may be the language server?) we crash and don't recover.

I've tested `JSON` files and `Reason` files and in both you get no syntax highlights at all, and then crash moments later. This is on the latest master, macOS Catalina, node 13 and esy 0.6.0.

I'm nipping out for a bit, but I can do some bisection to work out what commit caused it since it happens on the released builds and the dev builds so its not a release artifact issue.

bryphe:
Strange! I just tested on Catalina (the latest release, and the downloaded from `master`), and am not reproducing it.

It's interesting that it reproduces for both the released builds and dev builds. I wonder if there is a bug with node 13 and we happen to be jumping to the system one instead of the vendored version.

Some things that might help track it down:
- The log file (preferably with `ONI2_DEBUG=1`)
- The results of `--checkhealth`
- Any custom configuration

I'll also add some CLI commands to:
- Disable syntax highlighting (`--no-syntax-highlighting`)
- Disable extensions (`--no-extensions`)
- Disable configuration file (`--no-configuration`)

so that we can at least narrow down whether it is syntax highlighting vs extension host, or possibly configuration.

CrossR:
I actually only get this behaviour in my dev build, though someone on discord had mentioned syntax was dead on master in the release build. Weirdly for me, the dev build doesn't work and the release build does.

Check health is as follows:

```sh
┌─[16:49:40 MacBook (master $) ~/git/oni2]
└─╼ esy run run -f --checkhealth

[DEBUG]    +2ms Global : Setup: Looking for setupJson at: /Users/ryanc/git/oni2/_esy/default/store/i/oni2-5ca89004/bin/setup.json
[INFO]     +0ms Global : RUNNING CHECK: Verify camomile:datadir
[INFO]     +0ms Global :  -- RESULT: PASS
[INFO]     +0ms Global : RUNNING CHECK: Verify camomile:localedir
[INFO]     +0ms Global :  -- RESULT: PASS
[INFO]     +0ms Global : RUNNING CHECK: Verify camomile:charmapdir
[INFO]     +0ms Global :  -- RESULT: PASS
[INFO]     +0ms Global : RUNNING CHECK: Verify camomile:unimapdir
[INFO]     +0ms Global :  -- RESULT: PASS
[INFO]     +0ms Global : RUNNING CHECK: Verify node executable
[...]

PR #1103, opened 2019-12-23
Bugfix - Syntax Server: Proper exit code + logging forwarding

bryphe:
This fixes a few issues found while investigating #1097 :

- We were logging the `pid` instead of the process exit code
- The `Core.Log` strategy could potentially be problematic, because since both the main process and the syntax process get the `ONI2_LOG_FILE` environment variable, they could both end up writing to the same file. This is fixed by overriding the logging reporter, but I'm not sure if there is a better way - @glennsl might have an idea for a cleaner solution.
- We create a pipe for `stderr`, but are never listening to it / responding to it.

Unfortunately, I'm unable to repro #1097 currently, so I'm not sure this actually fixes the root cause - but the additional logging might help us drill in a bit further.


glennsl:
Having multiple processes write to the same file shouldn't be a problem as long as we open it in append only mode. We can do so using [`open_out_gen`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Stdlib.html#VALopen_out_gen) with the `Open_append` flag.

bryphe:
> Having multiple processes write to the same file shouldn't be a problem as long as we open it in append only mode. We can do so using open_out_gen with the Open_append flag.

Cool, I logged #1107 to track exploring this further.

PR #1105, opened 2019-12-24
Extensions - Use default localization for localized tokens

bryphe:
With our set of built-in extensions, like TypeScript & CSS, we'd see a lot of un-localized tokens like:
```
%displayName%
```
or 

showing up in the UI like:

image

This change picks up the `package.nls.json`, when available, for an extension, which specifies a set of tokens + default localizations. This starts adding a concept of a `LocalizedToken` and a `LocalizationDictionary`, which we can use to fully plumb through localizations later.


src/Extensions/LocalizationDictionary.re
```diff
@@ -0,0 +1,34 @@
+/*
```
bryphe: interface file

src/Extensions/LocalizedToken.re
```diff
@@ -0,0 +1,47 @@
+/*
+ * LocalizedToken.re
+ */
+
+[@deriving show]
+type t = {
```
bryphe: interface file

PR #1114, opened 2019-12-28
Test case for #1097: Try to create a regression test for 1097

bryphe:
This adds a test case to try to reproduce #1097 

@CrossR , would you mind trying to run this case and see if it fails?
`esy '@integrationtest' x Regression1097Test.exe`

Curious to see the result. Would also be interesting to see if the other integration tests pass (`esy '@integrationtest` run).


CrossR:
I'm getting a failure when running that, so the test does repro it for me, though obviously not on the CI...

bryphe:
> I'm getting a failure when running that, so the test does repro it for me, though obviously not on the CI...

Well, test case is good at least... 😅 I'll bring it in so we have a safety net on CI for this

PR #1117, opened 2019-12-28
Feature: Initial 'Sneak Mode'

bryphe:
This is a carryover from our V1 - a way to bridge the 'UI' world with pure-keyboard-accessibility.


src/UI/SneakView.re
```diff
@@ -0,0 +1,80 @@
...
+      color(theme.sneakForeground),
+      fontFamily(state.uiFont.fontFile),
+      fontSize(12),
+    ];
+    let highlightStyle = [
+      backgroundColor(theme.sneakBackground),
+      color(theme.sneakHighlight),
+      fontFamily(state.uiFont.fontFile),
+      fontSize(12),
+    ];
```
glennsl: Most of the benefit of having a `Styles` module is to know you have all the styles in one place, can find them easy and get a good overview. But of course that only works if you _do_ put all the styles in that module.

src/Model/Sneak.re
```diff
@@ -0,0 +1,118 @@
...
+  (model.prefix, remainder);
+};
+
+let reset = _sneak => {...initial, active: true};
+
+let hide = _sneak => initial;
+
+let isActive = sneaks => sneaks.active;
+
+let _filter = (prefix: string, sneak: sneak) => {
```
glennsl: I'm partial to using underscore to denote internal functions too, but there are some unintentional issues with it. One ise that syntax highlighting tends to grey them out, making them hard to visually scan for. Another problem is that the compiler uses a leading underscore as a hint that it is intentionally unused and won't give you any warning about it. That might eventually lead to a lot of unused code laying around and cluttering the place up. It can also lead to subtle errors with shadowing, although that's much less likely to occur with top-level definitions.

In any case, I'm not sure it's worth it with these issues. Or perhaps a different scheme would be better, like an `Internal` submodule.

bryphe: I like the idea of an `Internal` submodule 👍

src/Model/Sneak.re
```diff
@@ -0,0 +1,118 @@
+open Revery.Math;
+
+type callback = unit => unit;
```
bryphe: Create `rei`

src/UI/SneakRegistry.re
```diff
@@ -0,0 +1,32 @@
+open Oni_Core;
+type callback = unit => unit;
```
bryphe: Add `rei`

src/Model/Sneak.re
```diff
@@ -0,0 +1,120 @@
...
+  boundingBox: BoundingBox2d.t,
+  id: string,
+};
+
+[@deriving show({with_path: false})]
+type action =
+  | Initiated
+  | Discover([@opaque] list(sneakInfo))
+  | KeyboardInput(string)
+  | Stopped;
```
glennsl: You're still treating these basically as functions, despite naming them as events. The events described haven't occurred yet, but are assumed to be the result of processing these, hence essentially begging the question. This is going to cause confusion if you try to debug based on this information, assuming something has happened becasue the log says ti has when it fact it hasn't.

Let functions be functions. `Initiated` and `Stopped` can be ordinary functions, `start` and `stop`, taking and returning a `Sneak.t`, and be invoked from the actual events that should trigger it, `SomeButtonClicked` or `Command("sneak.stop")`.

`Discover` could just be named `Discovered` I think.

bryphe: Thanks! Still need to work on this.

I've updated the PR and refactored the actions a bit - hopefully it's a step in the right direction.

glennsl: Definitely, but `NoneAvailable` certainly isn't a very event-like name. Couldn't simply `Completed` work?

PR #1120, opened 2019-12-28
Fix stack overflow crashes with a large number of search results

### LINKED ISSUE #1113: Search (find in files) crashes when result length is large
For example when you try to search `a` in oni2 repo

glennsl:
Fixes #1113. Stack Overflows be gone.

I had to provide stack-safe implementations of both `List.concat` and `List.map`. It also still seems to lock up after a few hundred thousand search results. Perhaps we should impose some kind of limit, like vscode does. Not sure what exactly their limit is as it seems to vary by query, but it seems to be in the range of 25,000-50,000.

### DIFF
--- a/src/Core/Ripgrep.re
+++ b/src/Core/Ripgrep.re
@@ -2,6 +2,7 @@ open Rench;
 
 module Time = Revery_Core.Time;
 module List = Utility.List;
+module ListEx = Utility.ListEx;
 module Queue = Utility.Queue;
 
 module Match = {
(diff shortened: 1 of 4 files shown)


bryphe:
> Perhaps we should impose some kind of limit, like vscode does. Not sure what exactly their limit is as it seems to vary by query, but it seems to be in the range of 25,000-50,000.

This makes sense to me. Having 25k search results does not seem very useful - it's really a prompt that you need to refine your query. I think capping it there would be reasonable.

PR #1121, opened 2019-12-28
[WIP] Add to PATH Command.

CrossR:
Progress to add an "Add to Path" command for Mac/Linux.

 - [X] Copy over script and adopt for mac.
 - [ ] Adopt for Linux (`tar.gz`? `AppImage`?)
 - [ ] Verify on both.
    - [x] Mac
    - [ ] Linux
  - [x] Move to own action.

For now, I've jammed it into the KeyDisplayer, just so I can deal with the more awkward side of actually getting it working.

I think its now working on mac, in that the commands are valid and the script works to run the executable. However, it doesn't actually work when I make a release, but does in a dev build. I'm assuming there is some plist magic I need to deal with to say that we are now going to potentially request admin access or something.


CrossR:
>However, it doesn't actually work when I make a release, but does in a dev build. I'm assuming there is some plist magic I need to deal with to say that we are now going to potentially request admin access or something.

Was actually much simpler than that, I just had the resource path wrong.
Fixing that means it works on Mac now, though it still freezes the UI which is far from ideal.

I'll look at fixing that, but the Linux side may have to wait until next week as I need to be at home with my VMs to test that out properly.

Only other open question is where should the action live? I've moved it to a seperate action, but I can't think of a nice place it should live, so its still in an odd place for now.

bryphe:
This is awesome - thanks @CrossR !

> I'll look at fixing that, but the Linux side may have to wait until next week as I need to be at home with my VMs to test that out properly.

Sounds good. We can bring the Linux change in separately; and just use this PR for OSX.

> EDIT: Swapped to NodeTask now.

Cool! This should be async now 👍

bryphe:
> Only other open question is where should the action live? I've moved it to a seperate action, but I can't think of a nice place it should live, so its still in an odd place for now.

I think the place is reasonable for now, we can gather it up later. We might just want to swap it over to be an `Effect` (kind of like the `windowMoveEffect`): 
```
let addToPathEffect = Isolinear.Effect.create(~name="system.addToPath", () => {
    let _ =
      Oni_Extensions.NodeTask.run(
        ~scheduler=Scheduler.immediate,
        ~setup=Oni_Core.Setup.init(),
        "add-to-path.js",
      );
});
```

So that it gets logged / handled appropriately.

node/add-to-path.js
```diff
@@ -0,0 +1,73 @@
...
+        // TODO: Check this is valid for all use cases.
+        if (!fs.existsSync(imgPath)) {
+            imgPath = path.join(appDirectory, "assets", "images", "Onivim2.icns");
+        }
+
+        const options = { name: "Oni2", icns: imgPath };
+        let linkDest = "";
+
+        if (isMac()) {
+            // TODO: This isn't valid for dev builds, does that matter?
```
CrossR: This always links to a file over in the installed mac folder. Should it also work for a development build? I figured it wouldn't really be possible but perhaps I don't get the `esy` folder structure enough.

If not, it may make sense to gate the action and only make it accessible outside of dev builds. We need a gate in general to restrict this to mac/linux, so it doesn't appear in the menu on Windows.

bryphe: > This always links to a file over in the installed mac folder. Should it also work for a development build? I figured it wouldn't really be possible but perhaps I don't get the esy folder structure enough.

I think we'd want this to be relative to the executable - the `add-to-path.js` script should be in the same `Resources` folder as `run.sh`, so I believe we could just use `path.absolute(path.join(__dirname, "run.sh"))` here. This would't solve the dev build issue, but would help if the user was running it from another place (ie, if I run `esy create-release` -> run it from `open _release/Onivim2.app`.

> If not, it may make sense to gate the action and only make it accessible outside of dev builds. We need a gate in general to restrict this to mac/linux, so it doesn't appear in the menu on Windows.

Sounds good! For this PR - let's just gate this action for Mac-only.

src/Store/CommandStoreConnector.re
```diff
@@ -106,6 +106,12 @@ let createDefaultCommands = getState => {
         ~action=CopyActiveFilepathToClipboard,
         (),
       ),
+      Command.create(
+        ~category=Some("System"),
+        ~name="Add Oni2 to System PATH", /* TODO: This should be dynamic (Add/Remove) */
+        ~action=Command("system.addToPath"),
```
bryphe: We can gate the command here with `~enabled=Revery.Environment.os == Revery.Environment.mac`.

src/Store/CommandStoreConnector.re
```diff
@@ -106,6 +106,12 @@ let createDefaultCommands = getState => {
         ~action=CopyActiveFilepathToClipboard,
         (),
       ),
+      Command.create(
+        ~category=Some("System"),
+        ~name="Add Oni2 to System PATH", /* TODO: This should be dynamic (Add/Remove) */
```
bryphe: One thing that is a bit confusing is the script actually toggles the link, but the command is called 'Add'.

Perhaps we could just make the script always-add, and handle toggling here later? (We could add a 'Remove Oni2 from System Path' with an enabled check of seeing if the symlink exists, and augment our enabled check for 'Add' to check that the symlink _doesn't_ exist - but I'm OK deferring that).

bryphe:
I'll bring this in so we can get it in tonights build, and test it tomorrow 👍  Let me know if you have any concerns, @CrossR - thanks for all the work on this!

PR #1125, opened 2019-12-30
File Explorer - scroll to follow focus + spring + minor refactor

glennsl:
This implements scrolling to follow focus, and does so using a spring. It also refactors a few FsTreeNode functions to be more consistent with he rest, using string paths instead of node paths.

The scrolling logic has become a bit complicated, but could be simplified a bit with a spring hook that included a setter. It would also be nice to have a way to do "hard" transitions, like when scrolling with the mouse wheel, as it currently feels a bit sluggish.

### DIFF
--- a/src/Model/FileExplorer.re
+++ b/src/Model/FileExplorer.re
@@ -13,10 +13,10 @@ type t = {
 
 [@deriving show({with_path: false})]
 type action =
-  | TreeLoaded([@opaque] FsTreeNode.t)
-  | NodeLoaded(string, [@opaque] FsTreeNode.t)
-  | FocusNodeLoaded(string, [@opaque] FsTreeNode.t)
-  | NodeClicked([@opaque] FsTreeNode.t)
+  | TreeLoaded(FsTreeNode.t)
+  | NodeLoaded(FsTreeNode.t)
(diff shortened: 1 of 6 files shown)


bryphe:
> This implements scrolling to follow focus, and does so using a spring. It also refactors a few FsTreeNode functions to be more consistent with he rest, using string paths instead of node paths.

Very cool! This is helpful for maintaining context and having an idea of where files are relative to each other.

glennsl:
This should be good to go now. Still not particularly elegant, but I've encapsulated most of the nastiness in `useScroll` at least. The `setImmediately` function for the spring hook is also still a bit awkward as you have to simultaneously set the target value, or it'll just bounce back.

bryphe:
Tried out the latest - looks good to me @glennsl !

Something like this would be nice to have for the editor surface eventually... I wonder if we could reuse `useScroll` for that. The `setImmediately` functionality is helpful if the user is moving the scrollbars via mouse, but the 'spring' functionality would be nice for commands like `gg`/`G` to have context and feedback about the way the viewport moved. I guess the `itemHeight` could map to `lineHeight`, and the `viewportHeight` could map to the item height in pixels.

Very cool use of the hooks 👍

PR #1134, opened 2019-12-31
CI: Another attempt to remove console colors on windows

bryphe:
Trying to still resolve the CI hangs for Windows. A couple possibilities:
- An errant color / escape code is causing the terminal to pause and wait for input
- There is still a running process after closing

### DIFF
--- a/src/Core/Log.re
+++ b/src/Core/Log.re
@@ -202,10 +202,20 @@ let consoleReporter = {
   };
 };
 
+let winConsoleReporter = {
+  Logs.format_reporter(
+    ~app=Format.std_formatter,
+    ~dst=Format.std_formatter,
+    ()
+  );
+};
+
 let reporter =
   Logs.{
     report: (src, level, ~over, k, msgf) => {
-      let kret = consoleReporter.report(src, level, ~over=() => (), k, msgf);
+      let kret = Sys.win32 ? 
+      winConsoleReporter.report(src, level, ~over=() => (), k, msgf)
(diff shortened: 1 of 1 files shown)


src/Core/Log.re
```diff
@@ -202,10 +202,26 @@ let consoleReporter = {
   };
 };
 
+// Something is broken with the [consoleReporter] on Windows -
+// it can cause the terminal to hang. This happens on CI, but
+// also reproduces intermittently when run locally.
+// Switching to a simpler formatter solves the issue - but it'd be nice to have our
+// colors back on Windows!
```
glennsl: Interesting. This is the code that emits the colors codes, so the problem is likely here: 

https://github.com/dbuenzli/fmt/blob/master/src/fmt.ml#L729-L746

Like most things OCaml it has probably not been tested much on Windows, but if we can reproduce the problem on Windows with this it shouldn't be too hard to fix. It's not much code at all, so the possible error space seems small.

bryphe: Actually, I started testing on Windows - I ran `esy run -f --checkhealth` 10x, and it still hung on the last time. So I'm not 100% sure the log colors are the culprit!

Looking at process explorer - it seems like there are still some threads running even after calling `exit`. So I'm trying to figure out how to figure out how to narrow that down. I believe they are the threads associated with the syntax client. 

I might create a `ThreadHelper` that logs out when a thread starts, the id, and when it stops, and ensure that all created threads are closed prior to exiting in the health check. might give some clues...

bryphe:
Testing out an alternate fix in #1135

PR #1136, opened 2019-12-31
Fix #1089 - OSX: 'Couldn't get esy version' with RLS

### LINKED ISSUE #1089: Extension - Reason - OSX: 'Couldn't get esy version' error when run from Finder
In our OSX packaged builds, we're getting this error when using the reason language service: 
image

It's likely either an environment issue or a PATH issue - when running Onivim 2 from Finder, we might not have the environment context to discover `esy`. We might need to pick up / export the path environment for the exthsot.

bryphe:
__Issue:__ When opened from finder, on OSX, the `$PATH` isn't passed to the extension host - so `reason-language-server` can't find `esy`. The VSCode extensions expect to have some environment context like `$PATH`.

__Fix:__ Pull the `$PATH` out of the login shell. 

__Next steps:__
We might need to do additional work to expose the entirety of the environment, not just `$PATH`.

Fixes #1089


src/Core/ShellUtility.re
```diff
@@ -0,0 +1,27 @@
...
+    | None => Sys.win32 ? "cmd.exe" : "/usr/bin/bash"
+    };
+  };
+};
+
+let getShellPath = () => {
+  let res =
+    if (Sys.win32) {
+      Sys.getenv("PATH");
+    } else {
```
bryphe: Scope this to just OSX, for now.

PR #1141, opened 2019-12-31
Fix #1138 - vscode-python - Make 'author' parsing less strict

### LINKED ISSUE #1138: Extension - vscode-python: Error parsing `author`
From logs when loading `vscode-python`:
```
[ERROR]    +2ms Oni2.ExtensionScanner : Exception parsing /Users/bryphe/.config/oni2/extensions/ms-python-release/package.json : Failure("ExtensionManifest.t.author")
```

bryphe:
Fixes #1138

### DIFF
--- a/src/Extensions/ExtensionManifest.re
+++ b/src/Extensions/ExtensionManifest.re
@@ -15,6 +15,26 @@ module ExtensionKind = {
     | [@name "workspace"] Workspace;
 };
 
+module Author = {
+  [@deriving show]
+  type t = option(string);
+
+  let of_yojson_exn =
+    fun
+    | `String(v) => Some(v)
+    | v =>
+      Yojson.Safe.(
+        {
+          Util.member("name", v) |> Util.to_string_option;
+        }
+      );
+
+  let of_yojson = json =>
+    tryToResult(~msg="Error parsing author", () => of_yojson_exn(json));
+
+  let to_yojson = _author => `Null;
+};
+
 module Engine = {
   [@deriving (show, yojson({strict: false, exn: true}))]
   type t = {vscode: string};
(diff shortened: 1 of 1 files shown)


src/Extensions/ExtensionManifest.re
```diff
@@ -15,6 +15,26 @@ module ExtensionKind = {
...
+        {
+          Util.member("name", v) |> Util.to_string_option;
+        }
+      );
+
+  let of_yojson = json =>
+    tryToResult(~msg="Error parsing author", () => of_yojson_exn(json));
+
+  let to_yojson = _author => `Null;
+};
```
glennsl: Wow, that's a lot of ceremony for what's basically just [a single function](https://mattjbray.github.io/ocaml-decoders/decoders/decoders-yojson/Decoders_yojson/Basic/Decode/index.html#val-one_of) in a combinator-based API.

I wonder if we should consider using [`ocaml-decoders`](https://github.com/mattjbray/ocaml-decoders) instead. There are of course other benefits too, like decoding directly into the data structure we actually want, rather than using an intermediary representation for the JSON structure and having another layer of conversion code.

bryphe: I'm open to exploring it! It would be nice to streamline it - we have a lot of one-off JSON parsing logic like this all over.

glennsl: Do you have one which is just converted into a different data structure? I could convert that to `ocaml-decoders` to get a taste and show you how it's look.

bryphe: I imagine there are probably a _ton_ of opportunities here: https://github.com/onivim/oni2/blob/5d88ef8e359789c650698ba02e2c5daf7ad664a8/src/Extensions/ExtHostProtocol.re

That handles a lot of the JSON serialization <-> deserialization coming in/out of the extension host. A great place to simplify!

bryphe: I'll bring in this change for now, since it gets us to the next layer of issues with `vscode-python` - but interested to see how we can improve with `ocaml-decoders`.

Issue #707, opened 2019-08-27
filetype php


tofulm:
When I open a file .php with the last onivim2.appimage on linux ([7676fc5](https://github.com/onivim/oni2/commit/7676fc509955ebe40452b349bf229e792c2fe37f))
the filetype is not recognized.
When I try `:set filetype=php` It doesn't work any better.

CrossR:
We currently don't ship with any PHP syntax highlights, you can see the ones we do have over in https://github.com/onivim/oni2/tree/master/extensions.

A PR to add this should just be copying the extensions over from VSCode (#171, #636). That said, we are also looking at moving to `tree-sitter` powered syntax highlights as well, and PHP looks to have `tree-sitter` definitions, so we could use them in the future instead of the `textmate` ones.

We should probably raise a seperate issue to hook up `set ft`, since currently it doesn't do anything on the Oni side, where we'd rather send over a notification to Oni or something.

tofulm:
thanks

costra:
Why this enhancement was closed? I'm voting for PHP support, please!

CrossR:
I expect they installed it themselves locally!

I'll stick a PR in now for PHP, and if anyone else is wondering what the steps are it is as follows:

 - `git clone` VSCode
 - `cp -r vscode/extensions/LANG ~/git/oni2/extensions/`
 - Run Oni2 and just have a quick check with a test file.

Or, if its something more niche you think not worth commiting to the main repo, just copy to your extensions folder - https://onivim.github.io/docs/configuration/extensions#installing-extensions

PR #1155, opened 2020-01-02
Add PHP extension.

CrossR:
I slightly lied in the issue (https://github.com/onivim/oni2/issues/707), this wasn't as easy as just copying over.

It looks like the second part of the JSON was confusing Oni2, ([this part](https://github.com/microsoft/vscode/blob/af59519833a55eed13db8c9be4175395a400d3b2/extensions/php/package.json#L40-L51)). With that, we just treated it as HTML by the looks of things, rather than HTML. Removing it gives proper syntax in pure PHP files at least.

Is that known, or should I raise an issue?

### DIFF
--- a/extensions/php/.vscode/launch.json
+++ b/extensions/php/.vscode/launch.json
@@ -1,0 +1,18 @@
+{
+	"version": "0.2.0",
+	"configurations": [
+		{
+			"name": "Launch Extension",
+			"type": "extensionHost",
+			"request": "launch",
+			"runtimeExecutable": "${execPath}",
+			"args": [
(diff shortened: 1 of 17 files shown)


bryphe:
> It looks like the second part of the JSON was confusing Oni2, (this part). With that, we just treated it as HTML by the looks of things, rather than HTML. Removing it gives proper syntax in pure PHP files at least.

Ah, good catch - this is one gap we have in our textmate parser right now - it doesn't handle `embeddedLanguages` properly. Removing it seems reasonable, until we support that.

PR #1164, opened 2020-01-04
Fix #1160: Command Palette / Quickopen key bindings not working in some cases

### LINKED ISSUE #1160: Command Palette and Quick Open not opening on chord keypress
## What?
`Shift + CMD + P` and `CMD + P` do not trigger either of their UI elements to appear.
## Versioning
macOS 10.14.6
Onivim 2.0.3.0
## Notes
I posted to the discord under 'questions' it seems another user running Wayland is having the same issue.

bryphe:
__Issue:__ The format of our `keybindings.json` underwent a revamp to be closer in terms of compatibility with VSCode. 

However, in some older versions of Onivim 2, we dropped our default keybindings in a file - these would override the newer defaults, and some of the command names changed. Specifically:
- `quickOpen.open` was changed to `workbench.action.quickOpen`
- `commandPalette.open` was changed to `workbench.action.showCommands`

Fixes #1160 

__Fix:__ The JSON structure of the legacy format is different, so we can detect that case, and run an upgrader to address this change.

### DIFF
--- a/src/Input/Keybindings.re
+++ b/src/Input/Keybindings.re
@@ -1,6 +1,7 @@
 open Oni_Core;
 
 module List = Utility.List;
+module Result = Utility.Result;
 
 module Keybinding = {
   type t = {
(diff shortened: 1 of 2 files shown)


bryphe:
Thanks @szbergeron for your help troubleshooting!

Spirarel:
Thanks for attending to this so quickly!

Issue #80, opened 2019-02-23
Performance: Buffer Updates - Investigate & benchmark Piece Table approach


bryphe:
We currently have a sub-optimal strategy for handling buffer updates from Neovim. We do a lot of Array splicing which is slow for large buffers. We should start by benchmarking our existing performance characteristics.

In investigating various data structures that we could use to improve this:
- [Rope](https://en.wikipedia.org/wiki/Rope_(data_structure))
- [Gap buffer](https://en.wikipedia.org/wiki/Gap_buffer)
- [Piece Table](https://en.wikipedia.org/wiki/Piece_table)

The __Piece Table__ is the most appealing to me - it's an immutable data structure, so it's a very natural fit for our language and architecture. We don't necessarily need the undo faculty of it (as we defer that functionality to Neovim) - but we could periodically consolidate the immutable blocks when the application is idle - this would help offset some of the key downsides of the data structure (that it isn't self optimizing, that retrieving a byte involves iterating sequentially over the 'pieces').

bryphe:
The piece table should drastically reduce the cost of inserting into large buffers, as most of the cost of our previous strategy is because we incur lots of copying / memory cost in creating new arrays when splicing. Ideally, with the piece table strategy, the cost should be the same when inserting in a large buffer or a small buffer.

In the current state of the world, as measured by #96 , that is not the case:
 ```
+----------------------------------------------------------------------------------------------------------------------------------------------+
|  BENCHMARK                                      |  TIME               |  MINOR GC  |  MAJOR GC  |  MINOR ALLOC  |  PROMOTED  |  MAJOR ALLOC  |
|-------------------------------------------------+---------------------+------------+------------+---------------+------------+---------------|
|  Buffer: Add lines to empty buffer              |  1.22300958633      |  334       |  67        |  24359        |  1627      |  100002627    |
|-------------------------------------------------+---------------------+------------+------------+---------------+------------+---------------|
|  Buffer: Insert line in middle of small buffer  |  0.000998497009277  |  1         |  0         |  244029       |  25        |  25           |
|-------------------------------------------------+---------------------+------------+------------+---------------+------------+---------------|
|  Buffer: Insert line in middle of large buffer  |  0.546031951904     |  401       |  80        |  42349        |  3945      |  110005945    |
|-------------------------------------------------+---------------------+------------+------------+---------------+------------+---------------|
|  Buffer: Clear large buffer                     |  0.                 |  1         |  0         |  19029        |  25        |  25           |
```

PR #1170, opened 2020-01-05
Completions polish

### LINKED ISSUE #1057: UX - Polish: Completion Improvements
We have some code completion now!
image

...but there are several UX issues we need to fix:
- There is no way to cycle through the completion items. We should handle the `suggestNextItem` and `suggestPrevItem` commands.
- Only a max of five items are shown (a lot of the same issues as the Menu!) - we should show a scrollbar and the full set of items.
- There are some formatting issues with the control, as shown above - we should either expand or clamp the text.

glennsl:
Fixes #1057 

Implements selection of previous/next suggestion and adds scrolling through FlatList. No mouse interaction since I can't get it to respond to mouse events, but that doesn't seem all that important for this kind of UI anyway.

I also added an optional refactoring commit at the end. If you think this is too much for a single PR it can relatively easily be split into a separate PR, although it will of course still depend on this.


src/Model/CompletionMeet.rei
```diff
@@ -6,30 +6,26 @@ open CamomileBundled.Camomile;
...
 // the position where completions should be requested from, as well
 // as the 'base' - candidate text that should apply to the filtering of completion items.
-type t;
+type t = {
+  bufferId: int,
+  // Base is the prefix string
+  base: string,
+  // Meet is the location where we request completions
+  location: Location.t,
+};
```
glennsl: I don't see a good reason to hide this. The purpose of this data structure is pretty much just to bundle data together. Hiding it and adding constructor and accessor functions for everything is just extra boilerplate.

There are two motivation for doing encapsulation (at least that I can think of right now):

1. When you want to be able to change the data structure without breaking consumer code.
2. When you want to maintain an invariant internally, that is to restrict the possible values the data can gave.

Neither of these seem to apply in this case, and a few other cases I've seen, such as `Buffer`.

src/Model/Completions.re
```diff
@@ -71,51 +70,3 @@ let setFilter = (filter, model) => {
   filter: Some(filter),
   filtered: filterItems(Some(filter), model.all),
 };
-
-let reduce = (model, action: Actions.t) => {
```
glennsl: These have been moved to the `CompletionStore`. Having them all in one place makes it much easier to reason about what's going on.

src/Model/LanguageFeatures.re
```diff
@@ -134,10 +134,9 @@ let requestDocumentHighlights =
   |> DocumentHighlightProvider.request((buffer, location));
 };
 
-let requestCompletions =
-    (~buffer: Buffer.t, ~meet: CompletionMeet.t, ~location: Location.t, lf: t) => {
+let requestCompletions = (~buffer: Buffer.t, ~meet: CompletionMeet.t, lf: t) => {
   lf.completionProviders
-  |> CompletionProvider.request((buffer, meet, location));
+  |> CompletionProvider.request((buffer, meet, meet.location));
```
glennsl: The `location` argument seems like unnecessary duplication, since it's contained in `meet` and I can't see any good reason to request completions at a location other than the meet. `meet` and `location` a re carried together through a whole bunch of boundaries, and is unnecessary the entire way, but this was the simplest change from a user POV.

src/Store/CompletionStoreConnector.re
```diff
@@ -11,139 +11,192 @@ open Actions;
...
+    });
 };
 
-let applyCompletion = (state: State.t) =>
-  Isolinear.Effect.createWithDispatch(~name="vim.applyCompletion", dispatch => {
-    let maybeFocused =
-      Option.map(
-        i => state.completions.filtered[i],
-        state.completions.focused,
+module Actions = {
```
glennsl: this module clashes a bit with `Core.Actions`, but this contains actual actions, while `Core.Actions.t` are just messages, which might describe actions, but ought to and now often does describe events instead. So this is another case where the the `Actions` name causes a bit of confusion.

src/Store/CompletionStoreConnector.re
```diff
@@ -11,139 +11,192 @@ open Actions;
...
+              all,
+              focused:
+                model.focused == None && Array.length(filtered) > 0
+                  ? Some(0) : model.focused,
+              filtered,
+            },
+          state,
+        );
+      }
+
```
glennsl: I'm not entirely sure about the use of an `Actions` module here. It certainly makes this `updater` much clearer, and compositions well with `checkCompletionMeet`, but I'm not sure it'll work well more generally where it might just add indirection for little benefit.

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -406,55 +406,6 @@ let start = (extensions, setup: Core.Setup.t) => {
...
-        ~location,
-        languageFeatures,
-      );
-
-    Lwt.on_success(completionPromise, completions => {
-      dispatch(Model.Actions.CompletionAddItems(meet, completions))
-    });
-  };
-
-  let checkCompletionsEffect = state =>
```
glennsl: Moved to `CompletionsStore`.

src/UI/CompletionsView.re
```diff
@@ -68,192 +63,186 @@ let completionKindToIcon: option(Ext.CompletionItemKind.t) => int =
...
+    justifyContent(`Center),
+    border(~color=theme.editorSuggestWidgetBorder, ~width=1),
+    backgroundColor(theme.editorSuggestWidgetBackground),
+  ];
+
+  let detailText = (~editorFont: EditorFont.t, ~tokenTheme: TokenTheme.t) => [
+    textOverflow(`Ellipsis),
+    fontFamily(editorFont.fontFile),
+    fontSize(editorFont.fontSize),
+    Style.color(tokenTheme.commentColor),
```
bryphe: Removing the `backgroundColor` actually causes the text rendering quality to degrade - we don't do [gamma correction](https://www.freetype.org/freetype2/docs/text-rendering-general.html) when there is no `backgroundColor` - this causes the text to appear thinned out. Can be a significant degradation depending on the contrast between the background and foreground.

src/UI/CompletionsView.re
```diff
@@ -68,192 +63,186 @@ let completionKindToIcon: option(Ext.CompletionItemKind.t) => int =
...
+    top(int_of_float(lineHeight +. 0.5)),
+    Style.width(Constants.maxDetailWidth),
+    flexDirection(`Column),
+    alignItems(`FlexStart),
+    justifyContent(`Center),
+    border(~color=theme.editorSuggestWidgetBorder, ~width=1),
+    backgroundColor(theme.editorSuggestWidgetBackground),
+  ];
+
+  let detailText = (~editorFont: EditorFont.t, ~tokenTheme: TokenTheme.t) => [
```
bryphe: Is the detail text still showing?

__`master`:__
image

__`pr`:__
image

Also - if you compare the screenshots - you can see the difference the gamma-correction makes in the text rendering. The un-gamma-corrected text in the PR makes the text appear thin and washed out.

src/Store/KeyBindingsStoreConnector.re
```diff
@@ -145,12 +145,23 @@ let start = () => {
       },
       {
         key: "<CR>",
-        command: "insertBestCompletion",
-        condition: "suggestWidgetVisible" |> parseExp,
+        command: "acceptSelectedSuggestion",
+        condition:
+          "acceptSuggestionOnEnter && suggestWidgetVisible" |> parseExp,
```
bryphe: When / where does the `acceptSuggestionOnEnter` condition get set?

src/Store/ExtensionClientStoreConnector.re
bryphe: Thanks for consolidating this logic!

src/Store/CompletionStoreConnector.re
bryphe: Sounds good. Per our discussion - we plan on moving `Core.Actions` -> `Core.Events`

src/Model/CompletionMeet.rei
bryphe: Agree regarding `CompletionMeet`!

`Buffer`, though, the encapsulation makes sense I think - we'll likely pursue a different underlying implementation for the data structure (it is a naive array-of-strings right now, but that isn't the ideal data structure for performance) - long-term I picture us moving to a different structure, like a rope or piece table (IE, #80 ) - so having an abstraction like `getLine` means that we can change the underlying backing data store w/o needing the consumer to change the way they interact with `Buffer`s

src/UI/CompletionsView.re
glennsl: Ah, right,I think you've told me this before, but I forget. How hard is it to fix this properly in Revery? It's a real pain to pass the background color around when it's not static. Even worse with the new notifications in the StatusBar where the background and foreground colors are now animated.

src/Store/KeyBindingsStoreConnector.re
glennsl: It doesn't. Good catch. This was copied from the vscode keybindings, but I've forgotten to implement the configuration setting that would set this.

src/UI/CompletionsView.re
glennsl: Will look into it.

src/UI/CompletionsView.re
bryphe: It is definitely painful, and certainly shouldn't be necessary (and it isn't obvious!)

Skia handles this correctly for us. 

In the meantime, we could keep context of the last 'backgroundColor' we saw in the hierarchy and pass it down, but not sure if it is worth the trouble (since we'd pull it out anyway once Skia is in)

src/Model/CompletionMeet.rei
glennsl: There's no need to hide the entire data structure just because the structure of one field is prone to change. We can hide the structure of just that field instead. But yeah, that's absolutely a good reason for encapsulation.

(And also really interesting data structures to explore!)

src/UI/CompletionsView.re
glennsl: Is it worth the trouble to keep passing `backgroundColor` if the only issue is slightly degraded rendering quality, and if Skia will save us soon anyway?

src/UI/CompletionsView.re
glennsl: They're showing up for me. And even change when I move up and down the list. You don't seem to have any suggestion focused though. Not sure how you managed that.

src/Store/KeyBindingsStoreConnector.re
glennsl: Fixed in https://github.com/onivim/oni2/pull/1170/commits/67b1043f5bd2cbe268452fa2fd8b99c14c13e0b7

glennsl: It's a bit confusing to use the term "suggestion" in keybindings and configuration, and "completion". Is there a reason for using the latter instead?

src/Store/CompletionStoreConnector.re
bryphe: It's an interesting experiment! Thanks for exploring ways of improving this code.

src/UI/CompletionsView.re
bryphe: I see it now with the latest! It looks like this theme isn't showing the suggested highlight - but if I switch to another theme I can see it 🤔

bryphe:
A few bugs I noticed trying out the latest:

__Duplicated completions:__

If I open up a JavaScript file, and get suggestions, I see some duplication:
image

__Crash with certain completions:__

I found with some completions I'd get a crash:
image

I could repro this consistently if I typed `Log.i`, and scrolled down to `setLogFile`, and tried to complete it.

I get this callstack:
```
[ERROR]    +1ms Global : Exception Invalid_argument("index out of bounds"):
Raised by primitive operation at file "src/UI/CompletionsView.re", line 212, characters 53-65
Called from file "src/UI/EditorSurface.re", line 549, characters 10-25
```

__Completions sometimes don't show:__

While trying to reproduce the above, I noticed the completions wouldn't show for me around 8/10 times I tried. Sometimes, after typing `Log.i`, I'd get completions - but most of the time I wouldn't.

 However, I found if I moved the cursor (either by using the arrow keys or by clicking on the mouse and back), they would show up. I wonder if there is a difference in the timing of when we ask for completions? Did you see any intermittency in getting the completions when you were testing? If you can't reproduce - I can try and  get more details.

Having the scroll bar and being able to move through completions is awesome though - so nice to have that now!
image

src/UI/CompletionsView.re
bryphe: I think it is! The un-gamma-corrected text is really bad on some of my monitors, so I'd prefer to pass `backgroundColor` to address it.

> and if Skia will save us soon anyway

I just wish I knew how soon! There are still several blockers to work through before we can integrate it.

src/Store/KeyBindingsStoreConnector.re
bryphe: It's an unfortunate consequence of being VSCode-compatible - they actually use `suggest` and `completion` both!

In the extension host: 
```
	$provideCompletionItems(handle: number, resource: UriComponents, position: IPosition, context: modes.CompletionContext, token: CancellationToken): Promise<SuggestResultDto | undefined>;
	$resolveCompletionItem(handle: number, resource: UriComponents, position: IPosition, suggestion: modes.CompletionItem, token: CancellationToken): Promise<modes.CompletionItem>;
	$releaseCompletionItems(handle: number, id: number): void;|
```
https://github.com/onivim/vscode-exthost/blob/a25f426a04fe427beab7465be660f89a794605b5/src/vs/workbench/api/node/extHost.protocol.ts#L971

(And note they even use `suggestion: modes.CompletionItem` above).

But in the keybindings / configuration, they use `suggest`. I figured it was best to keep the 'code-facing' parts across the protocol consistent - and use completion up until the point where keybindings / configuration enter the picture.

Either way - there will be some inconsistency, either from keybindings -> our code, or our code -> extension host.

src/Core/ConfigurationParser.re
```diff
@@ -144,6 +144,19 @@ let configurationParsers: list(configurationTuple) = [
     "editor.matchBrackets",
     (s, v) => {...s, editorMatchBrackets: parseBool(v)},
   ),
+  (
+    "editor.acceptSuggestionOnEnter",
```
bryphe: Thanks for adding this!

src/UI/CompletionsView.re
```diff
@@ -68,192 +63,186 @@ let completionKindToIcon: option(Ext.CompletionItemKind.t) => int =
...
+      | None => React.empty
+      };
+    | None => React.empty
+    };
+
+  <View style={Styles.outerPosition(~x, ~y)}>
+    <Opacity opacity=Constants.opacity>
+      <View
+        style={Styles.innerPosition(~height, ~width, ~lineHeight, ~theme)}>
+        <FlatList
```
bryphe: Having the scroll bar and being able to move through completions is awesome - so nice to have that now!
image

🎉 🎉

src/UI/CompletionsView.re
glennsl: Alright, will do. I don't think ti's too bad of a fix here. The notifications UI will be more convoluted though.

src/Store/KeyBindingsStoreConnector.re
glennsl: Ah, thanks for explaining. We're using "completions" to talk about the feature generally though, in issues and when talking with users. We should probably use "suggestions" instead and just deal with the confusion ourselves.

glennsl:
Oh man, those are some nasty bugs!

> Did you see any intermittency in getting the completions when you were testing?

I've tested mostly with JavaScript completions, which seemed to work consistently. Although apparently I've missed the duplications. But when testing the details I did notice that it took a while for completions to show up. Figured that was just because it had to build or something before delivering completions. I'm used to merlin and rls being pretty flaky, so didn't think much about it.

I'll check these out right away.

glennsl:
> Issue 1: Duplicated completions

Hmm, not able to reproduce this even when doing exactly what the screenshot shows. Can you consistently reproduce it, and in different conditions?

> Issue 2: Crash with certain completions

Oops, fixed in https://github.com/onivim/oni2/pull/1170/commits/3dfa6086386a3c9d5f58eceefa020b0c9dbf7081. The problem was that the focus wasn't adjusted when resetting the completions after applying it. This invariant is now maintained entirely within `Completions`, and an interface added where `Completions.t` is made private so it has to be manipulated using only the operations provided.

> Issue 3: Completions sometimes don't show
> Issue 4: Completions persist pass that their 'meet'

These are the same issue. I botched the condition between narrowing and restarting. Fixed in https://github.com/onivim/oni2/pull/1170/commits/affdc60f3c4bfd8a26ec447f78143edaa64c2a21

bryphe:
> Hmm, not able to reproduce this even when doing exactly what the screenshot shows. Can you consistently reproduce it, and in different conditions?

Just tried again, and wasn't able to reproduce it - will keep an eye out though if I see it again.

src/Store/KeyBindingsStoreConnector.re
bryphe: Good point - it would make sense in the documentation to reference "suggestions" so that the configuration, documentation, and key bindings are consistent.

src/Model/Actions.re
```diff
@@ -30,13 +30,10 @@ type t =
   | CommandsRegister(list(command))
   // Execute a contribute command, from an extension
   | CommandExecuteContributed(string)
-  | CompletionStart([@opaque] CompletionMeet.t)
```
bryphe: Thanks for cleaning this up 👍

src/Store/CompletionStoreConnector.re
```diff
@@ -6,106 +6,151 @@
...
 
-open Actions;
+module Log = (val Log.withNamespace("Oni2.CompletionStore"));
 
-type lastCompletionMeet = {
-  completionMeetBufferId: int,
-  completionMeetLine: Index.t,
-  completionMeetColumn: Index.t,
-};
+module Effects = {
```
bryphe: I like this style of putting effects in an `Effects` module 👍

src/Store/QuickmenuStoreConnector.re
```diff
@@ -230,53 +232,31 @@ let start = (themeInfo: Model.ThemeInfo.t) => {
...
               ...state,
               focused:
-                Option.map(
-                  focused =>
-                    if (count == 0) {
-                      0;
-                    } else {
-                      (focused + 1) mod count;
-                    },
+                IndexEx.nextRollOverOpt(
```
bryphe: Good idea to abstract this logic into the `IndexEx` utilities

src/Store/CompletionStoreConnector.re
glennsl: Doesn't work if the effects depend on something passed in through `start` though, unfortunately :confused:

src/Store/QuickmenuStoreConnector.re
glennsl: Not entirely sure where to put them though. They operate on index-like values, but aren't actually `Index.t`s, sp `IndexEx` is a bit misleading. And the name is a bit kludgy in general. It's Gooed Enough For Now :tm: though.

PR #1180, opened 2020-01-07
Auto-Closing Pairs: Add syntax scope, respect 'notIn' configuration

bryphe:
This adds the concept of a `SyntaxScope` which is a simplified syntax scope of use in language features. In particular, many of the language features are gated by being in a string or comment (for example, in Reason, we'd want to disable the `"` auto-closing pair when in a string - otherwise closing a string gets really annoying!).

This wires up our syntax highlighting to provide that syntax scope. The next step is to leverage it in auto-closing pairs, so we account for that when necessary.

__TODO:__
- [x] Add 'languageConfiguration' that we can use to store the auto-closing pairs information, and other metadata, from extensions
- [x] Implement on-demand (cached) loading of languageConfiguration. Query when we call `Vim.input` to properly set up the auto closing information.
- [x] Add a function to get the current syntax scope from a buffer position. Given a line, byte, buffer id, and syntax tokens -> get scope. Run filter on applied rules when creating AutoClosing info for vim input.
- [x] __reason-libvim__ - change the ACP API to be a 'push' in the input, as opposed to requiring synchronization

I'm going to move the configuration / enabling out to test these latest changes before moving it out of the 'experimental' phase:
__Deferred:__
- Handle string parsing in `editor.autoClosingBrackets`. Start by supporting `languageDefined` and `never`.
- Hook up `editor.autoClosingBrackets` to gate auto-closing functionality


src/Core/SyntaxScope.re
```diff
@@ -0,0 +1,44 @@
...
+
+// Type defining a simplified 'syntax scope' of a token -
+// is it a comment, string, both, or neither?
+// This is important for features like auto-closing pairs -
+// we may not care to close certain characters in contexts like strings or comments.
+type t = 
+| None
+| String
+| Comment
+| StringAndComment;
```
glennsl: Maybe I'm missing something, but it seems more appropriate to model this as two `bool`s.  That would greatly simplify the two functions below, and probably other code as well.

src/Extensions/LanguageConfiguration.re
```diff
@@ -1,14 +1,16 @@
...
 open Oni_Core.Utility;
 module Json = Oni_Core.Json;
 
 module AutoClosingPair = {
   type scopes =
     | String
     | Comment
-    | Other(string);
+    | Other(string)
+    | None;
```
glennsl: It's not a good idea to use `None` as a custom constructor as it conflicts with `option`. Better to use `Nothing` or something like that.

src/Extensions/LanguageConfigurationLoader.re
```diff
@@ -16,6 +23,7 @@ let create = languageInfo => {
           loadResult
           |> Json.Decode.decode_value(LanguageConfiguration.decode)
           |> Stdlib.Result.map_error(Json.Decode.string_of_error)
+          |> Stdlib.Result.map_error(FunEx.tap(err => Log.error(err)))
```
glennsl: This can be eta reduced to `Stdlib.Result.map_error(FunEx.tap(Log.error))`

src/Extensions/LanguageConfigurationLoader.re
```diff
@@ -1,11 +1,18 @@
 open Oni_Core;
+open Oni_Core.Utility;
+
+module Log = (val Log.withNamespace("Oni2.LanguageConfigurationLoader"));
 
 type t = LazyLoader.t(LanguageConfiguration.t);
 
 let create = languageInfo => {
   let loadFunction = language => {
+    Log.tracef(m => m("Loading configuration for language: %s", language));
```
glennsl: Is this going to be run so frequently that it would be overwhelming with `debug` logging?

src/Extensions/LanguageConfiguration.re
```diff
@@ -0,0 +1,137 @@
...
+           | "comment" => Comment
+           | v => Other(v),
+         );
+
+    let scopeToList = scope |> map(v => [v]);
+
+    // Pulled this from: https://github.com/mattjbray/ocaml-decoders/blob/550c05c16384c9f8d2ab6a36eebc8e1a2684a7e0/src/decode.mli#L106
+    let (>>==::) = (fst, rest) => uncons(rest, fst);
+
+    let tuple =
```
bryphe: This was my first go at using our new JSON decoders - maybe there is a cleaner way @glennsl ? I wanted to parse `["a", "b"]` into `{ openPair: "a", closePair: "b" }`.

I pulled that `>>==::` operator from the docs, but have no idea what it means (perhaps it's a standard nomenclature somewhere).

I think the rest of it was pretty straightforward, though! I like the warm fuzzy feeling of having it covered by the type system... when it compiled, it worked exactly as I expected.

src/Extensions/LanguageConfigurationLoader.re
bryphe: I don't think so - that seems like a reasonable upgrade. Right now, it's on user input, but wouldn't be the bottleneck (and only happens once). I'll make that change

src/Extensions/LanguageConfiguration.re
bryphe: Good point - I realized this isn't necessary anyway. I'll take it out.

src/Extensions/LanguageConfiguration.re
glennsl: Yeah, I'm not a big fan of cryptic symbols, and it definitely doesn't help that `refmt` treats them as linebreaks.

I'd just decode it as a list, then `map` it using a function that matches exactly a two-element array:

```
list |> map(fun | [openPair, closePair] => { openPair, closePair } | _ => failwith("oh no!"))
```

Although it might be more appropriate to use `and_then`, `succeed` and `fail`:

```
list |> and_then(fun
            | [openPair, closePair] => succeed({ openPair, closePair })
            | _ => fail("oh no!")
          );
```

glennsl: Otherwise the decoders look great!

bryphe: Ah, thank you! I was looking for something like `and_then` and couldn't find it (was searching for a `bind`, `flatMap`, etc) - that's exactly what I needed. Much cleaner than what I had before, w/o the cryptic symbols.

Updated, and tests pass still pass 👍

glennsl: `and_then` is what Elm and Rust calls it, because it usually reads better. Not very intuitive when trying to find it though.

PR #1181, opened 2020-01-07
[WIP] feat(language-support): add UI for find-all references

bryphe:
We actually already have a basic API for find-all references, but we're not surfacing it in the UI anywhere.

This adds a find-all references pane, alongside the other panes we currently have:
image

__TODO:__
- [ ] Show line info for a buffer, when available
- [ ] Add shortcut key to toggle showing / hiding
- [ ] Add error notification when references are not available upon request

### DIFF
--- a/src/Model/Actions.re
+++ b/src/Model/Actions.re
@@ -105,6 +105,7 @@ type t =
   | QuitBuffer([@opaque] Vim.Buffer.t, bool)
   | Quit(bool)
   | RegisterQuitCleanup(unit => unit)
+  | ReferencesHotKey
   | SearchClearMatchingPair(int)
   | SearchSetMatchingPair(int, Location.t, Location.t)
   | SearchSetHighlights(int, list(Range.t))
--- a/src/Model/Pane.re
+++ b/src/Model/Pane.re
@@ -7,7 +7,8 @@
 [@deriving show({with_path: false})]
 type paneType =
   | Search
-  | Diagnostics;
+  | Diagnostics
+  | References;
 
 type t = {
   activePane: paneType,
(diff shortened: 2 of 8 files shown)


src/Model/Actions.re
```diff
@@ -105,6 +105,7 @@ type t =
   | QuitBuffer([@opaque] Vim.Buffer.t, bool)
   | Quit(bool)
   | RegisterQuitCleanup(unit => unit)
+  | ReferencesHotKey
```
zephraph, line 108: I'm somewhat coming from a place of ignorance here so excuse the question if it's misinformed. There are a relative few actions associated by name with a key press. Is that solely because there are relatively few actions triggered by key presses? Again, with little context, it would seem to me that most actions would benefit from not being directly associated to a particular usage. I.e. `FindReferences` vs `ReferencesHotKey`. 

I don't feel strongly about that, just looking to learn!

bryphe:
Moved to #2532

src/Model/Actions.re
bryphe: Hi @zephraph,

Just saw this - sorry for the late reply!

> There are a relative few actions associated by name with a key press. Is that solely because there are relatively few actions triggered by key presses? 

This is a good question. We moved from having all our actions in this top-level `Actions` module to having them owned by particular 'Features'. So instead of defining all these hot keys in top-level actions, we've moved them to the relevant features, like:

https://github.com/onivim/oni2/blob/e797a3a0f913e0a1b87907ac7bf7266e156281cb/src/Feature/Pane/Feature_Pane.re#L470

It's mainly to strive for more encapsulation - letting individual features be fully defined end-to-end (from their model, to the potential actions / commands, available keybindings, configuration) - and bundling it up into an [interface](https://github.com/onivim/oni2/blob/master/src/Feature/Pane/Feature_Pane.rei) we can compose with the rest of the editor.

Hope that helps answer your question!

PR #1182, opened 2020-01-07
Feature - UX - Semibold / italic font for UI

bryphe:
Test out moving from Selawik to Inter-UI, which has additional styles (like italics) that we can use in our UI: 

image

__TODO:__
- [x] Investigate incorrect text offset in vim commandline (This was fixed by @fanantoxa in #1231)
- [x] Try semibold styling for the 'mode' in the statusbar


bryphe:
Will give this a shot! The addition of italic / semibold / bold gives us some additional tools for improving our UI. We can switch the font down the road, but the plumbing is here at least.

PR #1184, opened 2020-01-08
[WIP] Sexy Notifications (ie. Notifications Revamp)

### LINKED ISSUE #849: UI: Notification experience improvements

glennsl:
Fixes #849

Peek 2020-01-08 05-26

This can be merged as-is, although there are still some issues and remaining work:

- [x] Pass along background color to all text elements in order to get proper gamma correction
- [ ] Context menu with Clear All action (turned out to be more pain to implement than expected, for various reasons)
- [ ] Animations will replay when older notifications are removed since keys aren't actually used during reconciliation (needs to be fixed upstream in Revery).
- [ ] Garbage collect expired notifications in the `useExpiration` hook.
- [ ] Refactor Pane modules


src/Core/Utility.re
```diff
@@ -894,3 +894,15 @@ module Json = {
...
+module ColorEx = {
+  open Revery;
+
+  let mix = (~start, ~stop, ~amount) =>
+    Color.{
+      r: (stop.r -. start.r) *. amount +. start.r,
+      g: (stop.g -. start.g) *. amount +. start.g,
+      b: (stop.b -. start.b) *. amount +. start.b,
+      a: (stop.a -. start.a) *. amount +. start.a,
+    };
```
glennsl: Should probably be upstreamed to Revery

src/UI/GlobalContext.re
```diff
@@ -24,7 +24,7 @@ type t = {
   setActiveWindow: (int, int) => unit,
   openEditorById: int => unit,
   closeEditorById: int => unit,
-  hideNotification: int => unit,
+  hideNotification: Notification.t => unit,
```
glennsl: Should be removed in favor of using `dispatch`

src/UI/StatusBar.re
```diff
@@ -14,32 +14,193 @@ open Oni_Model;
...
+          |> List.filter(item =>
+               !List.exists(((it, _t)) => equals(it, item), active)
+             );
+
+        if (untracked != []) {
+          let init = item => (item, time);
+          setActive(tracked => List.map(init, untracked) @ tracked);
+        };
+
+        // TODO: Garbage collection of expired, but on what condition?
```
glennsl: TODO

glennsl:
I've tried to get a context menu working, but am bumping up against a lot of rough edges. First with right-click not working (fixed in https://github.com/revery-ui/revery/pull/706), then a bunch of issues with value restriction errors caused by the brisk ppx, and now nested clickables not working.

I've therefore moved it [a separate branch], although still based on this, and recommend this be merged as-is since it might take a while to sort out these issues as well as the design of it.

bryphe:
Wow! 

This looks incredible @glennsl - such a huge improvement over what we had before. Awesome work!

src/Core/Utility.re
bryphe: Makes sense, this would be helpful in the `Color` module

bryphe:
> I've therefore moved it a separate branch, although still based on this, and recommend this be merged as-is since it might take a while to sort out these issues as well as the design of it.

Sounds good - thanks for fixing some of the initial issues (like the mouse click issue) - tracking the context menu in a separate branch sounds reasonable 👍 The notification experience w/o the context menu is still so much improved over what we have today.

src/Model/Notification.re
```diff
@@ -1,12 +1,29 @@
-open Actions;
+[@deriving show({with_path: false})]
+type t = {
```
bryphe: Thanks for cleaning up the types here and decoupling from Actions

src/UI/CustomHooks.re
```diff
@@ -0,0 +1,41 @@
+open Oni_Core;
+
+open Revery;
+open Revery.UI;
+
+module ColorEx = Utility.ColorEx;
+
+let colorTransition =
```
bryphe: Neat way to build on the `animation` hook 👍

glennsl: Should we upstream this to Revery too? It seems a little bit special-purpose, and should perhaps ideally be in some extended library for Revery.

PR #1188, opened 2020-01-08
Recognize all re files as Reason

imbsky:
This change enables syntax highlighting correctly and may improve readability on the GitHub.


bryphe:
Awesome, thank you @imbsky !

PR #1191, opened 2020-01-11
Cleanup syntax server process

### LINKED ISSUE #1144: Launching from Python with subprocess.run() or os.system() leaves language highlight server running HOT.
# What I did:

I have a [fman](https://fman.io/) extenstion for launching my current editor specified in a file [Open With Editor](https://github.com/raguay/OpenWithEditor). In order to use it with Oni2, I had to add a line to launch the editor using the commandline. Since the modules are written in Python, I used the subprocess.run() command using the path `/Applications/Onivim2.App/Contents/MacOS/Oni2`.

# What happened:

It launches the editor just fine and doesn’t take up much memory or cpu. But, when I exit the editor in any way (`qall`, `wqall`, etc), the process `Oni2_editor —syntax-highlight-service` goes berserk and has to be forceably killed. I didn’t realize this at first and ran several instances (about 4) in this manner. Four of those services were left running at 90% cpu (I had a terminal with htop filtering for Oni2). The system would not respond to anything at all and I had to force power down.

# What I would expect:

Upon exiting the editor, it should kill any subprocess that it launches! When a subprocess sees it’s orphaned, it should terminate itself and not go berserk.

# Recreating:

I’ve tried to recreate this on a command line, but it doesn’t happen. Not sure what about the execution process is going astray for the Python subprocess, but it will not mimic on a normal command line.

CrossR:
Fixes #1144 (Need to test on non-Windows)

Clean up the language server process when finished.

### DIFF
--- a/src/Store/SyntaxHighlightingStoreConnector.re
+++ b/src/Store/SyntaxHighlightingStoreConnector.re
@@ -95,6 +95,16 @@ let start =
         )
       });
 
+    let registerQuitCleanupEffect =
+      Isolinear.Effect.createWithDispatch(
+        ~name="syntax.registerQuitCleanup", dispatch =>
+        dispatch(
+          Model.Actions.RegisterQuitCleanup(
+            () => Oni_Syntax_Client.close(_syntaxClient),
+          ),
+        )
+      );
+
     let isVersionValid = (updateVersion, bufferVersion) => {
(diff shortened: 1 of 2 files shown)


bryphe:
> Fixes #1144 (Need to test on non-Windows)

Tested on my Macbook - working for me 👍

PR #1196, opened 2020-01-13
Logging renovation

glennsl:
This converts every instance of logging to use namespaced loggers, and goes through every log message to tweak them for consistency and log level granularity.

I've used roughly [the conventions suggested by @dbuenzli/logs](https://erratique.ch/software/logs/doc/Logs/index.html#usage):

- __Error__ if it WILL cause unexpected behaviour
- __Warn__ if it MIGHT cause unexpected behaviour, but also might not, i.e. it's suspicious, but not a definite problem.
- __Info__ if it is both understandable AND actionable by the end-user
- __Debug__ otherwise

There's lots of log messages I'm unsure about though, so please keep an extra eye on this during review.

I've also removed a few log messages that just reiterates a dispatched message.

### DIFF
--- a/src/Core/Cli.re
+++ b/src/Core/Cli.re
@@ -5,6 +5,9 @@
  */
 open Rench;
 
+module CoreLog = Log;
+module Log = (val Log.withNamespace("Oni2.Core.Cli"));
+
 type t = {
   folder: string,
   filesToOpen: list(string),
@@ -105,7 +108,7 @@ let parse =
     "",
   );
 
-  if (!Log.isPrintingEnabled()) {
+  if (!CoreLog.isPrintingEnabled()) {
     /* On Windows, detach the application from the console if we're not logging to console */
     Utility.freeConsole();
   };
--- a/src/Core/Cli.re
+++ b/src/Core/Cli.re
@@ -44,7 +44,7 @@ let show = (v: t) => {
 let noop = () => ();
 
 let setWorkingDirectory = s => {
-  Log.info("--working-directory - chdir: " ++ s);
+  Log.debug("--working-directory - chdir: " ++ s);
   Sys.chdir(s);
 };
 
--- a/src/Store/ExtensionClientStoreConnector.re
+++ b/src/Store/ExtensionClientStoreConnector.re
@@ -12,9 +12,7 @@ module Uri = Core.Uri;
 open Oni_Core.Utility;
 module Model = Oni_Model;
 
-module Log = (
-  val Core.Log.withNamespace("Oni2.ExtensionClientStoreConnector")
-);
+module Log = (val Core.Log.withNamespace("Oni2.Extension.ClientStore"));
 
 open Oni_Extensions;
 module Extensions = Oni_Extensions;
--- a/src/Store/ExtensionClientStoreConnector.re
+++ b/src/Store/ExtensionClientStoreConnector.re
@@ -7,12 +7,12 @@
  */
 
 open EditorCoreTypes;
-module Core = Oni_Core;
-module Uri = Core.Uri;
-open Oni_Core.Utility;
-module Model = Oni_Model;
+open Oni_Core;
+open Oni_Model;
 
-module Log = (val Core.Log.withNamespace("Oni2.Extension.ClientStore"));
+module Uri = Oni_Core.Uri;
+module Log = (val Log.withNamespace("Oni2.Extension.ClientStore"));
+module Option = Utility.Option;
 
 open Oni_Extensions;
 module Extensions = Oni_Extensions;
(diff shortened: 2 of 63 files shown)


src/Core/Cli.re
```diff
@@ -5,6 +5,9 @@
  */
 open Rench;
 
+module CoreLog = Log;
+module Log = (val Log.withNamespace("Oni2.Core.Cli"));
```
glennsl: I've added the project as an extra namespaces as well, since ti disambiguates a few things and even allows some namespaces to be shorter if some of the redundancy is removed.

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -218,15 +216,12 @@ let start = (extensions, setup: Core.Setup.t) => {
...
-      Oni_Model.Actions.LanguageFeature(
-        Model.LanguageFeatures.DefinitionProviderAvailable(
-          id,
-          definitionProvider,
-        ),
+      Actions.LanguageFeature(
+        LanguageFeatures.DefinitionProviderAvailable(id, definitionProvider),
       ),
     );
-    Log.infof(m => m("Registered suggest provider with ID: %n", provider.id));
```
glennsl: Removed because the action dispatched just before includes exactly the same information

src/Store/FontStoreConnector.re
```diff
@@ -136,7 +130,7 @@ let start = (~getScaleFactor, ()) => {
             };
 
           let onError = errorMsg => {
-            Log.error("setFont: Failed to load font " ++ fullPath);
+            Log.error("Failed to load font " ++ fullPath);
 
             dispatch(
               ShowNotification(
```
glennsl: Kept despite `ShowNotification` including the same message, because the log level is error, which is quite significant.

src/Store/KeyBindingsStoreConnector.re
```diff
@@ -202,7 +204,7 @@ let start = () => {
           | Some(fileName) => fileName
           };
         if (bufferFileName == configPath && cmd == Vim.Types.BufWritePost) {
-          Oni_Core.Log.info("Reloading key bindings from: " ++ configPath);
+          Log.info("Reloading key bindings from: " ++ configPath);
           dispatch(Actions.KeyBindingsReload);
```
glennsl: Kept despite the dispatched action because the log message includes additional information

src/Store/KeyBindingsStoreConnector.re
```diff
@@ -236,11 +238,7 @@ let start = () => {
...
         };
 
-      Log.info(
-        "Loading "
-        ++ string_of_int(List.length(keyBindings))
-        ++ " keybindings",
-      );
+      Log.infof(m => m("Loading %i keybindings", List.length(keyBindings)));
 
       dispatch(Actions.KeyBindingsSet(defaultBindings @ keyBindings));
```
glennsl: This might be a candidate for removal though. But not sure how significant key keybinding count is.

src/Store/VimStoreConnector.re
```diff
@@ -121,7 +121,7 @@ let start =
         | Info => ("INFO", Notification.Info)
         };
 
-      Log.infof(m => m("Message - %s [%s]: %s", priorityString, t, msg));
+      Log.debugf(m => m("Message - %s [%s]: %s", priorityString, t, msg));
 
       dispatch(ShowNotification(Notification.create(~kind, msg)));
```
glennsl: This might be removed, but contains some additional information. Unfortunately I have no idea what `t` is since the name is so indescriptive.

src/Store/VimStoreConnector.re
```diff
@@ -172,7 +172,7 @@ let start =
 
   let _ =
     Vim.Buffer.onModifiedChanged((id, modified) => {
-      Log.infof(m => m("Buffer metadata changed: %n | %b", id, modified));
+      Log.debugf(m => m("Buffer metadata changed: %n | %b", id, modified));
       dispatch(Actions.BufferSetModified(id, modified));
```
glennsl: Kept for now because `BufferSetModified` seems like a function call by action.

src/Store/KeyBindingsStoreConnector.re
bryphe: We should keep this, because we don't get a count from the keybindings - the count of keybindings is actual helpful for drilling into issues like https://github.com/onivim/oni2/pull/1164 (ie, were we able to load any bindings)

bryphe:
> Error if it WILL cause unexpected behaviour
Warn if it MIGHT cause unexpected behaviour, but also might not, i.e. it's suspicious, but not a definite problem.
Info if it is both understandable AND actionable by the end-user
Debug otherwise

We might want to add these guidelines to a doc somewhere, too! 👍

src/Store/VimStoreConnector.re
bryphe: Ah, sorry - `t` should be `title` (some messages include a title)

glennsl:
Added documentation of log priority levels in https://github.com/onivim/oni2/pull/1196/commits/392955252014855b2cc0e1c112dd87251e083451

Issue #958, opened 2019-11-20
Ubuntu build dependencies missing from build steps.


fanantoxa:
Can you help please with:
```
error: build failed with exit code: 1
  build log:
    # esy-build-package: building: reason-libvim@github:onivim/reason-libvim#f50c7da
    # esy-build-package: pwd: /home/fanantoxa/.esy/3________________________________________________________________/b/reason_libvim-1c3edde1
    # esy-build-package: running: 'refmterr' 'dune' 'build' '-p' 'libvim'
        discover src/c_flags.sexp,src/cxx_flags.sexp,src/flags.sexp
    -L/home/fanantoxa/.esy/3/i/libvim-8.10869.24-2060eb22/lib
        ocamlopt src/vim.cmxs (exit 2)
    (cd _build/default && /home/fanantoxa/.esy/3/i/ocaml-4.7.1004-a6ae015e/bin/ocamlopt.opt -w -40 -g -shared -linkall -I src -o src/vim.cmxs src/vim.cmxa)
    /usr/bin/ld: cannot find -lacl
    collect2: error: ld returned 1 exit status
    File "caml_startup", line 1:
    Error: Error during linking
    error: command failed: 'refmterr' 'dune' 'build' '-p' 'libvim' (exited with 1)
    esy-build-package: exiting with errors above...
    
  building reason-libvim@github:onivim/reason-libvim#f50c7da
esy: exiting due to errors above

```

But
```
sudo apt install acl
Reading package lists... Done
Building dependency tree       
Reading state information... Done
acl is already the newest version (2.2.52-3build1).
```

CrossR:
I think the package is actually `libacl1-dev` for Ubuntu.

(Assuming that works, we should update the build instructions to mention that on Ubuntu, since a specific dependency of Oni2 rather than Revery, since its only needed for `libvim`.)

fanantoxa:
@CrossR Another error after:
```
info esy build 0.5.7 (using package.json)
info building reason-libvim@github:onivim/reason-libvim#f50c7da@d41d8cd9
info building reason-libvim@github:onivim/reason-libvim#f50c7da@d41d8cd9: done
File "assets/configuration/dune", line 4, characters 8-18:
4 |  (files setup.json))
            ^^^^^^^^^^
Error: No rule found for assets/configuration/setup.json
error: command failed: 'refmterr' 'dune' 'build' '-p' 'Oni2' '-j4' (exited with 1)
esy-build-package: exiting with errors above...
[...]

PR #1199, opened 2020-01-14
Update PATH docs.

CrossR:
Just updates the CLI docs.

Hoping to get the Linux "Add to PATH" stuff sorted soon.


CrossR:
Also updated docs to fix #958.

bryphe:
Awesome, thanks @CrossR !

CrossR:
I'm actually adding some more docs here, around the ability to pass files/folders etc and Zen mode. So will merge once that is added.

PR #1201, opened 2020-01-14
Convert from in-house animation to Revery's animation framework

glennsl:
This also refactors the Hover functionality, removing the models, fixing the animation being reset on every cursor move, and making it more or less ready for multiple cursors.


bryphe:
I'm seeing an empty 'hover' square show up when testing this:

__Repro:__
- Open `Notifications.re`
- Move cursor (`jjjw`)

image

src/Store/LanguageFeatureConnector.re
```diff
@@ -103,56 +102,32 @@ let start = () => {
   let updater = (state: Model.State.t, action: Actions.t) => {
     let default = (state, Isolinear.Effect.none);
     switch (action) {
-    | Actions.Tick({deltaTime, _}) =>
```
bryphe: Nice to see this go away!

src/UI/HoverView.re
```diff
@@ -3,97 +3,154 @@
...
+  let maybeEditor =
+    state |> Selectors.getActiveEditorGroup |> Selectors.getActiveEditor;
+
+  switch (maybeEditor) {
+  | Some(editor) when hoverEnabled && state.mode != Vim.Types.Insert =>
+    switch (Buffers.getBuffer(editor.bufferId, state.buffers)) {
+    | Some(buffer) =>
+      <View style=Styles.container>
+        {editor.cursors
+         |> List.map(cursor =>
```
bryphe: Interesting, once we have multi cursors we'll have hover support for each individual one!

src/UI/HoverView.re
```diff
@@ -3,97 +3,154 @@
...
+    Configuration.getValue(c => c.editorHoverEnabled, state.configuration);
+
+  let State.{theme, editorFont, diagnostics, _} = state;
+  let maybeEditor =
+    state |> Selectors.getActiveEditorGroup |> Selectors.getActiveEditor;
+
+  switch (maybeEditor) {
+  | Some(editor) when hoverEnabled && state.mode != Vim.Types.Insert =>
+    switch (Buffers.getBuffer(editor.bufferId, state.buffers)) {
+    | Some(buffer) =>
```
bryphe: Related to the issue / screenshot I logged - seems like we're missing the case where there is no 'hover' - we're always showing at least a blank container w/ this change.

src/UI/QuickmenuView.re
```diff
@@ -60,31 +60,36 @@ let onInputClicked = cursorPosition =>
...
+
+    <View style={Styles.progressBarIndicator(~width, ~offset, ~theme)} />;
+  };
+
+  <View style=Styles.progressBarTrack> <indicator /> </View>;
+};
+
+let%component busyBar = (~theme, ()) => {
+  let%hook (time, _, _) =
+    Animation.(
```
bryphe: Cool!

src/UI/HoverView.re
glennsl: Oh man, how did I miss that! Fixed in https://github.com/onivim/oni2/pull/1201/commits/d8b825b7f1290e0e7bb923694bb7c9d606dc937e

src/UI/HoverView.re
glennsl: There's still the problem that it gets passed the x and y pixel coordinates of the cursor from EditorSurface. So right now the hover for every cursor would be rendered on top of the primary cursor. We'll need to pass in whatever is needed to calculate the pixel coordinates in `hoverItem`, and from a quick glance it seemed to need quite a few things.

PR #1202, opened 2020-01-14
feat(other): add oni2 command to PATH - Linux

CrossR:
Should work for the two Linux builds now, just want to give it a check with some pre-made builds and also check there is no regression on mac.

EDIT: Tested on mac and there is no regression. Just need to check the image issue.

### DIFF
--- a/node/add-to-path.js
+++ b/node/add-to-path.js
@@ -20,58 +20,63 @@ const isAddedToPath = () => {
     }
     return true
 }
-const removeFromPath = () => {
+
+const getOptions = () => {
+    let appDirectory = "";
+
+    if (isMac()) {
+        // Valid for an install from a dmg.
+        appDirectory = path.join(path.dirname(process.mainModule.filename), "..");
+    } else {
+        // Valid path for an AppImage or Linux tar.gz
+        appDirectory = path.join(path.dirname(process.mainModule.filename), "..", "..");
+    }
+
+    // Valid for an install from a dmg.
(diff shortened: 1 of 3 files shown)


node/add-to-path.js
```diff
@@ -14,63 +14,92 @@ const getLinkPath = () => (isMac() || isLinux() ? "/usr/local/bin/oni2" : "")
...
+        // Valid path when in a development build.
+        imgPath = path.join(appDirectory, "assets", "images", "Onivim2.icns")
+    }
+
+    if (!fs.existsSync(imgPath)) {
+        // Valid path for an AppImage or Linux tar.gz
+        imgPath = path.join(appDirectory, "..", "Onivim2.icns")
+    }
+
+    const options = { name: "Oni2", icns: imgPath }
```
CrossR: Need to check this...its not showing up in my VM, but that could just be the weird flavour of Ubuntu I'm running...should try in regular ol Ubuntu or something like CentOS or Arch.

CrossR: Looks like this doesn't matter, since I can't get an icon to show up on Ubuntu at all (I can't see anywhere one would go). Its set correctly for the valid locations at least.

z0al: I don't know what you're doing and I could be completely off, but `.icns` isn't a [supported format in Desktop Entries](https://wiki.archlinux.org/index.php/Desktop_entries#Common_image_formats), so maybe that also applies here.

lastthyme (approved):

lastthyme (approved):

PR #1205, opened 2020-01-15
Extract SearchPane to Feature_Search project

glennsl:
Also required moving LocationList and OniInput to a new Components project.

There's still plenty of room for improvement with the wiring, but as a first pass I think this turned out well. I'll look over it again after getting some distance, but this is a good time to comment on it.


src/Store/StoreThread.re
```diff
@@ -227,8 +242,15 @@ let start =
     // TODO: Wire this up properly
     let quickmenuSubs = quickmenuSubscriptionsUpdater(newState);
     QuickmenuSubscriptionRunner.run(~dispatch, quickmenuSubs);
-    let searchSubs = searchSubscriptionsUpdater(newState);
-    SearchSubscriptionRunner.run(~dispatch, searchSubs);
+
+    let searchDispatch = msg => dispatch(Model.Actions.Search(msg));
```
glennsl: With feature projects and a hierarchy we're going to be writing a lot of these wrapping function, so it's a good idea to try to make it as convenient as possible. There are a couple of ppxes that can compensate for the lack of first-class variant constructors to make this nicer:

```reason
// without ppxes and first-class vairant constructors
msg => dispatch(Msg.Search(msg))

// with ppx_pipebang (which inlines `|>`)
msg => msg |> Msg.Search |> dispatch

// with ppx_variants_conv and `>>` for function composition
dispatch >> Msg.search
```

Both of these are from Jane Street and require `base` during build, but fortunately not at runtime. Also, if we choose one, we might as well bring in the other one, and can have our pick at some other useful ones too.

src/Model/Actions.re
```diff
@@ -122,10 +122,7 @@ type t =
   | CopyActiveFilepathToClipboard
   | SearchStart
   | SearchHotkey
-  | SearchInput(string)
-  | SearchInputClicked(int)
-  | SearchUpdate([@opaque] list(Ripgrep.Match.t))
-  | SearchComplete
+  | Search(Feature_Search.msg)
```
bryphe: Excellent - your guidance to extract the 'Actions' out this way will help us support additional 'feature extractions' 👍

src/Store/Features.re
```diff
@@ -0,0 +1,55 @@
...
+
+  (state: State.t, dispatch) => {
+    quickmenuSubscriptions(dispatch, state)
+    |> QuickmenuSubscriptionRunner.run(~dispatch);
+
+    let searchDispatch = msg => dispatch(Search(msg));
+    searchSubscriptions(searchDispatch, state.searchPane)
+    |> SearchSubscriptionRunner.run(~dispatch=searchDispatch);
+  };
+};
```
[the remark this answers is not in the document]
bryphe: Cool!

src/UI/PaneView.re
```diff
@@ -38,13 +38,23 @@ let make = (~theme, ~uiFont, ~editorFont, ~state: State.t, ()) => {
        let childPane =
          switch (paneType) {
          | Pane.Search =>
-           <SearchPane
+           let onSelectResult = (file, location) =>
+             GlobalContext.current().dispatch(
+               Actions.OpenFileByPath(file, None, Some(location)),
+             );
+           let dispatch = msg =>
+             GlobalContext.current().dispatch(Actions.Search(msg));
```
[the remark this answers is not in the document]
bryphe: Sounds good to me. `GlobalContext` would be nice to have disappear

src/Store/StoreThread.re
bryphe: Cool! I've used `ppx_pipebang` before, it was pretty convenient, and I can confirm it works on Windows.

Issue #1148, opened 2020-01-02
Release: Bake build constants


bryphe:
For our welcome screen, for bug logging, and for information about the build, it'd helpful to have several constants available:
- `buildType` (`Development` | `Official`): What kind of build this is
- `commitId`: Commit built
- `version`: The version identifier, so it doesn't need to be copied in multiple places (like #1145 )

We could use something like [`ppx_getenv`](https://github.com/ocaml-ppx/ppx_getenv) if it works cross-platform.

PR #1210, opened 2020-01-16
Feature - Logging: Bundle in commitId / version

bryphe:
From #1185 - there were a few extra things I wanted to log. This brings in a generated commitId / version that we add to the log file (and means we don't have to hardcode the version string anymore!)

I also added an API to get the OSX version here: https://github.com/revery-ui/reason-sdl2/pull/35 which we can plumb through. The additional context will be useful for future logs.


bryphe:
Fixes #1148

PR #1211, opened 2020-01-16
Add ContextMenu to Notification Status Bar item

glennsl:
Although perhaps more importantly, this adds a ContextMenu component for general use.

### DIFF
--- a/src/Components/ContextMenu.re
+++ b/src/Components/ContextMenu.re
@@ -232,7 +232,14 @@ module Make = (()) => {
   module Anchor = {
     let component = React.Expert.component("Anchor");
     let make =
-        (~model as maybeModel, ~orientation=(`Bottom, `Left), ~onUpdate, ()) =>
+        (
+          ~model as maybeModel,
+          ~orientation=(`Bottom, `Left),
+          ~offsetX=0,
+          ~offsetY=0,
+          ~onUpdate,
+          (),
+        ) =>
       component(hooks => {
         let ((maybeRef, setRef), hooks) = Hooks.ref(None, hooks);
 
@@ -242,7 +249,11 @@ module Make = (()) => {
             let (x, y, _, _) =
               Math.BoundingBox2d.getBounds(node#getBoundingBox());
             let placement =
-              Some({x: int_of_float(x), y: int_of_float(y), orientation});
+              Some({
+                x: int_of_float(x) + offsetX,
+                y: int_of_float(y) + offsetY,
+                orientation,
+              });
 
             if (model.placement != placement) {
               onUpdate({...model, placement});
--- a/src/Components/ContextMenu.re
+++ b/src/Components/ContextMenu.re
@@ -1,0 +1,258 @@
+open Oni_Core;
+
+open Revery;
+open Revery.UI;
+open Revery.UI.Components;
+
+module Option = Utility.Option;
+
+module Constants = {
+  let menuWidth = 200;
+  // let maxMenuHeight = 600;
+};
+
+// TYPES
+
+type identity = int;
+
+type item('data) = {
+  label: string,
+  // icon: option(IconTheme.IconDefinition.t),
+  data: 'data,
+};
+
+type t('data) = {
+  identity,
+  x: int,
+  y: int,
+  originX: [ | `Left | `Middle | `Right],
+  originY: [ | `Top | `Middle | `Bottom],
+  items: list(item('data)),
+};
+
+let create = (~originX=`Left, ~originY=`Bottom, identity, items) => {
+  identity,
+  x: 0,
+  y: 0,
+  originX,
+  originY,
+  items,
+};
+
+// MENUITEM
+
+module MenuItem = {
+  module Constants = {
+    let fontSize = 12;
+  };
+
+  module Styles = {
+    open Style;
+
+    let bg = (~theme: Theme.t, ~isFocused) =>
+      isFocused ? theme.menuSelectionBackground : theme.menuBackground;
+
+    let container = (~theme, ~isFocused) => [
+      padding(10),
+      flexDirection(`Row),
+      backgroundColor(bg(~theme, ~isFocused)),
+    ];
+
+    // let icon = fgColor => [
+    //   fontFamily("seti.ttf"),
+    //   fontSize(Constants.fontSize),
+    //   marginRight(10),
+    //   color(fgColor),
+    // ];
+
+    let label = (~font: UiFont.t, ~theme: Theme.t, ~isFocused) => [
+      fontFamily(font.fontFile),
+      textOverflow(`Ellipsis),
+      fontSize(Constants.fontSize),
+      color(theme.menuForeground),
+      backgroundColor(bg(~theme, ~isFocused)),
+    ];
+
+    let clickable = [cursor(Revery.MouseCursors.pointer)];
+  };
+
+  let make:
+    'data.
+    (
+      ~item: item('data),
+      ~theme: Theme.t,
+      ~font: UiFont.t,
+      ~onClick: unit => unit,
+      unit
+    ) =>
+    React.element(React.node)
+   =
+    (~item, ~theme, ~font, ~onClick, ()) => {
+      let isFocused = false;
+
+      // let iconView =
+      //   switch (item.icon) {
+      //   | Some(icon) =>
+      //     IconTheme.IconDefinition.(
+      //       <Text
+      //         style={Styles.icon(icon.fontColor)}
+      //         text={FontIcon.codeToIcon(icon.fontCharacter)}
+      //       />
+      //     )
+
+      //   | None => <Text style={Styles.icon(Colors.transparentWhite)} text="" />
+      //   };
+
+      let labelView = {
+        let style = Styles.label(~font, ~theme, ~isFocused);
+        <Text style text={item.label} />;
+      };
+
+      <Clickable style=Styles.clickable onClick>
+        // onMouseOver={_ => setIsFocused(_ => true)}
+        // onMouseOut={_ => setIsFocused(_ => false)}
+
+          <View style={Styles.container(~theme, ~isFocused)}>
+            // iconView
+             labelView </View>
+        </Clickable>;
+    };
+};
+
+// MENU
+
+module Menu = {
+  module Styles = {
+    open Style;
+
+    let container = (~x, ~y, ~theme: Theme.t) => [
+      position(`Absolute),
+      top(y),
+      left(x),
+      backgroundColor(theme.menuBackground),
+      color(theme.menuForeground),
+      width(Constants.menuWidth),
+    ];
+  };
+
+  let component = React.Expert.component("Menu");
+  let make = (~model, ~theme, ~font, ~onItemSelect, ()) =>
+    component(hooks => {
+      let ((maybeRef, setRef), hooks) = Hooks.state(None, hooks);
+      let {x, y, originX, originY, items, _} = model;
+
+      let height =
+        switch (maybeRef) {
+        | Some((node: node)) => node#measurements().height
+        | None => List.length(model.items) * 20
+        };
+      let width = Constants.menuWidth;
+
+      let x =
+        switch (originX) {
+        | `Left => x
+        | `Middle => x - width / 2
+        | `Right => x - width
+        };
+
+      let y =
+        switch (originY) {
+        | `Top => y - height
+        | `Middle => y - height / 2
+        | `Bottom => y
+        };
+
+      (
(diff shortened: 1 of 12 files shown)


src/Components/ContextMenu.re
```diff
@@ -0,0 +1,257 @@
...
+  let create = () => {
+    incr(lastId);
+    lastId^;
+  };
+};
+
+[@deriving show({with_path: false})]
+type item('data) = {
+  label: string,
+  // icon: option(IconTheme.IconDefinition.t),
```
glennsl: `IconTheme` currently lives in `Model`. I'm not sure if `Core` is the right place for it, so for now we'll have to do without.

src/Components/ContextMenu.re
```diff
@@ -0,0 +1,257 @@
...
+        // TODO: BoxShadow apparently blocks mouse events. Figure out why before adding back
+        // <BoxShadow
+        //   boxShadow={Style.BoxShadow.make(
+        //     ~xOffset=-11.,
+        //     ~yOffset=-11.,
+        //     ~blurRadius=25.,
+        //     ~spreadRadius=0.,
+        //     ~color=Color.rgba(0., 0., 0., 0.2),
+        //     (),
+        //   )}>
```
glennsl: This is unfortunate. No idea what's going on here, but if this is uncommented the menu items stop receiving mouse events.

src/Components/ContextMenu.re
```diff
@@ -0,0 +1,257 @@
...
+  module Anchor = {
+    let component = React.Expert.component("Anchor");
+    let make =
+        (
+          ~model as maybeModel,
+          ~originX=`Left,
+          ~originY=`Bottom,
+          ~onUpdate,
+          (),
+        ) =>
```
glennsl: We could use mutation and internal bookkeeping instead of passing in `model` and `onUpdate` here, but then there wouldn't be a state update to trigger a re-render.

src/Components/ContextMenu.re
```diff
@@ -0,0 +1,257 @@
...
+
+module Make = (()) => {
+  let id = Id.create();
+
+  let init = items => {
+    id,
+    x: 0,
+    y: 0,
+    originX: `Left,
+    originY: `Bottom,
```
glennsl: It might be better to extract these into a separate `placement` type and wrap that in an `option` here instead of initializing with arbitrary defaults, since this might cause the menu to be rendered in the wrong place for a frame or so.

src/UI/Root.re
```diff
@@ -105,6 +119,22 @@ let make = (~state: State.t, ()) => {
...
+       let onItemSelect = item =>
+         GlobalContext.current().dispatch(ContextMenuItemSelected(item));
+
+       <ContextMenu.Overlay
+         theme
+         font=uiFont
+         model
+         onOverlayClick
+         onItemSelect
+       />;
```
glennsl: This is what renders the actual menu, above a clickable overlay that dismisses the menu.

src/UI/StatusBar.re
```diff
@@ -240,15 +249,32 @@ let textItem = (~background, ~font, ~theme: Theme.t, ~text, ()) =>
...
+      Actions.StatusBar(NotificationsContextMenu),
+    );
 
-  <item onClick>
+  <item onClick onRightClick>
+    <Notifications.ContextMenu.Anchor
+      originY=`Top
+      model=contextMenu
+      onUpdate=onContextMenuUpdate
+    />
```
glennsl: This is just used as a reference point for where to render the menu. Nothing visible is rendered. It also has the added benefit of "live-updating" and moving the menu along with the anchor on resize and such (unlike the sneaks, for example).

src/Model/Notifications.re
```diff
@@ -3,12 +3,16 @@ open Notification;
 
 type t = list(Notification.t);
 
+module ContextMenu =
+  Oni_Components.ContextMenu.Make({});
```
glennsl: The functor implementation is just to generate an identity to connect the anchor with the menu data. It could be generated manually and passed around instead, but this seemed more convenient.

src/Components/ContextMenu.re
```diff
@@ -0,0 +1,257 @@
...
+      backgroundColor(theme.menuBackground),
+      color(theme.menuForeground),
+      width(Constants.menuWidth),
+    ];
+  };
+
+  let component = React.Expert.component("Menu");
+  let make = (~model, ~theme, ~font, ~onItemSelect, ()) =>
+    component(hooks => {
+      let ((maybeRef, setRef), hooks) = Hooks.state(None, hooks);
```
glennsl: There's some unfortunate weak polymorphism issues when using the brisk-reconciler ppx, so we have to go old-school here. The ppx also has issues with type annotations, so `MenuItem` above can't use it either because it needs a type variable to be universally quantified.

src/Model/Actions.re
```diff
@@ -55,6 +56,9 @@ type t =
   | TextInput([@opaque] Revery.Events.textInputEvent)
   | HoverShow
   | ChangeMode([@opaque] Vim.Mode.t)
+  | ContextMenuUpdate([@opaque] ContextMenu.t(t))
```
bryphe: `ContextMenuUpdate` doesn't sound like an event to me (it sounds like a description of what _will_ happen) - it sounds like the 'function call` paradigm. Would `ContextMenuUpdated` make more sense?

src/Model/Actions.re
```diff
@@ -80,6 +84,7 @@ type t =
   | EditorScrollToColumn(EditorId.t, int)
   | ShowNotification(Notification.t)
   | HideNotification(Notification.t)
+  | ClearNotifications
```
bryphe: This doesn't seem like an 'Event' either - this seems to describe the _intent_, which is really just a different naming scheme for a function. I believe what we'd want is to describe the event that caused it, vs the intent.

It could be I'm missing nuance, or that this is just set up to fit into the existing notification actions `ShowNotification` and `HideNotification` (which also aren't very 'Event-like'). It might be that we could tackle those events holistically if we move notifications to a feature. 

Just want to double-check my understanding of the 'Event' model we are using-

src/Components/ContextMenu.re
```diff
@@ -0,0 +1,270 @@
...
+};
+
+[@deriving show({with_path: false})]
+type item('data) = {
+  label: string,
+  // icon: option(IconTheme.IconDefinition.t),
+  data: [@opaque] 'data,
+};
+
+type placement = {
```
bryphe: This `placement` is awesome! This will be useful for all kinds of hover / context menu like UI 💯

src/UI/Root.re
bryphe: Cool! Looks great

src/UI/StatusBar.re
```diff
@@ -222,12 +223,20 @@ let section = (~children=React.empty, ~align, ()) =>
...
+      ~onRightClick=?,
+      (),
+    ) => {
   let style = Styles.item(backgroundColor);
 
-  switch (onClick) {
-  | Some(onClick) => <Clickable onClick style> children </Clickable>
-  | None => <View style> children </View>
+  // Avoid cursor turning into pointer if there's no mouse interaction available
+  if (onClick == None && onRightClick == None) {
```
bryphe: Thanks for the comment here

src/UI/StatusBar.re
bryphe: Cool!

src/Components/ContextMenu.re
bryphe: Bummer, I'll look at bringing the 'style' property back so we can use it w/o interfering with click events

bryphe:
Looks awesome @glennsl !

image

Sorry about the shadow bug (that would help it 'pop' for sure) - i'll look at adding the 'box shadow' style back in Revery as a temporary workaround. Change looks great!

src/Model/Actions.re
glennsl: Yep, definitely should be. Good catch!

src/Model/Actions.re
glennsl: Yeah, this is just to be consistent with what was already there. Both `Notifications` and `StatusBar` should be refactored into feature projects I think, but I'm not entirely sure where the boundary between them will be yet. Which project the notification count and context menu should live in, which will impact all these actions. I think they might just become function though, to be triggered by events in other projects.

src/Components/ContextMenu.re
glennsl: The `originX` and `originY` terms don't seem very intuitive though. I need to find some better terminology for describing positioning relative to a point.

src/Components/ContextMenu.re
bryphe: Working on bringing it back here: https://github.com/revery-ui/revery/pull/719

This would let us add a `boxShadow` style to the root `<View />`, side-stepping this issue.

src/Components/ContextMenu.re
bryphe: Ah... interesting. I wonder if we'll hit more issues like this. Nice catch.

src/Model/Actions.re
bryphe: Cool! Starting to get the hang of it, hopefully 😅

src/Model/Actions.re
bryphe: Cool, that makes sense! As we start to untangle the obvious parts into features, maybe some of these will fall out.

src/Components/ContextMenu.re
glennsl: Probably. It's not the first time I've hit the type annotation issue. It's tracked here: https://github.com/briskml/brisk-reconciler/issues/53. There's also an issue with locally abstract types. Both of those are just omissions though, and should be relatively easy to fix. I'm not entirely sure about the weak polymorhpism issue as I haven't figured out the root cause and a repro yet.

src/Components/ContextMenu.re
```diff
@@ -142,14 +141,21 @@ module Menu = {
       backgroundColor(theme.menuBackground),
       color(theme.menuForeground),
       width(Constants.menuWidth),
+      boxShadow(
+        ~xOffset=-5.,
+        ~yOffset=-5.,
+        ~blurRadius=25.,
+        ~spreadRadius=-10.,
+        ~color=Color.rgba(0., 0., 0., 0.0001),
+      ),
```
glennsl: It works! But it's also kind of broken. The alpha value of the color doesn't seem to be respected, and there seems to be some built-in offset that has to be countered by specifying negative offsets.

I managed to tone it down a bit by adding a negative spread, but it still looks a bit off.

src/Components/ContextMenu.re
```diff
@@ -36,8 +36,7 @@ type item('data) = {
 type placement = {
   x: int,
   y: int,
-  originX: [ | `Left | `Middle | `Right],
-  originY: [ | `Top | `Middle | `Bottom],
+  orientation: ([ | `Top | `Middle | `Bottom], [ | `Left | `Middle | `Right]),
```
glennsl: This is much better terminology I think.

glennsl:
This should be good to go now. See the latest changes in https://github.com/onivim/oni2/pull/1211/files/3bc5690246786b4622eb9cfc28fe301790537212..9f8d77984c20c29990e34cb27dd05e9c57d31c71#diff-7fdbd38419fb3f5323ee3440ba789a9e

src/Components/ContextMenu.re
bryphe: > It works!

Hooray! 🎉 

> But it's also kind of broken. The alpha value of the color doesn't seem to be respected, and there seems to be some built-in offset that has to be countered by specifying negative offsets.

Ya, unfortunately there are bugs and quirks with the [current implementation](https://github.com/revery-ui/revery/blob/50e50c3a3dd56a2e835f427c530940b7e7b601de/src/Draw/GradientShader.re#L55). 

Skia will give us a more robust set of primitives for stuff like this!

src/Components/ContextMenu.re
bryphe: Looks pretty clear to me!

bryphe:
Just tried the latest - looks so much nicer with the drop shadow!

src/Components/ContextMenu.re
glennsl: Thanks for the pointer! Looks like fixing it so the alpha value is taken into account shouldn't be too hard, just pass in a vec4 instead of a vec3 and multiple the alpha value with the blur value. But the offset bug seems trickier. I can't se where that comes from. That's also less of an issue I think though, so it can definitely wait for Skia.

PR #1214, opened 2020-01-17
SCM: Add git extension

glennsl:
Also adds `enableProposedApi` to `ExtenionManifest` and `ExtHostInitData`, since without it the extension fails.

I imagine this can be trimmed down a whole lot, but I have no idea how to bundle the extension up into a shippable form. Please advise.


bryphe:
Awesome!

> Also adds enableProposedApi to ExtenionManifest and ExtHostInitData, since without it the extension fails.

Nice catch on this 👍 

> I imagine this can be trimmed down a whole lot, but I have no idea how to bundle the extension up into a shippable form. Please advise.

We have this problem with all our extensions currently - copy+pasting means we have a bunch of `map` and source files left over. I'll take a look and see if we can have a script to clean-up / trim this. But we can tackle that separately.

bryphe:
Oops, sorry, meant to let you merge! Thought I was on another PR 😬 But I guess it was going to go in anyway. 

I'll track the clean-up work here: https://github.com/onivim/oni2/issues/1215

PR #1217, opened 2020-01-17
Dependency: revery -> 4e5ea27

bryphe:
This commit of Revery has Skia, linked, but not yet used. But this will exercise packaging, shipping, and verifying the binaries w/ Skia


bryphe:
Picked up in #1225

PR #1218, opened 2020-01-18
Updates input model to handle keypress of Home/End buttons

fanantoxa:
Small improvements for input navigation:
 - Positions cursor to beginning on pressing `Home`
 - Positions cursor to the end on pressing `End`


bryphe:
This is a great improvement - thank you @fanantoxa !

PR #1220, opened 2020-01-18
Refactor Utility into separate project

glennsl:
To be able to split it into separate modules. This turned out to be a much bigger task than I thought, and should have been done much sooner, but the end result is pretty good I think. The boundary between Utility and Core isn't exactly clear though, and I had to create a `Core.Kernel` project too because `Utility` depended on several modules in `Core`.


glennsl:
Seems like Windows absolutely refuses to build. It fails consistently with the following error: 
```
C:/Users/VssAdministrator/.esy/3/i/reason_font_manager-2.0.1-6a83d740/lib\reason-font-manager\libFontManager_stubs.a(FontManagerWindows.o): In function `IsEqualGUID':
/usr/x86_64-w64-mingw32/sys-root/mingw/include/guiddef.h:161: multiple definition of `IsEqualGUID'
C:/npm/prefix/node_modules/esy/node_modules/esy-bash/.cygwin/usr/x86_64-w64-mingw32/sys-root/mingw/lib/libole32.a(dwzns00301.o):(.text+0x0): first defined here
```

bryphe:
Maybe related to this?
https://stackoverflow.com/questions/59275209/ole32-link-error-with-multiple-definitions-of-isequalguid

We like `-lole32` in Revery [here](https://github.com/revery-ui/revery/blob/8aa55f388a03a00a4e96d396086f1d88aaa924d1/src/Native/config/discover.ml#L28) and `-ldwrite` in `reason-font-manager` [here](https://github.com/revery-ui/reason-font-manager/blob/47198846be3b47f5b93a85e57bfd19b29b5c32d0/config/discover.ml#L31), but I'm not sure why these new projects would cause a break like this...

src/Core/Utility/Win32.re
```diff
@@ -0,0 +1 @@
+external freeConsole: unit => unit = "win32_free_console";
```
bryphe: One thing I noticed is that the `bindings.c` that implements this function should also be moved to `Core/Utility` - and the `(c_names bindings)` moved from `Core/dune` to `Core/Utility/dune`. But this unfortunately doesn't address the windows build error

bryphe:
I tried a bunch of different things:
- Coming up with a minimal repro: https://github.com/onivim/oni2/tree/repro/1220-minimal-build-failure
- Un-nesting the dependencies (I've had issues with nested-dependencies in the past, which is why our module hierarchy is flat): https://github.com/onivim/oni2/tree/experiment/1220/un-nest
- Tweaking the `-dwrite` flag from `reason-font-manager` and the `-ole32` flag from `revery` (ie, having revery specify both of them)

Not having any luck at coming up with a conclusive diagnosis or fix though. The change in the dependency hierarchy must be changing the link arguments in problematic way, but it's not clear to me why... I'm stuck at the moment, will have to come back to it later.

Do you have any ideas, @glennsl ?

src/Core/Utility/Win32.re
glennsl: Good catch! Fixed in https://github.com/onivim/oni2/pull/1220/commits/5001690541b9a04cb040a107fa0a9c640a698083

I find it a bit uncomfortable how little encapsulation there are with dependencies though, even more so with c dependencies.

glennsl:
> Maybe related to this?
https://stackoverflow.com/questions/59275209/ole32-link-error-with-multiple-definitions-of-isequalguid

I noticed this uses `esy` as well. Could it be a case of esy not sufficiently sandboxing different build processes?

> Do you have any ideas, @glennsl ?

Can you whither down the minimal repro even more? Sounds like it's an issue with Revery, so could you remove all other dependencies from `Kernel` and `Utility`? What happens if you remove the dependencies on `kernel` and `utility` in `core` and try to add them directly to `ui` and `store` instead? Does that still trigger the issue, or is it just when they're dependencies of `core`?

bryphe:
Hmm, I'm trying to break it up into smaller, more incremental changes - starting in #1244

glennsl:
Since the issue seems to be with Revery, have you tried moving just that and the modules that depend on it, back to Core? I don't think it's too much that depends on it.

glennsl:
Revery is used by `Setup` in `Kernel`, which I think was only use by some extensions functions in `Utility` which I've since moved out. And in `Utility` I think Revery is just used for `ColorEx.mix`, which has since been moved to Revery. So I don't think the dependency is actually needed at all!

I'll give this a shot. It'd be cool if it was actually that easy.

glennsl:
Alright, that ended up being a little bit more involved, as Setup was used by Camomile, which was used by Zed, which was used by VimEx. So now `VimEx` lives in `Core`, which is just a little bit more awkward than having it in `Utility` anyway.

glennsl:
Fingers crossed!

src/Core/Kernel/dune
```diff
@@ -6,7 +6,7 @@
         ppx_deriving.runtime
         ppx_deriving_yojson.runtime
         zed_oni
-        Revery
+        threads
```
glennsl: This was brought in transitively with Revery too it seems

glennsl:
That seems to have done the trick!

bryphe:
> Revery is used by Setup in Kernel, which I think was only use by some extensions functions in Utility which I've since moved out. And in Utility I think Revery is just used for ColorEx.mix, which has since been moved to Revery. So I don't think the dependency is actually needed at all!

@glennsl - awesome, great idea Ya, I think the problem is the Revery-as-a-transitive dependency, getting us some duplication in the `-lole32` and `-ldwrite` flags, which apparently causes problems. But glad there is a simple fix for it. Nice find!

Having the 'sub-modules' like this will be really cool. Being able to move out of the 'flat' module namespaces will help us have significantly better organization 👍

PR #1224, opened 2020-01-18
Fix #515: Performing actions on long lines is laggy

### LINKED ISSUE #515: Performing actions on long lines is laggy
~5000 characters line causes significant lag
~30000 line almost freezes the editor
The further you move the laggier it becomes. Toggling syntax highlighting does not affect anything.

OS: Windows 10

bryphe:
During the course of rendering the `EditorSurface`, there were several places where we were inadvertently traversing the entire 'line' string during the course of rendering. 

These included:
- Calls to `Zed_utf8.length`, which require traversing the byte string to get the UTF8 length
- Calls to `Zed_utf8.get` / `Zed_utf8.sub`, which require traversing up to the index point

We were calling these across `BufferViewTokenizer`, `EditorSurface`, and in rendering 'underlines' for diagnostics / definitions, which means we were making many unnecessary traversals across the same strings during the course of rendering.

(Note that calls to `String.length` are fast, because the string byte length is encoded in the header for string values in the OCaml runtime - but to get the UTF8 characters out of the byte array requires traversing the string).

We do need these UTF8 characters, but often, we don't need the entire line. So there are a few things we can do to fix this:
- Cache byte offsets for utf8 characters, so operations like `get` and `sub` can be very fast
- Only traverse / compute UTF8 byte offsets up to a bounded limit. Often, for rendering, we only need to care up to the max column we have visible.

This change refactors our 'BufferLine' model from just a raw string of bytes to something more abstract, in which we can shim in a caching layer to implement those functionalities.

__TODO:__
- [x] Add benchmarks for `BufferLine`
  - [x] `slowLengthUtf8`
  - [x] `boundedLengthUtf8`
  - [x] `getPositionAndWidth`
  - [x] `unsafeGetUChar`
- [x] Implement incremental caching strategy & test against benchmarks
- [x] Create `BufferLine` on buffer update, instead of on-demand with `getLine`
- [x] Synchronize indentation settings

__DEFERRED:__
- Fix `String.trim` in `IndentLineRenderer` (Will handle this in a separate PR)

Benchmark results:

__BEFORE:__
```
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|  BENCHMARK                                                          |  ITERATIONS |  TIME               |  MINOR GC  |  MAJOR GC  |  MINOR ALLOC  |  PROMOTED  |  MAJOR ALLOC  |
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  BufferViewTokenizer: Tokenize line, no tokens                      |  1000       |  3.65921807289      |  36        |  0         |  9188247      |  1860      |  1860         |
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  BufferLine: slowLengthUtf8                                         |  1000       |  0.029247045517     |  0         |  0         |  27           |  0         |  0            |
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  BufferLine: boundedLengthUtf8                                      |  1000       |  0.0294649600983    |  0         |  0         |  27           |  0         |  0            |
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  BufferLine: unsafeGetUChar                                         |  1000       |  0.0286679267883    |  0         |  0         |  27           |  0         |  0            |
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  BufferLine: getPositionAndWidth                                    |  1000       |  0.0321841239929    |  0         |  0         |  27           |  0         |  0            |
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  Buffer: Add lines to empty buffer                                  |  1000       |  5.19752502441e-05  |  1         |  0         |  19029        |  25        |  25           |
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  Buffer: Insert line in middle of small buffer                      |  1000       |  0.000346899032593  |  1         |  0         |  232041       |  25        |  25           |
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  Buffer: Insert line in middle of large buffer                      |  1000       |  1.5930390358       |  400       |  8         |  30103        |  3544      |  110005544    |
|---------------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  Buffer: Clear large buffer                                         |  1000       |  3.11925506592      |  666       |  14        |  28163        |  4582      |  200006582    |
```

__AFTER:__
```
|  BufferViewTokenizer: Tokenize line, no tokens                      |  1000       |  2.00271606445e-05  |  1         |  0         |  4031         |  25         |  25           |
|  BufferLine: slowLengthUtf8                                         |  1000       |  0.029205083847     |  0         |  0         |  27           |  0          |  0            |
|  BufferLine: boundedLengthUtf8                                      |  1000       |  0.000568866729736  |  1         |  0         |  50037        |  25         |  10026        |
|  BufferLine: unsafeGetUChar                                         |  1000       |  9.67979431152e-05  |  1         |  0         |  30030        |  25         |  25           |
|  BufferLine: getPositionAndWidth                                    |  1000       |  4.38690185547e-05  |  1         |  0         |  10030        |  25         |  25           |
|  Buffer: Add lines to empty buffer                                  |  1000       |  12.7578420639      |  4033      |  54        |  700052729    |  700000181  |  800001181    |
|  Buffer: Insert line in middle of small buffer                      |  1000       |  0.000385999679565  |  1         |  0         |  246031       |  25         |  25           |
|  Buffer: Insert line in middle of large buffer                      |  1000       |  1.32106399536      |  400       |  7         |  44097        |  11337      |  110013337    |
|  Buffer: Clear large buffer                                         |  1000       |  2.37255597115      |  666       |  13        |  33153        |  4589       |  200006589    |
```

One of the key benchmarks - the `BufferViewTokenizer` was a __multiple order of magnitude improvement (from 3s to less than a millisecond).__ This was a relatively modest bench too - working with a string that was only 1000 characters. This was part of the critical path for rendering large lines, so this will be a huge improvement in working with long lines in the editor.

On the flip side, one operation got more expensive - add lines to an empty buffer. This is due to the extra allocation / bookkeeping we do to the 'raw' strings. That benchmark was more aggressive - at 100k lines. I think the tradeoff is the right one to speed up editing, but we could look at improving that particular benchmark by deferring allocation until a line is actually accessed.


src/Core/BufferLine.re
```diff
@@ -0,0 +1,152 @@
...
+};
+
+let unsafeGetUChar = (~index, bufferLine) => {
+  Internal.resolveTo(~index, bufferLine);
+  let characters = bufferLine.characters;
+  switch (characters[index]) {
+  | Some({uchar, _}) => uchar
+  | None => raise(OutOfBounds)
+  };
+};
```
glennsl: This is an inaccurate use of `unsafe`, at least in regards to its conventional meaning in OCaml where memory/type unsafety and undefined behavior are considered unsafe, but raising an exception is not.

Not saying we should adopt OCaml's attitude to exceptions, but we might want to avoid using terminology in ways that conflict with its use in the ecosystem we're in, since that's likely to eventually cause confusion and mistakes.

`getUCharExn` might be an alternative name for this.

src/Core/BufferLine.re
```diff
@@ -0,0 +1,152 @@
...
+  // of the UTF8 string, if it was all 1-byte unicode characters (ie, an ASCII string).
+  let characters = emptyArray;
+  {indentation, raw, characters, nextByte: 0, nextIndex: 0, nextPosition: 0};
+};
+
+let empty = make(~indentation=IndentationSettings.default, "");
+
+let lengthInBytes = ({raw, _}) => String.length(raw);
+
+let slowLengthUtf8 = ({raw, _}) => ZedBundled.length(raw);
```
glennsl: It would be better for auto-completion to have "modifiers" at the end rather than at the beginning I think. For example, if you're looking for `length` functions, you're not going to find this if you start typing `BufferLine.len`.

src/Core/Buffer.re
```diff
@@ -90,14 +100,14 @@ let getUri = (buffer: t) => {
...
-};
+/*let getLineLength = (buffer: t, line: int) => {
+    let line = getLine(buffer, line);
+    String.length(line);
+  };*/
 
 let getNumberOfLines = (buffer: t) => Array.length(buffer.lines);
 
-let slice = (~lines: array(string), ~start, ~length, ()) => {
+let slice = (~lines: array('a), ~start, ~length, ()) => {
```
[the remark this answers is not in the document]
bryphe: Cool!

src/Core/BufferLine.re
```diff
@@ -0,0 +1,152 @@
...
+  raw: string,
+  // [characters] is a cache of discovered characters we've found in the string so far
+  mutable characters: array(option(characterCacheInfo)),
+  // nextByte is the nextByte to work from, or -1 if complete
+  mutable nextByte: int,
+  // nextIndex is the nextIndex to work from
+  mutable nextIndex: int,
+  // nextPosition is the graphical position (based on character width)
+  mutable nextPosition: int,
+};
```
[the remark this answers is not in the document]
bryphe: > This is going to break diffing, which depends on the structural equality of the lines of the new and old documents. 

Ah, good point - we still keep the `raw` string around (and `getLines` actually returns these raw strings - the syntax highlighting still uses it). There still is an extra cost (the `Array.map` to go from `array(BufferLine.t)` -> `array(string)` We actually still need the `array(string)` for SyntaxHighlighting

> Diffing is also currently done every render, although we might eventually want to do it async/time-sliced and only when the document changes.

Definitely! This will be important to maintain perf. We probably only need to recompute the diffs on `BufferUpdate` or when the original resource location changes. And even then - it doesn't  need to be an immediate computation (in other words, it shouldn't block the hot path of key press -> render) - it could be deferred until idle, handed off to be handled asynchronously, or process in time-sliced chunks. That would help keep the core render path as fast as we can.

src/Core/BufferLine.re
bryphe: Ah, good point. I'll switch to `Exn`. I also like your proposal of adding these as suffixes (`Exn`, `Slow`).

src/Core/BufferLine.re
bryphe: Makes sense, thanks!

PR #1227, opened 2020-01-18
[WIP] [Experiment] feat(language-support): keyword completion

bryphe:
This is a prototype that hooks into syntax highlighting to provide keyword completion (ie, completion when a language server isn't available, based on current syntax highlighting).

__TODO:__
- [x] Implement `sortText`, such that 'real' LSP completions are prioritized higher than syntax completions
- [ ] Merge syntax + LSP completions when there are duplicates
- [x] Implement a 'streaming' approach to completion. Right now, we wait for _all_ completion providers to report results before showing _any_ completions. We should instead show the current set of completions, along with a progress indicator that some are still loading.
- [ ] Implement tree-sitter keyword extraction
- [ ] Clean up / refactor `KeywordMap` and `KeywordCount`


src/Store/SyntaxHighlightingStoreConnector.re
```diff
@@ -18,6 +18,32 @@ module Protocol = Oni_Syntax.Protocol;
...
+  let create =
+      (
+        client: Oni_Syntax_Client.t,
+        languageInfo: Ext.LanguageInfo.t,
+        (buffer, _completionMeet, _location),
+      ) => {
+    buffer
+    |> Core.Buffer.getFileType
+    |> Option.bind(ft =>
+         Ext.LanguageInfo.getScopeFromLanguage(languageInfo, ft)
```
glennsl: Can be partially applied

src/Store/SyntaxHighlightingStoreConnector.re
```diff
@@ -18,6 +18,32 @@ module Protocol = Oni_Syntax.Protocol;
...
+       )
+    |> Option.map(scope => {
+         let promise =
+           Oni_Syntax_Client.requestKeywords(client, scope)
+           |> Lwt.map(v => List.map(toCompletionItem, v));
+         Some(promise);
+       })
+    |> Option.value(~default=None);
```
glennsl: This looks weird. Why is the option doubled up here?

src/Syntax_Server/State.re
```diff
@@ -97,7 +101,50 @@ let doPendingWork = state => {
...
+            )
+          },
+          discoveredKeywords,
+          scope,
+        )
+        |> Option.value(~default=acc);
+      },
+      state.keywordMap,
+      state.visibleBuffers,
+    );
```
glennsl: It's usually better to just name the function than have a large anonymous function effectively described by a comment.

src/Syntax_Server/State.re
```diff
@@ -148,6 +195,15 @@ let clearTokenUpdates = state => {
 
 let bufferUpdate =
     (~scope, ~bufferUpdate: BufferUpdate.t, ~lines: array(string), state) => {
+  let bufferIdToScope =
+    IntMap.update(
+      bufferUpdate.id,
+      fun
+      | None => Some(scope)
+      | Some(_) => Some(scope),
```
glennsl: Why not `_ => Some(scope)`?

bryphe:
Implemented in #2581

PR #1228, opened 2020-01-18
Fix #1215: Clean up extension collateral

### LINKED ISSUE #1215: Packaging: Clean up extra '.js.map' and other build collateral from extensions
We're shipping a bunch of extra '.js.map' files and other collateral along with our extensions. This is tracking cleaning it up, and adding a 'hygiene check' to verify we aren't keeping those around.

bryphe:
There are several files left over in our extensions folder that we don't need - `.js.map` files, unbuilt `.ts` files, etc.

This adds a script that removes those, as well as a hygiene check to help verify there are no left over files when bringing in new extensions.

Fixes #1215 and removes ~4MB of unnecessary files


CrossR:
Unrelated to the actual issue, but the extensions in general. Is it worth marking the whole extensions folder as generated in git attributes? To my knowledge its all vendored code, so we could mark it as such and have GitHub pick this up as a reason project even more strongly.

Marking `extensions/*` as `linguist-generated` (which supresses diffs too) or just `linguist-vendored`.

bryphe:
Good point @CrossR - all the code is vendored from elsewhere 👍 

It looks like we started it here, just for the `css` extension:
```
# Hide vendored extension code
extensions/css-language-features/* linguist-generated
```
(https://github.com/onivim/oni2/blob/master/.gitattributes)

I just tweaked it to include the entire extensions folder. Thanks for the suggestion!

PR #1229, opened 2020-01-19
Dependency: timber -> ae5de6c

glennsl:
In addition to using native printf to avoid crashing on Windows, this:
- appends to the log file, so multiple processes can safely log to the same file. An optional `truncate` argument can be passed to `setLogFile`, which the main process can use to start from scratch each time.
- timestamp added to the log file header
- uses black foreground for all log levels, since that seems to be most legible with normal terminal colors


bryphe:
I'll bring this in immediately so we can upgrade Revery to pick up boxshadow fixes

PR #1230, opened 2020-01-20
Warn when closing window with unsaved changes

### LINKED ISSUE #610: Windows: No warning about unsaved changes
On Windows 10, if I have a file open in which I have made changes I am not warned about this before closing the program. This means all changes are lost, without having that behaviour be confirmed by the user.

glennsl:
2020-01-20-124655_413x233_scrot

Fixes #610

The message is a bit unbalanced, both visually and in terms of "scanability". It would be better to lead with a very short question summarizing the decision to be made, and then expand on it. We can tweak this later though, but it would be nice to have rich text support to be able to more easily to so.

A neat keyboard interface is planned for a subsequent PR.

### DIFF
--- a/src/Components/MessageBox.re
+++ b/src/Components/MessageBox.re
@@ -1,0 +1,77 @@
+open Oni_Core;
+
+open Revery.UI;
+open Revery.UI.Components;
+
+type action('msg) = {
+  label: string,
+  msg: 'msg,
+};
+
+module Styles = {
+  open Style;
+
+  let container = (~theme: Theme.t) => [backgroundColor(theme.background)];
+
+  let message = [padding(10)];
+
+  let messageText = (~theme: Theme.t, ~font: UiFont.t) => [
+    fontFamily(font.fontFile),
+    color(theme.foreground),
+    backgroundColor(theme.editorBackground),
+    fontSize(14),
+  ];
+
+  let actions = [flexDirection(`Row)];
+
+  let buttonOuter = (~isHovered, ~theme: Theme.t) => [
+    isHovered
+      ? backgroundColor(theme.menuSelectionBackground)
+      : backgroundColor(theme.editorBackground),
+    flexGrow(1),
+  ];
+
+  let buttonInner = [padding(10)];
+
+  let buttonText = (~isHovered, ~theme: Theme.t, ~font: UiFont.t) => [
+    fontFamily(font.fontFile),
+    color(theme.foreground),
+    isHovered
+      ? backgroundColor(theme.menuSelectionBackground)
+      : backgroundColor(theme.editorBackground),
+    fontSize(14),
+    alignSelf(`Center),
+  ];
+};
+
+let%component button = (~text, ~onClick, ~theme, ~font, ()) => {
+  let%hook (isHovered, setHovered) = Hooks.state(false);
+
+  <Clickable onClick style={Styles.buttonOuter(~theme, ~isHovered)}>
+    <View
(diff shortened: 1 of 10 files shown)


src/Components/MessageBox.re
```diff
@@ -0,0 +1,71 @@
...
+};
+
+module Styles = {
+  open Style;
+
+  let container = (~theme: Theme.t) => [backgroundColor(theme.background)];
+
+  let message = [
+    padding(20),
+    paddingBottom(30) // I don't know why this is needed, but it is
```
glennsl: There's some bugginess with the content size calculation here which causes the content to overflow unless adding some extra padding.

src/Model/State.re
```diff
@@ -54,6 +54,7 @@ type t = {
   pane: Pane.t,
   searchPane: Feature_Search.model,
   focus: Focus.stack,
+  modal: option(Modal.t),
```
glennsl: We might want to combine the modal and context menu model, so that a modal and context menu can't both be visible at the same time. Have to merge the context menu PR before doing that though.

src/Model/Actions.re
```diff
@@ -126,6 +126,10 @@ type t =
   | Sneak(Sneak.action)
   | PaneTabClicked(Pane.paneType)
   | VimDirectoryChanged(string)
+  | WindowCloseBlocked
+  | WindowCloseDiscardConfirmed
+  | WindowCloseSaveAllConfirmed
```
glennsl: These are a bit awkward as they're essentially user intents. I also don't think it's necessarily a problem to have "action-like" messages, and with feature projects the encapsulation and lack of granularity might sometimes make event-like messages difficult. But enforcing event-like names is still very useful to stop thinking of them as functions.

This is also a reason why I'm leaning more towards using `msg` instead of `event`. Another reason being that the term is established and familiar from the Elm architecture.

src/Components/MessageBox.re
glennsl: Seems like the text wrapping is what's causing the issue here. If the text doesn't wrap then everything measures as expected.

bryphe: Ah, ya, it sounds like from our conversation there may be a bug with the `measure` method of the `TextNode` 🤔

src/Model/Actions.re
bryphe: Sounds good! It seems like we're converging towards a very Elm-like architecture (ie, `updater`, `subscription`, and `msg`).

src/Model/Buffers.re
```diff
@@ -44,24 +44,17 @@ let isModifiedByPath = (buffers: t, filePath: string) => {
...
-let ofBufferOpt = (f, buffer) => {
-  switch (buffer) {
-  | None => None
-  | Some(b) => Some(f(b))
-  };
-};
-
 let applyBufferUpdate = bufferUpdate =>
-  ofBufferOpt(buffer => Buffer.update(buffer, bufferUpdate));
+  Option.map(buffer => Buffer.update(buffer, bufferUpdate));
```
bryphe: Oh, nice, thanks for cleaning this up!

src/Model/State.re
bryphe: Cool. I think the context menu should be unblocked now (https://github.com/onivim/oni2/pull/1232 brought in the `boxShadow` styling)

bryphe:
Cool to see this in action! Thanks for all the work on this, @glennsl

PR #1231, opened 2020-01-20
Fixes cursor position for menu dialog with prefixes

fanantoxa:
Code it wasn't counting full input value since that cursor was off to the left for a size of prefix.
Screenshot from 2020-01-20 15-59-21

This fixes this bug. Now cursor
Screenshot from 2020-01-20 15-59-59
 in place:


fanantoxa:
@glennsl Updated!

bryphe:
Great find @fanantoxa !

PR #1233, opened 2020-01-20
feat(other): paste text from buffer to input with ctrl + v

fanantoxa:
@glennsl It works but I'm not sure that this is ideal implementation.
What to start with something. This kinda missing feature when you for example copied path of failed test from console (CI).

### DIFF
--- a/src/Components/InputModel.re
+++ b/src/Components/InputModel.re
@@ -27,6 +27,12 @@ let add = (~at as index, insert, text) => (
   index + String.length(insert),
 );
 
+let getClipboardText = () =>
+  switch (Sdl2.Clipboard.getText()) {
+  | Some(value) => value
+  | None => ""
+  };
+
 let handleInput = (~text, ~cursorPosition) =>
   fun
   | "<LEFT>" => (text, max(0, cursorPosition - 1))
@@ -35,5 +41,6 @@ let handleInput = (~text, ~cursorPosition) =>
   | "<DEL>" => removeAfter(cursorPosition, text)
   | "<HOME>" => (text, 0)
   | "<END>" => (text, String.length(text))
+  | "<C-v>" => (text ++ getClipboardText(), String.length(text))
   | key when String.length(key) == 1 => add(~at=cursorPosition, key, text)
   | _ => (text, cursorPosition);


fanantoxa:
@glennsl Sounds REasonable! :) 
OS in general I get the point how to do that and will try to implement.

The most problems that I actually experience with implementing new stuff is typing system.
It _sounds_ good but if you should already know _notes_.
I think code might be much easier to work with if we'd add some type annotations.
It takes a lot of time to figure out what coming from where. (even with many years experience in other languages incl. functional)

Is that any easy way to get a hint on variable type for sublimetext or print it to console or smth? )

glennsl:
I don't know much about sublime text, but it seems to use the same language server that vscode does, which supports type hints. In vscode it's triggered by hovering over the definition, or typing `gh` when the cursor is on it. I (almost?) never need to use that myself though, even when first getting into a codebase.

Any module that exports definitions should have an interface file with type definitions for all the exports, and we're slowly moving towards that goal but a lot are unfortunately still missing. Otherwise there are usually plenty of hints other than type annotations, but it can take some getting used to seeing those. If there aren't any hints, that pretty much means it actually can be anything. There are still plenty of places where the code is structured badly so that the hints are too far away or there's just too much going on, and while local type annotations can certainly help there, it's usually better to refactor the code so they're not needed, which is happening, but slowly.

I was brand new to the codebase just three months ago myself, so I definitely see where you're coming from. But sounds like you're new to the language too, which of course doesn't help. What I struggled with the most was finding where the modules came from, because of inconsistent conventions, some being from Revery, and some being submodules that aren't searchable by filename. And the lack of imports in Reason certainly doesn't help. This has also gotten better though. There are fewer submodules and the structure is steadily improving, but still far to go. We could also be more explicit about external modules, like explicitly referencing or aliasing `Revery.Color` instead of opening `Revery` and using `Color` directly.

But let me know if you come across code that is really hard to follow. Then maybe I can explain it to you in a way that teaches you some new patterns and idioms that'll make it easier to understand similar code. Or I might just agree that it's really bad and refactor it right away :)

fanantoxa:
@glennsl Could you please add label WIP for it?
I think It'd be nice to implement ctrl+c, ctrl+A, ctrl+X along with that. 
But here another bug issue that prevent implementation for any of those

fanantoxa:
This is already implemented.

PR #1236, opened 2020-01-21
Development / Documentation: Add instructions for working with exthost

bryphe:
This adds some documentation describing on how to test / work with the extension host (it's useful to add logging to it when things go wrong!)

Also, when I was testing locally, what I'd often do is remap the `setup.json` to point to my local vscode exthost build. I made that a bit easier in this PR by providing a `ONI2_EXTHOST` env variable that can be used to override this setting - that just needs to point to a `vscode-exthost` folder (should be built, of course!).


bryphe:
Put this together per our discussion earlier @glennsl - let me know if you have any feedback. Hopefully we can get some logging for that issue you were running into w/ the text content provider / SCM 🤞

PR #1239, opened 2020-01-21
Add neat keyboard interaction to MessageBox

glennsl:
Peek 2020-01-21 16-33


src/Store/LifecycleStoreConnector.re
```diff
@@ -60,21 +60,25 @@ let start = quit => {
     | Actions.Quit(force) => (state, quitAllEffect(state, force))
 
     | WindowCloseBlocked => (
-        {...state, modal: Some(UnsavedBuffersWarning)},
+        {...state, modal: Some(Oni_UI.Modals.unsavedBuffersWarning)}
```
glennsl: I've added both the initial model and an `update` function to `Modals` in the UI project, which works since Store depends on UI. The only thing missing for `Modals` to be a feature module then is the model. So while this might seem a bit awkward at first, it'll be less of a transition to a proper feature project/module and we can omit an entire store and the boilerplate surrounding it just for this little thing.

bryphe: Sounds good! Thanks for calling it out 👍

PR #1240, opened 2020-01-22
SCM: Add diff algorithm, borrowed from camlp5

glennsl:
This adds the diff algorithm from camlp5 directly, instead of the Jane Street derivate because their interface is actually significantly different. This one essentially returns a list of adds and a list of deletes, while the Jane Street one returns a list of pointers to the lines that are the same in both. The former seems much easier to work with to me at least.

I've added a copyright notice and link to the original to the top of each file. Is that sufficient or should I add it to `ThirdPartyLicenses.txt` as well?

### DIFF
--- a/src/Core/Diff.ml
+++ b/src/Core/Diff.ml
@@ -1,7 +1,35 @@
- 
-(* camlp5r *)
-(* diff.ml,v *)
-(* Copyright (c) INRIA 2007-2017 *)
+(* This file has been adapted from camlp5:
+ * https://github.com/camlp5/camlp5/blob/master/ocaml_src/lib/diff.ml
+ *)
+
+(*
+ * Copyright (c) 2007-2017, INRIA (Institut National de Recherches en
+ * Informatique et Automatique). All rights reserved.
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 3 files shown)


src/Core/Diff.ml
```diff
@@ -0,0 +1,192 @@
...
+ * Copyright (c) 2007-2017, INRIA (Institut National de Recherches en
+ * Informatique et Automatique). All rights reserved.
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
```
CrossR: This part implies to me that we might need to add it to `ThirdPartyLicenses` (and perhaps surface that in the docs?).

glennsl: Yeah, I think you're right. Added in https://github.com/onivim/oni2/pull/1240/commits/2626a6cb56fe6ab4790889759e919f050b99dc42

bryphe: Perfect! Thanks for adding it.

PR #1244, opened 2020-01-23
Refactor: Utility into separate project (Piecewise - Part 1)

bryphe:
This creates a `Kernel` module and moves an initial set of files to it.


src/Core/Kernel/dune
```diff
@@ -0,0 +1,14 @@
...
+		str
+		zed_oni
+        Revery
+		fmt
+		fmt.tty
+		logs
+		logs.fmt
+		timber
+    )
+    (preprocess (pps lwt_ppx ppx_deriving_yojson ppx_deriving.show)))
```
glennsl: You're missing the deriving runtimes. I wonder if they're coming along with Revery though.

src/Core/Kernel/Kernel.re
```diff
@@ -0,0 +1,5 @@
+module CamomileBundled = CamomileBundled;
+module EnvironmentVariables = EnvironmentVariables;
+module Log = Log;
+module Setup = Setup;
+module ZedBundled = ZedBundled;
```
glennsl: You shouldn't need this I think, since everything's exported anyway.

bryphe:
Looks like this is no longer needed, since @glennsl got #1220 unblocked - thanks @glennsl ! Will close this in lieu of #1220

PR #1262, opened 2020-01-26
SCM: Editor diff markers

glennsl:
2020-01-26-083805_668x310_scrot

Part of the SCM extension integration, this adds diff makers to the editor and minimap showing the changes made to the current file relative to its checked-in or staged counterpart.

This implementation differ from vscode's in that they show the whole block as modified if there are both additions and deletions overlapping in said, while this implementation is more granular in a sense, showing a line as modified only if there's both an addition and deletion on that line, and otherwise treating them as stand-alone additons and deletions. For example, in the above screenshot. the deletion at the end of the first block would not be present in vscode.

This is functionally ready to be merged, although there;s still plenty of room for improvement:

- [ ] Find a good heuristic for choosing SCM provider
- [ ] Update original content when changed (ie. on staging or commit). As far as I can tell there is no notification from the git extension of this happening, and since it's a git uri it's not straightforward to use fswatch on it either.
- [ ] Investigate and improve heuristics of diff algorithm, in particular with regards to alignment of code blocks.
- [ ] Render an arrow marker for deletion, to make it more visible. I don't think this is possible with the current drawing primitives however (unless we use a font icon I guess). We might want to wait until skia has dropped for this one.
- [ ] Improve performance (although from what I can tell the impact is negligible in normal use cases)
  - [ ] Adapt diff algorithm to BufferLine interface to avoid allocation of an extra array
  - [ ] Generate markers only on buffer enter and update, rather than on every render
  - [ ] Generate markers asynchronously or time-sliced

### DIFF
--- a/bench/lib/EditorDiffMarkersBench.re
+++ b/bench/lib/EditorDiffMarkersBench.re
@@ -1,0 +1,112 @@
+open Oni_Core;
+open Oni_UI;
+open BenchFramework;
+
+// DATA
+
+module Data = {
+  let randomString = () =>
+    String.init(Random.int(100), _ =>
+      Char.chr(Char.code('a') + Random.int(29))
+    );
+
+  let randomBufferLine = () =>
+    BufferLine.make(~indentation=IndentationSettings.default, randomString());
+
+  let lines_10k_a = Array.init(10000, _ => randomString());
+  let lines_10k_b = Array.init(10000, _ => randomString());
+  let lines_100k = Array.init(100000, _ => randomString());
+
+  let buffer_10k_nochanges =
+    Buffer.ofLines(lines_10k_a) |> Buffer.setOriginalLines(lines_10k_a);
+
+  let buffer_10k_randomchanges =
+    Buffer.ofLines(lines_10k_a) |> Buffer.setOriginalLines(lines_10k_b);
+
+  let buffer_10k_onelineoriginal =
+    Buffer.ofLines(lines_10k_a)
+    |> Buffer.setOriginalLines([|randomString()|]);
+
+  let buffer_10k_onelinemodified =
+    Buffer.ofLines([|randomString()|])
+    |> Buffer.setOriginalLines(lines_10k_a);
+
+  let buffer_100k_nochanges =
+    Buffer.ofLines(lines_100k) |> Buffer.setOriginalLines(lines_100k);
+};
+
+// TESTS
+
+module Tests = {
+  let diff_10k_nochanges = () => {
+    let _ = EditorDiffMarkers.generate(Data.buffer_10k_nochanges);
+    ();
+  };
+
+  let diff_10k_randomchanges = () => {
+    let _ = EditorDiffMarkers.generate(Data.buffer_10k_nochanges);
+    ();
+  };
+
+  let diff_10k_onelineoriginal = () => {
+    let _ = EditorDiffMarkers.generate(Data.buffer_10k_onelineoriginal);
+    ();
+  };
+
+  let diff_10k_onelinemodified = () => {
+    let _ = EditorDiffMarkers.generate(Data.buffer_10k_onelinemodified);
+    ();
+  };
+
+  let diff_100k_nochanges = () => {
+    let _ = EditorDiffMarkers.generate(Data.buffer_100k_nochanges);
+    ();
+  };
+};
+
+// PLUMBING
+
+let setup = () => ();
+let options = Reperf.Options.create(~iterations=1000, ());
+
+bench(
+  ~name="EditorDiffMarkers: 10k lines, no changes",
+  ~options,
+  ~setup,
+  ~f=Tests.diff_10k_nochanges,
+  (),
+);
+
+bench(
+  ~name="EditorDiffMarkers: 10k lines, random changes",
+  ~options,
+  ~setup,
+  ~f=Tests.diff_10k_randomchanges,
+  (),
+);
+
+bench(
+  ~name="EditorDiffMarkers: 10k lines, 1 line original",
+  ~options,
+  ~setup,
+  ~f=Tests.diff_10k_onelineoriginal,
+  (),
+);
+
+bench(
+  ~name="EditorDiffMarkers: 10k lines, 1 line modified",
+  ~options,
+  ~setup,
+  ~f=Tests.diff_10k_onelinemodified,
+  (),
+);
+
+let options = Reperf.Options.create(~iterations=100, ());
+
+bench(
+  ~name="EditorDiffMarkers: 100k lines, no changes",
+  ~options,
+  ~setup,
+  ~f=Tests.diff_100k_nochanges,
+  (),
+);
--- a/src/Store/ExtensionClientStoreConnector.re
+++ b/src/Store/ExtensionClientStoreConnector.re
@@ -463,6 +463,54 @@ let start = (extensions, setup: Setup.t) => {
       )
     });
 
+  let getOriginalUri = (bufferId, path, providers: list(SCM.Provider.t)) =>
+    Isolinear.Effect.createWithDispatch(~name="scm.getOriginalUri", dispatch => {
+      // Try our luck with every provider. If several returns Last-Writer-Wins
+      // TODO: Is there a better heuristic? Perhaps use rootUri to choose the "nearest" provider?
+      providers
+      |> List.iter((provider: SCM.Provider.t) => {
+           let promise =
+             ExtHostClient.provideOriginalResource(
+               provider.handle,
+               Uri.fromPath(path),
+               extHostClient,
+             );
+
+           Lwt.on_success(promise, uri =>
+             dispatch(Actions.SCM(SCM.GotOriginalUri({bufferId, uri})))
+           );
+         })
+    });
+
+  let getOriginalContent = (bufferId, uri, providers) =>
+    Isolinear.Effect.createWithDispatch(
+      ~name="scm.getOriginalSourceLines", dispatch => {
+      let scheme = uri |> Uri.getScheme |> Uri.Scheme.toString;
+      providers
+      |> List.find_opt(((_, providerScheme)) => providerScheme == scheme)
+      |> Option.iter(provider => {
+           let (handle, _) = provider;
+           let promise =
+             ExtHostClient.provideTextDocumentContent(
+               handle,
+               uri,
+               extHostClient,
+             );
+
+           Lwt.on_success(
+             promise,
+             content => {
+               let lines =
+                 content |> Str.(split(regexp_string("\n"))) |> Array.of_list;
+
+               dispatch(
+                 Actions.SCM(SCM.GotOriginalContent({bufferId, lines})),
+               );
+             },
+           );
+         });
+    });
+
   let updater = (state: State.t, action) =>
     switch (action) {
     | Actions.Init => (
@@ -488,10 +536,18 @@ let start = (extensions, setup: Setup.t) => {
         changeWorkspaceEffect(path),
       )
 
-    | Actions.BufferEnter(bm, fileTypeOpt) => (
-        state,
-        sendBufferEnterEffect(bm, fileTypeOpt),
-      )
+    | Actions.BufferEnter(metadata, fileTypeOpt) =>
+      let eff =
+        switch (metadata.filePath) {
+        | Some(path) =>
+          Isolinear.Effect.batch([
+            sendBufferEnterEffect(metadata, fileTypeOpt),
+            getOriginalUri(metadata.id, path, state.scm.providers),
+          ])
+
+        | None => sendBufferEnterEffect(metadata, fileTypeOpt)
+        };
+      (state, eff);
 
     | Actions.NewTextContentProvider({handle, scheme}) => (
         {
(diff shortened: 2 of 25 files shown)


src/Store/ExtensionClientStoreConnector.re
```diff
@@ -455,6 +463,54 @@ let start = (extensions, setup: Setup.t) => {
       )
     });
 
+  let getOriginalUri = (bufferId, path, providers: list(SCM.Provider.t)) =>
+    Isolinear.Effect.createWithDispatch(~name="scm.getOriginalUri", dispatch => {
+      // Try our luck with every provider. If several returns Last-Writer-Wins
+      // TODO: Is there a better heuristic? Perhaps use rootUri to choose the "nearest" provider?
```
glennsl: TODO

src/UI/EditorDiffMarkers.re
```diff
@@ -0,0 +1,123 @@
...
+
+           | DeletedBefore
+           | DeletedAfter
+           | Unmodified => DeletedAfter
+           }
+         );
+       };
+
+       markers;
+     });
```
glennsl: This could certainly be better commented, and probably better structured.

src/UI/EditorSurface.re
```diff
@@ -242,6 +250,9 @@ let%component make =
         )
       : 0.0;
 
+  let gutterWidth =
+    lineNumberWidth +. Constants.diffMarkerWidth +. Constants.gutterMargin;
```
glennsl: The lack of pretty much any kind of structure in this module makes it very difficult to know if every case has been covered. I'm hoping that just replacing the use of `lineNumberWidth` with this should do the trick, but I don't really think this is an improvement over what was either.

Ideally there would be separate "canvases" for separate parts, so that measures accumulate structurally rather than being pieced together from loose parts wherever needed. If there was a gutter component or function, for example, then the parts both available to it and exposed by it would be greatly reduced, and hence reduce the overall complexity. `lineNumberWidth` would most likely only be needed within the gutter, and the gutter width is probably the only measure needing to be exposed.

src/UI/Minimap.re
```diff
@@ -178,74 +189,71 @@ let%component make =
...
           );
         };
 
         let cursorPosition = Editor.getPrimaryCursor(editor);
+
         /* Draw cursor line */
         Shapes.drawRect(
           ~transform,
-          ~x=0.,
+          ~x=Constants.leftMargin,
```
glennsl: Here I've also added a margin, as that makes the minimap stand out a bit more from the editor surface.

src/UI/EditorDiffMarkers.rei
```diff
@@ -0,0 +1,27 @@
...
+
+[@deriving show]
+type t = array(marker)
+
+and marker =
+  | Modified
+  | Added
+  | DeletedBefore
+  | DeletedAfter
+  | Unmodified;
```
glennsl: These are only exposed for testing purposes. I wish there was a better way of testing that didn't require breaking encapsulation.

bryphe: Ah, right, I wish I had this too - sometimes I want to expose types just for tests, but not as part of the 'public' API.

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -455,6 +463,54 @@ let start = (extensions, setup: Setup.t) => {
...
+               handle,
+               uri,
+               extHostClient,
+             );
+
+           Lwt.on_success(
+             promise,
+             content => {
+               let lines =
+                 content |> Str.(split(regexp_string("\n"))) |> Array.of_list;
```
bryphe: Do we need to handle Windows-style line ending (CRLF) here?

src/Store/ExtensionClientStoreConnector.re
bryphe: Thanks for calling this out! I think last-writer-wins is OK for now. Too bad there isn't a concept of 'precedence' with the SCM providers.

src/UI/EditorDiffMarkers.re
```diff
@@ -0,0 +1,129 @@
...
+type t = array(marker)
+
+and marker =
+  | Modified
+  | Added
+  | DeletedBefore
+  | DeletedAfter
+  | Unmodified;
+
+let generate = buffer =>
```
bryphe: It would be nice to have a benchmark for this algorithm, so we could get a sense of the relative performance (especially for very large files - like 10k + lines).

bryphe: It seems like this algorithm will always need to traverse the full set of buffer lines - which worries me in terms of render performance

src/UI/EditorSurface.re
bryphe: Definitely - it'd be great to have the line numbers (and gutter) as a separate component. That would help significantly.

bryphe:
What's the best way to test this out? I tried making some changes to files (ie, open `NodeTask.re`, make changes). I can see the 'getOriginalUri' effect is being executed, but I'm not getting any content back. 

Is it expected to be hooked up to the git provider at this point, or is there some other way I should try it out?

src/Store/ExtensionClientStoreConnector.re
glennsl: Isn't that covered by `\n`? Windows is just `\r\n` right? How do you handle it elsewhere?

src/UI/EditorDiffMarkers.re
glennsl: Relative to what? I can certainly add a benchmark case for it, but not sure what to compare it to initially.

glennsl:
> Is it expected to be hooked up to the git provider at this point, or is there some other way I should try it out?

Yeah, it should use git and it should update immediately when you make a change in the buffer. No need to save.

> I tried making some changes to files (ie, open NodeTask.re, make changes). I can see the 'getOriginalUri' effect is being executed, but I'm not getting any content back.

Odd, this works for me. Are you seeing any weird messages? If you `--trace` do you see the JSON-RPC message being sent, and if there's any response?

src/UI/EditorDiffMarkers.re
bryphe: Sorry, relative was the wrong choice of words! I really want to get a sense of the performance 'limits' - in other words, at what size of buffer do we need to start worrying about a non-negligible impact on rendering perf? 

The cases that are most interesting to me are the extreme 'large file' cases, ie:
- 10k line file, diffed against another 10k line file
- 100k line file, diffed against another 100k line file

It'd also be interesting to see cases where there is a large disparity:
- 1 line file diffed against a 10k line file, and vice versa

Really I just want to get a sense of the timing performance (ie, how many ms to diff a 100k file)?

bryphe:
> Odd, this works for me. Are you seeing any weird messages? If you --trace do you see the JSON-RPC message being sent, and if there's any response?

Strange! I've attached a log of tracing with the latest on this branch:
[log.txt](https://github.com/onivim/oni2/files/4125660/log.txt)

A couple things I noticed:
- I see a `$registerSourceControl` for the test provider only:
```
[TRACE] JSON-RPC RECV: {"jsonrpc":"2.0","method":"host/msg","params":{"type":4,"reqId":24,"payload":{"rpcName":"MainThreadSCM","methodName":"$registerSourceControl","args":[0,"test","Test",null]}}}
[INFO]  Oni2.Store.dispatch (SCM NewProvider {handle = 0; id = "test"; label = "Test"; rootUri = None})
[TRACE] JSON-RPC SEND: {"method":"ext/msg","params":{"type":12,"reqId":24,"payload":{}}}
```
... although I also see an unregister:
```
[TRACE] JSON-RPC RECV: {"jsonrpc":"2.0","method":"host/msg","params":{"type":4,"reqId":31,"payload":{"rpcName":"MainThreadSCM","methodName":"$unregisterSourceControl","args":[0]}}}
[INFO]  Oni2.Store.dispatch (SCM LostProvider {handle = 0})
[TRACE] JSON-RPC SEND: {"method":"ext/msg","params":{"type":12,"reqId":31,"payload":{}}}
```

and nothing for the git provider.

- I see the `scm.getOriginalUri` effect - but I don't see it actually execute the JSON-RPC message being sent. This makes me suspect that maybe we don't have any providers?

It doesn't look like the `git` extension ever gets activated for me on Mac - so perhaps that's the problem. Were there any special dependencies you had to install to get it working?

src/UI/EditorDiffMarkers.re
glennsl: Alright, will do!

glennsl:
> ... although I also see an unregister

Yeah, that should be there. To test the unregistration and to not have it mess interfere with actual scm extensions.

> I see the scm.getOriginalUri effect - but I don't see it actually execute the JSON-RPC message being sent. This makes me suspect that maybe we don't have any providers?

Indeed, that does sound like it.

> It doesn't look like the git extension ever gets activated for me on Mac - so perhaps that's the problem. Were there any special dependencies you had to install to get it working?

No, I did nothing special. From the logs it seems like it fails on not finding `which`. So as I understand we'll need to add dependencies like this to "our" `package.json` and run `node install-dependencies.js` right? Could you try that?

```
[DEBUG] Oni2.Extensions.NodeProcess [49579|bootstrap-fork.js] Failed to activate extensions: Error: Cannot find module 'which'
[DEBUG] Oni2.Extensions.NodeProcess [49579|bootstrap-fork.js] [error] Activating extension 'vscode.git' failed: Cannot find module 'which'.
[DEBUG] Oni2.Extensions.NodeProcess [49579|bootstrap-fork.js] Activating extension `vscode.git` failed:  Cannot find module 'which'
```

glennsl:
Benchmark results are in:
```
|  EditorDiffMarkers: 10k lines, no changes                           |  1000       |  9.14890289307      |  1001      |  30        |  78824181     |  78680665   |  460842665    |
|  EditorDiffMarkers: 10k lines, random changes                       |  1000       |  9.0534529686       |  1000      |  31        |  78824186     |  78601945   |  460763945    |
|  EditorDiffMarkers: 10k lines, 1 line original                      |  1000       |  4.32785820961      |  500       |  14        |  39497101     |  39320665   |  210397665    |
|  EditorDiffMarkers: 10k lines, 1 line modified                      |  1000       |  3.90272212029      |  500       |  13        |  39501096     |  39321637   |  190396637    |
|  EditorDiffMarkers: 100k lines, no changes                          |  1000       |  117.832752943      |  6068      |  236       |  781193163    |  781111465  |  4078281465   |
```

100k lines is definitely over the line, but 10k lines seems acceptable for now, although definitely not ideal. What do you think @bryphe?

src/Store/ExtensionClientStoreConnector.re
bryphe: Right, Windows could be`\r\n`. The concern I had would be that, our `Buffer.t` / `BufferLine.t` don't encode any newlines. 

So if we trim on the `\n`, there might still be a trailing `\r` that would show up as a diff (an extra character) - since the buffer/bufferlines we use don't store any newlines.

bryphe:
> No, I did nothing special. From the logs it seems like it fails on not finding which. So as I understand we'll need to add dependencies like this to "our" package.json and run node install-dependencies.js right? Could you try that?

Ah, good catch! That could definitely be it. I added it (and other deps I found) in this PR: https://github.com/onivim/oni2/pull/1277

After that - the `vscode.git` extension is being activated _and_ I can see diff markers on OSX now! 🎉 

image

> 100k lines is definitely over the line, but 10k lines seems acceptable for now, although definitely not ideal. What do you think @bryphe?

Great set of benchmarks @glennsl ! Thanks for adding those! Helps a lot in terms of understanding the performance characteristics.

The budget I usually work with for things that are on the hot path for rendering is ~1-2ms (so that we can stay < 16ms, which is 60 FPS). Our rendering today by itself often exceeds 16 ms, unfortunately, which means we have to be even more conscientious at the moment. 

For the 10k diff, amortized across those 1000 runs, that's about 9ms to run the diff - that can be a pretty significant additional latency, especially since our current rendering isn't super efficient. 

What might be reasonable for now is to 'cap it' - only render the diffs if the line count is below a certain limit, like 1-2k (which would be in the 1-2ms budget if it scales linearly). Once we optimize (ie, only process diffs on update, only process diffs up to the visible point, or offload it / time slice it), we could increase that limit.  Does that sound good?

glennsl:
> After that - the vscode.git extension is being activated and I can see diff markers on OSX now!
:tada: 

> What might be reasonable for now is to 'cap it' - only render the diffs if the line count is below a certain limit, like 1-2k

Done. I think that's very reasonable.

bench/lib/EditorDiffMarkersBench.re
```diff
@@ -0,0 +1,112 @@
+open Oni_Core;
+open Oni_UI;
+open BenchFramework;
+
+// DATA
+
+module Data = {
```
bryphe: I like how you organized this benchmark

src/Store/ExtensionClientStoreConnector.re
glennsl: Forgot about this. Fixed in https://github.com/onivim/oni2/pull/1262/commits/58c237fe70d692b356ee6171b23af8eb52276501

PR #1263, opened 2020-01-26
[docs] Fix broken link to system requirements in FAQ

glennsl:


glennsl:
I'll just merge this as I feel pretty confident it'll be accepted. What I'm less sure about is whether the removal of the installation doc from the sidebar in #905 was intentional, and whether it should be added back.

PR #1264, opened 2020-01-26
Selection behavior for inputs

fanantoxa:
Selection behavior for input field:
selection

Implements input ability to have selection
Implements Shift + [Left, Right, Home, End] for selection
Implements Shift + Control + [Left, Right] for selection of word boundaries
Implements [Left, Right, Home, End, char insertion] to deal with selection

This is not done fully done yet
- [x]  Fix handling Right on selection when cursor at the end
- [x]  Fix  `?`, `/` to work with it (missing default selection position now) ? Might not applicable
- [x] More testing to cover other menus that uses Input component
- [x] Refactor rendering code
- [x] Refactor to have separate `Selection` module for combined cursor position and selection`
- [x] Tune w`Wild` menu (`?`, `/`) t have only input
- [x] Fix Feature_SCM input


fanantoxa:
@bryphe , @glennsl 
Hi! This is not fully done yet.
But I wanted to share with you this part already to omit useless work:
 - Is that way as I implemented look good? Should I continue?
   - (I think every Input should have ability since this is default behavior for any similar imput withing any OS/other editors)
- Is that good please to implement it?
- I have trouble with handling right click when cursor already at the end so I can't clear selection in this situation
    - I think this related to `quickmenuCursorEnd`: https://github.com/onivim/oni2/blob/master/src/Core/Input.re#L14
     But I didn't get what it was created for and can I just remove that

src/Components/InputModel.re
```diff
@@ -1,5 +1,33 @@
...
+let findPrevWordBoundary = (text, cursorPosition) => {
+  let finalIndex = 0;
+  let index = ref(max(cursorPosition - 1, finalIndex));
+
+  while (index^ > finalIndex && !separatorOnIndex(text, index^ - 1)) {
+    index:= index^ - 1;
+  };
+
+  index^;
+};
```
glennsl: It would probably be better to use `Str.search_forward` and `Str.search_backward` for this. Or if that doesn't fit or perform well enough, generalize and extract this into `Utility.StringEx`.

src/Components/InputModel.re
```diff
@@ -14,26 +42,167 @@ let slice = (~start=0, ~stop=?, str) => {
...
+    });
+
+  let newIndex = switch (nextWord^) {
+   | Some(-1) => length
+   | Some(index) => index
+   | None => length
+   } ;
+
+  newIndex;
+}
```
glennsl: This code is really confusing to me. Why is there a distinction between `None` and `Some(-1)` when they seem to be treated exactly the same? Why is `nextWord` initialized with `None` and then immediately set to `Some(-1)`?

What is the difference between this and `findNextWordBoundary`? They seem very similar, except the default case is different. It seems like they could be combined into one function returning an `option`, deferring the decision of what to do with no match to the consumer.

src/Components/InputModel.re
```diff
@@ -14,26 +42,167 @@ let slice = (~start=0, ~stop=?, str) => {
...
+  index + String.length(insert)
+);
+
+let moveCursorLeft = (text, cursorPosition, _) => {
+  let leastCursorPoition = max(0, cursorPosition - 1);
+
+  (text, leastCursorPoition, leastCursorPoition);
+};
+
+let moveCursorToSelectionBeginnig = (text, cursorPosition, selectionPosition) => {
```
glennsl: Same typo again

test/Components/InputModelTests.re
```diff
@@ -0,0 +1,992 @@
...
+
+let runInputHandler = (~text=testString, position, selection, key) => {
+  InputModel.handleInput(
+    ~text=text,
+    ~cursorPosition=position,
+    ~selectionPosition=selection,
+  key);
+};
+
+describe("InputModel#handleInput", ({describe, _}) => {
```
glennsl: This is great. Thanks for adding this!

src/Components/InputModel.re
fanantoxa: I'm sorry. This was an my attempt to implement `findNextWordBoundary` but it didn't workout. Will remove.

glennsl: No worries, that certainly does explain the unfinished look of it :D Worth noting that this would also have been caught if this had an interface file.

My suggestion to make the `find*WordBoundary` functions return an option instead still stands though. It's not obvious to me from the name what the default would be.

fanantoxa:
@glennsl 
> Why do you think it needs to be configurable? It doesn't really seem preference-based.

This is mostly English word boundaries but people might have few more on other languages.
For ex. in Spanish they could use `¿` Mark

> Definitely. The only large structural change I might want to see is to encapsulate the selection model in a Selection submodule. See my comment below on that.
> These ought to be combined in a selection object I think. I believe a common model is to have cursorPosition named the focus and selectionPosition the anchor of a selection. That might also be a good opportunity to gather a lot of this functionality into a Selection submodule to encapsulate it.

I need some help here. I don't clearly see how to encapsulate it.
Can we have a short chat/call about that you can find me as `i80286#4671` on Discord.
Really need some help here. Functional patterns kinda new for me.

Also if you can tell me about those `interface file`

src/Components/Input.re
```diff
@@ -72,45 +72,15 @@ let getStringParts = (index, str) => {
...
-
 module Constants = {
   let cursorWidth = 2;
+  let selectionOpacity = 0.75;
 };
 
 module Styles = {
   let defaultPlaceholderColor = Colors.grey;
   let defaultCursorColor = Colors.black;
+  let defaultSelectionColor = Color.hex("#42557b");
```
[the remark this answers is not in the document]
fanantoxa: :+1:

src/Components/Input.re
```diff
@@ -127,11 +97,13 @@ let%component make =
                 ~style=Styles.default,
                 ~placeholderColor=Styles.defaultPlaceholderColor,
                 ~cursorColor=Styles.defaultCursorColor,
+                ~selectionColor=Styles.defaultSelectionColor,
                 ~placeholder="",
                 ~prefix="",
                 ~isFocused,
                 ~value,
                 ~cursorPosition,
+                ~selectionPosition,
```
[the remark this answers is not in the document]
fanantoxa: I need help here. Can find a time to discuss please?

src/Components/InputModel.re
fanantoxa: So this is really big question. I'd better leave it for now here because it's not handling Unicode right no. Current implementations will work with Unicode easily.
This something I really what to discuss and work on it after this PR.

In short all this thing should tolerate Unicode since filenames, search queries, etc. might be in non english (Ukrainian, Chinese, Japanise,  German..)

I'd better Create an Issue about Updating this to handle Unicode and own it.

src/Components/Input.re
```diff
@@ -72,45 +72,15 @@ let getStringParts = (index, str) => {
...
-      }
-    );
-  (newString, cursorPosition);
-};
-
-let addCharacter = (word, char, index) => {
-  let (startStr, endStr) = getStringParts(index, word);
-  (startStr ++ char ++ endStr, String.length(startStr) + 1);
-};
-
```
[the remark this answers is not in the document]
bryphe: Nice catch @fanantoxa !

src/Components/InputModel.re
bryphe: That would be a huge help @fanantoxa ! We're missing proper Unicode support (and some other stuff, like font fallbacks) in those places. 

I believe the [Oniguruma](https://github.com/kkos/oniguruma) regex engine we use can support Unicode (and we have an OCaml/Reason binding for it here: https://github.com/onivim/reason-oniguruma) - so could be an option to think about later too.

src/Components/InputModel.re
```diff
@@ -28,13 +56,113 @@ let add = (~at as index, insert, text) => (
...
+
+  (newText, Selection.collapse(~text=newText, focus));
+};
+
+let handleInput = (~text, ~selection, key) => {
+  switch (key, Selection.isCollapsed(selection)) {
+  | ("<LEFT>", true) => (
+      text,
+      Selection.collapseRelative(~text, ~selection, Left(1)),
+    )
```
glennsl: I really like how simple these all are. It makes it really easy to get a good overview of what it does for every given input. The function applications might be clearer in some cases though, but I've addressed that elsewhere

src/Components/Selection.re
```diff
@@ -0,0 +1,75 @@
...
+  create(~text, ~anchor=offset, ~focus=offset);
+};
+
+let collapseRelative =
+    (~text: string, ~selection: t, direction: directionVariant): t => {
+  switch (direction) {
+  | Left(offset) => collapse(~text, rangeStart(selection) - abs(offset))
+  | Right(offset) => collapse(~text, rangeEnd(selection) + abs(offset))
+  };
+};
```
glennsl: This seems more like creating a new selection relative to the old one, than collapsing it. Or alternatively collapse an move, but then I think it should remain as two operations.

src/Components/Selection.re
```diff
@@ -0,0 +1,75 @@
...
+  abs(selection.focus - selection.anchor);
+};
+
+let rangeStart = (selection: t): int => {
+  min(selection.focus, selection.anchor);
+};
+
+let rangeEnd = (selection: t): int => {
+  max(selection.focus, selection.anchor);
+};
```
glennsl: It's not clear to me what these functions mean, even after seeing their implementation. What is a "range" here? To me a "range" implies at least a start and an end, i.e. two numbers, or it could be mean the elements between the start and end, but I can't understand how a single number can be a range unless one of the ends are implied.

src/Components/Selection.re
```diff
@@ -0,0 +1,75 @@
...
+
+let isCollapsed = (selection: t): bool => {
+  selection.anchor == selection.focus;
+};
+
+let collapse = (~text: string, offset: int): t => {
+  create(~text, ~anchor=offset, ~focus=offset);
+};
```
glennsl: The name of this indicates that it will collapse a selection, when in fact it creates a new selection that is collapsed. Perhaps just rename it to `collapsed`?

src/Components/InputModel.re
```diff
@@ -28,13 +56,113 @@ let add = (~at as index, insert, text) => (
...
+};
+
+let addCharacter = (key, text, selection) => {
+  let (newText, focus) = add(~at=Selection.focus(selection), key, text);
+
+  (newText, Selection.collapse(~text=newText, focus));
+};
+
+let replacesSelection = (key, text, selection) => {
+  let (textSlice, sliceSelection) = removeSelection(text, selection);
```
glennsl: `sliceSelection` sounds like a function name. I think this should be `selectionSlice` instead?

glennsl:
Apologies for the delay in reviewing this. Github was giving me problems. There might therefore also be duplicate comments. I've found and deleted a few already.

src/UI/QuickmenuView.re
```diff
@@ -187,7 +187,10 @@ let make =
          | EditorsPicker => React.empty
          | _ => <input />
          }}
-        <dropdown />
+        {switch (variant) {
+         | Wildmenu(_) => React.empty
+         | _ => <dropdown />
+         }}
```
glennsl: This isn't right though. The Wildmenu has autocompletion for commands and lots of other things.

glennsl: You need to be more specific. I think this should be OK though:

```suggestion
        {switch (variant) {
         | Wildmenu(SearchForward | SearchBackward) => React.empty
         | _ => <dropdown />
         }}
```

src/Components/InputModel.re
```diff
@@ -1,6 +1,34 @@
 open Oni_Core;
 open Utility;
 
+let wordSeparators = " ./\\()\"'-:,.;<>~!@#$%^&*|+=[]{}`~?";
```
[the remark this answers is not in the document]
fanantoxa: @glennsl Actually it basic OS interactions with text selection. This exact stop symbols I took from sublimetext.
They are +- the same for Win/Linux (maybe Mac, don't know).
It should be good for now. I whanned to keep it stupid simple to implement Selection feature.
Next task I'll take is changing inputs to work with Unicode. So I'll make something better there.

src/Components/InputModel.re
```diff
@@ -28,13 +56,113 @@ let add = (~at as index, insert, text) => (
...
-  | key when String.length(key) == 1 => add(~at=cursorPosition, key, text)
-  | _ => (text, cursorPosition);
+let removeCharBefore = (text, selection) => {
+  let (textSlice, _) = removeBefore(Selection.focus(selection), text);
+
+  (
+    textSlice,
+    Selection.collapseRelative(~text=textSlice, ~selection, Left(1)),
+  );
+};
```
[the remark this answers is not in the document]
fanantoxa: @glennsl It for separating methods that work directly with string. 
It'll be easier to update them later for Unicode.

src/Components/InputModel.re
fanantoxa: :+1:

src/Components/Selection.re
```diff
@@ -0,0 +1,75 @@
+[@deriving show({with_path: false})]
+type t = {
+  anchor: int,
+  focus: int,
+};
+
+type directionVariant =
+  | Left(int)
+  | Right(int);
```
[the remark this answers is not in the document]
fanantoxa: 2 reasons:
 - I didn't know if someone will define another `direction` type and it'll brake. But this type is `Selection.direction` so we safe here.
 - My thought was that adding `..Variant` at the end will make better error since it'll be straight indication that this is variant type.

src/Components/Selection.re
```diff
@@ -0,0 +1,75 @@
...
+  {anchor: safeAnchor, focus: safeFocus};
+};
+
+let anchor = (selection: t): int => {
+  selection.anchor;
+};
+
+let focus = (selection: t): int => {
+  selection.focus;
+};
```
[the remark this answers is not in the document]
fanantoxa: Didn't know. Thank you!

src/Components/Selection.re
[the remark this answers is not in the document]
fanantoxa: it should be `rangeWidth`

src/Components/Selection.re
fanantoxa: I was inspired this thing https://developer.mozilla.org/en-US/docs/Web/API/Selection
For browser it does make since it mutate object.
In this case I agree. `collapsed` make much more sence.

src/Components/Selection.re
fanantoxa: I think concept of `collapseRelative` have a lot of advantages. 
It take `selection` and just collapse thing relatively in direction.

renaming it to `collapseAndMove` will be less representative from my point.
And introducing `move`  will require to add some logic in place where we will use it to calculate offset (but this will turn out to just using collapsed method)

So I'd better leave this thing here. 
But If you would insist I can change it :)

src/Store/QuickmenuStoreConnector.re
```diff
@@ -149,30 +149,30 @@ let start = (themeInfo: ThemeInfo.t) => {
...
 
-    | QuickmenuInputClicked(cursorPosition) => (
-        Option.map(state => Quickmenu.{...state, cursorPosition}, state),
+    | QuickmenuInputClicked(selection) => (
+        Option.map(state => Quickmenu.{...state, selection}, state),
         Isolinear.Effect.none,
       )
 
-    | QuickmenuCommandlineUpdated(text, cursorPosition) => (
+    | QuickmenuCommandlineUpdated(text, focus) => (
```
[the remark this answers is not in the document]
fanantoxa: It's not a selection object. It is `int` value came from `Vim.CommandLine.onUpdate`
I can change it back to `cursorPosition` or if you can find better name for it.
https://github.com/onivim/oni2/blob/master/src/Store/VimStoreConnector.re#L429

src/UI/QuickmenuView.re
fanantoxa: Nice to know that I can use `|` with variants.Thx!

src/Components/Selection.re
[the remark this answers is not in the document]
glennsl: Not entirely sure what you mean about the last point, but ordinary variants generally do produce  better error messages than polymorphic variants. The difference isn't all that significant for polymorphic variants with a small number of constructors though. It might even be better, since it will list the possibilities instead of just referring to the type name.

src/Components/Selection.re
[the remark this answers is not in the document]
glennsl: I still don't know what "range" means though...

src/Store/QuickmenuStoreConnector.re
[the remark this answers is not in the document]
glennsl: Oh, sorry, I must have looked at the wrong constructor. `cursorPosition` seems more correct though, since we're still in VIM domain here.

src/Components/InputModel.re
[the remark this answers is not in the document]
glennsl: Cool. Was just curious. If this was going to stay I'd like to have a comment pointing to the source, but no reason to bother if you're going to change it right away.

src/Components/InputModel.re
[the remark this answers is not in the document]
glennsl: Then the names should reflect that I think. As it is now, just from looking at the names, they seem to both do the same thing.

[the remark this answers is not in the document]
glennsl: And of course this applies to all similar cases as well, `add`/`addChar` etc..

src/Components/Selection.re
glennsl: > And introducing move will require to add some logic in place where we will use it to calculate offset (but this will turn out to just using collapsed method)

No, `move` would just move the selection. Increasing or decreasing both `anchor` and `offset` the same amount. No collapsing, that's a different operation.

> It take selection and just collapse thing relatively in direction.

That's just `collapse` in my mind. There is no "absolute" collapse, it's always relative, so I think the right approach is to just name it `collapse` and then pass in a variant that specify where it should collapse to. Either left, right, anchor or focus.

From a use POV, I'd prefer an API that looked something like this instead of `collapseRelative`:

```
selection |> collapse(`Left) |> move(1)
```

or 

```
selection |> collapseTo(`Anchor) |> move(`Right, 1)
```

Do you see how this kind of compositional API can be both simpler _and_ more flexible?

fanantoxa: I've found better solution. I just removed those `relative` things. They kinda redundant.

fanantoxa:
@glennsl Done and done) 
(test are failing on `MacOS: Validate Release` not sure what that mean.)

glennsl:
> (test are failing on MacOS: Validate Release not sure what that mean.)

That means we try again :)

src/Components/InputModel.re
```diff
@@ -1,6 +1,34 @@
 open Oni_Core;
 open Utility;
 
+let wordSeparators = " ./\\()\"'-:,.;<>~!@#$%^&*|+=[]{}`~?";
+
+let separatorOnIndex = (index, text) => {
+  String.contains(wordSeparators, text.[index]);
+};
```
glennsl: This can raise an exception since it indexes into the string without doing any bounds-checking. Not a big problem since it's just used internally, but ti should still be named to indicate so. The convention is to add an `Exn` suffix:

```suggestion
let separatorOnIndexExn = (index, text) => {
  String.contains(wordSeparators, text.[index]);
};
```

src/Components/Selection.re
glennsl: I'd prefer the simplicity and flexibility of a compositional API, because code tends to evolve, but yeah, that solution works fine :+1:

src/Components/Selection.re
```diff
@@ -0,0 +1,53 @@
...
+
+let withinLength = (text: string, position: int): int => {
+  let withinStart = max(position, 0);
+  let withinBoth = min(withinStart, String.length(text));
+
+  withinBoth;
+};
```
glennsl: We have a utility function for this already:

https://github.com/onivim/oni2/blob/b7238298362d4ecce8402739fc2e377e043edf11/src/Core/Utility/IntEx.re#L1-L11

src/Components/InputModel.re
fanantoxa: Done

src/Components/Selection.re
```diff
@@ -0,0 +1,53 @@
...
+  abs(selection.focus - selection.anchor);
+};
+
+let rangeStart = (selection: t): int => {
+  min(selection.focus, selection.anchor);
+};
+
+let rangeEnd = (selection: t): int => {
+  max(selection.focus, selection.anchor);
+};
```
glennsl: Suggestion for better names, since I still don't understand what "range" means in this context. I use `Left` and `Right` instead of `Start` and `End` because the latter is at best inaccurate. The "start" of the selection is the anchor, and the "end" is the node.

```suggestion
let length = (selection: t): int => {
  abs(selection.focus - selection.anchor);
};

let offsetLeft = (selection: t): int => {
  min(selection.focus, selection.anchor);
};

let offsetRight = (selection: t): int => {
  max(selection.focus, selection.anchor);
};
```

src/Components/Selection.re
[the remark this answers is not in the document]
fanantoxa: Ok. I need a help here then.
So I don't what to do `open Selection`
And I don't really thing that `selection.Selection.anchor` will be better than  `Selection.anchor(selection)`
Could you please tell me more detailed how you see this, and how I can achieve it. 
(should easy but I'm still new to the language, so I can miss some concepts)

glennsl:
I'm off to be now though, so I won't be able to merge when the CI is done. Feel free to do it for me @bryphe.

glennsl:
🎉

bryphe:
Thanks for all the hard work on this, @fanantoxa !! (and all your help reviewing @glennsl )

Such a huge UX improvement 🥂 🎉

PR #1266, opened 2020-01-27
[docs] Add reminder to run `esy @test install` before `esy @test run`

glennsl:
Many seem understandably confused by the bad error message when running `esy @test run` without first running `esy @test install`, so a reminder in the docs seems like a good idea.


bryphe:
Thanks @glennsl  ! Ya, makes sense to clarify this 👍

PR #1267, opened 2020-01-27
[docs] Add installation page back in the sidebar

glennsl:
It was removed in #905 for non-obvious reasons, possibly unintentional.


bryphe:
Wow, thanks for catching this @glennsl !

> It was removed in #905 for non-obvious reasons, possibly unintentional.

Definitely unintentional. Glad you found it!

PR #1270, opened 2020-01-28
Key Displayer revamp

### LINKED ISSUE #1265: Key Displayer doesn't properly handle control keys
It displays `LEFT`, `RIGHT` like `T`, `t`.
Same issue for other control keys like Home, End, Backspace, etc.

Some screen record can be found here: 
https://github.com/onivim/oni2/pull/1264

glennsl:
Fixes #1265 and quite a few other issues and missing cases, and removes the use of `Actions.Tick` in favor of `useExpiration`.


src/Input/Parser.re
```diff
@@ -1,50 +1,69 @@
...
+        let buffer = Buffer.create(String.length(key));
+        let significand =
+          Str.matched_group(2, key) |> bracketize |> toFriendlyName;
+
+        let hasControl = Str.string_match(Regexes.control, key, 0);
+        let hasShift = Str.string_match(Regexes.shift, key, 0);
+        let hasAlt = Str.string_match(Regexes.alt, key, 0);
+        let hasCommand = Str.string_match(Regexes.command, key, 0);
+
+        if (hasCommand) {
```
bryphe: Cool! Thanks for improving the logic here

src/Model/KeyDisplayer.re
```diff
@@ -7,96 +7,97 @@
...
         ];
       }
     };
 
-  let ret = {...v, active: true, presses};
+  let ret = {...model, isActive: true, presses};
 
   // Also filter out old key presses, while we're here
-  update(time, ret);
+  collectGarbage(time, ret);
```
bryphe: I like this more explicit name better than `update` - for a second though I thought it was referring to the actual OCaml garbage collector though (ie, `GC.full_major`). Maybe `removeExpired` would be clearer?

src/Model/KeyDisplayer.rei
```diff
@@ -7,43 +7,27 @@
...
 type groupedPresses = {
-  exclusive: bool,
+  isExclusive: bool,
   time: float,
   keys: list(string),
 };
 
-type t;
+type t =
+  pri {
```
bryphe: Neat, I always forget about `pri`. Seems like a good use case here.

src/Store/KeyDisplayerConnector.re
```diff
@@ -10,36 +10,27 @@ module Model = Oni_Model;
...
-    | Actions.Tick(_) =>
-      if (KeyDisplayer.getActive(state.keyDisplayer)) {
-        let keyDisplayer = KeyDisplayer.update(getTime(), state.keyDisplayer);
-        let newState = {...state, keyDisplayer};
-        (newState, Isolinear.Effect.none);
-      } else {
-        (state, Isolinear.Effect.none);
-      }
-    | action =>
+    | Actions.Tick(_) => (state, Isolinear.Effect.none)
```
bryphe: I guess we can remove this `Actions.Tick` usage completely now?

bryphe:
Ah, I was testing this while recording some gifs and noticed a bug:

If I type `gg`, I see:
image

If I type `50`, I see:
image

So it seems like there is some difference in how it is batching up the inputs. I'd expect to see either `5` on one line, and then `0` on another, or `50` together - but not an extra `5`.

src/Model/KeyDisplayer.re
glennsl: Yeah that does sound better. Good suggestion!

glennsl:
Oh. Yeah the expiration hook does a structural  equality check and makes a copy of it to display, so if it changes it'll cause duplicates. We'll need some kind of unique identifier to deduplicate it then, but that should be easy enough to fix.

src/Store/KeyDisplayerConnector.re
glennsl: Done.

glennsl:
Grouping should be fixed, Tick has been removed and I've also moved KeyDisplayer to Components.

src/Store/StoreThread.re
```diff
@@ -209,9 +201,10 @@ let start =
     Features.updateSubscriptions(setup, newState, dispatch);
 
     onAfterDispatch(action);
-  };
+    runEffects();
+  }
 
-  let runEffects = () => {
+  and runEffects = () => {
```
glennsl: dispatching relied on `runEffects` being run on every tick, so when I removed the ticking I had to execute `runEffects` here instead. All tests pass, so I don't think anything relied on it being run continuously.

glennsl:
Oh, integration tests don't pass, but there's no error being printed so I guess it crashes? Struggling to get more information on it, but it looks like it's going into an infinite loop running the effects.

glennsl:
Going back to running effects on every tick seems to fix it, so I guess we're still doing that then.

src/Components/KeyDisplayer.re
```diff
@@ -0,0 +1,130 @@
...
+let initial: t = {groups: [], isEnabled: false};
+
+module Constants = {
+  let duration = 2.5;
+  let maxGroupingInterval = 0.3;
+};
+
+// UPDATE-ish
+
+let remvoeExpired = (time, model) => {
```
fanantoxa: @glennsl Could a typo in `remvoeExpired`

src/Input/Parser.re
```diff
@@ -1,50 +1,69 @@
...
+        if (hasAlt) {
+          Buffer.add_string(buffer, "Alt + ");
+        };
+        if (hasShift) {
+          Buffer.add_string(buffer, "Shift + ");
+        };
+
+        Buffer.add_string(buffer, significand);
+
+        Buffer.contents(buffer);
```
fanantoxa: @glennsl This actually might be `command` for mac, `Windows` or `Super` for Linux.
Best would be to have different names for all OS's but `Super` might fit all if not doable right now
(for ex. `Command` will hurt me a lot since I didn't use to work on Mac, it going to be misread as `Control`

Also would be better to have the save order of control keys as on:
https://github.com/onivim/oni2/blob/master/src/Input/Handler.re#L85

glennsl: Yeah, good point. We could switch this based on OS. But is the Windows key every spelled out entirely, or is it always just `Win`?

I think we should keep the ordering though, as this is much more natural. But perhaps we should change the order in `Handler`. I don't know if the order is significant for vim though.

src/Components/KeyDisplayer.re
glennsl: Indeed, good catch!

glennsl: Done.

src/Input/Parser.re
glennsl: And Done.

bryphe:
Just tried the latest - looks great! 🎉 Thanks @glennsl !

PR #1273, opened 2020-01-28
Fix #1272: Evaluate experimental.viml commands in correct order

### LINKED ISSUE #1272: viml commands are interpreted in reverse order
`experimental.viml` commands in `configuration.json` are processed in reverse order. This also happens with `files.exclude`.

e.g.

```json
{
  "experimental.viml": [
    "do this first",
    "do this second"
  ]
}
```

would be evaluated as `do this second`, `do this first`

tristil:
This was due to a bug in parseStringList in ConfigurationParser. This
also affected `files.exclude`.

### DIFF
--- a/src/Core/ConfigurationParser.re
+++ b/src/Core/ConfigurationParser.re
@@ -6,6 +6,7 @@
 open Kernel;
 open ConfigurationValues;
 open LineNumber;
+open Utility;
 
 let parseBool = json =>
   switch (json) {
@@ -29,13 +30,10 @@ let parseFloat = json =>
 let parseStringList = json => {
   switch (json) {
   | `List(items) =>
-    List.fold_left(
-      (accum, item) =>
-        switch (item) {
-        | `String(v) => [v, ...accum]
-        | _ => accum
-        },
-      [],
+    List.filter_map(
+      fun
+      | `String(v) => Some(v)
+      | _ => None,
       items,
     )
   | `String(v) => [v]
--- a/test/Core/ConfigurationParserTests.re
+++ b/test/Core/ConfigurationParserTests.re
@@ -195,6 +195,22 @@ describe("ConfigurationParser", ({test, describe, _}) => {
     };
   });
 
+  test("list of strings", ({expect}) => {
+    let configuration = {|
+     { "experimental.viml": ["first thing", "second thing", "third thing"] }
+     |};
+
+    switch (ConfigurationParser.ofString(configuration)) {
+    | Ok(v) =>
+      expect.list(Configuration.getValue(c => c.experimentalVimL, v)).toEqual([
+        "first thing",
+        "second thing",
+        "third thing",
+      ])
+    | Error(_) => expect.bool(false).toBe(true)
+    };
+  });
+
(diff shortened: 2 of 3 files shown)


src/Core/ConfigurationParser.re
```diff
@@ -32,7 +32,7 @@ let parseStringList = json => {
     List.fold_left(
       (accum, item) =>
         switch (item) {
-        | `String(v) => [v, ...accum]
+        | `String(v) => List.rev_append(accum, [v])
```
glennsl: This reverses `accum` every time you append to it, so while it'll work for two items, with more items it will alternately append to the beginning and end. It is also very inefficient.

I think this could be solved by just using `List.fold_right` instead of `List.fold_left`, but even just reversing it after the fold would be more efficient. But this can also just use `List.filter_map`, which would be much simpler.

glennsl: Something like 

```reason
List.filter_map(fun | `String(v) => Some(v) | _ => None, items);
```

src/Core/ConfigurationParser.rei
```diff
@@ -9,3 +9,5 @@ let ofJson: Yojson.Safe.t => result(Configuration.t, string);
 let ofString: string => result(Configuration.t, string);
 
 let ofFile: string => result(Configuration.t, string);
+
+let parseStringList: [> | `List(list([> | `String('a)])) | `String('a)] => list('a);
```
glennsl: Could this be tested without exposing internals? Seems like every configuration item is optional, so you could just pass in `experimental.viml` or `files.exclude` and then check the final parsed value, couldn't you?

src/Core/ConfigurationParser.re
tristil: thanks. resolved.

src/Core/ConfigurationParser.rei
tristil: You're right. tested at a higher level.

tristil:
@glennsl Ready for re-review

src/Core/ConfigurationParser.re
```diff
@@ -29,15 +30,7 @@ let parseFloat = json =>
...
-    List.fold_left(
-      (accum, item) =>
-        switch (item) {
-        | `String(v) => [v, ...accum]
-        | _ => accum
-        },
-      [],
-      items,
-    )
+      List.filter_map(fun | `String(v) => Some(v) | _ => None, items);
```
bryphe: Nice catch @tristil !

test/Core/ConfigurationParserTests.re
```diff
@@ -195,6 +195,22 @@ describe("ConfigurationParser", ({test, describe, _}) => {
     };
   });
 
+  test("list of strings", ({expect}) => {
```
bryphe: Thanks for adding the test case!

glennsl:
@tristil Could you run `esy format`, commit and push the changes? That'll fix the CI failure

tristil:
@glennsl done

glennsl:
All green. Thanks @tristil !

PR #1277, opened 2020-01-30
Dependency - Extensions (NPM): Add 'which' and other deps for vscode git

bryphe:


bryphe:
> It just occurred to me why this wasn't a problem for me. which is a Windows polyfill for what is a standard utility on Unix.

It is weird... I was actually testing on OSX (that's where that first `log.txt` was from). I'm not sure why `which` is required there. But after I added `which` - there were a few other dependencies needed that I also added here.

I didn't actually get a chance to test this on Windows yet - I'll add an integration test to just validate the `vscode.git` extension gets activated. This will ensure that it has all the deps cross-platform.

bryphe:
Added test, and it passed! Need to test out the diff markers on Windows now...

PR #1279, opened 2020-01-30
Editor refactor - Part 1: Feature project

glennsl:
This creates two new feature projects, `Feature_LanguageSupport` and `Feature_Editor` with associated test projects, and moves a few other modules around to untangle the dependencies. `EditorSurface` has also been parameterized with a bunch of new options in order to get rid of the dependency on `State.t` and other app-specific functionality.

TODO:
- [x] Delete a few remaining modules from Oni_Model that have been copied to Feature_Editor
- [x] Move tests to separate test projects

Next steps:
- Move reducers and stores/connectors to feature projects
- Refactor the god function in EditorSurface
- Abstract away the implementation-specific models from LanguageSupport
- Create an `OniEditor` module in `UI` to abstract away all the configuration fetching and other setup.


src/UI/EditorGroupView.re
```diff
@@ -115,17 +117,122 @@ let make = (~state: State.t, ~windowId: int, ~editorGroup: EditorGroup.t, ()) =>
...
+            completions={state.completions}
+            tokenTheme={state.tokenTheme}
+            definition={state.definition}
+            shouldRenderWhitespace
+            showMinimapSlider
+            hoverDelay
+            isHoverEnabled
+            shouldRenderIndentGuides
+            shouldHighlightActiveIndentGuides
+          />;
```
glennsl: A few extra parameters have been added to get rid of the dependency on `State.t`. These are mostly configuration values, fetched above, and some callbacks to dispatch messages.

src/Model/EditorReducer.re
```diff
@@ -0,0 +1,108 @@
+open Oni_Core;
```
glennsl: This was extracted from `Editor` because it depends on `Actions`, but should be moved back once it has its own `msg` type, along with a bunch of other reducers and stores/connectors.

src/UI/EditorGroupView.re
bryphe: Cool! Thanks for making these dependencies explicit. This is a great step forward to a more-generally-reusable editor component.

PR #1282, opened 2020-02-01
SCM: File Explorer decorations

glennsl:
2020-02-01-124912_177x135_scrot

This adds Source Control decorations to the file explorer, ie. embellishments showing the source control resource state of files. It also adds some list theme colors and uses those properly.

There are some uncertainties here though:

I'm not sure how vscode updates the SCM. Currently we execute `git.refresh` on `BufferUpdate`, which isn't great because it happens too frequently and of course only works for git. There is unfortunately no save action/event though, so `BufferUpdate` seems like the best approximation for now.

I'm also not sure what some of the values we get actually are. The type of the payload as defined in the protocol, of which I have yet to figure out the first two, is:

```
export type DecorationData = [number, boolean, string, string, ThemeColor, string];
```

I don't know how to handle multiple decorations for the same file. Currently we just pick the last one, and I think it would be rare to have multiple as this mechanism seems to just be used for SCM. But the git extension itself registers two decoration providers, so who knows...

TODO:
- [ ] Add "letter" indicator, as in vscode (should we spend real estate on this?)
- [ ] Add tooltip
- [ ] Batch decoration requests?
- [ ] Add proper support for custom theme colors instead of hard-coding them (probably a separate PR as part of a Theme revamp)

I don't think any of the above are blockers, so this can be merged as-is.

### DIFF
--- a/src/Core/Theme.re
+++ b/src/Core/Theme.re
@@ -76,6 +76,12 @@ type t = {
   editorFindMatchHighlightBackground: Color.t,
   editorIndentGuideBackground: Color.t,
   editorIndentGuideActiveBackground: Color.t,
+  listActiveSelectionBackground: Color.t,
+  listActiveSelectionForeground: Color.t,
+  listFocusBackground: Color.t,
+  listFocusForeground: Color.t,
+  listHoverBackground: Color.t,
+  listHoverForeground: Color.t,
   menuBackground: Color.t,
   menuForeground: Color.t,
   menuSelectionBackground: Color.t,
@@ -140,6 +146,12 @@ let default: t = {
   editorOverviewRulerBracketMatchForeground: Color.hex("#A0A0A0"),
   editorActiveLineNumberForeground: Color.hex("#737984"),
   editorSelectionBackground: Color.hex("#687595"),
+  listActiveSelectionBackground: Color.hex("#495162"),
+  listActiveSelectionForeground: Color.hex("#FFFFFF"),
+  listFocusBackground: Color.hex("#495162"),
+  listFocusForeground: Color.hex("#FFFFFF"),
+  listHoverBackground: Color.hex("#495162"),
+  listHoverForeground: Color.hex("#FFFFFF"),
   scrollbarSliderBackground: Color.rgba(0., 0., 0., 0.2),
   scrollbarSliderActiveBackground: Color.hex("#2F3440"),
   editorIndentGuideBackground: Color.hex("#3b4048"),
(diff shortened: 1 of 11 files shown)


src/Core/Theme.re
```diff
@@ -431,3 +473,16 @@ let getColorsForMode = (theme: t, mode: Vim.Mode.t) => {
...
+  | "gitDecoration.addedResourceForeground" => Some(Color.hex("#81b88b"))
+  | "gitDecoration.modifiedResourceForeground" => Some(Color.hex("#E2C08D"))
+  | "gitDecoration.deletedResourceForeground" => Some(Color.hex("#c74e39"))
+  | "gitDecoration.untrackedResourceForeground" => Some(Color.hex("#73C991"))
+  | "gitDecoration.ignoredResourceForeground" => Some(Color.hex("#8C8C8C"))
+  | "gitDecoration.conflictingResourceForeground" =>
+    Some(Color.hex("#6c6cc4"))
+  | "gitDecoration.submoduleResourceForeground" => Some(Color.hex("#8db9e2"))
+  | _ => None
+  };
```
glennsl: These are currently hard-coded. We need a mechanism for contributed theme colors

src/Extensions/ExtHostClient.re
```diff
@@ -298,6 +336,46 @@ let provideCompletions = (id, uri, position, client) => {
...
 };
 
+let provideDecorations = (handle, uri, client) => {
+  let decodeItem =
+    fun
+    | (
+        _requestId,
+        `List([
+          `Int(_),
+          `Bool(_),
```
glennsl: I don't know what these are yet. But unless there's some actual functionality we're missing, we might as well just ignore them.

src/Model/FileExplorer.re
```diff
@@ -8,7 +8,8 @@ type t = {
   isOpen: bool,
   scrollOffset: [ | `Start(float) | `Middle(float) | `Reveal(int)],
   active: option(string), // path
-  focus: option(string) // path
+  focus: option(string), // path
+  decorations: StringMap.t(list(SCMDecoration.t)),
```
glennsl: decorations are stored directly on the `fileExplorer` in a `StringMap` that maps paths to decorations instead of on the nodes themselves, since the nodes might not have been loaded yet! It's also easier to update them when you don't have to search a tree that may or may not be there etc.

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -526,7 +550,10 @@ let start = (extensions, setup: Setup.t) => {
 
     | Actions.BufferUpdate(bu) => (
         state,
-        modelChangedEffect(state.buffers, bu),
+        Isolinear.Effect.batch([
+          modelChangedEffect(state.buffers, bu),
+          executeContributedCommandEffect("git.refresh"),
+        ]),
       )
```
glennsl: `git`refresh` hard-coded on `BufferUpdate`. It works, for now.

src/Model/FileExplorer.re
bryphe: Thanks for the explanation! Makes sense 👍

src/UI/FileTreeView.re
```diff
@@ -126,15 +157,19 @@ let make =
...
+
         <nodeView
           isFocus={Some(node.path) == focus}
           isActive={Some(node.path) == active}
           font
           theme
           node
-        />
-      }
+          ?decorations
```
bryphe: Neat, always forget about this syntax

bryphe:
> There is unfortunately no save action/event though, so BufferUpdate seems like the best approximation for now.

Good point - we should have a `BufferSaved` for sure.

> Add tooltip

Cool! This would be useful all over our UI...

> Add proper support for custom theme colors instead of hard-coding them (probably a separate PR as part of a Theme revamp)

Seems like you've already done some thinking about this. It would be nice to not have to re-parse the colors every render!

Looks great @glennsl - the SCM story is really coming together 🔥 🔥  Thank you!

PR #1292, opened 2020-02-04
Fix #1238 - File Explorer (Windows): Nodes not expanding

### LINKED ISSUE #1238: File Explorer - Windows: FocusNodeLoaded gets in infinite loop
When opening a file, it seems the 'FocusNodeLoaded' action gets in an infinite loop (the logic that expands out the focused node in the file explorer).

Simple repro on Windows:
- Open Onivim 2
- Open Quickopen (`Control+p`)
- Type `Notification` 
- Open `Notification.re`

Here's a gif of the repro:
editor

I suspect that the logic may not be accounting for the pathing in Windows - ie, the different path characters (`/` vs `\`). 

I also wonder if there should be some resiliency - ie, try X times to focus the node, and stop on failure. 

The editor is still usable, because the event is dispatched every frame, but it triggers a re-render every time which will hurt performance / battery life.

@glennsl - do you have any pointers on where to start looking at this?

bryphe:
Fix #1238 

The `Path.relativeAt` API wasn't correctly handling Windows-style root paths (ie, `D:\`) - this adds some test cases to exercise it and a fix.


test/Core/Utility/PathTests.re
```diff
@@ -0,0 +1,73 @@
...
+    expect.string(Path.toRelative(~base="/", "/Applications/Onivim2.app")).
+      toEqual(
+      "Applications/Onivim2.app",
+    );
+
+    // Windows-style paths
+
+    expect.string(Path.toRelative(~base="D:", "D:\\Onivim2")).toEqual(
+      "Onivim2",
+    );
```
glennsl: How does this pass CI? It fails for me locally because `Filename.dir_sep` is `/`, so it tried to remove `D:/`, which obviously doesn't work, so it returns the original string. This test should fail on anything other than Windows.

bryphe: Hmm, good question, I'm looking into this now. I figured since it was green it was good, but I can repro the failure locally too.

bryphe: I added a new test project as part of that change, and I forgot that there is a _separate_ test project I needed to add it for CI. So `esy @test run` fails, but `esy @test run-ci` passes, since it wasn't running that test...

Opened a PR here: https://github.com/onivim/oni2/pull/1296

bryphe: Sorry you hit this @glennsl ! Would've been confusing on top of the 4.08 changes... hope it didn't take too much time from you.

glennsl: No worries. The change was small and very recent, so didn't take much to figure out at all.

PR #1296, opened 2020-02-05
Bugfix: Fix Path.toRelative unit test failure

bryphe:


src/Core/Utility/Path.re
```diff
@@ -11,8 +11,8 @@ let hasTrailingSeparator = path => {
   };
 };
 
-let toRelative = (~base, path) => {
```
glennsl: This is code added just for testing, with a default that will not be tested. Wouldn't it be better make the tests platform-specific so we test all the actual code paths on all relevant platforms instead?

bryphe: That's a reasonable concern - made the suggested change in the latest comment

PR #1306, opened 2020-02-07
Experiment: Editor smooth scroll

bryphe:
This was an experiment I had for recording some gifs - using our spring hook to implement a smooth scroll on the editor surface (like we have for our file explorer). This is gated by an `experimental.editor.smoothScroll` configuration flag.

With `experimental.editor.smoothScroll` set to `true` in this PR:
smooth-scroll3


src/Feature/Editor/EditorSurface.re
```diff
@@ -302,9 +306,24 @@ let%component make =
...
+    Hooks.spring(
+      ~target=editor.scrollY,
+      ~restThreshold=100.,
+      ~enabled=smoothScroll,
+      scrollSpringOptions,
+    );
+  let%hook (scrollX, _setScrollXImmediately) =
+    Hooks.spring(
+      ~target=editor.scrollX,
+      ~restThreshold=100.,
```
bryphe: I used a high `restThreshold` to prevent a disconcerting 'jiggling' effect when the spring would be settling at the end of the animation. The font rendering would clamp to the nearest vertical pixel, with some lines being offset and some not transiently.

bryphe:
@glennsl , I'm happy to wait to merge this until after your `EditorSurface` work is done - I don't want you to have to keep dealing with merge conflicts. Let me know what you'd prefer!

bryphe:
> Tested it out, and paging up and down, jumping to line and such is very nice, but scrolling with the mousewheel is really jerky. And when the spring is running the cursor is kind of "stuck", not scrolling with the content.

Good points! The cursor being 'stuck' is fixed now. There's still the problem with the mousewheel and mouse scrolling - I think the best bet would be to disable this when the scroll gesture came from the mouse.

I'll bring it in gated by the 'experimental' configuration setting - before we make it first-class we'll need to address these mouse issues 👍

PR #1308, opened 2020-02-07
SCM: Initial Pane Experience

glennsl:
2020-02-07-224422_267x215_scrot
Currently just lists the resource in each group. Clicking on a resource opens the file, and it updates on every `BufferUpdate`, as with the file explorer decorations. It does not properly handle multiple providers/repositories yet though.

This also adds hover styles and active indicators to the dock items.

### DIFF
--- a/src/Components/FontIcon.re
+++ b/src/Components/FontIcon.re
@@ -8,6 +8,19 @@ open Revery.UI;
 open Oni_Core.CamomileBundled.Camomile;
 module ZedBundled = Oni_Core.ZedBundled;
 
+module Styles = {
+  open Style;
+
+  let text = (~fontFamily, ~fontSize, ~color, ~backgroundColor, ~margin) => [
+    Style.fontFamily(fontFamily),
+    Style.fontSize(fontSize),
+    Style.color(color),
+    Style.backgroundColor(backgroundColor),
+    Style.margin(margin),
+    textWrap(Revery.TextWrapping.NoWrap),
+  ];
+};
+
 let codeToIcon = icon => ZedBundled.singleton(UChar.of_int(icon));
 
 let make =
@@ -22,14 +35,11 @@ let make =
     ) =>
   <Text
     text={codeToIcon(icon)}
-    style=[
-      Style.fontFamily(fontFamily),
-      Style.fontSize(fontSize),
-      Style.color(color),
-      Style.backgroundColor(backgroundColor),
-      Style.margin(margin),
-      Style.height(fontSize |> int_of_float),
-      Style.width(fontSize |> int_of_float),
-      Style.textWrap(Revery.TextWrapping.NoWrap),
-    ]
+    style={Styles.text(
+      ~fontFamily,
+      ~fontSize,
+      ~color,
+      ~backgroundColor,
+      ~margin,
+    )}
   />;
--- a/src/Components/FontIcon.re
+++ b/src/Components/FontIcon.re
@@ -11,12 +11,10 @@ module ZedBundled = Oni_Core.ZedBundled;
 module Styles = {
   open Style;
 
-  let text = (~fontFamily, ~fontSize, ~color, ~backgroundColor, ~margin) => [
+  let text = (~fontFamily, ~fontSize, ~color) => [
     Style.fontFamily(fontFamily),
     Style.fontSize(fontSize),
     Style.color(color),
-    Style.backgroundColor(backgroundColor),
-    Style.margin(margin),
     textWrap(Revery.TextWrapping.NoWrap),
   ];
 };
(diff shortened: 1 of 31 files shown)


src/UI/SCMPane.re
```diff
@@ -0,0 +1,118 @@
...
+  let%hook (isHovered, setHovered) = Hooks.state(false);
+  let onMouseOver = _ => setHovered(_ => true);
+  let onMouseOut = _ => setHovered(_ => false);
+
+  let base =
+    Option.first_some(
+      Option.map(provider.rootUri, ~f=Uri.toFileSystemPath),
+      Option.map(workspace, ~f=w => w.workingDirectory),
+    )
+    |> Option.value(~default="/");
```
glennsl: Tried out Base a bit here. The labelled argument convention is nice, as it allowed putting the main argument first, which helps the type inference so the function argument doesn't have to be annotated. Otherwise I found the available functions a bit lacking compared to containers. `first_some` is not as nice as `choice` and `or_` in containers for example.

src/UI/SCMPane.re
```diff
@@ -0,0 +1,118 @@
...
+  let {theme, uiFont as font, workspace, _}: State.t = state;
+
+  let groups = {
+    open Base.List.Let_syntax;
+
+    let%bind provider = state.scm.providers;
+    let%bind group = provider.resourceGroups;
+
+    return((provider, group));
+  };
```
glennsl: Base does not support the 4.08 binding operators, but does have its own let syntax ppx, `ppx_let`, which I much prefer as it uses actual words instead of cryptic symbols.

src/Components/FontIcon.re
```diff
@@ -8,28 +8,22 @@ open Revery.UI;
...
   <Text
     text={codeToIcon(icon)}
-    style=[
-      Style.fontFamily(fontFamily),
-      Style.fontSize(fontSize),
-      Style.color(color),
-      Style.backgroundColor(backgroundColor),
-      Style.margin(margin),
-      Style.height(fontSize |> int_of_float),
-      Style.width(fontSize |> int_of_float),
```
glennsl: Removed these, `backgroundColor` is no longer needed, `width` and `height` messed with centering as it wasn't possible to control the alignment of the icon within this box, and `margin` doesn't really belong here (any more than any other style at least) and can be added when needed just by wrapping it in another view.

bryphe:
This looks awesome, @glennsl !

> This also adds hover styles and active indicators to the dock items.

So cool. Great addition!

src/UI/SCMPane.re
bryphe: Sounds good! I have some experience with this from `esy`, which also used `ppx_let` extensively (and had `Let_syntax` for `result` and `option` types).

src/UI/SCMPane.re
bryphe: > The labelled argument convention is nice, as it allowed putting the main argument first, which helps the type inference so the function argument doesn't have to be annotated.

Oh, interesting. That makes sense.

>  Otherwise I found the available functions a bit lacking compared to containers. first_some is not as nice as choice and or_ in containers for example.

That's too bad. I like the ergonomics of `or_` and `choose` as well. I guess we could either switch to containers, or still have some `OptionEx` wraps around the `base` methods. Thanks for taking it for a test run.

src/Components/FontIcon.re
bryphe: Glad to see these go away!

src/Core/Utility/ListEx.re
```diff
@@ -50,3 +50,21 @@ let rec sublist = (beginning, terminus, l) =>
       [h, ...tail];
     };
   };
+
+// Like Array.prototype.splice in JavaScript, but for lists
+let splice = (~start, ~deleteCount, ~additions, source) => {
```
bryphe: Cool!

src/UI/Dock.re
```diff
@@ -1,11 +1,50 @@
...
+    ),
+    backgroundColor(
+      isHovered ? theme.menuSelectionBackground : Colors.transparentWhite,
+    ),
+  ];
+};
+
+let%component item = (~onClick, ~theme: Theme.t, ~isActive, ~icon, ()) => {
+  let%hook (isHovered, setHovered) = Hooks.state(false);
+  let onMouseOver = _ => setHovered(_ => true);
```
bryphe: Thanks for adding this. Levels up the UX!

test/Core/Utility/UtilityTests.re
```diff
@@ -2,36 +2,213 @@ open TestFramework;
...
+        toEqual([
+        1,
+        9,
+        2,
+        3,
+      ])
+    );
+
+    test("replace first", ({expect}) =>
+      expect.list(
```
bryphe: Nice set of test cases!

Issue #1157, opened 2020-01-03
Extensions - Merge user configurations


bryphe:
We've done work to set up extension configuration (#1106, #1111, #1119), but we're missing the last step of merging the user's configuration settings with the extension defaults.

This is blocking the Rust extension (and likely several others that require a configured toolchain, like the C++ extension)

Reported from Discord: https://discordapp.com/channels/417774914645262338/424012160696582144/662783028912717824

glennsl:
Should be fixed by https://github.com/onivim/oni2/pull/1489

PR #1320, opened 2020-02-10
Developer: Add command to show the list of workspaceFolders

bryphe:
This adds a command to show the list of `vscode.workspace.workspaceFolders`, for debugging (to show the workspace info from the perspective of a plugin):
workspaces

FYI @glennsl - added this based on our convo today.

I took a look at the git extensions, and I think this is an interesting spot:
https://github.com/onivim/vscode-exthost/blob/8ebec0403c19a038932368df8c89e4676e923dd2/extensions/git/src/model.ts#L69

It uses the following APIs:
- `vscode.workspace.onDidChangeWorksapceFolders`, which we should support today
- `vscode.window.onDidChangeVisibleTextEditors`, which isn't currently supported (this might be how it decides to refresh?)
- `vscode.workspace.onDidChangeConfiguration` - not yet supported (planning on implementing this with #1157 )
- `vscode.workspace.createFilesystemWatcher` - not implemented yet (another case where we need `esy-fswatch`/`reason-fswatch`)

### DIFF
--- a/development_extensions/oni-dev-extension/extension.js
+++ b/development_extensions/oni-dev-extension/extension.js
@@ -80,7 +80,7 @@ function activate(context) {
            latestText = e.document.getText().split(os.EOL).join("|");
         }
 
-		//vscode.window.showInformationMessage('Changed!');
+        //vscode.window.showInformationMessage('Changed!');
         const document = e.document;
         if (document && path.basename(document.uri.fsPath) == "test.oni-dev") {
            collection.set(document.uri, [{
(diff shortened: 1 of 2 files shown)


glennsl:
The fswatcher is especially interesting.  Perhaps it just refreshes itself, and the behaviour seen with refresh only happening when the window gets focus is just because vscode for some reason holds on to the fswatch events until it does.

bryphe:
> Perhaps it just refreshes itself, and the behaviour seen with refresh only happening when the window gets focus is just because vscode for some reason holds on to the fswatch events until it does.

Hmm, maybe. I'm also curious about the `vscode.window.onDidChangeVisibleTextEditors` - it's possible that gets dispatched anytime an editor changes state:https://code.visualstudio.com/api/references/vscode-api#window.visibleTextEditors

PR #1324, opened 2020-02-11
Editor Refactor - Part 2

glennsl:
- Splits EditorSurface up into separate modules
- Extracts some common drawing primitives in `Draw`
- Gathers common rendering information in `Draw.context`

Paints object are currently just allocated wherever needed to get the thing to work. It might be a good idea to allocate one per primitive instead, to be used across calls and updated with properties passed in as plain function arguments.


src/Feature/Editor/IndentLineRenderer.re
```diff
@@ -65,34 +67,28 @@ let paint = Skia.Paint.make();
...
 
   let indentationWidthInPixels =
-    float_of_int(indentationSettings.tabSize) *. fontWidth;
+    float(indentationSettings.tabSize) *. context.charWidth;
 
-  let l = ref(startLine);
-
-  while (l^ < endLine) {
-    let line = l^;
+  for (line in startLine to endLine - 1) {
```
glennsl: I've swapped out a few while loops that just looped over a counter to use for loops instead, since that's kind of their thing.

glennsl: There's still some while loops below here though, which uses a bunch of unnecessary `ref`s. The idiomatic way of doing a loop that aborts on a certain condition is to raise an exception when the condition is met. Which might seem silly, but exceptions are really cheap in OCaml.

src/Feature/Editor/Draw.re
```diff
@@ -7,18 +7,47 @@ open Oni_Core;
...
+    (
+      ~canvasContext,
+      ~width,
+      ~height,
+      ~scrollX,
+      ~scrollY,
+      ~lineHeight,
+      ~editorFont: EditorFont.t,
+    ) => {
+  let font = Revery.Font.load(editorFont.fontFile) |> Stdlib.Result.get_ok;
```
glennsl: Shouldn't `EditorFont.t` contain the font instead of the font file path? This really isn't the right place to deal with font fallback and such.

bryphe: Definitely, that would be a nice simplification here.

src/Feature/Editor/Draw.re
```diff
@@ -0,0 +1,250 @@
...
+      );
+    let (endOffset, _) =
+      BufferViewTokenizer.getCharacterPositionAndWidth(
+        ~viewOffset=leftVisibleColumn,
+        text,
+        endC,
+      );
+    let length = max(float(endOffset - startOffset), 1.0);
+
+    let paint = Skia.Paint.make();
```
bryphe: I believe this will regress some of the perf work we did for Skia. Allocations like this end up pretty expensive (they build up promoted memory and make more work for the GC).

To see the impact - we have a `drawText` benchmark in Revery:
```
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|  BENCHMARK                                                     |  ITERATIONS |  TIME               |  MINOR GC  |  MAJOR GC  |  MINOR ALLOC  |  PROMOTED  |  MAJOR ALLOC  |
|----------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  Draw: drawText                                                |  10000      |  0.110551118851     |  1         |  0         |  230784       |  25        |  25           |
|----------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
```

This is with a pre-allocated `Skia.Paint.make`, and then if we allocate one every time:
```
|----------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|
|  Draw: drawText                                                |  10000      |  0.118616104126     |  2         |  0         |  320790       |  74165     |  74165        |
|----------------------------------------------------------------+-------------+---------------------+------------+------------+---------------+------------+---------------|

```

The promoted memory goes from `25` -> `74165`, and it's about a 9% increase in time. It adds up in our 'hot paths'.

Definitely going to have to do another round of perf measurement / optimization after this. With Skia, I can at least put together an end-to-end benchmark which will help measure and guide optimizations.

src/Feature/Editor/IndentLineRenderer.re
bryphe: The `for` loops are cleaner, thanks for switching!

src/Feature/Editor/GutterView.re
```diff
@@ -0,0 +1,201 @@
...
+        *. context.charWidth
+        /. 2.;
+
+  let lineNumberTextColor =
+    isActiveLine
+      ? theme.editorActiveLineNumberForeground
+      : theme.editorLineNumberForeground;
+  let paint = Skia.Paint.make();
+  Skia.Paint.setColor(paint, Revery.Color.toSkia(lineNumberTextColor));
+  Skia.Paint.setTextEncoding(paint, Utf8);
```
bryphe: These calls are actually a bit expensive since they haven't been optimized (ie, with `[@noalloc]`), which is why they were factored out previously

src/Feature/Editor/EditorSurface.re
```diff
@@ -356,641 +166,97 @@ let%component make =
...
-          width={int_of_float(layout.bufferWidthInPixels)}
-          theme
-        />
-      </View>
-    </View>
-    minimapLayout
-    hoverElements
-    <View style=verticalScrollBarStyle>
+  let (gutterWidth, gutterView) =
+    <GutterView
```
bryphe: Thanks for factoring this `<GutterView />` out! Much cleaner to have it separate. I think this can actually help us chip away at the `EditorLayout` too.

glennsl:
Yeah, I am somewhat aware of these performance considerations. Which is why I suggested moving the paint object allocations to the primitives and reusing them across calls. That would make the optimization an implementation detail, as it should be, instead of adding another layer of complexity to this.

But I wasn't sure if that was the only concern, and as you point the "setter" functions aren't optimized either. That could be alleviated just by adding more primitives though.

bryphe:
> Which is why I suggested moving the paint object allocations to the primitives and reusing them across calls

You mean the primitives in `Draw.re`, right? Like `underline`, `ruler`, etc? It seems like we're part way there, as we do allocate the paint to those primitives.... we just don't reuse them across calls - that would be easy to fix.

> But I wasn't sure if that was the only concern, and as you point the "setter" functions aren't optimized either.

Ya, the `paint` for the rectangles (`underline`, `drawRect`, `range`) is simple - we only ever change the color. But for text, there are several other attributes we set (encoding, antialias, lcd render text, size, typeface, etc). It'd be ideal for performance if we didn't have to set those for every `Draw.text` call. 

> That could be alleviated just by adding more primitives though.

Cool, it would be great if we didn't have to make that trade-off for `drawText`. Do you have an example of what you have in mind? I can see how we could limit the setting of `setTextEncoding`, `setLcdRenderText`, and `setAntiAlias` with this approach, but it seems like we'd still need to set the text size / typeface more often than we do today.

glennsl:
> Do you have an example of what you have in mind? 

I was thinking just `Draw.utf8Text` and `Draw.glyphText`, but it could of course be more granular if needed. That would make it much less error-prone as well. `fontSize` and `fontFamily` could be "memoized" in a sense. If it hasn't changed since last time, don't set it again.

Alternatively we could attach the paint objects to the `context`, since that's what specifies the `editorFont` and `fontSize`. or we could provide an equivalent of vim's `.`, "repeat the last operation".

There's lots of possibilities, It doesn't seem like a great idea to optimize what can and should be done at a lower level (e.g. with `[@noalloc]`, if possible) though. But if we still do, I think we should also consider how easy it would be to undo.

bryphe:
> It doesn't seem like a great idea to optimize what can and should be done at a lower level (e.g. with [@noalloc], if possible) though.

I agree... as long as we are actually implementing the optimizations. Otherwise - how can we be sure we're correct in our assumption that it can and should be done in another layer? My concern is that, if our assumptions are wrong - we could design ourselves into a slow-performing corner, and end up having to re-work it anyway.

Performance is a feature - an important feature for Onivim 2. And we do still have a lot of work to get to where we need to be. Would prefer not to regress perf if we can help it.

glennsl:
Sure, but it doesn't seem like this would make it any harder to get there if absolutely needed. And if it is needed, the effort required to re-work it both now and at that point still seems small compared to the overhead of adding layers of complexity by default. But I might have misunderstood something.

I would think encapsulating the optimizations in the primitives would get us at least most of the way there anyway though.

glennsl:
I also thought the `[@noalloc]` optimization was more certain though.

bryphe:
> Sure, but it doesn't seem like this would make it any harder to get there if absolutely needed. And if it is needed, the effort required to re-work it both now and at that point still seems small compared to the overhead of adding layers of complexity by default. But I might have misunderstood something.

We do need it, now. Otherwise when do we start caring about perf? See #1327 - we're still not meeting our performance goals (my goal is a consistent ~__8ms__ render time - giving us plenty of time to do other stuff, like diffs, syntax highlights, interop with exthost, etc - and meet the 60FPS/16ms goal).

A critical aspect of getting that predictable performance is to minimize allocations and removing unnecessary overhead in our hot paths (like the editor surface).

This change takes us a bit of a step back:

__Before this change:__
```
|  EditorSurface - Rendering: Minimal state                           |  10         |  0.18224811554      |  6261      |  0         |  8818099      |  77915      |  2156235      |
|  EditorSurface - Rendering: 1000 Lines state                        |  10         |  0.169266939163     |  6261      |  0         |  8818099      |  77915      |  2156235      |
|  EditorSurface - Rendering: 1000 Lines state - With indent guides   |  10         |  0.171715974808     |  6261      |  0         |  8818169      |  77915      |  2156235      |
|  EditorSurface - Rendering: 100000 Lines state                      |  10         |  0.166203022003     |  6261      |  0         |  8818099      |  77915      |  2156235      |
|  EditorSurface - Rendering: 100000 Lines state - No minimap         |  10         |  0.0929520130157    |  1451      |  0         |  2394109      |  38715      |  527365       |
```

__After this change:__
```
|  EditorSurface - Rendering: Minimal state                           |  10         |  0.243696928024     |  6261      |  0         |  9235343      |  354105     |  2432425      |
|  EditorSurface - Rendering: 1000 Lines state                        |  10         |  0.187000989914     |  6261      |  0         |  9027766      |  203143     |  2281463      |
|  EditorSurface - Rendering: 1000 Lines state - With indent guides   |  10         |  0.194977998734     |  6261      |  0         |  9027836      |  203143     |  2281463      |
|  EditorSurface - Rendering: 100000 Lines state                      |  10         |  0.181714057922     |  6261      |  0         |  9027766      |  203143     |  2281463      |
|  EditorSurface - Rendering: 100000 Lines state - No minimap         |  10         |  0.0943629741669    |  1451      |  0         |  2600276      |  162384     |  651034       |
```

The timing measurements aren't super useful (for the reasons you called out in #1328), but the allocations are a good proxy for the runtime performance. This adds around 5% in terms of our minor allocations, and a significant increase in our promoted bytes (2-4x depending on the benchmark). The allocations, in general, tend to be a good proxy for performance - ie, if we reduce allocations by 5%, we get ~5% better render time in real-world testing, and vice versa.

Here's the memory profile pulled out before/after:

| Benchmark | Minor alloc | Promoted alloc | Major alloc |
| --- | --- | --- | --- |
| EditorSurface - Minimal State - Before | 8818099      |  77915      |  2156235 |
| EditorSurface - Minimal State - After |  9235343 __ (~5% increase)__    |  354105 __(350% increase)__     |  2432425  __(13% increase)__    |

I think we could get pretty far by just caching the SKPaint objects... I don't think that would add significant complexity or maintainability cost either. I'll open a PR to try this out and compare the results.

bryphe:
Opened this PR to experiment with caching paints. Adding some perf numbers there too: https://github.com/glennsl/oni2/pull/4

glennsl:
> Otherwise when do we start caring about perf?

We always do, but this can't be the only consideration. If there are alternatives which do significantly better for other considerations while only marginally worse, if at all, in terms of perf, I think that should be tried first. As do you it seems, so I'm not entirely sure what we're arguing about.

> The allocations, in general, tend to be a good proxy for performance - ie, if we reduce allocations by 5%, we get ~5% better render time in real-world testing, and vice versa.

Thanks for explaining. But is that for minor allocations, major allocations, either or both? And what's the significance of the promotion measure? (I understand it's a measure of allocations moved from the minor to major heap, but not the implications of that in terms of performance.)

> I think we could get pretty far by just caching the SKPaint objects like you called out... I don't think that would add significant complexity or maintainability cost either. I'll open a PR to try this out and compare the results.

Great! Yeah I don't think it does either.

bryphe:
> We always do, but this can't be the only consideration. If there are alternatives which do significantly better for other considerations while only marginally worse, if at all, in terms of perf, I think that should be tried first. As do you it seems, so I'm not entirely sure what we're arguing about

Ya, totally, I do think we're on the same page here 😄 

> But is that for minor allocations, major allocations, either or both? And what's the significance of the promotion measure? 

So for some extra context - when I profile, it's often the case that the time we spend in the garbage collector is at the top of our trace.

image

This means we're allocating way too much as we render and giving the GC a heavy workload. This is probably why measuring allocations is a good proxy for us; it is one of the main bottlenecks for perf and so reducing the overhead of the GC by minimizing allocations directly correlates to reducing that bottleneck. 

It can be difficult to completely separate minor/promoted/major, because in general, if you reduce minor allocations, then the promoted+major will go down too. Minor allocations get 'promoted' after surviving multiple rounds of minor allocations. In other words - if there are lots of minor allocations, triggering lots of collections, then there are more promotions. And minor collections tend to be frequent.
 
But my understanding is that:
- Minor allocations and minor GC are the cheapest
- There is a cost to promoting from minor -> major (moving memory)
- The major collections are more expensive

For optimization - I just focus on reducing allocations during runtime, when possible / in a critical path. This reduces minor, promoted, and major together.

In the trace I grabbed, expanding the GC node:
image

It seems we are actually spending more time in minor collections than major slices. I think that `caml_empty_minor_heap` actually does promotion too (it calls some methods like `caml_oldify`)

> Great! Yeah I don't think it does either.

Excellent! Seems like we're both happy with that solution.

glennsl:
```
|  EditorSurface - Rendering: 100000 Lines state                      |  10         |  0.166203022003     |  6261      |  0         |  8818099      |  77915      |  2156235      |
|  EditorSurface - Rendering: 100000 Lines state - No minimap         |  10         |  0.0929520130157    |  1451      |  0         |  2394109      |  38715      |  527365       |
```

Am I reading this wrong, or is the minimap allocating 2-3 times more than the rest of the editor? Because that seems a bit nuts...

glennsl:
I did a bit of refactoring on top of your PR, which shouldn't have affected the allocations much since they're just moved around, but for some reason moving the shaping into `drawShapedText` in https://github.com/onivim/oni2/pull/1324/commits/386c02f06089575f20bbabdf613866f9675684d4 adds 0.5% to minor allocations. And I'm struggling to see why, since it looks to me like ti does exactly the same work the same number of times, just in a different function.

For reference, the starting point from your PR was 8910226 minor allocations.

bryphe:
> Am I reading this wrong, or is the minimap allocating 2-3 times more than the rest of the editor? Because that seems a bit nuts...

It's actually correct! The minimap is expensive to render right now - because of all the tokenization we do on every line. The string manipulations that occur during tokenization (`sub` and friends) are heavy allocators. This is definitely a case where we should try and not do that work on every render, and save those allocations.

> I did a bit of refactoring on top of your PR, which shouldn't have affected the allocations much since they're just moved around, but for some reason moving the shaping into drawShapedText in 386c02f adds 0.5% to minor allocations.

Thank you @glennsl ! Really appreciate the help + investigation.

Re the 0.5% difference - I think that is within reasonable limits, especially if the other metrics are unchanged. The `minor_words` (minor allocations) is actually an estimate in native code programs (from the [Gc](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Gc.html)) module notes:
```
**Number of words allocated in the minor heap since the program was started. This number is accurate in byte-code programs, but only an approximation in programs compiled to native code.**
```

My guess is this isn't a 'real' allocation, so I'm OK leaving it as-is. Thank you for calling it out and testing it, though!

> Edit: I also tried "memoizing" the context to avoid setting the font size and typeface if it hasn't changed, but that didn't really do much.

Cool, doesn't sound like that optimization is worth it then - thanks for trying it 👍

PR #1327, opened 2020-02-12
SCM: Extract Feature Project and Refactor

glennsl:
This also moved the SCM stuff out of ExtHostClient into a separate module.

There's also an awkwardly named `SCMModels` module in `Oni_Core` containing shared types because `Feature_SCM` currently depends on `Extensions` for the `provideOriginalUri` effect. I'm not so sure this is the right approach though. Looking at it now it would seem better to have the dependency go the other way, to have `Extensions` depend on `Feature_SCM` and be an implementation of the SCM protocol, but also allow other implementations without needing to change `Feature_SCM`. But  even `SCMModels` is still pretty hard-wired to the `Extensions` implementation, so we'd have to decouple that too somehow.


src/Feature/SCM/Feature_SCM.rei
```diff
@@ -0,0 +1,64 @@
+[@deriving show]
+type t = {providers: list(Oni_Core.SCMModels.Provider.t)};
```
glennsl: This shouldn't be exposed, but is currently needed to prevent `git.refresh` with not repositories, which should go away soon enough I think.

src/Model/State.re
```diff
@@ -26,12 +26,13 @@ type t = {
   contextMenu: option(ContextMenu.t(Actions.t)),
   mode: Vim.Mode.t,
   completions: Completions.t,
+  configuration: Configuration.t,
+  decorationProviders: list(DecorationProvider.t),
```
glennsl: The decoration providers were moved out of SCM because it's more closely related to the FileExplorer. Once there's a feature project for that, this should probably be moved there.

bryphe:
>  Looking at it now it would seem better to have the dependency go the other way, to have Extensions depend on Feature_SCM and be an implementation of the SCM protocol, but also allow other implementations without needing to change Feature_SCM. But even SCMModels is still pretty hard-wired to the Extensions implementation, so we'd have to decouple that too somehow.

Yes, this is tricky! Thanks for taking the first step towards decoupling this.

glennsl:
Actually, I think I got it right the first time. Consumers of `Feature_SCM` shouldn't need to know about `Oni_Extensions`. That should be abstracted away. Instead I've moved the SCM models to `Oni_Extensions.SCM` and then re-export from `Feature_SCM`. Which still makes the models still shaped after the extension models, but that's fine for now I think.

Eventually we might want to split `Feature_SCM` into a service project that deals with the extension and other providers, and a feature project that just deals with the general SCM protocol, but there's no point doing that until we know what shape we need that protocol to have.

bryphe:
> Actually, I think I got it right the first time. Consumers of Feature_SCM shouldn't need to know about Oni_Extensions. That should be abstracted away. Instead I've moved the SCM models to Oni_Extensions.SCM and then re-export from Feature_SCM. Which still makes the models still shaped after the extension models, but that's fine for now I think.

Sounds reasonable to me 👍

glennsl:
Oops.

> I think we removed that warning in the unit test, but maybe we do the same for the integration tests?

I'd rather have the warning enabled on `@test` as well. If you don't want an `open` to produce a warning, just use `open!` instead.

bryphe:
> I'd rather have the warning enabled on @test as well. If you don't want an unused open to produce a warning, just use open! instead.

Perfect. I run `@test` more frequently than `@integrationtest` - so as long as `@test` covers a superset of the errors on `@integrationtest`, I'm set.

Didn't know about `open!`, TIL

PR #1328, opened 2020-02-12
Perf: Add end-to-end benchmark for EditorSurface rendering

bryphe:
A deficiency in our current `<EditorSurface />` benchmarks is that the actual draw text / draw rect calls (where we do most of our work!) is not reflected at all in the benchmarks today.

We rely heavily on our benchmarks to understand the impact of changes, and to ensure that the 'hot paths' stay performant. This change leverages a new API in Revery - a `HeadlessWindow` - that allows us to render to a software canvas w/o needing a GPU. This lets us exercise the entire end-to-end rendering.

To troubleshoot, we can pass `ONI2_BENCH_SCREENSHOT=1` environment variable - this will cause the UI tests to output screenshots like this:

thousand

This can help verify that rendering is actually occurring as we expect.

Initial measurements:
```
|  EditorSurface - Rendering: Minimal state                           |  10         |  0.18224811554      |  6261      |  0         |  8818099      |  77915      |  2156235      |
|  EditorSurface - Rendering: 1000 Lines state                        |  10         |  0.169266939163     |  6261      |  0         |  8818099      |  77915      |  2156235      |
|  EditorSurface - Rendering: 1000 Lines state - With indent guides   |  10         |  0.171715974808     |  6261      |  0         |  8818169      |  77915      |  2156235      |
|  EditorSurface - Rendering: 100000 Lines state                      |  10         |  0.166203022003     |  6261      |  0         |  8818099      |  77915      |  2156235      |
|  EditorSurface - Rendering: 100000 Lines state - No minimap         |  10         |  0.0929520130157    |  1451      |  0         |  2394109      |  38715      |  527365       |
```

This will help us to make quantitative, data-driven decisions about performance, and easily test & validate the impact of changes in the rendering layer.

Next steps:
- We should incorporate some of the benchmarks @CrossR mentioned (ie, a large JSON file).
- We need a benchmark with syntax highlighting
- We should save snapshots of the performance results, so we can compare build-over-build and ensure there aren't performance regressions. The timing measurement can be flaky for this, but the memory measurements (minor alloc, minor collections, major collections, major alloc) are deterministic and tend to be good proxies for timing measurements in OCaml. It'd be better than having no checks, at least!


glennsl:
> ...that allows us to render to a software canvas w/o needing a GPU.

Isn't that going to be a bit imprecise? The GPU and CPU architectures are so different that it doesn't seem unlikely that some changes could improve performance on the GPU despite decreasing it on the CPU. Or the other way around.

bryphe:
> Isn't that going to be a bit imprecise? The GPU and CPU architectures are so different that it doesn't seem unlikely that some changes could improve performance on the GPU despite decreasing it on the CPU. Or the other way around.

In terms of timing - true.  However, the main goal I want with this is to keep track of our _overhead_ in terms of drawing - what price are we paying in the Reason/OCaml layer, independently of Skia? 

It turns out the __memory allocations__ happen to be deterministic, a fairly direct measurement of the overhead we have on top of the Skia drawing APIs, and a good proxy for actual runtime performance. So those are what I really care about here. Timing measurements aren't very useful for this case; but the memory measurements are! In the profiling I did - I found that the % improvement in allocations was correlated with a % improvement in render time. That's what drove some of the optimizations to reduce allocations.

In a perfect world - we'd have 0 allocations during rendering, and no GC involvement - absolutely minimal overhead except for the work Skia is doing. (Probably not possible to get all the way to 0 in a functional language... but we are allocating so much that could be optimized).

This helps us quantify how close we are to that, and whether we are moving towards or away from that direction.

bryphe:
I'll bring this in so I can merge it into #1324 , and open another PR to test out the cached paints.

PR #1330, opened 2020-02-13
File Explorer: Add tooltips

glennsl:
I'd add a screenshot but, you know, it's a tooltip...

### DIFF
--- a/src/Components/Tooltip.re
+++ b/src/Components/Tooltip.re
@@ -1,0 +1,87 @@
+open Oni_Core;
+open Revery;
+open Revery.UI;
+
+module Constants = {
+  let delay = Time.ms(400);
+};
+
+// MODEL
+
+type t = {
+  text: string,
+  x: float,
+  y: float,
+};
+
+// TOOLTIP
+
+module Tooltip = {
+  module Styles = {
+    open Style;
+
+    let tooltip = (~theme: Theme.t, ~x, ~y) => [
+      position(`Absolute),
+      left(int_of_float(x)),
+      top(int_of_float(y)),
+      backgroundColor(theme.menuBackground),
+    ];
+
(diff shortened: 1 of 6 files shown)


src/Components/Tooltip.re
```diff
@@ -0,0 +1,109 @@
...
+      };
+    };
+  };
+
+  let setTooltip = tooltip => debouncedSetTooltip(Some(tooltip));
+  let clearTooltip = () => debouncedSetTooltip(None);
+
+  let%component make = (~theme, ~font, ()) => {
+    let%hook (current, setCurrent) = Hooks.state(None);
+    internalSetTooltip := setCurrent;
```
glennsl: I've used a simpler mechanism here to communicate between the trigger/hotspot/anchor and the overlay components while still triggering a re-render.

In the context menu I used the state to pass information between them and trigger re-render. Here I use a hooks instead, and then assign the hook's setter to a global ref, which can then be called by the trigger component, through some convenience functions to cover over the rough edges. Much more contained, and works pretty well I think.

bryphe:
> I'd add a screenshot but, you know, it's a tooltip...

Maybe I'm easily excited but I think it looks awesome! 😎 

image

This will be very useful to have - I can think of lots of places it would be helpful and aid in discoverability.

src/Components/Tooltip.re
bryphe: Neat!

src/Components/Tooltip.rei
```diff
@@ -0,0 +1,16 @@
+open Oni_Core;
+open Revery.UI;
+
+module Overlay: {
```
bryphe: I like the API you designed for this (and using the `Trigger` module, but `included`, so the interface is streamlined) - seems like it'll be very easy to add in other spots, too.

PR #1335, opened 2020-02-15
SCM: Add input box with commit on enter

glennsl:
This isn't terribly useful yet, since by default you need to stage changes to be able to commit them. I haven't figured out how to do that yet, and I initially thought it would then commit all changes by default, but apparently I have `git.enableSmartCommit` configured which changes that behavour.


package.json
```diff
@@ -224,7 +224,7 @@
     "@opam/decoders-yojson": "*"
   },
   "resolutions": {
-    "reason-jsonrpc": "github:bryphe/reason-jsonrpc#95f2427",
+    "reason-jsonrpc": "onivim/reason-jsonrpc#ebadf6d",
```
glennsl: Seems like this got lost somewhere, therefore re-doing it so that we can get the trace logging back.

src/Extensions/ExtHostClient.rei
```diff
@@ -55,11 +159,12 @@ let start:
   ) =>
   t;
 let activateByEvent: (string, t) => unit;
-let executeContributedCommand: (string, t) => unit;
+let executeContributedCommand:
+  (~arguments: list(Core.Json.t)=?, string, t) => unit;
```
glennsl: Added `~arguments`, since that's needed for some commands.

src/Extensions/ExtHostClient.rei
```diff
@@ -55,11 +159,12 @@ let start:
...
 let acceptWorkspaceData: (Workspace.t, t) => unit;
 let addDocument: (Protocol.ModelAddedDelta.t, t) => unit;
 let updateDocument:
-  (Core.Uri.t, Protocol.ModelChangedEvent.t, bool, t) => unit;
+  (Core.Uri.t, Protocol.ModelChangedEvent.t, ~dirty: bool, t) => unit;
```
glennsl: Labelled the `~dirty` argument, since `bool` alone is almost void of meaning and should pretty much never be used as a positional function argument. Either label it or use a variant type of some kind that says what it actually does.

src/Extensions/ExtHostClient.rei
```diff
@@ -1,8 +1,112 @@
...
+
+module SCM: {
+  // MODEL
+
+  [@deriving show]
+  type command = {
+    id: string,
+    title: string,
+    arguments: list(Core.Json.t),
+  };
```
glennsl: I'm not really sure where this belongs. In vscode it's in a module called `modes`, along with a lot of other "dumb" data structures, but that name tells me pretty much nothing of its purpose or scope.

src/Store/ExtensionClient.rei
```diff
@@ -0,0 +1,7 @@
+open Oni_Core;
+open Oni_Model;
+open Oni_Extensions;
+
+let create:
+  (~extensions: list(ExtensionScanner.t), ~setup: Setup.t) =>
+  (ExtHostClient.t, Isolinear.Stream.t(Actions.t));
```
glennsl: I split the ExtHostClient creation out into a separate module, so it can be instantiated by `StoreThread` and be used outside the `ExtensionClientStore`.

src/Store/ExtensionClientStoreConnector.re
```diff
@@ -386,31 +63,29 @@ let start = (extensions, setup: Setup.t) => {
...
             );
 
-          let uri = Buffer.getUri(v);
-
           ExtHostClient.updateDocument(
-            uri,
+            Buffer.getUri(buffer),
             modelChangedEvent,
-            true,
+            ~dirty=Buffer.isModified(buffer),
```
glennsl: Hard-coding this to true caused the ext host to think every open file was dirty, including the welcome screen. And as a consequence it refused to commit.

src/Feature/SCM/Feature_SCM.rei
```diff
@@ -60,10 +65,17 @@ module Effects: {
...
+type outmsg =
+  | Effect(Isolinear.Effect.t(msg))
+  | Focus
+  | Nothing;
+
+let update: (ExtHostClient.t, model, msg) => (model, outmsg);
 
 let handleExtensionMessage:
-  (~dispatch: msg => unit, Oni_Extensions.SCM.msg) => unit;
+  (~dispatch: msg => unit, ExtHostClient.SCM.msg) => unit;
```
glennsl: Moved the `SCM` module into `ExtHostClient`, as it fit better there and allows `ExtHostClient.t` to be abstract again.

src/Extensions/ExtHostClient.rei
bryphe: Excellent!

src/Extensions/ExtHostClient.rei
bryphe: Good point. Without a label you have to dive into the implementation to figure it out. Thanks for fixing this.

src/Store/ExtensionClient.rei
bryphe: Cool, that sounds good.

src/Store/ExtensionClientStoreConnector.re
bryphe: Nice catch!

bryphe:
Looks great @glennsl !

image

Thanks for all the improvements you made on the road to implementing this. The SCM sidebar is really coming along!

glennsl:
I've now also figured out where the stage and revert commands come from. Just need to read the `contributes.menus` section form the manifest.

PR #1339, opened 2020-02-17
Dependency: Downgrade to ocaml 4.8.0; update lockfiles

bryphe:
`reason-language-server` doesn't currently support OCaml 4.9.0 (there's an open PR here: https://github.com/jaredly/reason-language-server/pull/351), which causes problems when working on Onivim's code from Onivim. 

This downgrades to 4.8.0 as a temporary workaround. However, better long-term fixes would be:
- Help get the RLS fix in for 4.9.0
- Experiment with picking up the [`ocaml-lsp`](https://github.com/ocaml/ocaml-lsp) binary, if available.


tcoopman:
Picking up the ocamllsp would be awesome!

PR #1340, opened 2020-02-18
Refactoring: New isolinear API

bryphe:
As part of factoring out the terminal integration into [revery-terminal](https://github.com/revery-ui/revery-terminal), I saw that there ended up being some duplicate concerns:
- Having a way to subscribe to an `onStateChanged` event
- Having a subscription model
- Queuing and flush effects

I moved these common pieces into `isolinear` - to streamline Onivim 2, as well as the revery terminal project.

I've also been wanting to bring over the `Subscription` model @glennsl introduced into isolinear - it fixes a major architectural gap we had in Onivim 2, and is a convenient model for listening to external effects.


src/Store/StoreThread.re
```diff
@@ -244,29 +261,22 @@ let start =
...
     Isolinear.Stream.connect(dispatch, inputStream);
   let _: Isolinear.Stream.unsubscribeFunc =
     Isolinear.Stream.connect(dispatch, vimStream);
   let _: Isolinear.Stream.unsubscribeFunc =
     Isolinear.Stream.connect(dispatch, editorEventStream);
-  let _: Isolinear.Stream.unsubscribeFunc =
-    Isolinear.Stream.connect(dispatch, syntaxStream);
   let _: Isolinear.Stream.unsubscribeFunc =
     Isolinear.Stream.connect(dispatch, extHostStream);
-  let _: Isolinear.Stream.unsubscribeFunc =
```
bryphe: There were several cases where we weren't using the stream at all (or it could be replaced by using an Effect-with-dispatch). I removed these to remove noise and make it easier to refactor later without touching all of these pieces.

The remaining items that are using the stream functionality could/should be replaced with subscriptions.

src/Store/SyntaxHighlightingStoreConnector.re
```diff
@@ -16,158 +16,218 @@ module Ext = Oni_Extensions;
...
+type subscriptionParams('msg) = {
+  id: string,
+  languageInfo: Ext.LanguageInfo.t,
+  setup: Core.Setup.t,
+  onStart: Oni_Syntax_Client.t => 'msg,
+  onClose: unit => 'msg,
+  onHighlights: list(Oni_Syntax.Protocol.TokenUpdate.t) => 'msg,
+};
+
+module Subscription =
```
bryphe: This first step of refactoring migrates from using the `Isolinear.Stream.create` to an actual managed `Subscription`.

The next step would be to factor this out to a `Service`, including the effects.

src/Store/LifecycleStoreConnector.re
```diff
@@ -26,16 +24,17 @@ let start = quit => {
     let anyModified = Buffers.anyModified(state.buffers);
     let canClose = force || !anyModified;
 
-    Isolinear.Effect.create(~name="lifecycle.quitAll", () =>
+    Isolinear.Effect.createWithDispatch(~name="lifecycle.quitAll", dispatch =>
       if (canClose) {
+        dispatch(Actions.ReallyQuitting);
         List.iter(h => h(), handlers);
```
bryphe: Before, we used this lifecycle handler to close the syntax client. But now we need an action to go through to turn the subscription 'off' so that we can close it. I added this `ReallyQuitting` action for that purpose, but maybe there is a cleaner way.

src/Store/StoreThread.re
bryphe: > I don't think streams are really necessary at all. 

Agreed! 

> For an easier immediate simplification though, it might be an idea to just use a single stream, and pass its dispatch function into the stores instead of having the stores all return their own stream.

Sounds good.

bryphe:
Thanks for reviewing this and the corresponding isolinear changes, @glennsl ! Still room for improvement, but baby steps... I like your idea of switching from a first-class module to a record - I think that could improve the API further.

> Since it seems to always run, if initially enabled, and isn't updated declaratively but through function calls, it looks like it's just a very elaborate way to break the dispatch dependency cycle. Though I might be missing something of course.

One main benefit I see is that it plugs into the `Features` / `Services` model you proposed - in particular, those functions like `notifyBufferUpdate` can be moved to 'Effects', and it can be wholly decoupled from the store - the interactions from Onivim being the updater that dispatches the effects, and hooking up a subscriber.

image

> I do see a potential benefit if notifyBufferEnter/update could be made data-driven instead. That would make it declarative and remove all the complexity you've added just to be able to make those calls.

I wasn't sure what you meant by this - do you mean not have the `syntaxClient` passed to them? I was thinking the next step - as part of refactoring this to a service - would be to remove the state from knowing about `syntaxClient` at all, and have that managed in the Syntax service. Is that similar to what you had in mind?

> I see you've solved the single subscription runner problem though.

Hopefully! I built off your work to add subscriptions in Onivim 2, of course. So I hope it'll be easy to map our current subscriptions over to this as well.

> I'm not sure I understand why handedOffInstance is needed, even if I understand what it's used for. 

I can relate... I didn't understand why this was needed in the ReactMini component update (here: https://github.com/reasonml/reason-react/blame/1333211c1ea4da7be61c74084011e23137075ede/ReactMini/src/React.re#L85) as well, until I ran into the problem myself 😅 

The issue was that, since we have this heterogenous list of subscriptions - when we have two of them: 

```
    | (Subscription(sub1, oldMapper), Subscription(sub2, newMapper)) =>
      let {
        config: (module ConfigOld),
        params as _oldParams,
        state as oldState,
        handedOffInstance as oldH,
      } = sub1;
      let {
        config: (module ConfigNew),
        params as newParams,
        state as _newState,
        handedOffInstance as newH,
      } = sub2;
```

There's no guarantee that `oldParams` and `newParams` are the same type, or that `oldState` and `newState` are the same type.  We checked that string key is equal, prior to get here - so if we get here, they _should_ be the same type (although, not guaranteed - could easily set up subscriptions that are differents but the same string key). So no way for the type system to know that `newState` and `oldState` are the same!

But we want to check for the case where they are the same type, and if they are, 'update' the subscription with the old state. If they aren't the same type, we should dispose the old one and and init the new one.

I didn't think there was a way to do this without `Obj.magic`. I thought I was going to have reach for that - since the type system doesn't know that `oldState` and `newState` are the same. But I remembered seeing that trick in ReactMini and it finally clicked - could use the same thing here. The `ref` is [shared by all instances of the same subscription type](https://github.com/bryphe/isolinear/blob/54765d532f7f32480df59ae1108cf59dc8955198/src/Sub_internal.re#L77), so if we set it in the old one, and the new one sees it - we know they are the same type (and that we can set it on the new one). If we set a value on the old one's ref, and don't see it in the new one's ref, we know they must be different types.

It's pretty sneaky and I can't take credit for it, but it solved my problem here. If it still doesn't click - you can try removing the handedOffInstance and getting this `update` method to compile: <https://github.com/bryphe/isolinear/blob/54765d532f7f32480df59ae1108cf59dc8955198/src/SubscriptionRunner.re#L63> - everything else works fine w/o the handed off instance - it's just that case where we want to pass the state of one subscription to another in a type safe way, _without the type checker being able to know that they are the same type_.

There might be better or other ways to do this! I'm not an expert. But I was glad I didn't need `Obj.magic` in this case. 

>  It might also be made simpler by not using a first-class module, but it already seems simpler than what I had implemented! 

Definitely! I see this as the next step.

> Kudos!

Thanks @glennsl ! This was really enabled by your work to bring subscriptions into our architecture, by your ideas for the next architectural steps, and the `Subscription` modules you built that I based these off of.

src/Store/SyntaxHighlightingStoreConnector.re
```diff
@@ -16,158 +16,218 @@ module Ext = Oni_Extensions;
...
-    (languageInfo: Ext.LanguageInfo.t, setup: Core.Setup.t, cli: Core.Cli.t) => {
-  let (stream, dispatch) = Isolinear.Stream.create();
-
-  if (!cli.shouldSyntaxHighlight) {
-    let updater = (state, _action) => (state, Isolinear.Effect.none);
-    (updater, stream);
-  } else {
-    let onHighlights = tokenUpdates => {
-      dispatch(Model.Actions.BufferSyntaxHighlights(tokenUpdates));
+type subscriptionParams('msg) = {
```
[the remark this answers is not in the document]
bryphe: Good point, this isn't needed right now. I'll remove it, and then we can revisit when we actually move it to a service.

src/Model/Actions.re
```diff
@@ -29,6 +29,9 @@ type t =
   | BufferSetIndentation(int, [@opaque] IndentationSettings.t)
   | BufferSetModified(int, bool)
   | BufferSyntaxHighlights([@opaque] list(Protocol.TokenUpdate.t))
+  | SyntaxHighlightingDisabled
+  | SyntaxHighlightingEnabled
```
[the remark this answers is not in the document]
bryphe: Good point, I'll remove these for now

src/Model/Actions.re
```diff
@@ -18,7 +18,7 @@ module Diagnostic = Feature_LanguageSupport.Diagnostic;
 
 [@deriving show({with_path: false})]
 type t =
-  | Init
+  | Init([@opaque] Cli.t)
```
glennsl: Why not just pass this as a parameter to `State.create` instead?  The would eliminate the need to initialize the state with a default that might not make sense, and might even introduce subtle timing bugs.

It should also be more explicit. Don't pass the entire Cli object when all you need is a single value from it. Doing so obscures the actual dependency and adds an unnecessary dependency on the Cli object. Feature stores shouldn't need to know about the existence of a CLI interface.

bryphe: > Why not just pass this as a parameter to State.create instead? The would eliminate the need to initialize the state with a default that might not make sense, and might even introduce subtle timing bugs.

I tried this and it had its own set of issues, because of the way we use state in `GlobalContext` right now. We'd need to plumb the Cli.t through the singleton, so I'm not sure it's much better. Removing the `GlobalContext` would help this...

But maybe I'm missing something. Feel free to push up a change if you see a cleaner way to do this!

> Don't pass the entire Cli object when all you need is a single value from

In the meantime, I went this route, and just simplified the `Cli.t` to only have a syntaxHighlightEnabled parameter.

glennsl:
> One main benefit I see is that it plugs into the Features / Services model - in particular, those functions like notifyBufferUpdate can be moved to 'Effects', and it can be wholly decoupled from the store - the interactions from Onivim being the updater that dispatches the effects, and hooking up a subscriber in the subscribers function.

Yeah, I can see that. It's still not a great interface for a subscription though, but more on that below.

> I wasn't sure what you meant by this - do you mean not have the syntaxClient passed to them? I was thinking the next step - as part of refactoring this to a service - would be to remove the state from knowing about syntaxClient at all, and have that managed in the Syntax service. Is that similar to what you had in mind?

`syntaxClient` should be an implementation detail, so your suggestion of moving to a Service sounds good. That's not what I meant here though.

Data-driven means that you pass it the current buffer so it can listen to changes on that, instead of calling `notifyBufferEnter` when it changes. It seems like that should be trivial to implement for `BufferEnter`, but `BufferUpdate` might be trickier to do, at least in a performant way.

It's the same way component rendering works. You don't call `EditorView.update` on `BufferUpdate`, but change the state and then re-render based on the new state. Ideally that's how subscriptions should work too, as pure functions of state. That isn't always going to be possible, of course, but doing so as much as possible makes for a simpler API that's easier to reason about, where looking at the params should be enough to determine behaviour, without needing to consider the effects applied to it, which order the effects were applied in, if the params changed in between the effects and so on.

> ....the type system doesn't know that oldState and newState are the same

Why does the runner need to know about the state at all? It didn't before, and was just kept internal to the provider. Why can't it still be an implementation detail?

src/Model/Actions.re
glennsl: > Removing the GlobalContext would help this...

Ah, yeah that's a wart I'd like to get rid of too. Having this on `Init` isn't that bad until we can get rid of it. We'll want to eventually get rid of `Init` entirely too I think, so can clean it up then.

bryphe:
> Data-driven means that you pass it the current buffer so it can listen to changes on that, instead of calling notifyBufferEnter when it changes. It seems like that should be trivial to implement for BufferEnter, but BufferUpdate might be trickier to do, at least in a performant way.

I think I understand what you are saying - but let me check - something like?
```
let subscriptions = (state) => {
   List.map((buffer) => Service_Syntax.Sub.highlights(~buffer))
  |> Isolinear.Sub.flatten;
}
```

As you mentioned, it's not clear to me how this model would work performantly in the BufferUpdates model. But I do think this refactoring helps us take a baby step towards that direction, so we can start to improve the interface for syntax highlights further. 

I agree that it'd preferable to have less effects, but we might need to think more about how to actually make that happen - unless you have a concrete suggestion for it.

>  Ideally that's how subscriptions should work too, as pure functions of state.

Yes, makes sense!

>  but doing so as much as possible makes for a simpler API that's easier to reason about, where looking at the params should be enough to determine behaviour, without needing to consider the effects applied to it, which order the effects were applied in, if the params changed in between the effects and so on.

Right, that's why subscriptions are so nice.

> Why does the runner need to know about the state at all? It didn't before, and was just kept internal to the provider. Why can't it still be an implementation detail?

The subscriptions, because they interact with the messy-mutable-outside-world, sometimes need to keep track of some per-instance external state (like the `state` hook in react elements, but for subscriptions). 

I use this in the terminal, for example, to track some subscriptions that we remove on dispose: https://github.com/revery-ui/revery-terminal/blob/2463427542b13d17d732e1f100a0e4ba6138882f/src/bin/Terminal.re#L205

You could scope it globally, with a hashtable of id to per-instance state, but I think it's cleaner if we have general way to carry per-instance state, and not make every subscription re-invent it.

glennsl:
> I think I understand what you are saying - but let me check - something like?

Yep!

> I agree that it'd preferable to have less effects, but we might need to think more about how to actually make that happen - unless you have a concrete suggestion for it.

I don't. And I think the `notifyBufferUpdate` might be a necessary evil. But having the syntax client in the state is really ugly. And either way this makes the subscription effectively a singleton and will break if multiple subscriptions are instantiated, so ideally that would be enforced by it somehow too.

> You could scope it globally, with a hashtable of id to per-instance state, but I think it's cleaner if we have general way to carry per-instance state, and not make every subscription re-invent it.

Hmm. while I do see the benefit I'm still not sure I agree. Mostly just because it makes the runner implementation very complicated and hard to understand. But since it's already there I also don't have a pressing need to remove it. Perhaps we can come up with a better implementation some day.

bryphe:
> I don't. And I think the notifyBufferUpdate might be a necessary evil. But having the syntax client in the state is really ugly. And either way this makes the subscription effectively a singleton and will break if multiple subscriptions are instantiated, so ideally that would be enforced by it somehow too.

I think we're on the same page, the next steps I see for this refactoring would be:
1) Move this subscription + effects to a `Service_Syntax` project.
2) Refactor `syntaxClient` to be purely managed by the `Service_Syntax` project, without needing to be in the state at all
3) Change the granularity of the subscription from a singleton to per-buffer highlighting
4) Refactor some of the effects to be managed by the subscription - some candidates would be:
- configurationChangeEffect - this could be managed by carrying the last-pushed `configuration` in the state, and sending via `params` - if they are different, the subscription can handle reconciling that
- likewise for theme change effect
- the buffer enter / visibility changes could also be managed by the subscription, which the per-buffer granularity would enable

This would leave us really with only the buffer update effect as the messy one - and one effect would be significantly easier to reason about than many effects. It would also help us clean up a messy block in the `updater` where, for a bunch of different types of view changes, we send visibility updates.

However, I'd prefer to tackle this in a separate change - I'm working on the terminal service pieces now which depend on this.

glennsl:
Sounds good!

> Refactor syntaxClient to be purely managed by the Service_Syntax project, without needing to be in the state at all

"purely managed" is a ppor choice of words I think, since it's certainly not very "pure". I'd use "encapsulated", and I'm pretty sure that's what you mean as well.

When dealing with side-effects, I think encapsulation is the right approach. Haskell encapsulates side-effects in the IO monad, for example. Encapsulation allows everything else to be pure and easy to reason about, and makes it relatively easy to remove side-effects from the equation to reason about the whole with just a few assumptions about the behaviour of the side-effects. Both figuratively and literally, in the sense that you can easily mock it out for testing.

tl;dr: I like encapsulation!

bryphe:
> "purely managed" is a ppor choice of words I think, since it's certainly not very "pure". I'd use "encapsulated", and I'm pretty sure that's what you mean as well.

Ah yes, you're right, encapsulated is a better description.

> When dealing with side-effects, I think encapsulation is the right approach. Haskell encapsulates side-effects in the IO monad, for example. Encapsulation allows everything else to be pure and easy to reason about, and makes it relatively easy to remove side-effects from the equation to reason about the whole with just a few assumptions about the behaviour of the side-effects. Both figuratively and literally, in the sense that you can easily mock it out for testing.

Definitely!

PR #1344, opened 2020-02-19
Feature - Terminal - Part 1: Initial extension host integration

bryphe:
This change adds an `ExtHostClient_Terminal` module, in a similar vein to the `ExtHostClient_SCM` module.

It exposes a set of messages and a parser for unpacking `MainThreadTerminalService` messages from the extension host to the editor, as well as a `Request` module to help with making requests to the `ExtHostTerminalService` module.

Part 2 will create a `Service_Terminal` module - a service that exposes this as a subscription and effects on top of this (with the `ExtHostClient` as a dependency).

### DIFF
--- a/src/Extensions/ExtHostClient_Terminal.re
+++ b/src/Extensions/ExtHostClient_Terminal.re
@@ -1,0 +1,96 @@
+open Oni_Core;
+
+module Log = (val Log.withNamespace("Oni2.Extensions.Terminal"));
+
+// MODEL
+module ShellLaunchConfig  = {
+  [@deriving show({with_path: false})]
+  type t = {
+    name: string,
+    executable: string,
+    arguments: list(string),
+  };
+}
+
+// UPDATE
+
+type msg =
+  | SendProcessTitle({
+      terminalId: int,
+      title: string,
+    })
+  | SendProcessData({
+    terminalId: int,
+    data: string,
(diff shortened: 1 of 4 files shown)


src/Extensions/ExtHostClient_Terminal.re
```diff
@@ -0,0 +1,101 @@
...
+    });
+
+let handleMessage = (~dispatch, method, args) =>
+  switch (method) {
+  | "$sendProcessTitle" =>
+    switch (args) {
+    | [`Int(terminalId), `String(title)] =>
+      dispatch(SendProcessTitle({terminalId, title}))
+    | _ => Log.error("Unexpected arguments for $sendProcessTitle")
+    }
```
glennsl: I haven't gotten that far in the SCM project yet, but I've wanted to try using ocaml-decoders for this, since I think that'll remove the repetitive args pattern match and catch all error.

bryphe: Good to know! I was curious about this looking at the `ExtHostClient_SCM` module (which I modeled this after).

bryphe: I'll hold off on converting for now too... but it sounds like it would be a useful refactoring to reduce the repetitive error handling

PR #1346, opened 2020-02-20
feat: Reset focus index when the input changes

Sloff:
Small change that resets the focused element when the text in the search field changes


Sloff:
Just as a side note, this is my first PR and the `license/cla` thing didn't take me through the process 
¯\\_(ツ)_/¯

bryphe:
Hi @Sloff ,

Thanks for the improvement here! Looks great 👍 

> Just as a side note, this is my first PR and the license/cla thing didn't take me through the process

Ah, appreciate you letting me know. Thought maybe something was broken in the tool but realized we have it set up so that small changes (<10 lines) don't need the process. Not perfect but a reasonable metric, I think. 

Thanks again for looking at this and fixing it! 💯

PR #1347, opened 2020-02-20
SCM: Switch to BufferSaved action to refresh

bryphe:
The SCM logs are pretty noisy on update, which was making it tough to keep track of the terminal work.

It sounds like the intention was to refresh on save, but the Action we had for that wasn't wired up or used, so we hooked up to BufferUpdate instead of save. The downside is the noisy logs, and potentially a lot of work done on each key press.

This brings in https://github.com/onivim/reason-libvim/pull/94, so that we have a `Vim.Buffer.onWrite` event we can listen to, that then dispatches the `BufferSaved` action, that triggers the git refresh effect.

### DIFF
--- a/src/Model/Actions.re
+++ b/src/Model/Actions.re
@@ -16,9 +16,11 @@ module CompletionItem = Feature_LanguageSupport.CompletionItem;
 module LanguageFeatures = Feature_LanguageSupport.LanguageFeatures;
 module Diagnostic = Feature_LanguageSupport.Diagnostic;
 
+type initOptions = {syntaxHighlightingEnabled: bool};
+
 [@deriving show({with_path: false})]
 type t =
-  | Init
+  | Init([@opaque] initOptions)
   | ActivityBar(ActivityBar.action)
   | BufferHighlights(BufferHighlights.action)
   | BufferDisableSyntaxHighlighting(int)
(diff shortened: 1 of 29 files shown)


src/Model/Actions.re
```diff
@@ -27,7 +27,7 @@ type t =
   | BufferEnter([@opaque] Vim.BufferMetadata.t, option(string))
   | BufferUpdate([@opaque] BufferUpdate.t)
   | BufferRenderer(BufferRenderer.action)
-  | BufferSaved([@opaque] Vim.BufferMetadata.t)
+  | BufferSaved([@opaque] int)
```
glennsl: Why make the `int` opaque?

src/Store/VimStoreConnector.re
```diff
@@ -214,6 +214,12 @@ let start =
       dispatch(Actions.BufferSetModified(id, modified));
     });
 
+  let _: unit => unit =
+    Vim.Buffer.onWrite(id => {
+      Log.debugf(m => m("Buffer written: %n", id));
+      dispatch(Actions.BufferSaved(id));
```
glennsl: The log here seems superfluous, when `dispatch` is going to log the action anyways.

src/Model/Actions.re
bryphe: Ah, good point, I left that in after switching to the `int`. Removed now

PR #1352, opened 2020-02-21
Update Esy Version Required

leematos:
0.6.0 is the latest version currently available. I currently am looking to try building and it appears 0.6.2 is not a thing yet?

Not sure what the right answer is, but making an assumption here :)

### DIFF
--- a/docs/docs/for-developers/building.md
+++ b/docs/docs/for-developers/building.md
@@ -10,7 +10,7 @@ sidebar_label: Building from Source
 
 - Install [Git](https://git-scm.com/)
 - Install [Node](https://nodejs.org/en)
-- Install [Esy](https://esy.sh) (__0.6.0__ is required)
+- Install [Esy](https://esy.sh) (__0.6.2__ is required `npm install -g esy@0.6.2`)
 - __Windows-only__: Run `npm install -g windows-build-tools` (this installs some build tools that aren't included by default on Windows)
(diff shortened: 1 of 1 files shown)


bryphe:
Hi @leematos ,

Thanks for the PR!

Did you see an error message when trying to install esy 0.6.2? (`npm install -g esy@0.6.2`).

I see 0.6.2 as the latest published version on NPM:
image

leematos:
weird! I don't really use NPM so my flow following the docs was:

- Prerequisites
  - click link for esy https://esy.sh/
      - See that the website shows 0.6.0 as latest
     - Look at repo releases and tags: https://github.com/esy/esy/releases 
        - See 0.6.1 as listed, but conflicting with esy website

Install 0.6.0, and I'm doing a build now, and submitted this PR to try and coalesce the difference. If 0.6.2 is real and npm installable, It could help to update the docs to just point to the npm install command instead?

bryphe:
Oh, interesting, good point... I think those must be out-of-date. 

I logged an issue on the esy repo to track: https://github.com/esy/esy/issues/1064

bryphe:
And ya, perhaps updating the documentation to show the install command directly - `npm install -g esy@0.6.2` would be clearer.

leematos:
Edited! <3

bryphe:
Excellent, thank you @leematos !

PR #1353, opened 2020-02-21
Feature - Terminal - Part 3: Terminal renderer

bryphe:
This picks up the common model / renderer utilities from https://github.com/revery-ui/revery-terminal and picks up to be used for Onivim 2's terminal rendering.

Note that Onivim 2 supports terminals on Windows as well, because we leverage the extension host (node-pty) for the PTY layer.

Progress:
image

This wires up the `$onProcessData` from our extension host and pipes it into `ReveryTerminal` to be handled. Most of the heavy lifting is handled by Revery Terminal for the state management - this mainly plumbs through the `$onProcessData` into a revery terminal instance (managed by the terminal subscription) - augments the state with `ReveryTerminal.Screen.t` and `ReveryTerminal.Cursor.t`, and then hands it off to `ReveryTerminal.render` to actually draw.

__TODO:__
- [x] Factor terminal renderer out
- [x] Fix revery-terminal compilation on Windows
- [x] Validate and test on Windows


src/Service/Terminal/Service_Terminal.re
```diff
@@ -49,6 +59,29 @@ module Sub = {
...
           };
 
+        let onEffect = eff =>
+          switch (eff) {
+          | ReveryTerminal.ScreenResized(screen)
+          | ReveryTerminal.ScreenUpdated(screen) =>
+            dispatch(TerminalScreenUpdated({id: params.id, screen}))
+          | ReveryTerminal.CursorMoved(cursor) =>
+            dispatch(TerminalCursorMoved({id: params.id, cursor}))
+          // TODO: Handle output
```
bryphe: This will be coming in the next PR, for handling focus / input keys in the terminal.

src/Store/VimStoreConnector.re
```diff
@@ -649,6 +649,24 @@ let start =
       | Some(_) => dispatch(Actions.BufferEnter(metadata, fileType))
       | None => ()
       };
+
+      if (StringEx.startsWith(~prefix="oni://terminal", filePath)) {
```
bryphe: Need to think about a way to generalize this as we add more buffer renderers.

src/UI/EditorGroupView.re
```diff
@@ -228,6 +228,32 @@ let make = (~state: State.t, ~windowId: int, ~editorGroup: EditorGroup.t, ()) =>
             shouldHighlightActiveIndentGuides
           />;
         | BufferRenderer.Welcome => <WelcomeView state />
+        | BufferRenderer.Terminal({id}) =>
+          let maybeFont =
```
bryphe: I'll refactor this to a separate component, and clean up the font handling, before bringing this in.

src/Service/Terminal/Service_Terminal.re
```diff
@@ -14,6 +15,14 @@ type msg =
...
       id: int,
       title: string,
+    })
+  | TerminalScreenUpdated({
+      id: int,
+      screen: [@opaque] ReveryTerminal.Screen.t,
+    })
+  | TerminalCursorMoved({
+      id: int,
+      cursor: [@opaque] ReveryTerminal.Cursor.t,
```
glennsl: Can there be other screens and cursors in this context than terminal screens and terminal cursors?

src/Service/Terminal/Service_Terminal.re
```diff
@@ -63,14 +96,25 @@ module Sub = {
...
             | SendProcessPid({terminalId, pid}) =>
-              dispatch(ProcessStarted({id: terminalId, pid}))
+              dispatchIfMatches(
+                terminalId,
+                ProcessStarted({id: terminalId, pid}),
+              )
+            | SendProcessData({terminalId, data}) =>
+              if (terminalId == params.id) {
+                let _: int = ReveryTerminal.write(~input=data, terminal);
+                ();
```
glennsl: You can also do `ReveryTerminal.write(~input=data, terminal) |> (ignore: int => unit);`. Not saying that's better, but might be, considering `refmt`s behaviour.

What is the ìnt` anyway?  Chars written? I find the API is usually better if it has a separate function returning some status value if that's needed (which it almost never is), while the "plain" one just returns `unit`.

bryphe: Indeed, it's returning the chars actually written. I agree, we should be able to tuck this completely under the `ReveryTerminal` abstraction - https://github.com/revery-ui/revery-terminal/pull/5

bryphe: Fixed this now 👍

src/Service/Terminal/Service_Terminal.re
bryphe: Nope, you're right, the `Terminal` is superfluous - removed!

PR #1360, opened 2020-02-23
Bugfix: Don't clear notifications with <ESC>

bryphe:
As part of our older notification experience (before @glennsl did the awesome UX revamp in #1184 ), we had set it up so that pressing `<ESC>` when a notification was open would clear it out.

This is still wired up, but I think it makes less sense in the current experience, since the notifications are much less obtrusive - it's not obvious that pressing 'esc' is removing them unless you watch the notification counter carefully.


glennsl:
There's apparently an integration test fir it as well, which is of course now failing 😆

PR #1361, opened 2020-02-23
Make when expression parser (almost) fully compatible with vscode

glennsl:
The old menhir parser was too strict and well-defined, and also a bit overkill and difficult to get good error messages out of. Adding more features and reaching full compatibility would therefore have been difficult, perhaps even impossible, because all the weird edge cases would have had to be made explicit in the lexer and parser.

This PR therefore replaces the menhir parser with a hand-built one, translated reasonably directly from vscode (https://github.com/microsoft/vscode/blob/7fc5d9150569247b3494eb3715a078bf7f8e9272/src/vs/platform/contextkey/common/contextkey.ts). This makes the parser really bad, the code itself pretty and, and the semantics vague and weird, but the weirdness should the same as vscode's. Trying to emulate the weirdness with a better design would probably not have lead to better code anyway. And since the translation maps so closely to vscode, future changes should be easy to incorporate.

Feature/wise, this adds:
- `==` and `!=` expressions.
- `=~` (regex) expressions (see the caveat below).
- Compatiblity fixes to comparison semantics.
- Support for a wider range of variable names, including `config.` variables.
- Non-strict parsing. It should never fail.
- Probably a bunch of other less intentional compatibility fixes, like precedence semantics etc.

There is however still one known incompatibility. This uses `ocaml-re`s `Re.Pcre` module for the regexes, which isn't entirely compatible with the ECMAScript regexes that vscode uses. Most importantly, it does not support flags. Although we probably could emulate that, at least the case insensitive flag. But since it's not needed for the `git` extension, and there might be better solutions, I haven't implemented it yet.


src/Model/ContributedMenu.re
```diff
@@ -0,0 +1,10 @@
+type item = {
+  label: string,
+  msg: Actions.t,
+  condition: WhenExpr.t,
+};
+
+type t = {
+  id: string,
+  teims: list(item),
+};
```
glennsl: Oops, this shouldn't really be here.

src/Store/InputStoreConnector.re
```diff
@@ -128,8 +128,9 @@ let start = (window: option(Revery.Window.t), runEffects) => {
 
     let getValue = v =>
       switch (Hashtbl.find_opt(currentConditions, v)) {
-      | Some(variableValue) => variableValue
-      | None => false
+      | Some(true) => WhenExpr.Value.True
+      | Some(false)
+      | None => WhenExpr.Value.False
       };
```
glennsl: We'll need to generalize this as well, to keep track of the contexts globally, and also be able to extend it contextually.

test/Core/WhenExpr/WhenExprTests.re
```diff
@@ -0,0 +1,65 @@
+open TestFramework;
+
+let re = (pattern, str) => {
+  let re = pattern |> Re.Pcre.re |> Re.compile;
+  Re.execp(re, str);
+};
+
+describe("WhenExpr", ({describe, _}) => {
```
bryphe: Thanks for bringing over the tests, too!

src/Core/WhenExpr/WhenExpr.re
```diff
@@ -0,0 +1,156 @@
...
+    | False => "false"
+    | String(str) => str;
+};
+
+[@deriving show({with_path: false})]
+type t =
+  | Defined(string)
+  | Eq(string, Value.t)
+  | Neq(string, Value.t)
+  | Regex(string, option(Re.re))
```
bryphe: Wow, you brought in Regex compatability too - nice!

I wonder how well the `Re` regular expression map to use the regular expression syntax used in VSCode. If we find that we aren't supporting some regular expression constructs that VSCode uses, we could see if Oniguruma handles those. No action needed; just something to think about.

src/Core/WhenExpr/WhenExpr.re
```diff
@@ -0,0 +1,156 @@
...
+          switch (deserializeValue(right)) {
+          | True => Defined(key)
+          | False => Not(Defined(key))
+          | value => Eq(key, value)
+          };
+        | _ => failwith("unreachable")
+        };
+      } else if (Re.execp(regex, str)) {
+        switch (Re.split(regex, str)) {
+        // This matches more than two "pieces", which should be a synatx error.
```
bryphe: Minor typo: `synatx` -> `syntax`

src/Core/WhenExpr/dune
```diff
@@ -0,0 +1,5 @@
+(library
```
bryphe: Thanks for factoring this out into its own module and generalizing it 👍

src/Store/InputStoreConnector.re
bryphe: That makes sense. It seems like we might want to factor out this whole function: https://github.com/onivim/oni2/blob/21f0fffb5c9343a7a6b11840c3bd43da053175fc/src/Store/InputStoreConnector.re#L19 into a common place - something that gives us the 'context' for a `State.t`, and then we can use it to evaluate expressions anywhere.

src/Core/WhenExpr/WhenExpr.re
```diff
@@ -0,0 +1,156 @@
...
+    | And(exprs) => List.for_all(eval, exprs)
+    | Or(exprs) => List.exists(eval, exprs)
+    | Not(expr) => !eval(expr)
+    | Value(value) => Value.asBool(value);
+
+  eval(expr);
+};
+
+module Parse = {
+  // Translated relatively faithfully from
```
bryphe: I think this was the right call to switch from menhir to manual parsing for this particular case. As you mentioned, it'll be easier to keep in sync later too.

src/Core/WhenExpr/WhenExpr.re
glennsl:
```suggestion
        // This matches more than two "pieces", which should be a syntax error.
```

src/Store/InputStoreConnector.re
glennsl: Yeah, I'm currently working on an overhaul of the command infrastructure, which would use `WhenExpr.t` as well to determine if it's enabled, so will do this as part of that.

src/Core/WhenExpr/WhenExpr.re
glennsl: Yeah, I couldn't find a good source for what "standards" oniguruma tries to support, but they're likely also mutually exclusive to some extent, so that vagueness didn't fill me with confidence either. Hopefully the regexes used just aren't all that advanced.

PR #1362, opened 2020-02-23
Use hook setter for inter-component comunication in ContextMenu

glennsl:
Uses the same hook setter trick used in Tooltip, which simplifies the API quite a bit. Also added support for multiple open menus at once in the component itself. Making sure there's only a single menu open is therefore an application-concern now, which is much cleaner.

### DIFF
--- a/src/Model/State.re
+++ b/src/Model/State.re
@@ -9,7 +9,6 @@ open Oni_Input;
 open Oni_Syntax;
 
 module Ext = Oni_Extensions;
-module ContextMenu = Oni_Components.ContextMenu;
 module KeyDisplayer = Oni_Components.KeyDisplayer;
 module Completions = Feature_LanguageSupport.Completions;
 module Diagnostics = Feature_LanguageSupport.Diagnostics;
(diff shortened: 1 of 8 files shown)


src/Model/State.re
```diff
@@ -9,21 +9,26 @@ open Oni_Input;
...
 module Definition = Feature_LanguageSupport.Definition;
 module LanguageFeatures = Feature_LanguageSupport.LanguageFeatures;
 module BufferSyntaxHighlights = Feature_Editor.BufferSyntaxHighlights;
 
+module ContextMenu = {
+  type t =
+    | NotificationStatusBarItem
+    | Nothing;
+};
+
```
glennsl: I think we might want to have a `Feature_ContextMenu` project for the application-specific model and store, but for now I just plopped the model in here. If you agree this is the right approach I can include that in this PR as well though.

src/UI/StatusBar.re
```diff
@@ -233,13 +234,34 @@ let notificationCount =
...
+    <ContextMenu
       orientation=(`Top, `Left)
-      offsetX=(-10) // correct for item padding
-      model=contextMenu
-      onUpdate=onContextMenuUpdate
-    />
+      offsetX=(-10)
+      items
+      theme
+      font // correct for item padding
```
glennsl: `refmt`...

src/Model/State.re
bryphe: That makes sense to me to have a `Feature_ContextMenu`. I'm not sure if it is pressing right now since the surface area is small.

I'll leave it up to you if you want to include it in this PR, or in a separate one down the road 👍

glennsl: We'll do it later then.

PR #1366, opened 2020-02-25
add scroll bar to SCM Pane (fix #1358)

### LINKED ISSUE #1358: UI - SCM Pane: No scrollbar on overflow
From 1:24 of this repro video: https://drive.google.com/file/d/1e8O9_tD6BkVTLURk04O9m4_7bX46qfBx/view?usp=sharing

When there are too many files, we don't have a proper overflow experience:

image

glennsl:
Fixes #1358 

The SCM UI is very preliminary, but no reason to not have a scrollbar when adding it is so trivial.

I'm also looking at fixing the extension list, but that's trickier because it involves the FlatList, with all its edge cases.

### DIFF
--- a/src/Feature/SCM/Feature_SCM.re
+++ b/src/Feature/SCM/Feature_SCM.re
@@ -529,7 +529,7 @@ module Pane = {
       return((provider, group));
     };
 
-    <View style=Styles.container>
+    <ScrollView style=Styles.container>
       <Input
         style={Styles.input(~font)}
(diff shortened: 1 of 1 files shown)


glennsl:
y u no CI? Have you added a threshold to it or something, @bryphe?

bryphe:
@glennsl - weird! No threshold... not sure why CI isn't kicking off. Looking into it.

Thanks for fixing this, and the help with the extension list too!

Issue #1370, opened 2020-02-26
VSCode themes not applied consistently (with how VSCode does it)


DaniRey:
When using a VSCode theme, certain elements are not styled the way VSCode does. This mainly concerns the Explorer, the Tabs and the bar on the bottom.

For Dark themes, this seems to be no issue, but for light themes, the contrast on those elements is problematic.

Please see attached Screenshot, taken with "Winter is Coming (Light)" from https://vscodethemes.com/e/johnpapa.winteriscoming applied to both editors.

![themes_not_applied_consistently](https://user-images.githubusercontent.com/7700482/75327435-ab383d00-587c-11ea-961b-ea93912a7b66.png)

tatchi:
@glennsl Couldn't this issue be closed now?

glennsl:
Ah, yes! Seems like I fat-fingered the close word on that PR. Thanks for the heads up!

DaniRey:
Thanks @glennsl for fixing this so quickly

PR #1380, opened 2020-02-28
Fix bad theme fallback colors, especially for light themes

glennsl:
Fixes #1073 

- The tab colors weren't picked up from the theme, now they are
- The treeview arrow color was hard-coded, now uses theme foreground color
- The theme used in #1370 depends a lot on fallback colors, which were mostly wrong in out theme settings. I've fixed up quite a few of them, but the theme architecture really needs an overhaul to be able handle this correctly.
- The dropshadow from the minimap seems like a custom style. It doesn't show up in any vscode theme I have, including the theme used in #1370.

I also brought in a few more theme colors for the tabs and activity bar, and cleaned up the tabs component a bit.

2020-02-28-132124_800x600_scrot

Edit: For comparison, a screenshot of vscode and oni2 before, from #1370:

themes_not_applied_consistently

### DIFF
--- a/src/Core/Theme.re
+++ b/src/Core/Theme.re
@@ -446,24 +446,20 @@ let ofColorTheme = (uiTheme, ct: Textmate.ColorTheme.t) => {
 
   let scrollbarSliderActiveBackground =
     getColor(
-      defaults.editorBackground,
-      [
-        "scrollbarSlider.activeBackground",
-        "menu.selectionBackground",
-        "list.activeSelectionBackground",
-      ],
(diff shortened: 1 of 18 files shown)


src/Core/Theme.re
```diff
@@ -51,10 +62,12 @@ type editorGutter = {
 };
 
 type t = {
-  background: Color.t,
```
glennsl: Remove this because it's not an actual theme color. We might want to use `panel.background` instead, but I've just used `editor.background` so far, which seems to be fine.

bryphe: Thanks @glennsl !

PR #1382, opened 2020-02-28
Initial syntax highlight refactor to Feature/Service model

bryphe:
This is an initial step towards the architecture for syntax highlighting we discussed in #1340 (although, it still doesn't implement it - just a baby step towards it!), as well as in preparation for the 'eager' syntax highlighting.

There are lots of syntax-highlight-like concepts floating around the code, so consolidating them in the Feature/Service model will have definite benefits.

This PR does the following:
- Moves `BufferSyntaxHighlights`, as well as initial updater + msgs, to `Feature_Syntax`
- Moves the syntax subscription to a `Service_Syntax`

This is an incremental step - there's still the larger work of changing the granularity of the `Syntax` subscription to buffer-granularity. In addition, the `Oni_Syntax_Client`/`Oni_Syntax_Server` projects should be sub-projects of `Service_Syntax`, and `Oni_Syntax` should be merged with `Service_Syntax.


src/Feature/Syntax/Feature_Syntax.re
```diff
@@ -95,3 +95,23 @@ let handleUpdate = (bufferUpdate: BufferUpdate.t, highlights: t) => {
     highlights,
   );
 };
+
+let update = (highlights: t, msg) =>
+  switch (msg) {
+  | TokensHighlighted(tokens) => setTokens(tokens, highlights)
+  | BufferUpdated(bu) when !bu.isFull => handleUpdate(bu, highlights)
```
glennsl: `bu` is ambiguous. It fits just as well for `Buffer` as `BufferUpdate`

```suggestion
  | BufferUpdated(update) when !update.isFull => handleUpdate(update, highlights)
```

src/Feature/Syntax/Feature_Syntax.re
```diff
@@ -95,3 +95,23 @@ let handleUpdate = (bufferUpdate: BufferUpdate.t, highlights: t) => {
...
+  | _ => highlights
```
glennsl: When you have a well-defined set of variants, you should enumerate all of them. Exhaustiveness checking is a very useful tool that allows you to have a "conversation" with the compiler when adding a new variant as part of adding functionality. I therefore usually consider the wildcard pattern a code smell.

src/Feature/Syntax/Feature_Syntax.re
bryphe: Thanks @glennsl ! First time using the suggestions feature... that's cool

src/Feature/Syntax/Feature_Syntax.re
bryphe: Good point. I implemented the exhaustive check and took out the wildcard.

PR #1384, opened 2020-02-28
Bugfix: Crash when deleting all lines of a file

bryphe:
__Issue:__ While working on #1383 and tested the fix, I saw a crash when opening a file, and then deleting all lines (`ggdG`).

I obtained a crash log which made it pretty easy to troubleshoot: 
```
(Invalid_argument "index out of bounds"):
Raised by primitive operation at file "src/Feature/Editor/EditorDiffMarkers.re", line 60, characters 18-54
Called from file "option.ml", line 24, characters 57-62
Called from file "src/Feature/Editor/EditorSurface.re", line 181, characters 8-42
Called from file "src/UI/EditorGroupView.re", line 204, characters 10-925
Called from file "src/UI/EditorLayoutView.re", line 54, characters 12-74
Called from file "list.ml", line 92, characters 20-23
Called from file "src/UI/EditorLayoutView.re", line 37, characters 2-583
Called from file "src/UI/EditorLayoutView.re", line 66, characters 17-46
...
```

It seems in the case where `markers` is a 0-length array, we'd crash with an out-of-bounds exception. This just checks for the length==0 case and skips the check.

In addition, added an integration test that exercises this.

### DIFF
--- a/src/Feature/Editor/EditorDiffMarkers.re
+++ b/src/Feature/Editor/EditorDiffMarkers.re
@@ -55,13 +55,11 @@ let generate = buffer =>
          );
 
        // if there are deleted lines past the end of the current document, mark the last lines as having deletes afterwards, or being modified
-       let markerLen = Array.length(markers);
-       if (markerLen > 0
-           && Array.length(deletes)
-           - shift^
-           - 1 > Array.length(adds)) {
-         markers[markerLen - 1] = (
-           switch (markers[markerLen - 1]) {
+       if (markers == [||]) {
+         [|DeletedBefore|];
+       } else if (Array.length(deletes) - shift^ - 1 > Array.length(adds)) {
+         markers[Array.length(markers) - 1] = (
+           switch (markers[Array.length(markers) - 1]) {
            | Modified
            | Added => Modified
 
@@ -70,9 +68,10 @@ let generate = buffer =>
            | Unmodified => DeletedAfter
            }
          );
+         markers;
+       } else {
+         markers;
        };
-
-       markers;
      });
 
 let markerPaint = Skia.Paint.make();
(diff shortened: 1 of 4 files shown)


src/Feature/Editor/EditorDiffMarkers.re
```diff
@@ -55,9 +55,13 @@ let generate = buffer =>
          );
 
        // if there are deleted lines past the end of the current document, mark the last lines as having deletes afterwards, or being modified
-       if (Array.length(deletes) - shift^ - 1 > Array.length(adds)) {
-         markers[Array.length(markers) - 1] = (
-           switch (markers[Array.length(markers) - 1]) {
+       let markerLen = Array.length(markers);
+       if (markerLen > 0
```
bryphe: I think this is safe, but just wanted to double-check!

`refmt` sure made it hard to parse that logic ☹️

glennsl: Seems safe, but not correct. If you've deleted the entire document there's definitely deletes past the end, so there should be a marker for that. So perhaps just special case it, `if (markers == [||]) { [|DeletedBefore|] } else { ... }`.

Also, `markerLen` sounds like it describes the length of a single marker, not the array of markers. I'm not sure I see the point of adding this indirection anyway though, since `Array.length` doesn't computer anything the cost should be the same.

bryphe: Oh, good catch, sorry I missed that case. Implemented the suggestion.

bryphe: Looks like `refmt` is happier this time.

bryphe:
> Oh, and generate should be a pure function, so there shouldn't be a need for an integration test. Unit testing should be sufficient.

I agree the unit testing would be sufficient for this particular case (and that it's preferable to reach for unit tests first) - but I'd like to keep the integration test, because this isn't the first time there have been crashes when deleting all lines (not related to diff markers - but they've cropped up both tweaking buffer updates and syntax highlights). I'm trying to cover/catch that broader class of errors.

PR #1390, opened 2020-02-29
Feature - Terminal: Allow setting separate font from editor

bryphe:
This allows setting separate fontFamily, fontSize, and smoothing options from the editor surface, via these configuration settings:

- `terminal.integrated.fontFamily`
- `terminal.integrated.fontSize`
- `terminal.integrated.fontSmoothing`


src/Service/Font/Service_Font.re
```diff
@@ -35,15 +35,15 @@ type msg =
   | FontLoaded(t)
   | FontLoadError(string);
 
-let requestId = ref(0);
-
-let setFont = (dispatch1, fontFamily, fontSize, smoothing) => {
+let setFont = (requestId, dispatch1, fontFamily, fontSize, smoothing) => {
```
glennsl: `dispatch1`? What does that mean?

Also seems like some labelled arguments might be good here. There's quite a few arguments now, and they're not very specifically typed.

src/UI/TerminalView.re
```diff
@@ -23,14 +23,14 @@ let make =
     (
       ~metrics: EditorMetrics.t,
       ~terminal: Feature_Terminal.terminal,
-      ~editorFont: Service_Font.t,
+      ~font: Service_Font.t,
```
glennsl: Not sure about having the font type be the "main" type of a service. I'm leaning towards services and features not having a `t` type at all, since it seem a bit ambiguous, but rather use `model` for the model, and in this case `font`.

Not a big deal though, just food for though.

src/Service/Font/Service_Font.re
bryphe: Ah, I think `dispatch1` was left over from porting from the previous font strategy over to `Service_Font`. Not sure why it has the `1` - I renamed it and change all the arguments to labelled.

src/UI/TerminalView.re
bryphe: Seems reasonable - it's not clear what `Service_Font.t` is (is it a service? is it a font?).

bryphe: Updated from `Service_Font.t` -> `Service_Font.font`

PR #1391, opened 2020-02-29
Feature - Syntax: Implement eager syntax highlighting

bryphe:
__Issue:__ After moving to our 'async` syntax highlighting - there would be a noticeable delta / flash-of-unstyled-content when opening buffers. This was because, in some cases, we'd render before round-tripping to the syntax service.

__Fix:__ Allocate a small budget of time for synchronous highlighting when first opening a buffer. This can be controlled via a few configuration settings:
- `syntax.eagerMaxLines` - the maximum number of lines to 'eagerly' (synchronously) highlight
- `syntax.eagerMaxLineLength` - the maximum length of lines. If we encounter a line longer than this setting, we will only eagerly/synchronously highlight up to the preceding line.

In any case, we allocate ~100ms as our eager highlight budget - although since we evaluate using the textmate strategy, per line, long lines can put us slightly over.


src/Core/ConfigurationParser.re
```diff
@@ -281,6 +281,22 @@ let configurationParsers: list(configurationTuple) = [
     "editor.zenMode.singleFile",
     (s, v) => {...s, zenModeSingleFile: parseBool(v)},
   ),
+  (
+    "syntax.eagerMaxLines",
+    (s, v) => {
```
glennsl: Not a fan of `s` and `v` here. Apparently `s` is somehow short for `configuration` and `v` is short for `json`. A good example of too short variable names becoming wildly misleading because they eventually lose their meaning.

src/Core/Constants.re
```diff
@@ -12,6 +12,10 @@ let defaultFontFamily = "FiraCode-Regular.ttf";
 let fontAwesomeRegularPath = "FontAwesome5FreeRegular.otf";
 let fontAwesomeSolidPath = "FontAwesome5FreeSolid.otf";
 
+let syntaxEagerMaxLines = 500;
+let syntaxEagerMaxLineLength = 1000;
+let syntaxEagerBudget = 0.25; /* 250 milliseconds */
```
glennsl: Do these really need to be global? We should move towards a configuration infrastructure that supports contributing configurations. We're going to need that for extensions anyway.

src/Store/SyntaxHighlightingStoreConnector.re
```diff
@@ -17,11 +17,10 @@ module NativeSyntaxHighlights = Oni_Syntax.NativeSyntaxHighlights;
 module Protocol = Oni_Syntax.Protocol;
 
 // TODO:
-// - Move effects to Service_Terminal
 // - Move updater to Feature_Terminal
 // - Change subscription granularity to per-buffer -
 // - this could help remove several effects!
-let start = (languageInfo: Ext.LanguageInfo.t) => {
+let start = (cli: Core.Cli.t, languageInfo: Ext.LanguageInfo.t) => {
```
glennsl: This seems too coarse-grained. Does the syntax highlighting service really need to know about the CLI interface?

src/Store/SyntaxHighlightingStoreConnector.re
```diff
@@ -43,6 +42,46 @@ let start = (languageInfo: Ext.LanguageInfo.t) => {
...
+
+    while (idx^ < len && ! limitExceeded^) {
+      if (String.length(lines[idx^]) > maxLineLength) {
+        limitExceeded := true;
+      };
+
+      incr(idx);
+    };
+
+    let numberOfLinesToHighlight = idx^;
```
glennsl: The idiomatic way of doing this is to use a for loop and an exception for the early termination. No refs needed, although an exception certainly isn't pure either. The purely functional approach would be to use a recursive helper function.

src/Core/ConfigurationParser.re
bryphe: Updated `s` -> `config` and `v` -> `json` to be more understandable

src/Core/Constants.re
bryphe: They're used in two places right now (two places we specify defaults - the default config in `ConfigurationValues.re` and `ConfigurationParser` defaults when parsing).

Supporting contributed configurations would be great! I'm on board with that.

Since we don't have it yet, though, I think there is a trade-off - make these not global and duplicate, or have them global in constants. I'd lean towards the latter to ensure they keep in sync. But would be great to revisit when we have a concept of contributed configurations.

src/Store/SyntaxHighlightingStoreConnector.re
bryphe: Switched to just use an `enabled` flag

src/Store/SyntaxHighlightingStoreConnector.re
bryphe: Switched to a recursive helper function in latest commit

src/Store/SyntaxHighlightingStoreConnector.re
```diff
@@ -53,18 +53,18 @@ let start = (cli: Core.Cli.t, languageInfo: Ext.LanguageInfo.t) => {
...
+        idx
+      } else if (String.length(lines[idx]) > maxLineLength) {
+        idx;
+      } else {
+        loop(idx + 1)
+      }
     };
 
-    let numberOfLinesToHighlight = idx^;
+    let numberOfLinesToHighlight = loop(0);
```
glennsl: Nice. This looks so much clearer to me.

You can use a scope here to prevent the `loop` function from polluting the namespace unnecessarily:

```reason
let numberOfLinesToHighlight = {
  let rec loop = ...

  loop(0);
};

src/Core/Constants.re
glennsl: Yeah, I think it makes sense to have them be global then.

glennsl: I'm working on revamping the theme infrastructure right now btw, and I think much of that can transfer to configuration as well.

bryphe: Cool @glennsl ! Look forward to seeing what you come up with.

PR #1392, opened 2020-03-01
Feature: Terminal filetype + show current process

bryphe:
This modifies the terminal to show the current active process, and show a file icon (as well as be set to `terminal` filetype):

__Before:__
image

__After:__
image


src/Store/BufferRendererReducer.re
```diff
@@ -4,6 +4,24 @@
...
+        ),
+      )
+        when terminal.id == id => {
+        id,
+        title,
+      }
+
+    | _ => terminal
+    };
+  };
```
bryphe: This `_` wildcard is a code-smell - I think to solve this, we'd need a way to have contributed buffer renderers (so that the terminal renderer could be scoped to `Feature_Terminal`). But maybe there is a cleaner way to handle this in the shorter-term.

@glennsl - let me know if you have ideas here.

glennsl: I don't think I've mentioned this to you yet, but was going to bring it up in our meeting tomorrow. There are a number of other issues with the BufferRenderer as well, like the possibility of having buffers exist without renderers and renderers without buffers, and the awkwardness of getting a sensible label out of them. I think we need to have an app-specific editor model, separate from the current editor model, that bundles the buffer, editor model and bufferrenderer, or some equivalent, so that they can't exist completely independent of each other. The renderer itself could then possibly just be a function exported by the features.

bryphe: Definitely, the model today for our `BufferRenderer` is awkward. We'll keep thinking about ways to improve this, but I agree with your concerns, and that bundling those concepts together makes sense.

> I think we need to have an app-specific editor model, separate from the current editor model, that bundles the buffer, editor model and bufferrenderer, or some equivalent, so that they can't exist completely independent of each other.

That sounds good to me. We'll need to think more about how to structure that and relate it to `Feature_Editor` and our current buffer model. If it was abstract, and not tied to the implementation of the editor, perhaps it could be bundled with the editor model. With what we have today - there can be many editors to one buffer, and it might make sense for them to have different renderers. For example, in implementing a markdown preview - we could have two side-by-side editors for a `README.md` - but one has the 'Editor' renderer and another has the 'Markdown Preview' renderer. Similar scenarios exist for a potential 'diff' view - so things we can keep in mind as we start to explore that functionality.

glennsl:
This triggers a bunch of warnings for unused variables and unbound record labels. How did this get through CI? Did you change the build command?

bryphe:
@glennsl - hmm, looking into it. I'm not getting any warnings with `esy build`, `esy '@test' build` or `esy '@integrationtest' build`. Which command did you run that shows the warnings?

glennsl:
`esy b dune build --watch --only-packages Oni2` is what I usually use. Submitting PR with fixes now.

bryphe:
Ah, I see. Thanks @glennsl !

I'm able to get the warnings running `esy b dune build @check`  as well. I think we should add this to the CI, as it seems the current set of build commands we run aren't catching this.

Issue #1343, opened 2020-02-19
Go to definition opening an empty file


Sloff:
Hi, when using "gd" to go to the definition of something it opens the correct file, but the file looks empty. When manually opening the file everything looks ok.

This happens with the latest release on Windows 10 and only when the definition is in a different file, it works when the definition is in the same file

Sloff:
PS, I've only tested this on Typescript files

Sloff:
Just retested, and it seems to be happening when in the same file as well

bryphe:
Ah, I think this is a bug with the way we are handling URLs on Windows. It looks like we're not properly converting to a windows-style path. Thanks for opening the issue!

PR #1396, opened 2020-03-02
fix: Correctly handle go to definition on Windows

Sloff:
This resolves issue #1343, but I'm unable to test it on Linux or Mac

### DIFF
--- a/src/Core/Uri.re
+++ b/src/Core/Uri.re
@@ -117,7 +117,9 @@ let fromMemory = path => fromScheme(~scheme=Scheme.Memory, path);
 let fromPath = path => fromScheme(~scheme=Scheme.File, path);
 
 let toString = (uri: t) => {
-  Scheme.toString(uri.scheme) ++ "://" ++ uri.path;
+  Scheme.toString(uri.scheme)
+  ++ "://"
+  ++ Internal.addSlash(uri.scheme, uri.path);
 };
 
 let toFileSystemPath = (uri: t) => {
--- a/src/Core/Uri.re
+++ b/src/Core/Uri.re
@@ -65,13 +65,13 @@ module Internal = {
     switch (scheme) {
     | File =>
       let pathLen = String.length(path);
-      if (pathLen > 2 && path.[0] == '/') {
-        let firstLetter = path.[1];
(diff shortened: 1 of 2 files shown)


bryphe:
Wow, this is awesome @Sloff , thank you for investigating and fixing this!

I tested on Mac - still works. I added a unit test to document the behavior, and confirmed it's still working on Windows / OSX after the change.

bryphe:
Looked at the CI failure - looks like one of our integration tests is failing with this change - the `LanguageCssTest`:

```
[INFO]   +101ms IntegrationTest :  WAITER (Validate a diagnostic showed up, since our current input is erroneous): Finished - result: false
[INFO]     +0ms IntegrationTest :  WAITER (Validate a diagnostic showed up, since our current input is erroneous): FAILED: D:\oni2\_esy\integrationtest\store\i\oni2-799ad8f9\bin\LanguageCssTest.exe
[ERROR]    +1ms Oni2.Exception : Exception "Assert_failure integration_test/lib/Oni_IntegrationTestLib.re:174:6":
Raised at file "integration_test/lib/Oni_IntegrationTestLib.re", line 174, characters 6-28
Called from file "integration_test/LanguageCssTest.re", line 57, characters 2-396
Called from file "integration_test/lib/Oni_IntegrationTestLib.re", line 179, characters 2-56
Called from file "integration_test/LanguageCssTest.re", line 10, characters 0-1023
```

I can reproduce it locally on Windows with `esy '@integrationtest' x LanguageCssTest.exe`.

bryphe:
I can reproduce this locally too - if I create a file `test.css` like: 
```
body {};
```

I'll get an error on `master` of "Do not use empty rulesets", but no error with this PR.

The diagnostics also use a URI to match up buffers / diagnostics from the extension host.

Sloff:
Thanks @bryphe for pointing me in the right direction. I've rather altered the code to only remove the leading slash when calling `toFileSystemPath`. With this Go To Definitions works on Windows, as well as the error message displays in the `.css` file.

bryphe:
Excellent! Thanks for the fix and the help @Sloff - works great!

Issue #1284, opened 2020-02-01
Sneak: navigation cut off, out of alignment, or showing when it shouldn't be


emythe:
Hi there - 

Distro: Arch
Plasma (5.17.5), X11, Nvidia propietary drivers
Resolution: 3440x1440 (scaling has no effect)

When I open Onivim2 I encounter the following 'Sneak' behaviour:

![sneak](https://user-images.githubusercontent.com/45769506/73596553-cda79680-451a-11ea-8c66-e37687aef062.png)

- Tab navigation is half out of view
- The navigation for the directory tree (and left of) is slightly too high
- Sneak displays shortcuts over empty space for files and directories that aren't expanded
- The Nord theme is active, although this happens in all of them as far as I can tell

This happens using the appimage (note: it crashes constantly for me and is unusable), and when built from source (doesn't crash and is usable).

Blackhawk-TA:
I fixed the first two bugs in #1400:
- Tab navigation is half out of view
- The navigation for the directory tree (and left of) is slightly too high

PR #1400, opened 2020-03-02
Fix: Remove y-axis offset of sneak items

Blackhawk-TA:
There is a Y-Axis offset for sneak items. This pushes the items partially out of the view as stated here: https://github.com/onivim/oni2/issues/1284
This PR removes the offset and fixes the first two bugs mentioned in the linked issue, which are:
- Tab navigation is half out of view
- The navigation for the directory tree (and left of) is slightly too high

I'm not sure if the offset is wanted. If it is, just reject the PR.

Old state:
grafik

New state:
grafik

### DIFF
--- a/src/UI/SneakView.re
+++ b/src/UI/SneakView.re
@@ -32,7 +32,7 @@ module Styles = {
     Style.[
       backgroundColor(theme.sneakBackground),
       position(`Absolute),
-      top(y - Constants.size / 2),
+      top(y),
       left(x + Constants.size / 2),
       Style.height(Constants.size),
       Style.width(Constants.size),


bryphe:
Thanks @Blackhawk-TA ! Really appreciate the PR.

I do think this is an improvement over what we have today 👍 

What would be nice longer-term is if we had a way to customize this - I think there a few elements where 'tweaking' the offset would help. In other words, having an `xOffset` and `yOffset` argument for the `Sneakable` component: https://github.com/onivim/oni2/blob/master/src/UI/Sneakable.re#L15 (could default to `0`).

Then, we'd include that information when we register the 'sneak' (https://github.com/onivim/oni2/blob/992828fa84125cd7915f50275c874541e48afadb/src/UI/SneakRegistry.re#L4) and when we render it, we'd consult it to set these styles.

But for now - removing that yoffset is definitely an improvement over the styling now. Thanks for the contribution!

PR #1401, opened 2020-03-02
Bugfix - Terminal: Handle characters like <C-c> in the terminal

bryphe:
We weren't handling key presses like `Control+C` in the terminal... this is a problem if you want to be able to kill running processes

### DIFF
--- a/src/Service/Terminal/Service_Terminal.re
+++ b/src/Service/Terminal/Service_Terminal.re
@@ -199,18 +199,50 @@ module Effect = {
     |> List.to_seq
     |> Hashtbl.of_seq;
 
+  let controlKeyRegex = Oniguruma.OnigRegExp.create("<C-([a-z])>");
+
+  let getControlKey = keyCandidate => {
+    Oniguruma.(
+      controlKeyRegex
+      |> Stdlib.Result.map(OnigRegExp.search(keyCandidate, 0))
+      |> Stdlib.Result.to_option
+      |> Utility.OptionEx.flatMap(matches =>
+           if (Array.length(matches) < 2) {
+             None;
+           } else {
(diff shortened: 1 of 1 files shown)


src/Service/Terminal/Service_Terminal.re
```diff
@@ -199,18 +199,50 @@ module Effect = {
...
+             if (String.length(key) == 1) {
+               Some(key.[0]);
+             } else {
+               None;
+             };
+           }
+         )
+    );
+  };
+
```
bryphe: This seems way more complex than it should be. It'd be nice to have a general key parser, instead of using the vim-style strings as our input currency.

glennsl: I talked about this with @fanantoxa a while ago. Not sure if they're working on it, or wanted to work on it, but I think we concluded that a simplified record-type mostly carrying the same information as the string, should work well. We can then easily serialize it at the vim boundary, while passing it around as a record elsewhere.

bryphe: That sounds great! Something like:
```
type key = {
   controlKey: bool,
   altKey: bool,
   shiftKey: bool,
   metaKey: bool,
   keyCode: int,  // SDL2 keycode
   scanCode: int, // Physical scan code
};
```

This would make all of our input handling much more flexible (ie, we could handle VSCode style bindings like `Command+P`)

glennsl: The problem is that we also generate presses from `textInput` events, which don't include key codes and scan codes. So those would have to be optional, and we might have an additional `key: string` instead.

bryphe: Good point - we'd need our model to accomodate the text case as well.

PR #1409, opened 2020-03-04
Pin reason-sdl2 to 2.10.3016

glennsl:
The latest version of reason-sdl2 (2.10.3018) apparently has a breaking change. This tries pinning it to the previous version to unbreak it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,7 +207,7 @@
     "esy-macdylibbundler": "*",
     "isolinear": "^3.0.0",
     "ocaml": "~4.7.0",
-    "reason-sdl2": "*",
+    "reason-sdl2": "2.10.3016",
     "reason-jsonrpc": "1.4.0",
     "reason-libvim": "8.10869.26000",
     "reason-native-crash-utils": "github:onivim/reason-native-crash-utils#ae1fd34",


bryphe:
Weird - I'm not sure why `reason-jsonrpc` is failing in this build. I just upgraded Revery to pick up some SDL2 fixes - I'm trying that out in #1410

glennsl:
We've figured it out in #esy on the ocaml discord. The problem is that dune 2.x doesn't "promote" install files to the source dir anymore, so we'll need to change the path to `_build/default`. Will send a PR to reason-jsonrpc shortly.

glennsl:
It's also hard to reproduce this locally, because if it's been built with `dune 1.x` the source dir will have been changed, and so the install files will remain after building with `2.x` as well, although they might be stale.

bryphe:
Ah, got it - thanks for the help investigating, @glennsl !

> It's also hard to reproduce this locally, because if it's been built with dune 1.x the source dir will have been changed, and so the install files will remain after building with 2.x as well, although they might be stale.

That is confusing... no wonder it would build locally but fail on CI

glennsl:
Obsoleted by #1410

PR #1413, opened 2020-03-04
Theme refactor - part 1

glennsl:
This includes the core infrastructure, the beginning of a theme feature project and a few converted projects and views. The goal here is mostly just to exercise the surface area of the API to get a feel for how it is to work with and organize.

TODO:
- [ ] Extension contributions
- [ ] Customkizations
- [ ] "Convert" old defaults - This is very time consuming because most of it is wrong and has to be looked up in vscode's source code.
- [ ] Move defaults to feature project contributions where it makes sense
- [ ] Convert old theme API usage


src/Core/ColorTheme.rei
```diff
@@ -0,0 +1,75 @@
...
+};
+
+// THEME
+
+type t = {
+  variant,
+  colors: Colors.t,
+};
+
+type resolver = {. color: string => Revery.Color.t};
```
glennsl: This could just be a function, or a record or fist-class module, but there's a few good reasons why I chose an object instead:

- It can easily be extended, unlike a function
- It's structural, which means in can also easily be extended _locally_, and unlike a record doesn't need to have its fields in scope.
- It's less syntactically heavy than a first-class module.

src/Feature/SCM/Feature_SCM.re
```diff
@@ -398,24 +398,32 @@ open Revery.UI.Components;
 
 module Input = Oni_Components.Input;
 
+module Theme = Feature_Theme;
+
+module Colors = {
+  let foreground = Theme.Colors.SideBar.foreground;
+  let background = Theme.Colors.SideBar.background;
+  let hoverBackground = Theme.Colors.List.hoverBackground;
+};
```
glennsl: Aliasing colors locally makes them more convenient to use, gives local contextual meaning, and also gives a nice overview of which colors are used in this module.

src/Feature/SCM/Feature_SCM.re
```diff
@@ -425,20 +433,20 @@ module Pane = {
...
-      color(theme.sideBarForeground),
+      color(theme#color(Colors.foreground)),
       textWrap(TextWrapping.NoWrap),
       textOverflow(`Ellipsis),
     ];
 
     let groupItems = [marginLeft(6)];
 
-    let item = (~isHovered, ~theme: Theme.t) => [
+    let item = (~isHovered, ~theme) => [
```
glennsl: No more type annotations :tada:

src/Feature/Terminal/Feature_Terminal.rei
```diff
@@ -55,3 +55,26 @@ let subscription:
...
+  let ansiWhite: string;
+  let ansiBrightBlack: string;
+  let ansiBrightRed: string;
+  let ansiBrightGreen: string;
+  let ansiBrightYellow: string;
+  let ansiBrightBlue: string;
+  let ansiBrightMagenta: string;
+  let ansiBrightCyan: string;
+  let ansiBrightWhite: string;
+};
```
glennsl: These are only exported because `TerminalView` for some reason is outside the feature project. Otherwise it can be kept completely private.

src/Feature/Terminal/Feature_Terminal.rei
```diff
@@ -55,3 +55,26 @@ let subscription:
...
+
+module Contributions: {let colors: ColorTheme.Defaults.t;};
```
glennsl: I have big plans for this module. The `colors` are picked up when initializing the state, which we could also do for configuration, commands etc.

src/Feature/Theme/Feature_Theme.re
```diff
@@ -0,0 +1,101 @@
...
+    Colors.Tab.defaults,
+    Colors.remaining,
+  ]
+  |> List.map(ColorTheme.Defaults.fromList)
+  |> ColorTheme.Defaults.unionMany;
+
+let initial = contributions => {
+  defaults: ColorTheme.Defaults.unionMany([defaults, ...contributions]),
+  theme: ColorTheme.{variant: Dark, colors: ColorTheme.Colors.empty},
+};
```
glennsl: The contributions are passed in here and combined with the local defaults to be stored in the state. It could also be stored outside the state, but would then have to be passed in alongside the model where needed.

src/Feature/Theme/Feature_Theme.re
```diff
@@ -0,0 +1,101 @@
...
+
+let fallback = keyName => {
+  Log.warnf(m => m("Fallback color used for: %s", keyName));
+  Revery.Colors.black; // TODO: Transparent? Magenta?
+};
```
glennsl: To avoid having to deal with an `option` in the case of a missing value, we use a fallback function that provides some stupid default. vscode will just return `undefined` when a color is missing. Unsure what does during rendering though, but I suspect the effect is much the same as what this fallback function provides.

src/Feature/Theme/Feature_Theme.re
```diff
@@ -0,0 +1,101 @@
...
+let update = (model, msg) => {
+  switch (msg) {
+  | TextmateThemeLoaded(variant, colors) => {
+      ...model,
+      theme: {
+        variant,
+        colors:
+          Textmate.ColorTheme.fold(
+            (key, color, acc) => {
+              Console.log(key ++ ": " ++ color);
```
glennsl: Oops. remove.

src/Feature/Theme/GlobalColors.re
```diff
@@ -0,0 +1,216 @@
...
+        {dark: hex("#2D2D2D"), light: hex("#ECECEC"), hc: unspecified},
+      ),
+      (
+        inactiveForeground,
+        {
+          dark: transparent(0.4, ref(activeForeground)),
+          light: transparent(0.4, ref(activeForeground)),
+          hc: hex("#fff"),
+        },
+      ),
```
glennsl: This is one of the more advanced definitions.

src/UI/StatusBar.re
```diff
@@ -23,6 +23,43 @@ module FontIcon = Oni_Components.FontIcon;
 module Diagnostics = Feature_LanguageSupport.Diagnostics;
 module Diagnostic = Feature_LanguageSupport.Diagnostic;
 module Editor = Feature_Editor.Editor;
+module Theme = Feature_Theme;
+
+module Colors = {
+  let transparent = Colors.transparentWhite;
```
glennsl: Ideally, instead of defining this here, this would be a key name corresponding to a contributed color. Unfortunately that's not feasible from the UI project. We'll need feature projects for that.

src/Core/ColorTheme.rei
```diff
@@ -0,0 +1,75 @@
...
+  type t;
+
+  let fromList: list((string, entry)) => t;
+
+  let get: (key, t) => option(entry);
+
+  let union: (t, t) => t;
+  let unionMany: list(t) => t;
+
+  module DSL: {
```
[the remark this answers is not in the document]
bryphe: Thanks for calling this out! 

What does `DSL` stand for - I think of `Domain-Specific Language` - is that the meaning here?

src/Core/ColorTheme.rei
bryphe: Neat! I'll be interested to see how this works.

src/Feature/SCM/Feature_SCM.re
bryphe: I like it 👍

src/Feature/Terminal/Feature_Terminal.rei
bryphe: Ah, ya, it should be in the feature project.

src/Feature/Theme/Feature_Theme.re
bryphe: I like the idea of using Magenta here - to make it very explicit when we miss one.

src/Feature/Terminal/Feature_Terminal.re
```diff
@@ -139,3 +139,89 @@ let subscription = (~workspaceUri, extHostClient, model: t) => {
...
+  let ansiBrightCyan = "terminal.ansiBrightCyan";
+  let ansiBrightWhite = "terminal.ansiBrightWhite";
+};
+
+// CONTRIBUTIONS
+
+module Contributions = {
+  open Revery;
+
+  let colors =
```
bryphe: Cool!

src/Feature/Theme/GlobalColors.re
bryphe: Cool! You've done a nice job expressing the different (unhandled) cases that come up in VSCode themes - like the references and derived theme colors.

src/Core/ColorTheme.rei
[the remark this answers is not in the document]
bryphe: Ah, I guess it is the 'domain-specific language' for specifying colors (computed, references, etc)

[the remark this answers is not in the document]
glennsl: Yep!

src/Feature/Theme/Feature_Theme.re
glennsl: Magenta it is then!

Apostolique: I expect some great memes will come from this.

src/UI/Tab.re
```diff
@@ -21,31 +21,92 @@ module Constants = {
...
+          (isGroupFocused ? Colors.unfocusedHoverBorder : Colors.hoverBorder)
+          |> theme#tryColor
+          |> Option.value(~default=unhovered);
+        } else {
+          unhovered;
+        };
+      };
+
+      borderBottom(~color, ~width=1);
+    };
```
glennsl: Did I mention that the vscode theming is insanely complex? Well, it's actually worse. In the UI code there's null checks everywhere special casing the absence of a theme color in various ways. The tab component alone is 1500 LOC, and the theming is scattered all over the place. Sometimes you have to look into several UI components to figure out the exact behaviour of a single color.

So these 30 lines of code above is basically just to figure out the color of the bottom border. I expect we might not want to be this rigorous everywhere though, and instead approximate by specifying sensible defaults instead of `null`/`unspecified` as vsode does. This is mostly just to make sure the API can handle it if we need to.

PR #1424, opened 2020-03-05
Proper mouse click handling for wild menu

fanantoxa:
Several updates to wild menu:

- [x] When clicking on menu input i handle cursor position only on front-end side.
  So it happen miss between what editor have and vim have.

### DIFF
--- a/src/Store/QuickmenuStoreConnector.re
+++ b/src/Store/QuickmenuStoreConnector.re
@@ -5,10 +5,12 @@
  */
 open Oni_Core;
 open Oni_Model;
+open Oni_UI;
 open Utility;
 
 module InputModel = Oni_Components.InputModel;
 module ExtensionContributions = Oni_Extensions.ExtensionContributions;
+module Selection = Oni_Components.Selection;
 
 module Log = (val Log.withNamespace("Oni2.Store.Quickmenu"));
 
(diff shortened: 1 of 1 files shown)


src/Store/QuickmenuStoreConnector.re
```diff
@@ -160,8 +162,33 @@ let start = (themeInfo: ThemeInfo.t) => {
...
-        Option.map(state => Quickmenu.{...state, selection}, state),
+    | QuickmenuInputClicked((newSelection: Selection.t)) => (
+        Option.map(
+          (Quickmenu.{variant, selection, _} as state) => {
+            switch (variant) {
+            | Wildmenu(_) =>
+              let transition = selection.focus - newSelection.focus;
+
+              if (transition > 0) {
+                for (x in 0 to abs(transition)) {
```
glennsl: This doesn't need to be `abs`. You know it's positive because you've just checked that it is.

src/Store/QuickmenuStoreConnector.re
```diff
@@ -160,8 +162,33 @@ let start = (themeInfo: ThemeInfo.t) => {
...
+                  GlobalContext.current().dispatch(
+                    Actions.KeyboardInput("<LEFT>"),
+                  );
+                };
+              } else if (transition < 0) {
+                for (x in 0 to abs(transition)) {
```
glennsl: Don't need `abs` here either. You could either use `-transition`, or `downto` instead of `to`.

fanantoxa:
@glennsl Updated PR description. Let's merge this)

PR #1426, opened 2020-03-06
allow saving of buffer with ctrl+s

jakeday:
Would something like this work for mapping ctrl+s to save? Thoughts on alternative ways to handle it?

### DIFF
--- a/src/Store/KeyBindingsStoreConnector.re
+++ b/src/Store/KeyBindingsStoreConnector.re
@@ -170,6 +170,11 @@ let start = () => {
         condition: "editorTextFocus" |> WhenExpr.parse,
       },
       {
+        key: "<C-S>",
+        command: "workbench.action.files.save",
+        condition: "insertMode" |> WhenExpr.parse,
+      },
+      {
         key: "<C-G>",
         command: "sneak.start",
         condition: WhenExpr.Value(True),
--- a/src/Store/KeyBindingsStoreConnector.re
+++ b/src/Store/KeyBindingsStoreConnector.re
@@ -170,9 +170,14 @@ let start = () => {
         condition: "editorTextFocus" |> WhenExpr.parse,
       },
       {
+        key: "<D-S>",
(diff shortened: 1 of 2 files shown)


src/Store/VimStoreConnector.re
```diff
@@ -842,6 +842,14 @@ let start =
       ();
     });
 
+  let saveEffect =
+    Isolinear.Effect.create(~name="vim.save", () => {
+      let _ = Vim.input("<esc>");
+      let _ = Vim.input("<esc>");
+      let cursors = Vim.input("w");
+      ();
+    });
```
glennsl: `w` in normal mode will just move the cursor. What we want is to emulate `:w` followed by enter. Something like this should work I think:

```suggestion
  let saveEffect =
    Isolinear.Effect.create(~name="vim.save", () => {
      let _ = Vim.input("<esc>");
      let _ = Vim.input("<esc>");
      let _ = Vim.input(":");
      let _ = Vim.input("w");
      let _ = Vim.input("<CR>");
      ();
    });
```

glennsl:
Otherwise this seems on point!

src/Store/VimStoreConnector.re
jakeday: @glennsl Good catch! I just updated the PR.

src/Store/KeyBindingsStoreConnector.re
```diff
@@ -169,6 +169,11 @@ let start = () => {
         command: "redo",
         condition: "editorTextFocus" |> WhenExpr.parse,
       },
+      {
+        key: "<C-S>",
+        command: "workbench.action.files.save",
+        condition: "insertMode" |> WhenExpr.parse,
```
glennsl: Do we not want this in any mode? i.e. `editorTextFocus` instead of `insertMode`.

jakeday:
@glennsl I added the mapping for macOS and switched the mode per your suggestion!

PR #1436, opened 2020-03-07
Configuration revamp - Part 1

glennsl:
Just to make sure I'm on the right track. This introduces a new configuration API, similar to the recent Theme revamp, but without central management of contributions. Instead the "schema" is entirely local, with generated getters instead of keys that have to be passed to a centralized getter function. I'm pretty sure we could convert the Theme API to this as well, to cut even more of the boilerplate.

So far only the Editor has been converted to the new API.

TODO:
- [x] Proper loading and initialization of configuration file
- [x] Cleanup
- [x] Remove opt-in default?

Nest steps:
- [ ] "Synchronization"
- [ ] Update/transformation mechanism
- [ ] Convert everything to the new API
- [ ] Contributed configuration from extensions


src/UI/EditorGroupView.re
```diff
@@ -210,25 +158,15 @@ let make = (~state: State.t, ~windowId: int, ~editorGroup: EditorGroup.t, ()) =>
...
             completions={state.completions}
             tokenTheme={state.tokenTheme}
             definition={state.definition}
-            shouldRenderWhitespace
-            showMinimapSlider
-            hoverDelay
-            isHoverEnabled
-            shouldRenderIndentGuides
-            shouldHighlightActiveIndentGuides
+            config={state.config.settings}
```
glennsl: The configuration collection has no application-specific type information, and so could be even be used for generic components.

src/Feature/Editor/EditorConfiguration.re
```diff
@@ -0,0 +1,79 @@
...
+  setting("editor.fontFamily", string |> default("FiraCode-Regular.ttf"));
+let fontSize = setting("editor.fontSize", int |> default(14));
+let largeFileOptimization =
+  setting("editor.largeFileOptimizations", bool |> default(true));
+let highlightActiveIndentGuide =
+  setting("editor.highlightActiveIndentGuide", bool |> default(true));
+let indentSize = setting("editor.indentSize", int |> default(4));
+let insertSpaces = setting("editor.insertSpaces", bool |> default(false));
+let lineNumbers = setting("editor.lineNumbers", lineNumbers |> default(`On));
+let matchBrackets = setting("editor.matchBrackets", bool |> default(true));
```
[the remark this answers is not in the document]
CrossR: I can't imagine any case where we wouldn't want a default, even if the default is just like `""` for a path etc, or a `None` equivalent for some options.

If its easier/simpler code and it forces us to always set a sane default, I can't see a downside.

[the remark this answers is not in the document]
glennsl: Yeah, that's what eventually started dawning on me as well.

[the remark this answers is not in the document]
bryphe: Makes sense to me to always have a default as well 👍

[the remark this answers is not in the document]
bryphe: I really like having the settings completely described in this way! 

I could see in the future - extending this to a friendly name + a description. This could be used a couple of ways:
- Auto-generate our documentation
- Provide completion / suggestions when editing our configuration.

Very cool @glennsl

src/Feature/Editor/EditorSurface.re
```diff
@@ -213,16 +202,14 @@ let make =
...
       bottomVisibleLine
-      shouldHighlightActiveIndentGuides
       mode
       isActiveSplit
       gutterWidth
       bufferWidthInCharacters={layout.bufferWidthInCharacters}
+      config
     />
-    {showMinimap
+    {Config.Minimap.enabled(config)
```
bryphe: Is there any performance impact to this, versus what we had before? For example, does it have to parse JSON everytime we ask for `Config.Minimap.enabled`? I'd be curious if we are getting any allocations here. If so, perhaps it'd be possible to memoize this.

src/Feature/Editor/SurfaceView.re
```diff
@@ -161,19 +160,19 @@ let%component make =
           ~cursorPosition,
           ~definition,
           ~bufferSyntaxHighlights,
-          ~shouldRenderWhitespace,
+          ~shouldRenderWhitespace=Config.renderWhitespace(config),
           ~bufferWidthInCharacters,
         );
 
-        if (shouldRenderIndentGuides) {
+        if (Config.renderIndentGuides(config)) {
```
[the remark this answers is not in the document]
bryphe: Cool!

src/UI/EditorGroupView.re
bryphe: Excellent 👌

src/Feature/Editor/EditorConfiguration.re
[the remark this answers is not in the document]
glennsl: Definitely. That would require them to be collected centrally somehow though. We could do that impurely by registering them in a global, otherwise we'd need to have a list of contributions.  The problem with the latter is that these must be parameterized by the value for the getter to be able to return the value appropriately typed. We don't need that centrally though, so we could either hide it using an existential type or just extract the non-parameterized data. Either way, it requires making another function call for each item before putting it in the contributions list.

src/Feature/Editor/EditorSurface.re
glennsl: There's definitely some overhead, so I certainly wouldn't recommend putting these into a tight loop. It doesn't need to parse the entire JSON, but does need to look it up by key (which is primarily by a precomputed hash, but also matches the string itself when the hash matches match in case of collision) and decode the value (which for primitive types is just a pattern match).

To avoid this overhead I think we could create a local configuration record that's memoized based on physical equality for the configuration collection. I'll do some testing to see if it's worth the bother though.

glennsl: Results are in!

Before this change:

```
|  EditorSurface - Rendering: Minimal state                           |  10         |  0.251815080643     |  6261      |  0         |  9172453      |  254607     |  2333952      |
|  EditorSurface - Rendering: 1000 Lines state                        |  10         |  0.176056146622     |  6261      |  0         |  8964854      |  103670     |  2181990      |
|  EditorSurface - Rendering: 1000 Lines state - With indent guides   |  10         |  0.175040006638     |  6261      |  0         |  8964924      |  103670     |  2181990      |
|  EditorSurface - Rendering: 100000 Lines state                      |  10         |  0.18078494072      |  6261      |  0         |  8964854      |  103670     |  2181990      |
|  EditorSurface - Rendering: 100000 Lines state - No minimap         |  10         |  0.097088098526     |  1451      |  0         |  2536854      |  62802      |  551452       |
```

After this change:

```
|  EditorSurface - Rendering: Minimal state                           |  10         |  0.248698949814     |  6261      |  0         |  9169698      |  254547     |  2290072      |
|  EditorSurface - Rendering: 1000 Lines state                        |  10         |  0.178162097931     |  6261      |  0         |  8962099      |  103610     |  2138110      |
|  EditorSurface - Rendering: 1000 Lines state - With indent guides   |  10         |  0.176610946655     |  6261      |  0         |  8962698      |  103730     |  2138230      |
|  EditorSurface - Rendering: 100000 Lines state                      |  10         |  0.183979034424     |  6261      |  0         |  8962099      |  103610     |  2138110      |
|  EditorSurface - Rendering: 100000 Lines state - No minimap         |  10         |  0.10581612587      |  1451      |  0         |  2534758      |  62862      |  551512       |
```

So it's actually an improvement, although not a significant one. 2% less major allocation for 1000 lines is the largest difference from the numbers I checked.

bryphe: Interesting. Those are great results! Thanks for checking.

src/Feature/Editor/EditorConfiguration.re
```diff
@@ -0,0 +1,75 @@
...
+open Config.Schema;
+
+module CustomDecoders: {
+  let whitespace: Config.Schema.decoder([ | `All | `Boundary | `None]);
+  let lineNumbers: Config.Schema.decoder([ | `On | `Relative | `Off]);
+  let time: Config.Schema.decoder(Time.t);
+} = {
+  open Json.Decode;
+
+  let whitespace =
```
bryphe: Sorry, I just realized this PR I merged collides and will need to be brought over, too

https://github.com/onivim/oni2/pull/1421

I'll hold off on any other configuration changes until after this PR-

glennsl: No worries. There's not much that can conflict directly. On the contrary, this looks like a great example and test case for a pretty advanced custom configuration setting!

glennsl:
Looked more into how extension configuration works, and of course it's much more complicated than I expected.

As described in the [API reference](https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration), there are eight "levels" a setting can be specified in, and these all need to be rendered into a single set that should be passed on to the extensions. Which means we need to collect all the defaults centrally.

So it seems very similar to how theme definition and resolution works, and I think we can make both a kind of hybrid of the current Theme and Configuration proposals:

```reason
// Definition
module Config = {
  let enabled = setting("whatamafob.enabled", bool, ~default=true);
};

// Contributions
module Contributions = {
  let settings = [
    Config.enabled.spec,
    ...
  ];
};

// Get
if (Config.enabled.get(settings)) { ...}

// Set
dispatch(Actions.SettingChanged(Config.enabled.set(false)));
```

This will give us much more flexibility without giving up much convenience. It's still nice and type-safe, and we can protect against forgetting to contribute a setting by logging an error if we try to retrieve a default that doesn't exist centrally, but can still return the default locally so we don't have to deal with `option`s and we don't have to crash either.

glennsl:
It also seems we have to support merging of object values, which will complicate the decoders/encoders a bit.

bryphe:
> As described in the API reference, there are eight "levels" a setting can be specified in, and these all need to be rendered into a single set that should be passed on to the extensions.

One note - we don't actually have to handle this rendering-into-a-single-set - we pass the extension host each piece individually, and it handles showing the right 'view' to the extensions:

We have this configuration `model` that represents the `defaults`, `user` settings, and `workspace` settings - and we pass those individually to the extension host:

https://github.com/onivim/oni2/blob/39477a2999c5f269126f4d74d711fabb4885afc9/src/Extensions/Configuration.re#L65

(It corresponds to the `IConfigurationData` on the VSCode side: https://github.com/microsoft/vscode/blob/c2a604a18c5d3696f62267feadfb1d7583bcbccd/src/vs/platform/configuration/common/configuration.ts#L140)

So we might not need to implement the full 'merge' logic, if VSCode handles it for us.

However, I believe that VSCode requires that the settings be expanded, ie, instead of:

```
{
   "terminal.integrated.fontFamily": "Menlo" 
}
```

it expects objects of the form:
```
{
   "terminal": {
        "integrated": {
            "fontFamily": "Menlo"
        }
   }
}
```

And we do that transformation here: https://github.com/onivim/oni2/blob/39477a2999c5f269126f4d74d711fabb4885afc9/src/Extensions/Configuration.re#L50

glennsl:
Oh damn, I missed that, because I didn't notice `IConfigurationInitData` extended `IConfigurationData`. That changes things...

I have been going for precomputing the flattened set of settings whenever one of the sources changes, since that's a nice optimization anyway and makes the getter very simple. But perhaps that's not the way to go then. And if we need to "explode" the keys, perhaps we should do that initially when the key with precomputed hash is created and build a proper lookup tree instead.

Anyway, thanks for setting me straight! I might have wasted a lot of time this week if you hadn't!

bryphe:
>  I missed that, because I didn't notice IConfigurationInitData extended IConfigurationData. That changes things...

Right, that is really confusing!

> I have been going for precomputing the flattened set of settings whenever one of the sources changes, since that's a nice optimization anyway and makes the getter very simple. But perhaps that's not the way to go then. And if we need to "explode" the keys, perhaps we should do that initially when the key with precomputed hash is created and build a proper lookup tree instead.

I think you're right - it'd be better to "explode" initially, and then have a way to lookup against that tree. Seems simpler than the reverse. It also means we'd be more robust in our configuration parsing, if someone handed us an "exploded" configuration - a case we don't handle correctly right now.

> Anyway, thanks for setting me straight! I might have wasted a lot of time this week if you hadn't!

Glad it helped! I got burned by this before too. Your work here will set us up well to tackle some of the blockers we have w/ configuration and the extension host.

src/Feature/Editor/EditorSurface.re
glennsl: After the last refactor:

```
|  EditorSurface - Rendering: Minimal state                           |  10         |  0.256767034531     |  6261      |  0         |  9172609      |  254597     |  2290122      |
|  EditorSurface - Rendering: 1000 Lines state                        |  10         |  0.190254926682     |  6261      |  0         |  8965010      |  103660     |  2138160      |
|  EditorSurface - Rendering: 1000 Lines state - With indent guides   |  10         |  0.191276073456     |  6261      |  0         |  8966318      |  103950     |  2138450      |
|  EditorSurface - Rendering: 100000 Lines state                      |  10         |  0.197165966034     |  6261      |  0         |  8965010      |  103660     |  2138160      |
|  EditorSurface - Rendering: 100000 Lines state - No minimap         |  10         |  0.11461186409      |  1451      |  0         |  2538268      |  63172      |  551822       |
```

Minor allocations have jumped back up. Major allocations still 2% lower. Still nothing significant.

PR #1438, opened 2020-03-07
Bugfix: prevent IndentationGuesser from considering a single space as indentation

glennsl:
Also a bit of cleanup

### DIFF
--- a/src/Core/IndentationGuesser.re
+++ b/src/Core/IndentationGuesser.re
@@ -9,25 +9,18 @@ module Constants = {
 };
 
 let getLeadingWhitespace = (s: string) => {
-  let len = String.length(s);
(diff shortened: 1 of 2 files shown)


src/Core/IndentationGuesser.re
```diff
@@ -3,26 +3,24 @@
...
-    if (c == ' ') {
-      incr(spaceCount);
-    } else if (c == '\t') {
-      incr(tabCount);
+  let rec loop = (i, spaces, tabs) =>
+    if (i >= String.length(s)) {
+      (spaces, tabs, false);
     } else {
-      foundChar := true;
+      switch (s.[i]) {
```
bryphe: Much nicer than the logic before - thanks!

PR #1439, opened 2020-03-07
Fix #1403: OSX - Terminal fails to load default shell

### LINKED ISSUE #1403: Terminal fails to load default shell
MacOS Catalina
Nightly build, commit: a5e6fb52,

When launched the terminal loads bash, rather than the default shell (for me this is fish).

(Fish will now load if I enter the fish command).

bryphe:
__Issue:__ We weren't properly detecting the default shell in OSX.

It's a bit complicated, because we can't always rely on `$SHELL` being present - for example, in the case where a user launches the application from Finder.

__Fix:__ We had some existing utilities already for acquiring the shell path (needed to get the environment for the VSCode extension host), so first of all, re-use those.

Then, expand the detection logic - if the `$SHELL` environment variable isn't specified, then we'll run the `dscl` command to figure out the user's setting for default shell.  Adapted from this stack overflow post:  https://stackoverflow.com/a/41553295


src/Core/ShellUtility.re
```diff
@@ -3,27 +3,66 @@ open Kernel;
...
+      {
+        // We assume if the $SHELL environment variable is specified,
+        // that should be the default.
+        switch (Sys.getenv_opt("SHELL")) {
+        | Some(v) => v
+        | None =>
+          switch (Revery.Environment.os) {
+          | Windows => "cmd.exe"
+          | Mac => Internal.discoverOSXShell()
+          | _ => "/bin/bash"
```
glennsl: I don't think `$SHELL` is entirely reliable on Linux either. Should probably use `Unix` to read the shell from `/etc/passwd`. Ideally you'd even do that before trying `$SHELL`.

src/Core/ShellUtility.re
```diff
@@ -3,27 +3,66 @@ open Kernel;
...
+
+      let len = String.length(userShell);
+      let slashIndex = String.index(userShell, '/');
+      Log.infof(m => m("dscl returned: %s", userShell));
+
+      String.sub(userShell, slashIndex, len - slashIndex);
+    }) {
+    | ex =>
+      Log.warn("Unable to run dscl to get user shell");
+      "/bin/bash";
```
glennsl: Isn't `zsh` the default shell on macOS now?

src/Core/ShellUtility.re
```diff
@@ -3,27 +3,66 @@ open Kernel;
...
+  (
+    lazy(
+      {
+        // We assume if the $SHELL environment variable is specified,
+        // that should be the default.
+        switch (Sys.getenv_opt("SHELL")) {
+        | Some(v) => v
+        | None =>
+          switch (Revery.Environment.os) {
+          | Windows => "cmd.exe"
```
CrossR: Unrelated to Mac, but the default on Windows has been powershell for the past few years.

That said, `esy` isn't happy in Powershell, so from an Oni2 point of view, it being `cmd` is more useful, but is a deviation from Code at least (Vim does default to cmd to my knowledge).

Guess this will be cleaned up when config options are added to pick a default terminal, as its also possible a Windows user would want to set it to a WSL term etc.

src/Core/ShellUtility.re
```diff
@@ -11,19 +13,38 @@ module Internal = {
...
+      let () = close_out(stdIn);
+      let () = close_in(stdErr);
+      ret;
+  };
+
+  let discoverLinuxShell = () =>  {
+    try({
+      let user = Sys.getenv("USER");
+      let userShell = runCommand("getent passwd " ++ user ++ " | awk -F: '{print $NF}'");
+      userShell
```
glennsl: Why not just use [`Unix.getpwuid`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html#VALgetpwuid)?

src/Core/ShellUtility.re
bryphe: Thanks, it is. I've updated the default to use `zsh`

src/Core/ShellUtility.re
bryphe: Added a method to pull the default Linux shell. I kept `$SHELL` as the first try - I'm assuming if you launched from a shell (populating the `$SHELL` variable), you'd want to preserve that as the shell to use in Onivim 2.

src/Core/ShellUtility.re
bryphe: Thanks @CrossR ! I switched the default to be `powershell` here too

src/Core/ShellUtility.re
bryphe: Oh, nice find - I didn't know about those APIs. Thanks for the suggestion 👍

glennsl: Np :) I mentioned it in our meeting last monday though ;)

src/Core/ShellUtility.re
glennsl: Yeah that seems reasonable. Good call.

src/Core/ShellUtility.re
```diff
@@ -1,43 +1,114 @@
...
+      |> OptionEx.or_lazy(() => {
+           switch (Revery.Environment.os) {
+           | Mac => Internal.discoverOSXShell()
+           | Linux => Internal.discoverLinuxShell()
+           | _ => None
+           }
+         })
+      |> Option.value(~default);
+    })
+  )
```
glennsl: I really like how you've structured this. Very readable. You can easily get a good overview at a glance here, and then dive into the details in well-defined and we--encapsulated functions elsewhere.

Also, nice use of `lazy`. That's a useful feature that's way too easy to forget actually exists.

bryphe: Thanks @glennsl !

PR #1440, opened 2020-03-07
Fix warnings introduced by #1392

glennsl:
Fix warnings introduced by #1392 which somehow mysteriously passed CI


bryphe:
Thanks @glennsl for catching and fixing these!

I was able to reproduce them with `esy b dune build @check`, so I added it to the CI scripts so we can block the build when these warnings are present. 

Sorry about the noise!

glennsl:
`esy b dune build @check` also uncovered an orphaned integration test, which is handy. I just removed it, as it's severely outdated and didn't even test what it claimed to test.

PR #1445, opened 2020-03-08
Dependency: revery-terminal -> 6b2f4e0

bryphe:
Pick up fixes for unicode characters and large writes, and bring in scrollback.


bryphe:
Obsoleted by #1446

PR #1450, opened 2020-03-09
Feature/add tab mapping

jakeday:
Taking a stab add adding the Visual Mode mapping and adding support for tab and shift tab to indent and outdent when in the visual mode.


src/Store/KeyBindingsStoreConnector.re
```diff
@@ -179,6 +179,16 @@ let start = () => {
         command: "workbench.action.files.save",
         condition: "editorTextFocus" |> WhenExpr.parse,
       },
+      {
+        key: "<TAB>",
+        command: "input.indent",
```
glennsl: `input` would refer to the text input component I think. VSCode has an`editor.action.indentLines` command, but no corresponding command for the reverse. But perhaps we could just invent that ourselves. Ideally this would work in both normal and visual modes though.

glennsl:
Also, you need to run `esy format` to appease the CI.

jakeday:
@glennsl  I don't know why the check failed, but it generated the artifacts. I tested them and they appear to be working correctly.

src/Store/VimStoreConnector.re
```diff
@@ -862,6 +862,22 @@ let start =
       ();
     });
 
+  let indentEffect =
+    Isolinear.Effect.create(~name="vim.indent", () => {
+      let _ = Vim.input(">");
+      let _ = Vim.input("g");
+      let _ = Vim.input("v");
```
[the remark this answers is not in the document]
jakeday: Well I'm glad I could help!

leavengood:
The Windows CI tasks seems to be flaky. I had a similar issue last night for my PR, but it worked fine this morning. You just need @glennsl or @bryphe to re-run the CI pipeline.

src/Store/KeyBindingsStoreConnector.re
```diff
@@ -179,6 +179,16 @@ let start = () => {
         command: "workbench.action.files.save",
         condition: "editorTextFocus" |> WhenExpr.parse,
       },
+      {
+        key: "<TAB>",
```
glennsl: Still wondering where this keybinding comes from, and leaning towards the default being `Ctrol+]`, since that's the default for vscode. After all, we aim to be vscode/vim hybrid first and foremot.

CrossR: I use `tab` and `shift-tab` in VSCode with no extensions to indent and unindent blocks of code, though I can only see the action bound to `Ctrl-[]` in the docs...but it does work, so not so sure.

glennsl: Is it a `vscodevim` thing perhaps?

CrossR: Don't think so! I use it on clean machines helping students out in programming labs, and its worked across every machine I've tried, so I think it must be default to VSCode since the only plugins they have installed is the Python one.

glennsl: Ah, yeah it doesn't work with vscodevim, but does work with a selection if I disable it. And of course it works if I put the cursor at the beginning of the line. So I guess that's just part of the "core" editor experience.

Should we have both bindings then? They aren't entirely identical since the "proper" keybinding will indent the line even without a selection, but that's closer to this behaviour anyway.

Also, on macOS the keybinding is `Ctrl+Alt+Cmd+8` to outdent and `Ctrl+Alt+Cmd+9` to indent.

leavengood: This is what I see for a fairly stock VSCode on macOS:

Indent:

Screen Shot 2020-03-10 at 5 37 26 PM

Outdent:

Screen Shot 2020-03-10 at 5 36 39 PM

There does actually seem to be a command that is just plain `outdent`, but not `indent`, at least when I searched.

Tab and Shift-Tab definitely also indent and outdent for me, for both selections and just normal lines.

leavengood: I tend to use `Cmd-]` and `Cmd-[`.

glennsl: Weird. I have no custom keybindings on my mac. These are the default ones. But I do have a Norwegian keyboard layout, where I need to type `Alt+8` and `Alt+9` to get eh square brackets, so the bindings are effectively `Ctrl+Cmd+[` and `Ctrl+Cmd+]`. But that's still weird. I have to look into how the default keybindings are chosen.

Anyway, I propose the following:
- Add keybindings for `C-[`/`C-]` and `D-[`/`D-]` to `editor.action.indentLines`/`outdentLines`
- Change the command of `Tab` and `S-Tab` to `indent` and `outdent`, but just have them trigger the same effect, effectively turning them into aliases.  If in the future we then need to differentiate their behaviour, we can do so fairly easily.

Does that sound ok @jakeday?

jakeday: @glennsl I added the mappings for the CTRL+] and CTRL+[ for indenting and outdenting. With the current mappings in this PR, we can support both typical use cases for users who will expect to be able to use these shortcuts regardless of which editor they are use to.

glennsl: Thanks @jakeday. Could you implement my second suggestion as well? Unless you don't think it's a good idea for some reason.

jakeday: @glennsl Are you wanting the command to be just "indent" or "editor.action.indent"?

glennsl: I'd like `Tab` to be just `indent` and `S-Tab` to be just `outdent`, since the latter actually exists as a keybinding in vscode. `indent` is oddly missing, so we'll just have to infer that one.

jakeday: @glennsl Done!

PR #1453, opened 2020-03-10
CI - OSX: Try building on 10.15, with 10.12 still as deployment target

bryphe:


bryphe:
Bummer - it looks like building on 10.15 means we get a `dyld` error on 10.13:

```
dyld: lazy symbol binding failed: Symbol not found: ____chkstk_darwin
  Referenced from: /Users/runner/runners/2.165.1/work/1/s/./_unpacked/Onivim2.app/Contents/MacOS/Oni2_editor
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: ____chkstk_darwin
  Referenced from: /Users/runner/runners/2.165.1/work/1/s/./_unpacked/Onivim2.app/Contents/MacOS/Oni2_editor
  Expected in: /usr/lib/libSystem.B.dylib
```

I see some related issues here: https://users.rust-lang.org/t/compile-rust-binary-for-older-versions-of-mac-osx/38695 
...but not clear how to fix it on our end. For now, to be safe, I'll stick with the 10.14 build, which seems to work OK on 10.13.

PR #1463, opened 2020-03-13
Theme revamp - part 2: Schema refactor

glennsl:
Refactors the ColorTheme API to (more or less) use the same `Schema` mechanism as introduced in #1436, as well as some other minor refinements.

### DIFF
--- a/src/UI/Tab.re
+++ b/src/UI/Tab.re
@@ -47,16 +47,13 @@ module Styles = {
         };
 
       if (isHovered) {
-        (
+        let color =
           isGroupFocused
-            ? Colors.unfocusedHoverBackground : Colors.hoverBackground
-        ).
-          tryGet(
-          theme,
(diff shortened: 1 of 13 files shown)


src/UI/Tab.re
```diff
@@ -62,9 +65,12 @@ module Styles = {
...
         } else {
-          theme#tryColor(
+          (
             isGroupFocused
-              ? Colors.activeBorderTop : Colors.unfocusedActiveBorderTop,
+              ? Colors.activeBorderTop : Colors.unfocusedActiveBorderTop
+          ).
+            tryGet(
+            theme,
           )
```
glennsl: Jesus christ `refmt`...

glennsl: The philosophy of `refmt`: Why use 2 lines when you can use 7!?

PR #1469, opened 2020-03-14
CI: Try removing ASAN

bryphe:
...see if the `-fsanitize=address`/`-lasan` is causing the linker failure on CI

### DIFF
--- a/integration_test/dune
+++ b/integration_test/dune
@@ -39,7 +39,6 @@
     (libraries
         libvim
         Oni_IntegrationTestLib
-        reason-native-crash-utils.asan
             ))
 
 (install


glennsl:
I picked #1467 over this, but keeping this open just in case.

bryphe:
Sounds good @glennsl ! #1467 would be preferred since it keeps ASAN around 👍 

And we have validation of running the app (at least the `--checkhealth` run) on both 16.04/18.04 - so we still have some coverage that the app runs on 16.04. Thanks for bringing it in!

bryphe:
Obsoleted by #1467

PR #1476, opened 2020-03-16
Bugfix: bring back the insert mode cursor

glennsl:
Removed in #1464. I don't know what I was thinking and the QA department is on vacation...


glennsl:
The link error on Ubuntu is back, but I'm a rebel so I'll merge anyway :metal:

PR #1478, opened 2020-03-18
Feature: Development setup in Docker

MasseGuillaume:


glennsl:
Thanks for the PR! I feel it's lacking a bit of rationale though. We already have [a dockerfile that sets up CentOS](https://github.com/onivim/oni2/blob/master/scripts/docker/centos/Dockerfile) for CI, as well as a CI box running Ubuntu that's set up here:

https://github.com/onivim/oni2/blob/c21d25e2dab622ecdc781e57e7558541e6becc54/azure-pipelines.yml#L55-L99

Not sure what this docker setup would add, other than an increased maintenance burden since if we accept this we'd ought to commit to keeping it up-to-date as well. So I would really appreciate an explanation of what benefit this brings that would outweigh the cost.

MasseGuillaume:
Yup I’m indeed I saw the docker image after the CI ran this PR. I can update the documentation to point to that Dockerfile instead.

The motivation: it’s an easy way to setup the dev environment. It isolates your system.

glennsl:
It's ultimately up to @bryphe to decide, but yeah I think offering the CentOS Dockerfile as an alternative way of setting up a dev environment, would be nice, I'd prefer a few sentences explaining why you might want that though, and which of the there steps it replaces. And also point out that it's an alternative, not a requirement for anything.

Dockerfile.dev
```diff
@@ -0,0 +1,35 @@
...
+RUN apt-get install -y make
+RUN apt-get install -y m4
+RUN apt-get install -y nasm
+RUN apt-get install -y libfontconfig1-dev
+RUN apt-get install -y libx11-dev
+RUN apt-get install -y libsdl2-dev
+RUN apt-get install -y libbz2-dev
+RUN apt-get install -y libgtk-3-dev
+
+RUN apt-get install -y opam
```
goetzc, line 23: All `apt` operations should be put in a single `RUN`, see: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

Something like:
```dockerfile
RUN apt-get update -q && apt-get install -y --no-install-recommends \
    yarn \
    git \
    …
    …
    …
 && rm -rf /var/lib/apt/lists/*
```

At the end the apt cache is no longer required and deleted, for a slimmer container image.

Same with the rest of the similar `RUN` instructions.

Dockerfile.dev
```diff
@@ -0,0 +1,35 @@
...
+
+RUN yarn global add esy@0.6.2
+
+WORKDIR /code
+
+RUN esy install
+RUN esy bootstrap
+RUN esy build
+
+CMD esy run
```
goetzc, line 35: On the `CMD` instruction, it is preferred to use arguments JSON notation.
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#cmd
https://github.com/hadolint/hadolint/wiki/DL3025
```dockerfile
CMD ["esy", "run"]
```

docs/docs/for-developers/building.md
```diff
@@ -155,3 +155,17 @@ From the `oni2` directory:
...
 - `npm install`
 - `npm start`
+
+### Docker
+
+```bash
+docker build --network=host -t oni2-dev -f Dockerfile.dev .
+# Creates a 12GB image !!!
+
+docker run -it \
```
goetzc, line 165: 👏🏼

Dockerfile.dev
```diff
@@ -0,0 +1,35 @@
+FROM node:13.10.1-slim
```
goetzc, line 1: Why use not use a LTS version? 14.15.4 is the latest LTS.
https://nodejs.org/en/download/releases/

goetzc (commented):
Thanks! 
Left some commends.

PR #1479, opened 2020-03-18
Refactor: Notification Feature Project

glennsl:
It's a bit clunky in use still, but I imagine notification effects will be invoked through out messages from feature projects instead of directly from the stores as is done now.

It's also possible to replace the effects with fire-and-forget functions and a subscription, which will make it more convenient to use in some places, but effects lead to a better architecture I think.


src/Store/ExtensionClient.re
```diff
@@ -330,7 +330,7 @@ let create = (~extensions, ~setup: Setup.t) => {
   };
 
   let onShowMessage = message => {
-    dispatch(Actions.ShowNotification(Notification.create(message)));
+    dispatch(Actions.ExtMessageReceived(message));
```
glennsl: Because notifications are now created through an effect rather than directly though a "function-call-action", we have to dispatch a more specific event-like action and handle that to emit an effect. Which is how it should have been done anyway, so a nice side-effect.

src/Store/Features.re
```diff
@@ -101,6 +105,28 @@ let update = (~extHostClient, ~configFile, state: State.t, action: Actions.t) =>
...
+  | TerminalFont(Service_Font.FontLoaded(font)) => (
+      {...state, terminalFont: font},
+      Isolinear.Effect.none,
+    )
+  | TerminalFont(Service_Font.FontLoadError(message)) => (
+      state,
+      Feature_Notification.Effects.create(~kind=Error, message)
+      |> Isolinear.Effect.map(msg => Actions.Notification(msg)),
+    )
+
```
glennsl: The terminal font should probably be part of the `Feature_Terminal` module, but that requires first moving the terminal view to the feature project. When that's done, however, this should be moved there.

Same with `EditorFont` above, but that's a longer road.

src/Store/ThemeStoreConnector.re
```diff
@@ -101,20 +100,8 @@ let start = (themeInfo: ThemeInfo.t) => {
...
-
-  let loadThemeByNameEffect = name =>
-    Isolinear.Effect.createWithDispatch(
-      ~name="theme.loadThemeByName", dispatch =>
-      loadThemeByName(name, dispatch)
-    );
-
   let onChanged = (newTheme, dispatch) =>
-    loadThemeByName(newTheme, dispatch);
+    dispatch(Actions.ThemeChanged(newTheme));
```
glennsl: Instead of callbacks invoking side-effects directly, dispatch a message and encapsulate the side-effect in an `Effect`.

src/Store/VimStoreConnector.re
```diff
@@ -138,18 +138,8 @@ let start =
...
-        | Error => ("ERROR", Notification.Error)
-        | Warning => ("WARNING", Notification.Warning)
-        | Info => ("INFO", Notification.Info)
-        };
-
-      Log.debugf(m => m("Message - %s [%s]: %s", priorityString, title, msg));
-
-      dispatch(ShowNotification(Notification.create(~kind, msg)));
+    Vim.onMessage((priority, title, message) => {
+      dispatch(VimMessageReceived({priority, title, message}))
```
glennsl: Here too, instead of making decisions in callbacks scattered all over the place, dispatch an event-like message and do the decision-making and side-effects in in the store/update function.

src/Store/ExtensionClient.re
bryphe: Thanks for fixing this!

src/Feature/Notification/Feature_Notification.re
```diff
@@ -0,0 +1,315 @@
+open Oni_Core;
+
+module Internal = {
```
bryphe: One thing I was thinking about was that we could remove the `id` generation from being effect-ful to actually being par tof the `model`: 
```
type t = {
   nextId: int,
   notifications: list(notification),
};
```

This keeps the id generation (and message adding) pure. Something I should've done originally!

Not a big deal; as it still encapsulated from the consumer, just something I was thinking about.

src/Feature/Notification/Feature_Notification.re
```diff
@@ -0,0 +1,315 @@
...
+
+let update = (model, msg) => {
+  switch (msg) {
+  | Created(item) => [item, ...model]
+  | Dismissed(item) => List.filter(it => it.id != item.id, model)
+  | AllDismissed => initial
+  };
+};
+
+// EFFECTS
```
bryphe: I like the way this is organized!

src/Feature/Notification/Feature_Notification.re
glennsl: Definitely. I don't think it matters much in practice though, since it's just used for optimization. I've also considered just using the current time, since that should also be unique, might be nice to have for other purposes as well, and would make this much simpler. I have a note on that, so I'll probably revisit this eventually, but for now I'll just leave it as-is.

PR #1481, opened 2020-03-18
Dependency: rely -> 3.2.1

glennsl:
This is a bit more involved than just bumping the dependency, because several other dependencies has `rely` as a production dependency pinned to an ancient version. Those dependencies therefore also needs to be fixed, which is done in:

- [x] https://github.com/CrossR/reasonFuzz/pull/12
- [x] https://github.com/onivim/reason-oniguruma/pull/10
- [x] https://github.com/onivim/reason-textmate/pull/52
- [x] https://github.com/onivim/reason-tree-sitter/pull/18

Also, while this compiles the tests don't actually run. They just exit with error code 2 and no output. I do not understand why, and am hoping a CI run might shed some light on it.


package.json
```diff
@@ -253,10 +250,12 @@
     "yarn-pkg-config": "esy-ocaml/yarn-pkg-config#db3a0b6",
     "rench": "bryphe/rench#a976fe5",
     "@esy-ocaml/reason": "facebook/reason#8f71db0",
-    "reason-textmate": "onivim/reason-textmate#98fbca5",
+    "reasonFuzz": "glennsl/reasonFuzz#d10bac6",
+    "reason-tree-sitter": "glennsl/reason-tree-sitter#a5b5221",
+    "reason-oniguruma": "glennsl/reason-oniguruma#18a8287",
+    "reason-textmate": "glennsl/reason-textmate#51bfa1f",
```
glennsl: TODO

bryphe:
> Also, while this compiles the tests don't actually run. They just exit with error code 2 and no output. I do not understand why, and am hoping a CI run might shed some light on it.

Exit code 2 is the exit code for an OCaml exception. I added this code to troubleshoot: 
```
Printexc.record_backtrace(true);

prerr_endline ("Starting test...")
Printexc.set_uncaught_exception_handler((exn, bt) => prerr_endline(Printexc.to_string(exn)));
```

And got this: 
```
Failure("Error trying to create a snapshot directory. Currently it is required to provide a valid (non empty) directory name for snapshots to live in. This will likely be made optional in a future release")
```

Looks like we need to update all our `TestFramework` modules to have an actual `snapshotDir`

glennsl:
> Looks like we need to update all our TestFramework modules to have an actual snapshotDir

Ah, yeah I have a branch which does that too, which I apparently forgot to include when I split this out, but that also still fails. And also doesn't print anything even if I add the exact same logging code. 

But pulling down this clearly works so... weird...

glennsl:
Looking at the snapshot dirs I used, it's no surprise it fails:

```reason
    Rely.TestFrameworkConfig.initialize({snapshotDir: "/test/Core/__snapshots__", projectDir: ""});
```

But this still emits exit code 2, so I would think it's an OCaml exception that should be logged witt the same logging code. Yet it does not...

Anyway, good catch. Thanks for fixing this!

PR #1483, opened 2020-03-18
CI - Linux: Reduce disk space usage to unblock Linux builds

bryphe:


bryphe:
I'm suspecting that this is from:
https://stackoverflow.com/a/55533602

It seems like, in the case where we run out of disk space, `ld` might return 1 _with no additional error information_. I saw a few runs where the disk space was maxed out, so this seems a likely culprit.

Trying to run `esy cleanup` to see if that helps...

PR #1489, opened 2020-03-21
Configuration: Merge and pass user configuration and defaults to extensions.

glennsl:
This works as-is, but is a bit hacky and only passes the configuration initially. That should still unblock a ton of extensions though, if we're really eager to do so.

TODO:
- [x] Initialize state properly
- [x] Notify extensions when configuration changes
- [ ] Optional: wire up extension defaults to the config resolver (if we want to use contributed configuration in our own code)


src/Store/ExtensionClient.re
```diff
@@ -138,41 +138,29 @@ module ExtensionDocumentSymbolProvider = {
...
+    |> Config.Settings.union(Config.Schema.defaults(config.schema))
+    |> Configuration.Model.fromSettings;
+
+  let user =
+    Config.Settings.fromList([
+      ("reason_language_server.location", Json.Encode.string(setup.rlsPath)),
+      ("terminal.integratied.enc.windows", Json.Encode.null),
+      ("terminal.integratied.enc.linux", Json.Encode.null),
+      ("terminal.integratied.enc.osx", Json.Encode.null),
+    ])
```
glennsl: This is one way we can hard-code settings now. Don't know if this is still the way we want to do it though.

src/Store/StoreThread.re
```diff
@@ -115,8 +115,16 @@ let start =
...
   let themeUpdater = ThemeStoreConnector.start(themeInfo);
 
+  // TODO: This is only necessary because the config file can't be passed to the initialization in order for the state to be properly initalized.
+  let config =
+    Feature_Configuration.update(
+      ~configFile=configurationFilePath,
+      state.config,
+      ConfigurationFileChanged,
+    );
+
```
glennsl: This is pretty bad, but is currently needed because the state is currently initialized in several different places and can't be passed any initial arguments because of that. I've bumped into this limitation several times now, so I might try to tackle that before merging this, so we can remove this hack.

glennsl:
Do you have any ideas what kind of change to the configuration could cause the terminal integration test to fail like this @bryphe?

bryphe:
> Do you have any ideas what kind of change to the configuration could cause the terminal integration test to fail like this @bryphe?

Hmm, the integrated terminal requires this config:
```
      (
        "terminal",
        `Assoc([
          (
            "integrated",
            `Assoc([
              (
                "env",
                `Assoc([
                  ("windows", `Null),
                  ("linux", `Null),
                  ("osx", `Null),
                ]),
              ),
            ]),
          ),
        ]),
      ),
```

Looks like there might be typos where this was moved -

src/Store/ExtensionClient.re
```diff
@@ -138,41 +138,29 @@ module ExtensionDocumentSymbolProvider = {
...
+    |> Config.Settings.unionMany
+    |> Config.Settings.union(Config.Schema.defaults(config.schema))
+    |> Configuration.Model.fromSettings;
+
+  let user =
+    Config.Settings.fromList([
+      ("reason_language_server.location", Json.Encode.string(setup.rlsPath)),
+      ("terminal.integratied.enc.windows", Json.Encode.null),
+      ("terminal.integratied.enc.linux", Json.Encode.null),
+      ("terminal.integratied.enc.osx", Json.Encode.null),
```
bryphe: There are typos here that are likely the cause of the terminal integration test failing:

`integratied` -> `integrated`
`enc` -> `env`

Also, you should be able to repro this locally just by opening up a terminal in the editor, w/o needing to run the integration test (might be quicker).

src/Store/StoreThread.re
bryphe: Thanks for fixing this in #1490

test/Core/Utility/KeyedTreeTests.re
```diff
@@ -0,0 +1,120 @@
...
+          ])
+        );
+
+      let actual = Tree.map(int_of_string, tree);
+
+      expect.ext.tree(actual).toEqual(expected);
+    })
+  });
+
+  describe("fold", ({test, _}) => {
```
bryphe: Nice set of test cases 👍

src/Store/ExtensionClient.re
glennsl: Ouch, lol, should probably have been able see that myself...

```suggestion
      ("terminal.integrated.env.windows", Json.Encode.null),
      ("terminal.integrated.env.linux", Json.Encode.null),
      ("terminal.integrated.env.osx", Json.Encode.null),
```

test/Core/Utility/TestFramework.re
```diff
@@ -1,7 +1,84 @@
+open Rely.MatcherTypes;
+
+open Kernel;
+
+// TREE EXTENSIONS
+
+type treeExtensions = {toEqual: KeyedStringTree.t(int) => unit};
+
+let treeExtensions = (actual, {createMatcher}) => {
```
bryphe: Cool, I always forget about extending the matchers... this is a great in-depth example

src/Store/StoreThread.re
glennsl: Fixed in https://github.com/onivim/oni2/pull/1489/commits/6d6d7f83fb3d3c562dde77d155443b1522a2ea65

src/Feature/Configuration/Feature_Configuration.re
```diff
@@ -14,49 +39,66 @@ let merge = model => {
...
     schema:
       Config.Schema.unionMany(
         contributions |> List.map(Config.Schema.fromList),
       ),
-    user: Config.Settings.empty,
+    user:
+      Internal.loadConfiguration(configFile)
+      |> Result.value(~default=Config.Settings.empty),
     merged: Config.Settings.empty,
   });
```
glennsl: This could have been implemented using init effects as well, and we'd usually want side-effects on init to be asynchronous anyway, but in this case I think blocking is the right approach. We don't want the "flash of unstyled content"-experience that might occur if we render before having loaded the configuration.

I also think we might want to pass in some kind of "configuration provider" instead of the configuration file directly, to give the consumer more control over loading, fallback etc. and better separation for testing. I'm just not sure how that would interact with the configuration commands, `ConfigurationFileChanged` event and such. Maybe we should have a separate `Feature_ConfigurationFile` project?

bryphe: Thanks for thinking about this!

> We don't want the "flash of unstyled content"-experience that might occur if we render before having loaded the configuration.

Definitely not.

> I also think we might want to pass in some kind of "configuration provider" instead of the configuration file directly, to give the consumer more control over loading, fallback etc. and better separation for testing

This would be great. We have a hacky-configuration-file-override in our integration tests: https://github.com/onivim/oni2/blob/9e84aedee6755919a95f11cdf773d4cb5e45b610/src/Store/ConfigurationStoreConnector.re#L14

It would be nice to generalize the 'provider' strategy and expose it at a higher-level.

>  I'm just not sure how that would interact with the configuration commands, ConfigurationFileChanged event and such. Maybe we should have a separate Feature_ConfigurationFile project?

I'm also curious how it would relate to the other configuration files we load:
- `keybindings.json`
- (soon) `persisted.json`

Would it make sense to have `Feature_KeybindingsFile` and `Feature_PersistedFile` - or should they be grouped / share common logic?

development_extensions/oni-dev-extension/extension.js
```diff
@@ -156,6 +160,27 @@ function activate(context) {
...
+
+    // Configuration
+
+    const rlsLocation = vscode.workspace.getConfiguration().get("reason_language_server.location");
+    console.error("Configured RLS location: ", rlsLocation);
+
+    const editorFontFamily = vscode.workspace.getConfiguration().get("editor.fontFamily");
+    console.error("Editor Font Family: ", editorFontFamily);
+
+    const dontCrash = vscode.workspace.getConfiguration().get("developer.oni.dont-crash");
```
bryphe: This variable name is confusing for me 😆 

It seems like it's just a test boolean? Maybe `developer.oni.config-test-bool`?

development_extensions/oni-dev-extension/extension.js
```diff
@@ -156,6 +160,27 @@ function activate(context) {
...
+
+    vscode.workspace.onDidChangeConfiguration(event => {
+      if (event.affectsConfiguration("developer.oni.dont-crash")) {
+        const dontCrash = vscode.workspace.getConfiguration().get("developer.oni.dont-crash");
+        console.error("Setting changed, Don't crash: ", dontCrash);
+      } else {
+        const dontCrash = vscode.workspace.getConfiguration().get("developer.oni.dont-crash");
+        console.error("Setting did not change, Don't crash: ", dontCrash);
+      }
+    });
```
bryphe: Do you think we could have a test case to exercise this? Perhaps a simple integration test - like: `ExtHostConfigurationChangedTest.re` Similar in form to: https://github.com/onivim/oni2/blob/master/integration_test/ExtHostCompletionTest.re - except we just need a way to pick up that the extension saw a configuration change.

We could add similar behavior to the `test-extension` - like have it emit some way we can detect the latest value it sees (ie, send a notification) - and then change the configuration, and wait on the notification response.

Alternatively - we do have direct unit tests for the `ExtHostClient` - that's a possibility too: https://github.com/onivim/oni2/blob/9e84aedee6755919a95f11cdf773d4cb5e45b610/test/Extensions/ExtensionClientTest.re#L83

Having a test case that covers this will be helpful when I go to upgrade the extension host, to make sure I don't regress anything

glennsl: Added in https://github.com/onivim/oni2/pull/1489/commits/daf2a3fc734c09e0207f1b7f4241371f58098891

src/Feature/Configuration/Feature_Configuration.re
glennsl: I abstracted the config file out behind a `getUserSettings` function for now, just to make easy to manipulate the settings for the integration test. But in the future perhaps a `Service_ConfigurationFIle` project that allows subscribing to a configuration file would be a good approach. Might try that out to get a feel for using `luv` for file watching as well.

integration_test/ExtConfigurationChangedTest.re
```diff
@@ -0,0 +1,45 @@
+open Oni_Core;
+open Oni_Model;
+open Oni_IntegrationTestLib;
+
+// This test validates:
+// - The 'oni-dev' extension gets activated
+// - That changes to configuration settings are properly propagated to extensions
+// - That extension messages are delivered as notifications Oni-side
+runTestWithInput(
+  ~name="ExtConfigurationChangedTest", (_input, dispatch, wait, _runEffects) => {
```
bryphe: Thanks for adding this test, @glennsl !

PR #1490, opened 2020-03-21
Refactor: Deduplicate state initialization

glennsl:
The state was being initialized in three separate places, `Oni2_editor`, `StoreThread` and `GlobalContext`. It is now only initialized and stored in `Oni2_editor` and access to the state has been removed entirely from `GlobalContext`.

I also wanted to centralize the initialization effects and remove `Actions.Init` so that the initial state and init effects can be co-located. That could be done relatively easily by allowing initial effects to be passed in to the store along with the initial state, but currently that seems like removing one temporary hack for another, without solving any actual needs yet. So I'd rather wait for that and make sure it's done properly.


src/Store/StoreThread.re
```diff
@@ -80,13 +81,8 @@ let start =
       cliOptions,
     );
 
-  let state = Model.State.create();
-
-  let latestState: ref(Model.State.t) = ref(state);
   let latestRunEffects: ref(option(unit => unit)) = ref(None);
 
-  let getState = () => latestState^;
-
```
glennsl: Instead of storing the state here as well, the `getState` function is just passed in from `Oni2_editor` instead.

src/UI/WildmenuView.re
```diff
@@ -1,16 +0,0 @@
...
-      ~configuration: Configuration.t,
-      ~theme: Theme.t,
-      (),
-    ) =>
-  <QuickmenuView
-    font={GlobalContext.current().state.uiFont}
-    theme
-    configuration
-    state
-  />;
```
bryphe: Thanks for removing this unused code 👍

PR #1492, opened 2020-03-21
Feature - Input: Add <C-w> and <C-h> handling to the input model

bryphe:
This adds support for `<C-h>` (which is like backspace) and `<C-w>` (delete word) gestures to the `InputModel`. I use these often in insert mode and I miss them in the search and quickopen experiernce.


test/Components/InputModelTests.re
```diff
@@ -228,23 +228,21 @@ describe("handleInputS#handleInput", ({describe, _}) => {
...
 
       expect.equal(expected, newSelection);
       expect.string(text).toEqual(
-        "Som interesting. Test. String. Isn't it? Maybe",
+        " interesting. Test. String. Isn't it? Maybe",
       );
     });
 
-    test("Doesn't remove character if cursor at the beginnig", ({expect, _}) => {
+    test("Doesn't remove anything if cursor at the beginnig", ({expect, _}) => {
```
glennsl:
```suggestion
    test("Doesn't remove anything if cursor at the beginning", ({expect, _}) => {
```

test/Components/InputModelTests.re
```diff
@@ -265,6 +263,48 @@ describe("handleInputS#handleInput", ({describe, _}) => {
...
+           let (text, newSelection) = runInputHandler(selection, key);
+
+           expect.equal(expected, newSelection);
+           expect.string(text).toEqual(
+             "Som interesting. Test. String. Isn't it? Maybe",
+           );
+         });
+
+         test(
+           "Doesn't remove character if cursor at the beginnig",
```
glennsl:
```suggestion
           "Doesn't remove character if cursor at the beginning",
```

(I can apparently see typos now, just as long as they're not mine)

test/Components/InputModelTests.re
```diff
@@ -265,6 +263,48 @@ describe("handleInputS#handleInput", ({describe, _}) => {
...
+    |> List.iter(key => {
+         test("Removes character on the left of cursor", ({expect, _}) => {
+           let selection = collapsedSelection(4);
+           let expected = collapsedSelection(3);
+
+           let (text, newSelection) = runInputHandler(selection, key);
+
+           expect.equal(expected, newSelection);
+           expect.string(text).toEqual(
+             "Som interesting. Test. String. Isn't it? Maybe",
```
glennsl:
```suggestion
             "Some interesting. Test. String. Isn't it? Maybe",
```

test/Components/InputModelTests.re
bryphe: Nice catch! Thanks @glennsl

test/Components/InputModelTests.re
bryphe: 😆 

We need to bring in a spell-checker extension! Would be a good test case for the exthost

PR #1496, opened 2020-03-23
Feature - Keybindings: Input sequences

bryphe:
This updates our key-bindings to be powered by [`editor-input`](https://github.com/onivim/editor-input), a library completely agnostic of Onivim 2.

This gives us a few new features:

__Support for both VSCode and Vim binding support__

Our current parsing logic was quirky and not very forgiving - sensitive to even the placement of shift characters (ie, `<C-S-P>` vs `<S-C-P>` would give different parse results).

The new parser, provided by `editor-input`, is much more forgiving:
- It allows for VSCode style bindings, ie, `Ctrl+j`
- Styles can be mixed and matched, even, like `Ctrl+j<C-k>`
- It still supports the keybindings defined by the previous format - our defaults do not require modification

__Support for key sequences__

`editor-input` supports key sequences, like:
```
  { "key": "jk", "command": "vim.esc", when: "insertMode" }
```
or:
```
  { "key": "<C-\\><C-N>", "command": "explorer.toggle", when: "editorTextFocus" }
```

Although it is not currently wired up to Vim (ie, `inoremap` for sequences is still problematic), this at least provides a way to hook up bindings for `jj` -> `esc` and `jk` -> `esc` via `keybindings.jon` - common bindings for Vim users.

Existing keybindings should work as-is, but now we can support some of these additional bindings, too.

along with the potential for others:
- True key chording
- Binding using physical scancodes
- User interface faculty for showing pending keybindings (ie, spacevim/spacemacs style key discovery)

__TODO:__
- [ ] Fix `KeybindingsTest` - use new API
- [ ] Fix `AltGr` handling on Windows
- [ ] Describe acceptable key formats in documentation


src/Input/Handler.re
```diff
@@ -99,15 +97,20 @@ let keyPressToString =
...
+    let shiftKey = Keymod.isShiftDown(keymod);
+    let altKey = Keymod.isAltDown(keymod);
+    let ctrlKey = Keymod.isControlDown(keymod);
+    let altGr = Keymod.isAltGrKeyDown(keymod);*/
+
+  /*let (altGr, ctrlKey, altKey) =
+    // TODO: Move to place where we convert revery key -> EditorInput.key
     switch (Revery.Environment.os) {
     // On Windows, we need to do some special handling here.
     // Windows has this funky behavior where pressing AltGr registers as RAlt+LControl down - more info here:
```
bryphe: Need to port this over to be handled by the new logic

PR #1497, opened 2020-03-23
Feature - Keybindings: Input sequences + editor-input

bryphe:
This updates our key-bindings to be powered by [`editor-input`](https://github.com/onivim/editor-input), a library completely agnostic of Onivim 2.

This gives us a few new features:

__Support for both VSCode and Vim binding support__

Our current parsing logic was quirky and not very forgiving - sensitive to even the placement of shift characters (ie, `<C-S-P>` vs `<S-C-P>` would give different parse results).

The new parser, provided by `editor-input`, is much more forgiving:
- It allows for VSCode style bindings, ie, `Ctrl+j`
- Styles can be mixed and matched, even, like `Ctrl+j<C-k>`
- It still supports the keybindings defined by the previous format - our defaults do not require modification

__Support for key sequences__

`editor-input` supports key sequences, like:
```
  { "key": "jk", "command": "vim.esc", when: "insertMode" }
```
or:
```
  { "key": "<C-\\><C-N>", "command": "explorer.toggle", when: "editorTextFocus" }
```

Although it is not currently wired up to Vim (ie, `inoremap` for sequences is still problematic), this at least provides a way to hook up bindings for `jj` -> `esc` and `jk` -> `esc` via `keybindings.jon` - common bindings for Vim users.

Existing keybindings should work as-is, but now we can support some of these additional bindings, too.

along with the potential for others:
- True key chording
- Binding using physical scancodes
- User interface faculty for showing pending keybindings (ie, spacevim/spacemacs style key discovery)

__TODO:__
- [x] Fix `KeybindingsTest` - use new API
- [x] Fix KeyDisplayer
- [x] Fix `AltGr` handling on Windows
- [x] Integration test for `jj` binding
- [x] Describe acceptable key formats in documentation


src/Input/Keybindings.re
```diff
@@ -84,9 +91,49 @@ module Internal = {
...
+  | Input.Execute(cmd) => Command(cmd)
+  | Input.Unhandled(key) => Unhandled(key)
+  | Input.Text(text) => Text(text);
+
+let mapEffects = List.map(mapEffect);
+
+let keyDown = (~context, ~key, bindings) => {
+  let (bindings, effects) = Input.keyDown(~context, ~key, bindings);
+
+  let mappedEffects = mapEffects(effects);
```
glennsl: `mapEffect`, `mapEffects` and `mappedEffects` is very non-descriptive. The most significant information, what it maps from and to, is missing.

src/Input/Keybindings.re
```diff
@@ -111,22 +158,117 @@ module Legacy = {
...
-       });
-  | _ => Error("Unable to parse keybindings - not a JSON array.")
+let keyToSdl =
+  EditorInput.Key.(
+    {
+      fun
+      | Escape => "Escape"
+      | Return => "Return"
+      | Up => "Up"
+      | Down => "Down"
```
glennsl: `Sdl` is really not a thing I expect to be able to convert a key to. Perhaps it would make sense to further qualify it as `SdlName`?

src/Input/Keybindings.re
```diff
@@ -111,22 +158,117 @@ module Legacy = {
...
+  | 0 => None
+  | x => Some(x);
+
+let getKeycode = keycodeStr => {
+  keycodeStr |> keyToSdl |> Sdl2.Keycode.ofName |> codeToOpt;
+};
+
+let getScancode = scancodeStr => {
+  scancodeStr |> keyToSdl |> Sdl2.Scancode.ofName |> codeToOpt;
+};
```
glennsl: `keycodeStr` and `scancodeStr` does not seem to be strings, as `keyToSdl` takes a variant.

src/Input/Keybindings.re
```diff
@@ -111,22 +158,117 @@ module Legacy = {
...
+      | NumpadDivide => "Keypad //"
+      | NumpadDecimal => "Keypad ."
+      | Character(c) => String.make(1, c) |> String.uppercase_ascii;
+    }
+  );
+
+let codeToOpt =
+  fun
+  | 0 => None
+  | x => Some(x);
```
glennsl: This function does more to obfuscate the consumer code than to make it more readable, as the name really doesn't say much. It would be better to just inline it IMO.

test/Input/KeybindingsTests.re
```diff
@@ -48,6 +48,9 @@ bindings: [
 |}
   |> Yojson.Safe.from_string;
 
+let contextWithEditorTextFocus =
+  [("editorTextFocus", true)] |> List.to_seq |> Hashtbl.of_seq;
```
glennsl:
```suggestion
  Seq.return(("editorTextFocus", true))|> Hashtbl.of_seq;
```

src/Input/Keybindings.re
```diff
@@ -84,9 +91,49 @@ module Internal = {
     };
 };
 
-let empty = [];
+module Input =
+  EditorInput.Make({
+    type context = Hashtbl.t(string, bool);
+    type payload = string;
```
glennsl: It's also very unclear for me what `payload` is. What is an example of a payload value in this context?

bryphe:
> Why is it called editor-input? 

It handles input management for an editor, which includes parsing keybindings as well as other functionality - handling key sequences, chords, etc. Do you have a concrete suggestion here?

> What's editor-specific about it?

It is about as editor-specific as `editor-core-types`, which is the precedence I was following.

glennsl:
> It is about as editor-specific as editor-core-types, which is the precedence I was following.

Right, that was created and named more because it was needed for functionality shared across various editor-specific components though, not because it necessarily offered editor-specific functionality. Although in the future it certainly could.

> It handles input management for an editor, which includes parsing keybindings as well as other functionality - handling key sequences, chords, etc.

None of that sounds particularly editor-specific though. I might very well want key sequences and chording for a git UI or DB management client as well.

> Do you have a concrete suggestion here?

Well, that depends on what it actually offers. I would definitely like to have a general-purpose keybinding parser library though, named accordingly, so maybe it would even make sense to split it up further?

bryphe:
@glennsl - re: the `editor-input` naming - I think we're seeing this from two different perspectives and I don't want us to get into another talk-past situation - let's table this particular item for now and we can discuss in our next weekly meeting.

I think it'll be more productive to discuss this particular item "in person" (or at least, "in video" 😄 )

test/Input/KeybindingsTests.re
bryphe: Cool!

src/Input/Keybindings.re
bryphe: Ah, ya, prior to https://github.com/onivim/editor-input/pull/6 they were `strings`, but missed updating these names once they came in. Fixed

src/Input/Keybindings.re
bryphe: Ah, `payload` is a poor name. I refactored it to `command` in this PR: https://github.com/onivim/editor-input/pull/9

`action` might also be a reasonable name. It's an object you associate with a key binding, that, when the key binding is engaged, it provides with an `Execute` effect.

For example, if you have a binding `j`, like:
```
let alwaysEnabled = _context => true;
let bindings = 
Input.empy |> Input.addBinding("j", alwaysEnabled, "jPressed");

let (_bindings, effects)  = Input.keyDown(~context, ~jKey, bindings);
// effects == [Execute("jPressed")]
```

src/Input/Keybindings.re
bryphe: I agree, but I don't have ideas for better names (except for something overly verbose like `mapEditorInputEffectToOnivim2Effect`, and I'm not sure that's great). 

Since it's an internal implementation detail, and there isn't a better name suggested, I'll leave it - but I moved it to the `Internal` module and tried to describe it via a comment.

src/Input/Keybindings.re
glennsl: Ah, yeah `command` makes sense I think!

src/Input/Keybindings.re
glennsl: It might be hard to name because it really doesn't seem to do anything "useful", other than perhaps avoid some technical limitation.

I wonder why the local `effect` type isn't just an alias of the one from `Input` though, as they seem to have identical definitions. That would avoid the need to map it at all.

bryphe: Good point. The challenge was that the `effect` type was tucked inside the functorized-`EditorInput.Make`. I extracted it out so that it could be aliased from `EditorInput`, which removed these extra maps 👍

src/Input/Keybindings.rei
```diff
@@ -2,10 +2,7 @@ type t;
 
 let empty: t;
 
-type effect =
-  | Command(string)
-  | Text(string)
-  | Unhandled(EditorInput.KeyPress.t);
+type effect = EditorInput.effect(string);
```
glennsl: You don't have to expose the alias in the interface if you don't want to. It can remain an implementation detail as long as you provide the full definition in the implementation, in addition to the alias.

So If in the implementation you have:

```
type effect = EditorInput.effect(string) =
  | Command(string)
  | Text(string)
  | Unhandled(EditorInput.KeyPress.t);
```

Then you can leave the interface as:

```suggestion
type effect =
  | Command(string)
  | Text(string)
  | Unhandled(EditorInput.KeyPress.t);
``

and it won't be a breaking change for consumers.

File: src/Store/VimStoreConnector.re [hunk 8]
```diff
@@ -504,9 +504,22 @@ let start =
...
+    && !String.equal(key, "<A-SHIFT>")
+    && !String.equal(key, "<D-SHIFT>")
+    && !String.equal(key, "<D->")
+    && !String.equal(key, "<D-A->")
+    && !String.equal(key, "<D-S->")
+    && !String.equal(key, "<C->")
+    && !String.equal(key, "<A-C->")
+    && !String.equal(key, "<SHIFT>")
+    && !String.equal(key, "<S-C->");
+  };
```
glennsl: This name is ambiguous. Does it check if it is or is not a "vim key"?

How about this (or the opposite)?

```suggestion
  let isVimKey  = key => {
    !String.equal(key, "<S-SHIFT>")
    && !String.equal(key, "<A-SHIFT>")
    && !String.equal(key, "<D-SHIFT>")
    && !String.equal(key, "<D->")
    && !String.equal(key, "<D-A->")
    && !String.equal(key, "<D-S->")
    && !String.equal(key, "<C->")
    && !String.equal(key, "<A-C->")
    && !String.equal(key, "<SHIFT>")
    && !String.equal(key, "<S-C->");
  };
```

src/Input/Keybindings.rei
bryphe: Hmm, I tried this, and I'm getting an error message:
```
# Unformatted Error Output:
# File "src/Input/Keybindings.re", line 11, characters 0-111:
# Error: This variant or record definition does not match that of type
#          string EditorInput.effect
#        They have different arities.


 ERROR  src/Input/Keybindings.re:11:0-14:35 

  8 ┆ 
  9 ┆ open EditorInput;
 10 ┆ 
 11 ┆ type effect = EditorInput.effect(string) =
 12 ┆ | Command(string)
 13 ┆ | Text(string)
 14 ┆ | Unhandled(EditorInput.KeyPress.t);
```

bryphe: In this commit: https://github.com/onivim/oni2/pull/1497/commits/6896763d1d93cdf0cea71ba11af90e1897052d9d

Perhaps I'm making a silly mistake? It would be nice to have this.

glennsl: Hmm, I guess this is a limitation of it being polymorphic. That's unfortunate.

Could you put it back in the functor and just instantiate that earlier? Or does that create other problems?

Otherwise it really isn't that important.

bryphe: > Could you put it back in the functor and just instantiate that earlier? Or does that create other problems?

I believe this would create other problems, because the functorized-module is completely internal. I'm not sure how the `rei` would pick up that value. I think it would work if the type was abstract (like `type effect;`), but not in the case where we want to specify variant.

> Or you could make this polymorphic as well I guess, although that would still be pretty weird from a consumer POV. 

Right, I was trying that out, but it seems pretty awkward. I think the `EditorInput.effect(string)` is the 'best' balance at the moment. The KeyBindings interface also uses `EditorInput.KeyPress.t`.

glennsl: > I'm not sure how the rei would pick up that value.

Why would it need to? That's what we're trying to avoid, isn't it? I'm thinking it'd be just like the commit you reference above, except you'd instantiate the functor before defining `effect`, so you can alias it to the now non-polymorphic `Input.effect`. The `rei` shouldn't need to change.

> The KeyBindings interface also uses EditorInput.KeyPress.t.

Right. You could probably alias and hide that type too of course, but again, no need to spend a lot of time on this. It's more just nice-to-have and of general interest I think.

bryphe: > I'm thinking it'd be just like the commit you reference above, except you'd instantiate the functor before defining effect, so you can alias it to the now non-polymorphic Input.effect. 

Ah, right - just tried it out and it seems to work 👍

lucis:
Hey, @bryphe. Should you update the [Keybindings docs](https://onivim.github.io/docs/configuration/key-bindings) perhaps?

Nice feature, btw!

PR #1499, opened 2020-03-24
Carry severity and extension id from extension message to notification

glennsl:
This creates the appropriate notification based on the severity of the message coming from the extension, and appends the extension id to the message to be able to identify where the message comes from.

It's a bit of an intermediary improvement, as we should probably enrich the notification with a `source` field and need to add support for commands and such eventually.

### DIFF
--- a/development_extensions/oni-dev-extension/extension.js
+++ b/development_extensions/oni-dev-extension/extension.js
@@ -94,14 +94,12 @@ function activate(context) {
         }
     }));
 
-    // The command has been defined in the package.json file
-    // Now provide the implementation of the command with  registerCommand
-    // The commandId parameter must match the command field in package.json
-    cleanup(vscode.commands.registerCommand('developer.oni.showNotification', () => {
-        // The code you place here will be executed every time your command is executed
+    cleanup(vscode.commands.registerCommand('developer.oni.showWarning', () => {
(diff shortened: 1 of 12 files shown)


test/Extensions/ExtensionClientTest.re
```diff
@@ -178,7 +190,7 @@ describe("ExtHostClient", ({describe, _}) => {
         );
 
       withExtensionClient(
-        ~onShowMessage,
```
bryphe: Thanks for removing this and switching it to the `dispatch` model 👍

development_extensions/oni-dev-extension/extension.js
```diff
@@ -94,14 +94,12 @@ function activate(context) {
...
 
-    // The command has been defined in the package.json file
-    // Now provide the implementation of the command with  registerCommand
-    // The commandId parameter must match the command field in package.json
-    cleanup(vscode.commands.registerCommand('developer.oni.showNotification', () => {
-        // The code you place here will be executed every time your command is executed
-
-        // Display a message box to the user
-        vscode.window.showInformationMessage('Hello from extension!');
+    cleanup(vscode.commands.registerCommand('developer.oni.showWarning', () => {
```
bryphe: Nice!

PR #1501, opened 2020-03-24
Remove assets/configuration/dune

glennsl:
After nuking my local repository dune-configurator has suddenly started complaining about there being no rule for "setup.json", referenced in this file. Since there doesn't seem to be any reason for this any longer it should probably be removed.


glennsl:
Or it's created by `esy boostrap`... Nevermind!

PR #1503, opened 2020-03-25
Feature: Comment Toggling

glennsl:
Peek 2020-03-25 19-31

### DIFF
--- a/src/Extensions/LanguageConfiguration.rei
+++ b/src/Extensions/LanguageConfiguration.rei
@@ -23,6 +23,8 @@ module AutoClosingPair: {
 type t = {
   autoCloseBefore: list(string),
   autoClosingPairs: list(AutoClosingPair.t),
+  lineComment: option(string),
+  blockComment: option((string, string)),
 };
 
 let default: t;
--- a/src/Store/VimStoreConnector.re
+++ b/src/Store/VimStoreConnector.re
@@ -744,6 +744,17 @@ let start =
         | _ => ()
         };
 
+        // Set configured line comment
+        editorBuffer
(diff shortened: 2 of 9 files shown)


src/Store/VimStoreConnector.re
```diff
@@ -744,6 +744,17 @@ let start =
...
+        editorBuffer
+        |> OptionEx.flatMap(Core.Buffer.getFileType)
+        |> OptionEx.flatMap(
+             Ext.LanguageConfigurationLoader.get_opt(languageConfigLoader),
+           )
+        |> OptionEx.flatMap((config: Ext.LanguageConfiguration.t) =>
+             config.lineComment
+           )
+        |> Option.iter(Vim.Options.setLineComment);
+
```
glennsl: This is in `synchronizeEffect`. Not sure if that's the best place to put it, but it's not a particularly expensive operation so this is probably fine.

src/Extensions/LanguageConfiguration.rei
```diff
@@ -23,6 +23,8 @@ module AutoClosingPair: {
 type t = {
   autoCloseBefore: list(string),
   autoClosingPairs: list(AutoClosingPair.t),
+  lineComment: option(string),
+  blockComment: option((string, string)),
```
bryphe: Nice!

src/Store/VimStoreConnector.re
bryphe: Seems like the best spot for now 👍 I'd like to eventually revisit the 'synchronization' and instead 'push' values into libvim... but for now, this is the right spot.

glennsl:
It was a long road for such a tiny feature, but I certainly learned a lot in the process!

PR #1504, opened 2020-03-25
Guess indentation on heuristically assumed paste into empty buffer

### LINKED ISSUE #1190: Unable to set tabstop after buffer is initially opened
There seems to be no way of setting `tabstop`/`shiftwidth` once a buffer has been opened, short of restarting Oni entirely (since there seems to be no way to actually close a buffer either) after changing the default or changing the content of the file so it detects it correctly. This is rather inconvenient, although not a terribly common occurrence.

Being able to properly close a buffer would go a long way to alleviate the inconvenience. We could also re-run the indentation guesser on certain conditions, like a paste into an empty buffer, or on command. Or we could just have a command to set it manually or have it respect `:setlocal ...`, whichever is easiest.

glennsl:
Fixes #1190 (with a bit of help from #1465)

### DIFF
--- a/src/Store/IndentationStoreConnector.re
+++ b/src/Store/IndentationStoreConnector.re
@@ -67,31 +67,33 @@ let start = () => {
    * Otherwise, it sets indentation values for a buffer,
    * if it hasn't been done yet.
    */
-  let checkIndentationEffect = (state: State.t, bufferId) => {
-    let maybeBuffer = Buffers.getBuffer(bufferId, state.buffers);
-    switch (maybeBuffer) {
-    | Some(buffer) =>
-      if (!Buffer.isIndentationSet(buffer)) {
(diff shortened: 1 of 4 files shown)


src/Store/IndentationStoreConnector.re
```diff
@@ -67,31 +67,33 @@ let start = () => {
...
+          state,
+          checkIndentationEffect(state, buffer),
+        )
+      | _ => (state, Isolinear.Effect.none)
+      }
+
+    | Actions.BufferUpdate({oldBuffer, newBuffer, _})
+        when
+          Buffer.getNumberOfLines(oldBuffer) <= 1
+          && Buffer.getNumberOfLines(newBuffer) > 2 => (
```
glennsl: Said heuristic here

PR #1505, opened 2020-03-26
Feature - Terminal Normal Mode - Part 1: Input / mode context

bryphe:
This is the first part of the 'terminal-normal-mode' implementation.

This augments our idea of a `Mode` - libvim/reason-libvim are agnostic of any sort of terminal.

In other words - 'terminal insert' mode completely bypasses Vim, and when we switch to terminal-normal mode, we switch to a read-only buffer, with contents provided by the terminal screen.

This means we need a way to represent 'modes' that is a superset of libvim/reason-libvim - so this adds an augmented mode type in Core, as well a 'manager' (with design inspired by the FocusManager) - that takes the current state and lets us know the real mode Onivim 2 is in.

This is primarily a refactoring to lay the foundation for creating the 'terminal-normal' mode, but there are a couple of functional differences:
- The `normalMode` and `terminalFocus` context keys have been added.
- When using a terminal, the mode is now correctly shown as 'Terminal Insert' (instead of 'Normal')


src/Model/Selectors.re
```diff
@@ -67,3 +67,23 @@ let getActiveConfigurationValue = (state: State.t, f) => {
     Configuration.getValue(~fileType, f, state.configuration);
   };
 };
+
+let getActiveTerminalIdOpt = (state: State.t) => {
```
glennsl: I don't think it's necessary to add an `Opt` suffix. That convention comes from the standard library where the implied default is to raise an exception. The default ought to be to return an `option` though, and instead specify `Exn` if it raises. And since we don't have the historical baggage of the standard library, I think we should follow the more modern convention.

src/Model/ModeManager.re
```diff
@@ -0,0 +1,15 @@
+let current: State.t => Oni_Core.Mode.t =
+  (state: State.t) =>
+    if (Selectors.terminalIsActive(state)) {
+      TerminalInsert;
+    } else {
+      switch (state.mode) {
```
glennsl: I think it would be a good idea to rename `state.mode` to `state.vimMode`. Otherwise this is going to cause a lot of confusion I think.

src/Model/Selectors.re
```diff
@@ -67,3 +67,23 @@ let getActiveConfigurationValue = (state: State.t, f) => {
...
+       | _ => None
+       }
+     );
+};
+
+let terminalIsActive = (state: State.t) =>
+  state
+  |> getActiveTerminalIdOpt
+  |> Option.map(_ => true)
+  |> Option.value(~default=false);
```
glennsl:
```suggestion
  getActiveTerminalIdOpt(state) == Some(true);
```

bryphe: Much nicer!

src/Model/Selectors.re
bryphe: Yes, good point! I'd prefer the default to be the one that is strongly-typed and backed by the type system - to ensure we're handling errors and not just falling through and crashing.

I started a rough draft of a 'style guide' to document some of these guidelines.

src/Model/ModeManager.re
bryphe: Good idea 👍

src/Model/Selectors.re
glennsl: Nice. Looks great!

PR #1506, opened 2020-03-26
Fix #496: Crash when pasting from commandline

### LINKED ISSUE #496: Paste into command window freezes program
I had `set syntax=html` in my clipboard, pressed `:`, and pasted into the command window that showed up. The program just locked up.

Fresh clone of project today and just compiled (First time to get it to compile!).

bryphe:
__Issue:__ When pasting via `<C-v>` in the command line, we'd crash.

__Defect:__ `<C-v>` in the command line actually engages the 'insert-literal' behavior, which blocks for input (causing a hang). This root cause is addressed in https://github.com/onivim/libvim/pull/191 , removing that functionality (we can bring it back in the Onivim 2 input layer if it is desired).

__Fix:__ In the Onivim 2 side, hook up our paste command to work with the command line too.

Related to #1233 , and might not be necessary at all with that change (especially if we move more of the command line behavior to Onivim 2)

__TODO:__

- [x] Depends on https://github.com/onivim/libvim/pull/191 
- [x] ...and being brought in via `reason-libvim`

Fixes #496

### DIFF
--- a/docs/docs/configuration/key-bindings.md
+++ b/docs/docs/configuration/key-bindings.md
@@ -124,7 +124,7 @@ Onivim-specific contexts:
 | Context Name | True When | 
 | --- | --- |
 | `insertMode` |  The active editor is in `insert` mode |
-| `commandLineMode` | The Vim commandline is open |
+| `commandLineFocus` | The Vim commandline is open |
 | `menuFocus` | A pop-up menu has focus |
 
 ## Commands
--- a/docs/docs/configuration/key-bindings.md
+++ b/docs/docs/configuration/key-bindings.md
@@ -124,7 +124,7 @@ Onivim-specific contexts:
 | Context Name | True When | 
 | --- | --- |
 | `insertMode` |  The active editor is in `insert` mode |
-| `commandLineFocus` | The Vim commandline is open |
+| `commandLineMode` | The Vim commandline is open |
 | `menuFocus` | A pop-up menu has focus |
 
 ## Commands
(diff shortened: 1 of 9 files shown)


docs/docs/configuration/key-bindings.md
```diff
@@ -124,7 +124,7 @@ Onivim-specific contexts:
 | Context Name | True When | 
 | --- | --- |
 | `insertMode` |  The active editor is in `insert` mode |
-| `commandLineFocus` | The Vim commandline is open |
```
bryphe: I didn't actually see `commandLineFocus` used anywhere

src/Store/InputStoreConnector.re
```diff
@@ -58,6 +58,10 @@ let conditionsOfState = (state: State.t) => {
     Hashtbl.add(ret, "editorTextFocus", true);
   | (false, Vim.Types.Visual) => Hashtbl.add(ret, "visualMode", true)
   | (false, _) => Hashtbl.add(ret, "editorTextFocus", true)
+  | (true, Vim.Types.CommandLine) =>
+    // LEGACY: Support `commandLineFocus` too.
+    Hashtbl.add(ret, "commandLineFocus", true);
+    Hashtbl.add(ret, "commandLineMode", true);
```
glennsl: Why add `commandLineMode` if it's exactly the same as the existing `commandLineFocus`?

bryphe: To be consistent with the other modes - we'll eventually have:
- `insertMode`
- `normalMode`
- `visualMode`
...

It seems inconsistent... although I guess it's debatable if `commandLine` _should_ be a mode. If we decide to build a 'first-class' command-line mode (like, powered by our text input model + menu infra, with completion, etc), I see that as a good argument for `commandLineFocus` instead though. I'll switch it back.

glennsl: Ah, I see. Yeah, from a normal user POV `commandlineFocus` makes much more sense I think, but from a vim POV `commandlineMode` definitely makes more sense. I'd lean towards favoring the user POV too though.

PR #1508, opened 2020-03-27
Docs: Start 'Moving Around' section

bryphe:


CrossR:
Not blocking per say, but we need to actually setup the packaging of all/some subset of the runtime files (which I think includes vimtutor, but also help text etc).

Its a worth while thing to do just for the help text though, so should be done regardless.

bryphe:
> Not blocking per say, but we need to actually setup the packaging of all/some subset of the runtime files (which I think includes vimtutor, but also help text etc).

Exactly! We  need a plan for how to include these runtime files. Right now, the _only_ thing we bring in from `libvim` is the compiled library.

glennsl:
> Runtime: We need to point libvim to these runtime files. We'll need to start hooking up the runtime path files (so that libvim knows where to look to find the docs.

For the help files, sure. But the tutor is really just a normal text file in a normal buffer, isn't it? Except you can't save it of course. And we already have assets, so it seems like we could just bundle it as any other asset, add a non-vim command to open it. Then the last part is just is just preventing the buffer from being saved, which I think I just saw a PR for in libvim, perhaps not entirely coincidentally.

We might want to change this when we pull in the rest of the runtime files of course, but if we really want the tutor now, it seems to me like it would already be really easy to do.

bryphe:
> And as we already have other assets, it seems like we could just bundle it the same way and add a non-vim command to open it.

Sounds like a good compromise!

> Then the last part is just is just preventing the buffer from being saved, which I think I just saw a PR for in libvim, perhaps not entirely coincidentally.

Right, we can set the read-only flag now via `Vim.Buffer.setReadOnly` API

bryphe:
I'll track bringing in the vimtutor here: https://github.com/onivim/oni2/issues/1513

PR #1511, opened 2020-03-28
Refactor - Editor: Per-editor font setting

bryphe:
This is a refactoring in preparation for terminal-normal mode - we currently have two different font settings (one for the editor, and one for the terminal). 

However, for the terminal, it's a bit jarring if we're switching back and forth between those fonts - without any changes, we'd switch from the terminal font to the editor font when entering 'terminal normal mode'.

This is an incremental refactoring - to allow editors to have individual settings for fonts. This decouples the font-sizing from the `EditorMetrics.t` (which hopefully we can get rid of completely, soon - and allow `Editor.t` to stand on its own!)

### DIFF
--- a/src/Feature/Editor/Editor.re
+++ b/src/Feature/Editor/Editor.re
@@ -71,8 +71,7 @@ let getId = model => model.editorId;
 let getLineHeight = editor => editor.font.measuredHeight;
 let getCharacterWidth = editor => editor.font.measuredWidth;
 
-let pixelPositionToLineColumn =
-    (view, pixelX, pixelY) => {
+let pixelPositionToLineColumn = (view, pixelX, pixelY) => {
   let line = int_of_float((pixelY +. view.scrollY) /. getLineHeight(view));
   let column =
     int_of_float((pixelX +. view.scrollX) /. getCharacterWidth(view));
@@ -101,8 +100,7 @@ let getVerticalScrollbarMetrics =
   {thumbSize, thumbOffset, visible: true};
 };
 
-let getHorizontalScrollbarMetrics =
-    (view, availableWidth) => {
+let getHorizontalScrollbarMetrics = (view, availableWidth) => {
   let totalViewWidthInPixels =
     float_of_int(view.maxLineLength) *. getCharacterWidth(view);
   let availableWidthF = float_of_int(availableWidth);
--- a/src/Feature/Editor/Editor.re
+++ b/src/Feature/Editor/Editor.re
@@ -19,7 +19,6 @@ type t = {
   viewLines: int,
   cursors: [@opaque] list(Vim.Cursor.t),
   selection: [@opaque] VisualRange.t,
-
   font: [@opaque] Service_Font.font,
 };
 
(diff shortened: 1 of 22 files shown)


src/Model/EditorMetricsReducer.re
```diff
@@ -7,16 +7,14 @@ open Feature_Editor;
...
-      pixelWidth,
-      pixelHeight,
-    }
-  | EditorFont(Service_Font.FontLoaded({measuredHeight, measuredWidth, _})) => {
-      ...v,
-      lineHeight: measuredHeight,
-      characterWidth: measuredWidth,
-    }
+  | EditorGroupSetSize(_, {pixelWidth, pixelHeight}) =>
+    EditorMetrics.{pixelWidth, pixelHeight}
```
bryphe: Next step would be to push `pixelWidth` and `pixelHeight` into the editor, so we can remove `EditorMetrics.t` completely.

src/Model/EditorGroups.rei
```diff
@@ -9,4 +9,6 @@ let getActiveEditorGroup: t => option(EditorGroup.t);
 let isActive: (t, EditorGroup.t) => bool;
 let isEmpty: (int, t) => bool;
 
-let reduce: (t, Actions.t) => t;
+let reduce: (~defaultFont: Service_Font.font, t, Actions.t) => t;
+
+let setBufferFont: (~bufferId: int, ~font: Service_Font.font, t) => t;
```
bryphe: This isn't used yet, but the plan is to leverage this in the `VimStoreConnector`'s `updater` method - when we get the command to switch to terminal-normal mode, we'd set the font for all editors related to that buffer to the terminal font.

It's a bit hacky (and not quite correct - if the user changes the terminal font, the editors in terminal normal mode won't get updated). The problem is the `Feature_Editor` doesn't know anything about terminals or terminal normal mode. We might need to add a way to have some Onivim-specific context (like, should it use the terminal font or editor font?) so that we can update it correctly.

src/Feature/Editor/Editor.re
```diff
@@ -66,25 +68,27 @@ let getPrimaryCursor = model =>
 
 let getId = model => model.editorId;
 
-let pixelPositionToLineColumn =
-    (view, metrics: EditorMetrics.t, pixelX, pixelY) => {
-  let line = int_of_float((pixelY +. view.scrollY) /. metrics.lineHeight);
+let getLineHeight = editor => editor.font.measuredHeight;
```
glennsl: @lessp has been working on a line height setting I believe, and so removing the line height property entirely is pulling the rug out underneath that work a bit.

lessp: This is fine, I haven't started any real work yet, more familiarising myself with the code base. 👍 Thanks for the consideration though! @glennsl

bryphe: Ah sorry @lessp (and thanks @glennsl for checking!)

This would actually be a good reference for where we use this - we use the `measuredHeight` as a proxy for line height, but it'd be to have a real configuration for this.

We'd want an additional `lineHeight` that we set per-editor (via a configuration), and we'd update our `getLineHeight` to account for it, ie:
```
let getLineHeight = editor => editor.font.measuredHeight *. editor.lineHeight;
```
(although we'd need to special case the `0` case - and just use `editor.font.measuredHeight` for that).

PR #1514, opened 2020-03-28
Feature - Terminal: Normal-mode transition

bryphe:
This implements a transition from normal-mode to insert-mode in the terminal (and vice versa).

New key-bindings to support this:
- <kbd>Control</kbd><kbd>\\</kbd>, <kbd>Control</kbd><kbd>N</kbd> - switch to terminal normal mode
- <kbd>A</kbd>, <kbd>I</kbd>, <kbd>O</kbd>, etc - switch back to terminal insert mode.

This implements the transition from insert <-> normal, but it's a bit jarring (since the font changes and the terminal colors are not preserved):

__TODO:__
- [x] Remove non-blank lines from terminal

__NEXT STEPS:__
- Add terminal section to docs
- Preserve colors from terminal when switching from insert -> normal
- Integrate with https://github.com/onivim/oni2/pull/1511 to use terminal font setting


src/Model/BufferRenderer.re
```diff
@@ -14,6 +14,7 @@
 type terminal = {
   title: string,
   id: int,
+  insertMode: bool,
```
bryphe: Deciding where to manage this state was a bit tricky (whether the terminal is in insert mode or not).

A few possibilities:
- We could store on the `Buffer`, but the buffer doesn't know about terminals...
- We could store it on the `Editor`, but the editor doesn't know about terminals...

For now, I put this in the `Terminal` buffer renderer state - when it is in insert mode, we use the terminal renderer. When it is in normal mode, we use the editor surface renderer. We also have to orchestrate with the buffer management in Vim to ensure the buffer is populated with the latest lines - we handle this in the transition from insert mode -> normal mode.

glennsl: Seems to me like a good place to put it. Are there any issues with having it here?

src/Service/Terminal/Service_Terminal.rei
```diff
@@ -36,3 +36,5 @@ module Sub: {
 module Effect: {let input: (~id: int, string) => Isolinear.Effect.t(msg);};
 
 let handleExtensionMessage: ExtHostClient.Terminal.msg => unit;
+
+let getScreenOpt: int => option(ReveryTerminal.Screen.t);
```
glennsl: Probably from a merge (which makes PRs a bit hard to review sometimes, btw), but as we've discussed recently the `Opt` is unnecessary.

src/Store/VimStoreConnector.re
```diff
@@ -959,6 +991,43 @@ let start =
...
+      let maybeTerminalId =
+        maybeBufferId
+        |> Option.map(id =>
+             BufferRenderers.getById(id, state.bufferRenderers)
+           )
+        |> OptionEx.flatMap(
+             fun
+             | BufferRenderer.Terminal({id, _}) => Some(id)
+             | _ => None,
+           );
```
glennsl: This is a pattern I';ve seen a few times. Might be worth having a `BufferRenderer.ifTerminal` or something like that. Or maybe it's a sign of larger structural issues. No biggie though, but seems inconvenient if it happens frequently.

glennsl: Actually, the number of `option`s being juggled here does suggest there are larger structural issues. That certainly doesn't mean the solution is in any way obvious, but I find there's usually a better way to organize things when it gets this hairy. Although it also tends to require a change in the model to more accurately reflect the actual domain.

src/Model/BufferRenderer.re
bryphe: Nope! I think this is the best place, for now. It just goes back to some of the issues we discussed about the buffer renderers in general (ie, they exist as an independent entity of buffers and editors - so I think those problems and caveats apply).

src/Service/Terminal/Service_Terminal.rei
bryphe: Ah, right, I added this prior to our discussion earlier - fixed now 👍

src/Store/VimStoreConnector.re
bryphe: > Actually, the number of options being juggled here does suggest there are larger structural issues.

Right, I thjink this is true - I think this goes back to our discussion about the relationship between the Buffer Renderer <-> Buffer (and <-> Editor).

I'm thinking that a better way to model this would be to have the 'renderer' associated with the 'editor'. An editor <-> renderer relationship is always 1-to-1, whereas a buffer to renderer might not be (ie - for something like markdown preview - we'd have two editors associated with the same markdown file - but one would use the default editor renderer, and one would use a markdown-preview-renderer). That could cut down one of these `option`s being juggled.

The other thorn is that the editor has a reference to a `bufferId`, that may or may not exist - so we still have two `option`s. Need to think more about having our domain model more accurately reflect this. Because the editor <-> buffer relationship is 1-to-many, it doesn't make sense for the editor to store the buffer (nor the buffer to the list of editors referencing it) - so we might need some higher-level model to accurately represent the domain.

But maybe if we do that first-level of renderer refactoring (move the renderer into some context the editor knows about it), and continue to refactor around this area, something will pop out...

bryphe: I'll leave this as-is for now - but there are definitely domain-level issues we need to improve here

glennsl: Yep, no problem.

I really should have recognized that this was the bugger/edit/renderer issue we've talked about. I like your thinking though.

> Because the editor <-> buffer relationship is 1-to-many, it doesn't make sense for the editor to store the buffer (nor the buffer to the list of editors referencing it)

Is it really 1-to-many? Or is it either 1-to-1 or 1-to-2, or some other finite set of possible states? If so, we can certainly model that. If not, we could separate the model we store from the "context" we use for rendering and updating, which could be derived from the stored model but maintain the invariant that an editor cannot exist without a buffer. Either way this will likely require large architectural changes to work well, so it's definitely a long game.

bryphe: > Is it really 1-to-many? Or is it either 1-to-1 or 1-to-2, or some other finite set of possible states?

Good question. The buffer <-> editor relationship is 1 to some finite limit - for example, for each split, we have a new editor pointing to potentially to the same buffer. If you create 5 splits for the same file, you'd have 5 editors for one buffer (in our current model).

A renderer <-> editor should be 1 <-> 1 at any given time, but could change or be toggled.

> which could be derived from the stored model but maintain the invariant that an editor cannot exist without a buffer

This would be great. We don't have this invariant modeled in our current state.

Thanks for the thoughts! Hopefully we can figure out a good plan for this to remove some of these warts.

PR #1517, opened 2020-03-28
Feature - Terminal: Use terminal font + background / foreground color

bryphe:
When switching to "Terminal Normal" mode, this uses the terminal font and terminal background/foreground color (overriding the default editor colors), to help smooth the transition. It also disables the diff markers in this mode:

### DIFF
--- a/src/Feature/Editor/EditorSurface.re
+++ b/src/Feature/Editor/EditorSurface.re
@@ -113,6 +113,9 @@ let minimap =
 
 let make =
     (
+      ~showDiffMarkers=true,
+      ~backgroundColor: option(Revery.Color.t)=?,
+      ~foregroundColor: option(Revery.Color.t)=?,
       ~buffer,
       ~onDimensionsChanged,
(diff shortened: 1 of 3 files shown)


src/Feature/Editor/EditorSurface.re
```diff
@@ -113,6 +113,9 @@ let minimap =
 
 let make =
     (
+      ~showDiffMarkers=true,
+      ~backgroundColor: option(Revery.Color.t)=?,
+      ~foregroundColor: option(Revery.Color.t)=?,
```
bryphe: This might be a bit confusing - these optional parameters override some of the configuration / theme settings that come in. @glennsl  - are you OK with this model? Or do you have another approach you'd prefer?

glennsl: I don't see any major issues with it, so I think it's fine. Clever use of the precomputed colors too. This should also work fine after extracting the editor as an app-independent component I think, once we get around to doing that.

PR #1518, opened 2020-03-28
Experiment: Smooth cursor animation

bryphe:
This adds an experimental cursor animation (via the `experimental.editor.cursorSmoothCaretAnimation` boolean flag):

__Next steps:__
- A nice addition would be a subtle motion-blur effect - by calculating the delta between the originalX and Y (as a proxy for velocity), we could lower the opacity and do a _very slight_ stretch to simulate a motion blur.


bryphe:
Thanks for the feedback, @glennsl !

> But at very short distances, like a single line or a character, it feels excruciatingly unresponsive. 

Good point. 

> So my recommendation would be to turn it off for single-position movements, and otherwise try to make the average speed rather than the movement time constant.

Agreed... I made some tweaks to the `Hooks.transition` function to support dynamically specifying a delay function, based on the current value and new target value. I used a simple step function for now (if the distance is too short, just make it teleport). Could be improved with a better function for calculating duraiton.

jordwalke:
What's the relationship between smooth scrolling and smooth cursor? Are they tied together? Is it possible to only enable the smooth scrolling on jump, but disable the smooth cursor movement?
As for the actual caret movement, it might be possible to achieve the same effect (of showing the user where the caret moved) by moving it instantly, but showing a very faint and brief wind/contrail behind it. Hard to make it look classy, but could be a feature that even the pros would leave enabled if unobtrusive and if it doesn't slow down the actual movement.

CrossR:
There is separate options for the two animations:

```json
"experimental.editor.smoothScroll": true,
"experimental.editor.cursorSmoothCaretAnimation": true,
```

I think the minimap does need a look over for the scrolling animations though, it gets a bit hectic when you are doing some movements like `Ctrl-d`.

PR #1519, opened 2020-03-28
Feature - Terminal: Preserve cell colors in transition to normal mode

bryphe:
This converts the terminal cell colors to syntax highlights when switching from insert -> normal mode:

It'll look even better when combined with #1517 (which preserves the font and uses the terminal background / foreground colors)

### DIFF
--- a/src/Feature/Syntax/Feature_Syntax.rei
+++ b/src/Feature/Syntax/Feature_Syntax.rei
@@ -32,6 +32,11 @@ let setTokensForLine:
 
 let update: (t, msg) => t;
 
+// [ignore(~bufferId, syntax)] marks a buffer to be ignored.
+// The only syntax highlight adjustment will come from explicit
+// calls to [setTokensForLine];
+let ignore: (~bufferId: int, t) => t;
+
 let subscription:
   (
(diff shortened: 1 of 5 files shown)


src/Feature/Syntax/Feature_Syntax.rei
```diff
@@ -32,6 +32,11 @@ let setTokensForLine:
 
 let update: (t, msg) => t;
 
+// [ignore(~bufferId, syntax)] marks a buffer to be ignored.
+// The only syntax highlight adjustment will come from explicit
+// calls to [setTokensForLine];
+let ignore: (~bufferId: int, t) => t;
```
glennsl: This is a very imperative API. Can the flag not be passed in with the subscription instead?

bryphe: Agreed, it would be nice to not have this imperative API at all.

 Unfortunately the syntax highlight integration (even though it has a subscription) is still imperative - the subscription manages the lifecycle of the syntax highlight service, as opposed to having per-buffer granularity (which may help streamline this). Another place we need a deeper refactor...

PR #1527, opened 2020-03-30
Feature: Write failure dialog

### LINKED ISSUE #1295: Cannot save files that have been externally modified
# Description

When a file opened in Oni is modified by another program, Oni fails to save.

## Steps to reproduce

- Create a file and write stuff in it.

```bash
mkdir oni-bug && cd oni-bug
echo 'Some text' > file.txt
```

- Run Oni and open the `file.txt` file.

file txt - oni-bug - Onivim 2_095

- Change the file externally.

```bash
echo 'Some more text' >> file.txt
```

- Change the file from inside Oni.

-file txt - oni-bug - Onivim 2_096

- Try and save the file.

-file txt - oni-bug - Onivim 2_097

## Observed result

Nothing happens. The file is not saved, and the "unsaved changes" indicator is still visible.

## Expected result

In the same scenario, Vim issues a warning telling the user that the file was changed, and prompts the user to overwrite or abort.

vi file txt_099

# General info

- OS: Ubuntu 18.04.3
- Oni2 version: 2019-02-05 nightly build (commit 04f242db)

glennsl:
Fixes #1295

When a write fails, a dialog is shown presenting a choice between overwriting and discarding the local changes.

2020-03-30-161829_490x112_scrot

Depends on https://github.com/onivim/libvim/pull/194 for `:w!` to work properly.

Also, in the future we'll probably want to offer a diff option with the ability to merge the diverging documents.

On a more technical note, this creates both a vim service project and a feature project for modals, with the latter being a bit messy because the modals ought to be contributed from other feature projects I think, but currently there's no such projects for these to live in.


src/Model/FocusManager.re
```diff
@@ -21,6 +23,8 @@ let pop = (focusable: Focus.focusable, state: State.t) =>
 let current = (state: State.t) =>
   if (Sneak.isActive(state.sneak)) {
     Focus.Sneak;
+  } else if (state.modal != None) {
+    Focus.Modal;
```
glennsl: The modal focus is now also enforced by the presence of a modal, rather than being manually controlled.

bryphe: Nice!

src/Service/Vim/Service_Vim.re
```diff
@@ -0,0 +1,19 @@
+let forceReload = () =>
+  Isolinear.Effect.create(~name="vim.discardChanges", () =>
+    Vim.command("e!")
+  );
+
```
bryphe: Thanks for starting this service 👍  Looks great

package.json
```diff
@@ -207,7 +207,7 @@
     "ocaml": "~4.7.0",
     "reason-sdl2": "^2.10.2020",
     "reason-jsonrpc": "1.5.0",
-    "reason-libvim": "onivim/reason-libvim#3ead103",
+    "reason-libvim": "onivim/reason-libvim#ec703ce",
```
bryphe: Bringing in the `libvim` update here: https://github.com/onivim/reason-libvim/pull/110

PR #1529, opened 2020-03-30
Feature: Reload active buffer on external change

glennsl:
This introduces a file watcher service, implemented with Luv, and uses it to watch all open buffers. But currently only updates the active buffer when changed on disk, as libvim takes care of the rest. This is however the first small step towards implementing a large set of functionality that requires a file watcher API.

Next steps:
- Add a banner, or some other buffer-specific notification, to modified buffers that have changed on disk, to warn that saving the buffer will fail.
- There's a bug where the watcher it stops responding to changes after a while or a certain sequence of events. It would be useful to have more information to be able to discern a pattern however.


src/Service/FileWatcher/Service_FileWatcher.rei
```diff
@@ -0,0 +1,8 @@
+[@deriving show]
+type event = {
+  path: string,
+  hasRenamed: bool,
+  hasChanged: bool,
+};
+
+let watch: (~path: string, ~onEvent: event => 'msg) => Isolinear.Sub.t('msg);
```
glennsl: This is the ideal subscription API IMO. Instead of offering a subscription with a hard-coded `msg` type that has to be mapped by the consumer to their own `msg` type, this take a set of mapping functions and does the mapping internally. Unfortunately this still requires a kludgy internal `msg` type, and fixing it requires using a record instead of fa first-class module I think. But this awkwardness only affects providers, so no rush.

src/bin_editor/Oni2_editor.re
```diff
@@ -133,15 +133,15 @@ if (cliOptions.syntaxHighlightService) {
...
-      Tick.interval(
-        _dt =>
-          if (isDirty^) {
-            update(<Root state=currentState^ />);
-            isDirty := false;
-          },
-        Time.seconds(0),
-      );
+    let tick = _dt => {
+      let _: bool = Luv.Loop.run(~mode=`NOWAIT, ());
```
glennsl: This is all the wiring needed to integrate Luv into the app loop.

bryphe: Cool! Very straightforward 👍

src/Service/FileWatcher/Service_FileWatcher.rei
bryphe: Makes sense, I like that better too - simpler for the consumer to use, versus needing the `Isolinear.Sub.map` everywhere.

src/Service/FileWatcher/Service_FileWatcher.re
```diff
@@ -0,0 +1,76 @@
...
+
+    let update = (~params as path, ~state, ~dispatch as _) =>
+      if (path != state.path) {
+        // Should be unreachable because the path is the id
+        failwith(
+          "unreachable",
+        );
+      } else {
+        state;
+      };
```
bryphe: Why is the conditional necessary here? Would it make sense to just return `state`?

src/Service/FileWatcher/Service_FileWatcher.re
```diff
@@ -0,0 +1,76 @@
...
+      switch (state.watcher) {
+      | Some(watcher) => ignore(Luv__FS_event.stop(watcher): result(_))
+      | None => ()
+      };
+    };
+  });
+
+let watch = (~path, ~onEvent) =>
+  WatchSubscription.create(path)
+  |> Isolinear.Sub.map(event => onEvent(event));
```
bryphe: Is there any sort of test we could have to validate the `WatchSubscription` works cross-platform? For example, create a temporary file, create a subscription, modify the file, and validate `onEvent` gets called?

src/Service/FileWatcher/Service_FileWatcher.re
glennsl: Yeah, definitely. I started with just the `failtwith`, because I couldn't, and still can't, see any hood reason for update to get called if `params` hasn't changed. Then of course that crashed, so it morphed into this just to assure myself that it didn't do other weird things. But I'm pretty assured now, so no reason to keep it.

src/Service/FileWatcher/Service_FileWatcher.re
glennsl: Good idea! Added in https://github.com/onivim/oni2/pull/1529/commits/464511e2c465170cd3ad66dddb3430f3c3831f65

integration_test/FileWatcherTest.re
```diff
@@ -0,0 +1,40 @@
...
+  Isolinear.Testing.SubscriptionRunner.Make({
+    type msg = unit;
+  });
+
+Runner.run(~dispatch=() => (), ~sub, Runner.empty);
+
+// Modify file on disk
+let oc = open_out(tempFilePath);
+Printf.fprintf(oc, "bar");
+close_out(oc);
```
bryphe: Awesome! Thanks @glennsl

bryphe: And it passes on Windows too 🎉

bryphe:
I saw an error pop up looking at the integration_test logs:
```
[TRACE]    +7ms Oni2.Service.FileWatcher : Starting file watcher for oni://Welcome
[ERROR]    +2ms Oni2.Service.FileWatcher : 'oni://Welcome': no such file or directory
```
It looks like the file watcher doesn't like our `oni://` paths - perhaps we should ignore those?

glennsl:
Yeah, I did look for a field on `Buffer` that specifies whether it's actually a file or not, but there doesn't seem to be one. We could just filter out those starting with `oni://` of course, but is that a reliable way to do it? What about newly created file buffers, that have proper file names but just haven't been saved to disk yet?

bryphe:
> Yeah, I did look for a field on Buffer that specifies whether it's actually a file or not, but there doesn't seem to be one. We could just filter out those starting with oni:// of course, but is that a reliable way to do it?

Most reliable way to check at the moment would be to use this - 
https://github.com/onivim/oni2/blob/master/src/Core/BufferPath.re and only watch Buffers that resolve to a `FilePath`. We could consider moving it into `Buffer` (or a helper that uses it to implement a function that let us check if it has a backing file).

> What about newly created file buffers, that have proper file names but just haven't been saved to disk yet?

That's a good question. Would `libuv` support this, and give a change event on creation? Or just ignore it?

glennsl:
Perhaps by using the `FS_poll` API. Or, worst case, we could create our own polling using `Timer` and `stat`.

There seems to be a bigger problem here though. `FileWatcherTest` fails on macOS. Haven't been able to figure out why yet, but there doesn't seem to be any errors. Currently checking if I can reproduce it using `Luv's` test framework.

glennsl:
Alright, I've fixed the macOS test failure, which was just because of the artificial nature of testing. Adding a delay did the trick.

I've also added filtering using `BufferPath` and a retry mechanism for non-existing files. The caveat on the latter is that it won't detect changes occurring in the period up to the retry interval, which is currently set to one second. I also tried falling back to polling, which had the same issue. We could also try something more sophisticated like watching the parent directory, which should pick up the creation immediately I think.

glennsl:
Now watching directory instead. Works much better!

bryphe:
> Alright, I've fixed the macOS test failure, which was just because of the artificial nature of testing. Adding a delay did the trick.

Awesome, thanks for all your investigation into this! Very excited to have `libuv` hooked up and validated with tests.

> Now watching directory instead. Works much better!

Great idea!

PR #1534, opened 2020-03-31
Fix #538 - Show version info

### LINKED ISSUE #538: "Version" command does not work
In opening #537, I attempted to run `:version` to report the version of the software. 

The first time I tried this, it froze and crashed.
The second time (and all subsequent attempts), it does not do anything that I can tell.

Note it does come up as an autocomplete.

OS: Windows 10

bryphe:
This hooks up the `:version` command to show a screen that displays the version of Onivim 2, as well as various components:

image

__Potential Next steps:__
- Add a sneak-able button to copy and paste
- Allow editing in read-only mode (switch to a buffer in visual mode)
- Add additional version info:
  - Harfbuzz compiled / linked version
  - libvim compilation flags

Fixes #538

### DIFF
--- a/src/UI/VersionView.re
+++ b/src/UI/VersionView.re
@@ -125,14 +125,14 @@ let make = (~state: State.t, ()) => {
   let header = HeaderView.make(~state);
 
   <View style={Styles.container(~theme)}>
-    <header text="Onivim 2:" />
+    <header text="Onivim 2" />
     <version name="Version" version=Oni_Core.BuildInfo.version />
     <version name="Commit" version=Oni_Core.BuildInfo.commitId />
     // spacer
-    <header text="OCaml:" />
+    <header text="OCaml" />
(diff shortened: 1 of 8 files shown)


src/UI/VersionView.re
```diff
@@ -0,0 +1,154 @@
...
+    ];
+
+    let versionText = (~theme: Theme.t, ~fontFile, ~fontSize) => [
+      fontFamily(fontFile),
+      Style.fontSize(fontSize),
+      color(theme.foreground),
+    ];
+
+    let spacer = Style.[flexGrow(1)];
+    let tempSpacer = Style.[flexGrow(1)];
```
glennsl: Why are these distinguished? What's temporary about the `tempSpacer`?

bryphe: Oops, good question... I was using this to hack while tweaking the styles. Removed and factored out some inline styles 👍

PR #1542, opened 2020-04-01
Bugfix: Cursor background default

glennsl:
**Issue**: The text under the cursor wasn't being rendered (visibly) with some themes

**Defect**: The default cursor background color was unspecified, which is according to vscode's default, but where vscode has view code that handles this, we just fall back to transparent.

**Fix**: Set the default to the opposite of the foreground color. This is what vscode does in its view code, and this just does using a default computed function instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -239,7 +239,7 @@
     "@opam/easy-format": "1.3.1",
     "@opam/lwt_ppx": "1.2.2",
     "@opam/merlin-extend": "0.4",
-    "revery": "glennsl/revery#ef05d3a",
+    "revery": "revery-ui/revery#6fd117a",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -239,7 +239,7 @@
     "@opam/easy-format": "1.3.1",
     "@opam/lwt_ppx": "1.2.2",
     "@opam/merlin-extend": "0.4",
-    "revery": "revery-ui/revery#4f08701",
+    "revery": "glennsl/revery#ef05d3a",
```
glennsl, line 242: TODO

glennsl: Done.

PR #1552, opened 2020-04-03
Refactor: Context Keys

glennsl:
This creates a new context keys data structure and schema, so that context keys can be defined as contributions and extended with context-specific context keys (:roll_eyes:).

### DIFF
--- a/src/Core/WhenExpr/ContextKeys.re
+++ b/src/Core/WhenExpr/ContextKeys.re
@@ -1,0 +1,57 @@
+module Lookup = Kernel.KeyedStringMap;
+
+module Value = {
+  [@deriving show({with_path: false})]
+  type t =
+    | String(string)
+    | True
+    | False;
+
+  // Emulate JavaScript semantics
+  let asBool =
+    fun
+    | True => true
+    | False => false
+    | String("") => false
+    | String(_) => true;
+
+  // Emulate JavaScript semantics
+  let asString =
+    fun
+    | True => "true"
+    | False => "false"
+    | String(str) => str;
+};
+
+module Schema = {
+  type entry('model) = {
+    key: string,
+    get: 'model => Value.t,
+  };
+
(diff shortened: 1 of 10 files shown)


src/Model/ContextKeys.re
```diff
@@ -0,0 +1,85 @@
...
+        fun
+        | Some({variant: EditorsPicker, _}) => true
+        | _ => false,
+      ),
+    ],
+  );
+
+let editors =
+  fromList([
+    bool("insertMode", state =>
```
bryphe, line 29: Much nicer than what we had before!

src/Core/WhenExpr/ContextKeys.re
```diff
@@ -0,0 +1,59 @@
...
+  // Emulate JavaScript semantics
+  let asBool =
+    fun
+    | True => true
+    | False => false
+    | String("") => false
+    | String(_) => true;
+
+  // Emulate JavaScript semantics
+  let asString =
```
bryphe, line 19: Cool, we support string context now!

src/Core/WhenExpr/WhenExpr.rei
```diff
@@ -9,6 +9,30 @@ module Value: {
   let asString: t => string;
 };
 
+module ContextKeys: {
+  module Schema: {
```
bryphe, line 13: Great API!

src/Core/WhenExpr/ContextKeys.re
bryphe: Is there any context key we're using this now? Or is this in preparation for the additional SCM functionality?

glennsl: This was added when I revamped `WhenExpr`. It's just been moved here. It's currently just used in unit tests, but will be needed for the SCM functionality shortly.

Issue #1513, opened 2020-03-28
Docs / Learning: Bundle vimtutor


bryphe:
Since we don't have an interactive tutorial, and our docs are sparse, we should at least be bundling the existing vim tutor.

- Include the `vimtutor` as a runtime asset.
- Add a command to open the `vimtutor`
- Set the `readonly` flag on `vimtutor`, so that it can't be saved.

Additional things to think about:
- We should hook up a Vim command as an entry point as well (ie, `:tutor`, `:vimtutor`)
- We should add this to the welcome screen, so a user knows it is available

Further out - we could think about adding Onivim-specific functionality to it (ie, quick-open, command-pallete, sneak-mode), or dressing it up with a custom buffer renderer.

PR #1553, opened 2020-04-03
Feature: Add Vim Tutor

glennsl:
This adds Vim's tutor as an asset, as well as a command to open it. Addresses #1513 

There's a few problems with it still though:
- [x] The read only flag causes a warning when the user first modifies the buffer
- [ ] Upon closing Onivim, the unsaved files warning triggers because the tutor is flagged as modified. We should either check whether modified buffers are read only, or not mark read only buffers as modified in the first place.
- [x] The formatting is off because the tutor uses 8-space tabs but Oni guesses 3 spaces. I've tried setting it in `openTutorEffect` immediately after open, but unfortunately it doesn't stick.
- [ ] Licensing? Are we covered by the license for libvim, or do we need to cite it separately when including parts of it directly?

We might also want to adapt the tutor a bit to our needs. I couldn't at a glance see anything that was entirely incorrect though. Even `:!` commands sort of work.

### DIFF
--- a/assets/tutor/dune
+++ b/assets/tutor/dune
@@ -1,0 +1,4 @@
+(install
+    (section bin)
+    (package Oni2)
+    (files tutor))
--- a/assets/tutor/tutor
+++ b/assets/tutor/tutor
@@ -1,0 +1,784 @@
+
+===============================================================================
+=    W e l c o m e   t o   t h e   V I M   T u t o r    -    Version 1.7      =
+===============================================================================
+
+     Vim is a very powerful editor that has many commands, too many to
+     explain in a tutor such as this.  This tutor is designed to describe
+     enough of the commands that you will be able to easily use Vim as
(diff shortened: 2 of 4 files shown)


glennsl:
The tutor is now copied to a temp file, which avoids the read only warning. You'll still get an unsaved warning if you close Onivim with the file modified though, but that doesn't seem as silly as getting it for read-only buffers.

I also tried setting the indentation directly on the buffer. The problem is that we don't have an Onivim buffer at that point, just a vim buffer. I tried dispatching `BufferSetIndentation`, but that probably gets processed before `BufferEnter`. Also tried dispatching a `BufferEnter` manually, but guessing that gets overwritten by the "real" `BufferEnter` that comes later. So... still tricky.

glennsl:
The linker error is back @bryphe :disappointed:

PR #1566, opened 2020-04-05
refactor(search): integrate Fzy

CrossR:
There is more to do to clean up the Oni2 side of the code, but its nice to have it working so that I can verify that the full pipeline is working, and this helped a lot in flushing out various issues in the C FFI code, especially intermittent freezes (Using `return` instead of `CAMLreturn` in one spot 🤦‍♂) and just getting to grips with making lists there.

Alongside cleaning up the code that is already here, I'm also interested in seeing if any of the limits/pre-matching steps we have can be changed. If the pre-matching code does change, then the current (scary) assumption I make about the lists being the same size will obviously need to change, and the comments @bryphe made on discord will be useful, since I'll have to start matching up items with results. Though I'll worry about that if the performance looks like it makes sense to do that.

EDIT: Oh there may also be some license stuff to change, though it is MIT licensed (I should add licenses to the libs in the chain), but should probably still list it somewhere in here.


bryphe:
> Using return instead of CAMLreturn in one spot 🤦‍♂

Ouch, nice catch on that - the FFI can be brutal!

> EDIT: Oh there may also be some license stuff to change, though it is MIT licensed (I should add licenses to the libs in the chain), but should probably still list it somewhere in here.

Sounds great - an entry for [`fzy`](https://github.com/jhawthorn/fzy) in our https://github.com/onivim/oni2/blob/master/ThirdPartyLicenses.txt should be sufficient 👍

bryphe:
> I'm also interested in seeing if any of the limits/pre-matching steps we have can be changed. If the pre-matching code does change, then the current (scary) assumption I make about the lists being the same size will obviously need to change, and the comments 

I didn't quite understand this piece - does this mean we can be assured we get the same results, because our 'pre-filtering' logic removes any items that wouldn't have a match? If that's the case - it does sound fragile

bryphe:
Awesome to see the progress here and see everything work end-to-end from `esy-fzy` -> `reason-fzy` -> `reasonFuzz` -> `oni2` 😄 Thanks for all your hard work to integrate this, @CrossR !

CrossR:
>I didn't quite understand this piece - does this mean we can be assured we get the same results, because our 'pre-filtering' logic removes any items that wouldn't have a match? If that's the case - it does sound fragile

Exactly, right now we get lucky really, but its obviously very fragile to just assume that pre-filter step does everything such that the list going in and the list coming out are the same, just with scores attached.

I'll start some work on some matching code, so that I can match between the two lists, which should be much better. After that, I can then more easily try some changes and see if and how certain things affect the performance.

CrossR:
>I'll start some work on some matching code, so that I can match between the two lists, which should be much better. After that, I can then more easily try some changes and see if and how certain things affect the performance.

Was being dumb, I already added the index to each of the items, I just needed to surface that `original_index` from the C layer and use that to pull the original item out of the input list. Think that is the quickest way of doing this matching, and removes the scary zipping code.

That way I can sort the items by score in the C layer and just match up at the end, rather than sorting twice, once by index and once by score.

bryphe:
> That way I can sort the items by score in the C layer and just match up at the end, rather than sorting twice, once by index and once by score.

Perfect! 👍

src/Core/Filter.re
```diff
@@ -12,38 +12,25 @@ type result('a) = {
...
-    let line = format(item);
-    let match = pathIndexMatch(~line, ~pattern);
+  let search = (query, format, items) => {
+    let searchStrings = List.map(item => format(item), items);
+    let results = Fzy.fzySearchList(searchStrings, query, ());
+    let lookup = (result: Fzy.Result.t) =>
+      List.nth(items, result.original_index);
 
-    (match, item);
+    List.map(result => (lookup(result), result), results);
```
bryphe, line 30: It looks like this logic might be O(N^2) - each of the `List.nth` calls requires a traversal of the list to pull the item out. (So for each item we `map`, we have to do a nested traversal via the `List.nth` call).

I wonder if it would be quicker to do the following:

```
let itemArray = items |> List.to_array;
let lookup = (result: Fzy.Result.t) => itemArray[result.original_index];
```

There's the extra copy of the array at the beginning (so a space trade-off).... but it means in terms of timing we wouldn't need to traverse the list for every item. If we have a benchmark in place, it would be interesting to see the comparison of the two approaches, at least.

bryphe:
Looks like it's almost set - thanks for implementing the full mapping, @CrossR (should be more resilient even if the pre-filter gets out of sync). I just had one question about the mapping, and it seems like the only other blocker is the space issues: https://github.com/onivim/oni2/pull/1585 

Can't wait to have this in! 👍

src/Core/Filter.re
CrossR: I'll give it a benchmark and see!

I've also realised that because this is pre-sorted, I don't need to do a final pass into `makeResult`, I can just do that inline here, removing another `List.map`.

CrossR: There isn't much in it, but it looks like `List.nth` is slightly better:

image

glennsl: Why is it a list in the first place? I assume it originates as an array, and I think it's going to be converted into an array a bit down the line anyway, so it might as well stay an array the whole way. Arrays can also modify items in-place, so you might save some allocation that way as well.

One of the things I noticed, and think I improved, when  I overhauled the filtering some time ago was several instances of unnecessary conversion between lists and arrays. And if given an array instead of a list it might be that more of the pipeline should work on an array. But either way, there's no need to do the list conversion prematurely, as that tends to lead to more back-and-forth, like it was before.

CrossR: In one place (Completions) it is converted back to an array so it would probably make sense there.

The other place is the `FilterJob` where I'm less sure about changing it, since the internals of the queue used there is also all list based, as well as the rest of `FilterJob`. I think you implemented the bits around there (at least the Queue), so not sure if you have any ideas if its worth refactoring it to return an array, or use an array internally (though that sounds slower).

glennsl: Right, yeah that Queue implementation definitely needs a list. I think an implementation using a list of arrays with a pointer into the "current" array could be as performant, although it would definitely be messier and would rely on the assumption that the array won't be mutated. A larger refactor of FitlerJob to be array-based might even be beneficial, although I doubt it would be more maintainable.

Whether this would be worth it depends largely on the size of the arrays I think. If they're on the order of 10 items it probably isn't. And judging by the performance of using `List.nth` they probably aren't very big.

bench/lib/FilterJobBench.re
```diff
@@ -26,17 +26,32 @@ let largeAmountOfItems =
...
+let job = getJob(largeAmountOfItems);
 let runJob = () => Job.doWork(job) |> (ignore: Job.t(_) => unit);
 
+let completeJob = () => {
+  let job = ref(getJob(smallerNumberOfItems))
+  while (! Job.isComplete(job^)) {
+    job := Job.doWork(job^);
+  }
+}
+
```
CrossR, line 54: Added this to more easily benchmark changes to the full (I think this is the full matching?) matching, rather than a single run of `doWork`.

Obviously knowing how fast `doWork` is useful but its not much help tuning the full performance (tuning limits or the pre-fuzzy step), since turning the limits up makes 1 take longer but you do less and removing the fuzzy makes it quicker etc.

CrossR:
image

Result from latest benchmark, assuming that its showing what I want it to show (a full example of a search). 

I had a look into removing the pre-Fzy filtering step, but I think it would be a bit more involved (if I understand what is happening in FilterJob) so could be something to follow up on again later. I expect the current biggest issue for issues like #1579 is the streaming of the results, and if / how they are cached between runs.

CrossR:
I'm going to use this branch a bit more this week and pending a review (and fixing the merge conflicts/pulling in proper packages rather than depending on PRs), I think this is done.

Previously `reasonFuzz` wasn't made into an actual package, so we just added a resolution. Should I be doing an actual release, to help tidy the resolutions up? I've not looked into what is needed for it, but I assume there is some docs somewhere.

bryphe:
> I'm going to use this branch a bit more this week and pending a review (and fixing the merge conflicts/pulling in proper packages rather than depending on PRs), I think this is done.

Sounds good to me!

> Previously reasonFuzz wasn't made into an actual package, so we just added a resolution. Should I be doing an actual release, to help tidy the resolutions up?

I'm OK with just a resolution for now to keep it simple (ie, we use just a resolution for `reason-libvim`). If we end up in using in multiple places - we could publish a versioned NPM package.

bench/lib/FilterJobBench.re
bryphe: Makes sense!

glennsl:
Why is `ReasonFuzz` even needed? Other than having a bunch of tests, it seems to do little more than to re-export `Fzy` from `reason-fzy`.

CrossR:
>Why is ReasonFuzz even needed? Other than having a bunch of tests, it seems to do little more than to re-export Fzy from reason-fzy.

Oh that is true now! I had originally done that to try and unify the interface (call one command and just change the matcher you want, so it made sense to be in one place), but that went out the window since the search methods are inherently different (array/list based + only return matches vs line based and always return), so it felt wrong to combine them.

It has been nice to have them all in one place for ease of benchmarking, but for here, I don't think we need anything from reasonFuzz. I'll drop back to just `reason-fzy`.

bryphe:
Sounds like this pretty much ready to go!

I'm going to do some stability testing this week (using it in my day-to-day work), just to make sure there aren't any crashes.

bryphe:
I'm testing this branch right now, and I'm seeing one intermittent crash - I wonder if the `fzy` is changing the sort of highlight text / positions we're getting?

```
(Invalid_argument "String.sub / Bytes.sub"):
Called from file "string.ml" (inlined), line 47, characters 2-23
Called from file "src/Components/HighlightText.re", line 25, characters 17-56
Called from file "src/Components/HighlightText.re", line 27, characters 13-40
Called from file "src/Components/HighlightText.re", line 27, characters 13-40
Called from file "src/Components/HighlightText.re", line 27, characters 13-40
Called from file "src/Components/HighlightText.re", line 27, characters 13-40
Called from file "src/Components/HighlightText.re", line 27, characters 13-40
Called from file "src/Components/HighlightText.re", line 27, characters 13-40
Called from file "src/Components/HighlightText.re", line 27, characters 13-40
Called from file "src/Components/HighlightText.re", line 27, characters 13-40
Called from file "src/Components/HighlightText.re", line 27, characters 13-40
Called from file "src/Components/HighlightText.re", line 30, characters 4-30
Called from file "src/UI/QuickmenuView.re", line 141, characters 6-72
Called from file "src/Components/FlatList.re", line 78, characters 8-23
Called from file "list.ml", line 92, characters 20-23
Called from file "list.ml", line 92, characters 32-39
Called from file "list.ml", line 92, characters 32-39
Called from file "list.ml", line 92, characters 32-39
Called from file "list.ml", line 92, characters 32-39
Called from file "src/Components/FlatList.re", line 82, characters 4-41
Called from file "src/Components/FlatList.re", line 176, characters 4-125
```

CrossR:
> 
> 
> I'm testing this branch right now, and I'm seeing one intermittent crash - I wonder if the `fzy` filter strategy is changing the sort of highlight text / positions we're getting?
> 
> ```
> (Invalid_argument "String.sub / Bytes.sub"):
> Raised at file "stdlib.ml", line 30, characters 20-45
> Called from file "string.ml" (inlined), line 47, characters 2-23
> Called from file "src/Components/HighlightText.re", line 25, characters 17-56
> Called from file "src/Components/HighlightText.re", line 27, characters 13-40
> Called from file "src/Components/HighlightText.re", line 27, characters 13-40
> Called from file "src/Components/HighlightText.re", line 27, characters 13-40
> Called from file "src/Components/HighlightText.re", line 27, characters 13-40
> Called from file "src/Components/HighlightText.re", line 27, characters 13-40
> Called from file "src/Components/HighlightText.re", line 27, characters 13-40
> Called from file "src/Components/HighlightText.re", line 27, characters 13-40
> Called from file "src/Components/HighlightText.re", line 27, characters 13-40
> Called from file "src/Components/HighlightText.re", line 27, characters 13-40
> Called from file "src/Components/HighlightText.re", line 30, characters 4-30
> Called from file "src/UI/QuickmenuView.re", line 141, characters 6-72
> Called from file "src/Components/FlatList.re", line 78, characters 8-23
> Called from file "list.ml", line 92, characters 20-23
> Called from file "list.ml", line 92, characters 32-39
> Called from file "list.ml", line 92, characters 32-39
> Called from file "list.ml", line 92, characters 32-39
> Called from file "list.ml", line 92, characters 32-39
> Called from file "src/Components/FlatList.re", line 82, characters 4-41
> Called from file "src/Components/FlatList.re", line 176, characters 4-125
> ```

Managed to have a look into this now. It looks like the array search strategy breaks if there is too many items. I've added a test of 100000 items that all match, and check that they all return sensible results (i.e. length is correct, they start properly and the highlights make sense).

For the list search, everything works fine. For the array search, it breaks over 80000ish items and starts storing messed up results. I'm inclined to think that Fzy etc is all fine, its just the FFI stuff and how I'm storing into an array. If I add a cut off at 80000 its fine, 90000 its not. I'll try and get some time to work out why that is.

Would also explain why I didn't see it, since I was only working on smaller repos at the mo.

src/Model/FilterJob.re
```diff
@@ -135,35 +136,28 @@ module Make = (Config: Config) => {
...
-            [item, ...acc];
-          } else {
-            acc;
-          };
-        },
-        filtered,
-        items,
-      );
+    // TODO: Understand why this is needed.
+    let filtered = items;
```
CrossR, line 151: If this isn't included, the results get a bit strange (i.e. just getting and passing back `filtered` unchanged, or setting it to `[]` etc). By strange I mean mostly correct, with a few missing bits and being generally nondeterministic. Including this means I get the same results as the CLI `fzy` (or equivalent based on score). Just need to poke through the Job code to understand why that is the case.

src/Model/FilterJob.re
```diff
@@ -135,35 +136,28 @@ module Make = (Config: Config) => {
...
+    // TODO: Should the mergeSortedList include a limit? Faster to quick return
+    // rather than merge then stop.
     let ranked =
-      filtered
-      |> ListEx.firstk(Constants.maxItemsToFilter)
-      |> Filter.rank(filter, format);
+      items
+      |> Filter.rank(filter, format)
+      |> ListEx.mergeSortedList(compareItems, ranked)
+      |> ListEx.firstk(Constants.maxItemsToFilter);
```
CrossR, line 160: I probably don't want this, since I'm doing extra work for no reason. Should integrate early stopping into the merging.

src/Model/FilterJob.re
```diff
@@ -83,6 +83,7 @@ module Make = (Config: Config) => {
     let currentMatches = ListEx.firstk(Constants.maxItemsToFilter, filtered);
 
     // If the new query matches the old one... we can re-use results
+    // TODO: This breaks new queries! The letter highlights will be wrong.
```
CrossR, line 86: Not to do with me that I can see, but searching for say `vecto` and letting it complete then adding `r` to make `vector` will show the results for `vector` but with `vecto` highlighted.

src/Model/FilterJob.re
```diff
@@ -135,35 +136,28 @@ module Make = (Config: Config) => {
...
-
-          if (Filter.fuzzyMatches(explodedFilter, name)) {
-            [item, ...acc];
-          } else {
-            acc;
-          };
-        },
-        filtered,
-        items,
-      );
```
CrossR, line 160: I'm not convinced we need this now, with how fast `fzy` is.

bryphe: Nice! It'll be great to simplify this

src/Model/FilterJob.re
bryphe: I wonder if `fzy` is fast enough that we could just scrap this whole block?
```
    if (pending.filter != ""
        && Filter.fuzzyMatches(oldExplodedFilter, filter)
        && List.length(currentMatches) < Constants.maxItemsToFilter) {
...
}
```

It's really just an optimization when the query is refined to not have a jarring set of results, but it might not be necessary anymore.

src/Model/FilterJob.re
bryphe: That sounds like a good idea!

mabuchner:
Please don't give up on this merge request, I think, it would be super useful to have it in Onivim 2.

CrossR:
Don't worry, not given up! Just been very busy with conferences and what not the past few months which has limited my time to work on it. It is pretty much there, just needs the merge conflicts fixing and me to use it for a week or so to check for bugs.

bryphe:
Ok, I merged up the latest and fixed warnings - hopefully build works now! 

There's a few issues I can see that need to be fixed prior to merging:

1) Some `FilterJob` unit tests are failing:

image

2) Highlights are not correctly applied when refining a query (but strangely, they get updated on backspace):



3) I see a non-optimal match for `EditorSurface`:

image

I'm not sure if this is because it's removing path separators, or if there is non-determinism (or just not working as I expect)? I'd expect that the `src/Feature/Editor/EditorSurface.re` would be the best match here.

Let me know if you'd like help investigating any of these, @CrossR !

CrossR:
I think issues 2) and 3) you were seeing with Fzy were both related to the "re-use the old results if we can" case. This was causing some results to stick around with their old score, which would break sorting etc, and also cause the highlights to lag behind. This case was only when you make a search more specific, not less, so backspace would break you out of that loop.

We actually had a spoke about this before! Just hadn't implemented it yet. Not checked 1) yet, but there will be a bunch more tests that "fail" now since we reset the full search on query change now, and the tests will look for that.

bryphe:
Woohoo, we have a green build @CrossR ! Ready to merge this when you are :+1:

CrossR:
I'm fine to merge whenever! I've been using this on my work machine and laptop in various amounts for the past 2-3 weeks continuously and haven't hit any issues, so *hopefully* that means its nice and stable.

After its merged (well, this weekend probably) I'm planning on making a meta issue or a project, just to collate some of the remaining work on the Quick open experience as a whole, as this improves the search results a lot, but there is still some UI/UX changes to go!

bryphe:
> I'm fine to merge whenever! I've been using this on my work machine and laptop in various amounts for the past 2-3 weeks continuously and haven't hit any issues, so hopefully that means its nice and stable.

Awesome! I'll merge it right now. I've been using it on-and-off too and it's been working well :+1: Thanks for the hard work on this.

> I'm planning on making a meta issue or a project, just to collate some of the remaining work on the Quick open experience as a whole, as this improves the search results a lot, but there is still some UI/UX changes to go!

Sounds great!

PR #1580, opened 2020-04-08
Fix insertSpaces default value

alarsyo:
[As stated in the docs](https://github.com/onivim/oni2/blob/master/docs/docs/configuration/settings.md), the default value for `insertSpaces` should be true, not false.


bryphe:
Great catch @alarsyo - thank you for fixing this!

PR #1584, opened 2020-04-09
Documentation - Fix Terminal Normal Command

sghill:
I noticed the video and docs were out of sync for going to terminal normal mode. The video has the right command, so this change updates the [integrated terminal docs](https://onivim.github.io/docs/using-onivim/integrated-terminal#normal-mode) to match.


bryphe:
Thanks for catching this, @sghill ! Appreciate it 👍

PR #1585, opened 2020-04-10
CI - Linux: More disk space issues

bryphe:
Looking at recent CI runs - I'm seeing several where the linux machines are running out of space:

```
collect2: error: ld returned 1 exit status
File "caml_startup", line 1:
Error: Error during linking
    ocamlopt integration_test/AddRemoveSplitTest.exe (exit 2)
```

Another fix in the vein of #1576 and #1483

### DIFF
--- a/.ci/clean-linux-deps.yml
+++ b/.ci/clean-linux-deps.yml
@@ -7,3 +7,9 @@ steps:
   # https://github.com/actions/virtual-environments/blob/master/images/linux/scripts/installers/dotnetcore-sdk.sh
   - script: rm -rf /usr/share/dotnet
     displayName: 'Remove .NET dependencies and SDK'
+  # Remove Haskell dependencies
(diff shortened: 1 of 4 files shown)


CrossR:
Perhaps stupid question, but I can't seem to see the answer...does the pre-installed software count towards the disk limit?

Looking over the stuff that the machines comes with (https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md), I would guesstimate there is like 15-20GB of stuff there we don't use (android SDKs must be 10GB+ on their own, and GHC is a big binary so that is probably pretty heavy too, having that many versions) and if we remove it, we'd gain a lot more space.

bryphe:
> Perhaps stupid question, but I can't seem to see the answer...does the pre-installed software count towards the disk limit?

Good call @CrossR ! I think it does... I'll give a shot at removing some of these deps 👍

bryphe:
We got a green build! Thanks for the suggestion @CrossR - you saved the day 🎉

PR #1586, opened 2020-04-10
Fix typo in Modal Editing 101

dhruska:


bryphe:
Ouch, thanks @dhruska for catching that!

PR #1590, opened 2020-04-10
Configuration: Add error notification when there is a parse failure

### LINKED ISSUE #1638: syntax highlighting dissapears - possibly due to broken config file
Syntax highlighting has dissapearred for me on oni, possibly related to a broken config (preferences) file as discussed last week on discord. I can't see an obvious mistake but here is my config below:

```
{
  "workbench.colorTheme": "One Dark Pro",
  "editor.detectIndentation": true,
  "editor.largeFileOptimizations": true,
  "editor.highlightActiveIndentGuide": true,
  "editor.indentSize": 4,
  "editor.insertSpaces": false,
  "editor.lineNumbers": "on",
  "editor.matchBrackets": true,
  "editor.minimap.enabled": true,
  "editor.minimap.maxColumn": 120,
  "editor.minimap.showSlider": true,
  "editor.renderIndentGuides": true,
  "editor.renderWhitespace": "all",
  "editor.rulers": [],
  "editor.tabSize": 4,
  "editor.zenMode.hideTabs": true,
  "files.exclude": [ "_esy", "node_modules" ],
  "workbench.activityBar.visible": true,
  "workbench.iconTheme": "vs-seti",
  "vim.useSystemClipboard": true:
}
```

bryphe:
__Issue:__ Brought up in Discord - if there is an error in configuration (like a missing comma), there is no notification or information presented to the user to help in understanding what the problem is.

__Defect:__ We were logging an error, but not adding a notification

__Fix:__ On error, add a notification as well.

Additional items:
- [x] Investigate why themes / syntax aren't loading correctly when the configuration is invalid
- [x] Add test case for yojson error parsing
- [x] Hook up to dispatch / clear diagnostics

Fixes #1638


src/Core/Utility/JsonEx.re
```diff
@@ -53,6 +53,31 @@ let getKeys = json => {
...
+         if (Array.length(matches) < 5) {
+           None;
+         } else {
+           let line = OnigRegExp.Match.getText(matches[1]) |> int_of_string;
+           let startByte =
+             OnigRegExp.Match.getText(matches[2]) |> int_of_string;
+           let endByte =
+             OnigRegExp.Match.getText(matches[3]) |> int_of_string;
+           let message = OnigRegExp.Match.getText(matches[4]);
+           Some((line, startByte, endByte, message));
```
glennsl, line 75: This isn't a very usable return value. You could use a `Location.t` to improve it a bit, but why not move it to `Core.Json` and do it properly with a record? Not sure this belongs in `Utility` in the first place, as it seems to fit in well there.

bryphe: I moved it to `Core.Json` as an `Error.t = { range: Range.t, message: string }` - it is more ergonomic to use.

PR #1591, opened 2020-04-11
Bugfix: Missing or incorrect statusbar/notification colors

glennsl:
This fixes two separate issues:

The first issue was just the wrong color being used for the foreground of statusbar text items. Easy fix.

The second issue is not so straightforward unfortunately. The root cause was two-fold. First, the defaults weren't registered, which is easy to fix, but exposes a problem with the approach of flattening the theme, that we no longer distinguish between an unspecified color and a missing default, which makes it easier to forget to register new defaults. This is more of a problem with the development process than with end-user experience however. The second cause was that all the defaults were left unspecified, so even if they were registered it wouldn't actually have made any difference.

There is a deeper problem here though, and a decision that needs to be made to resolve it. The notification color keys we've been using have been [deprecated since February 2018](https://code.visualstudio.com/updates/v1_21#_new-theme-colors-and-removal-of-old-colors). And deprecated in this case pretty much just means ignored, I think. I can find no reference to them in the current vscode code base. There's also no direct equivalent in the [current set of colors](https://code.visualstudio.com/api/references/theme-color#notification-colors). The only kind-specific notification colors are for icon foregrounds. The defaults for these in turn reference some colors used for the editor, but these also just specify the foreground, as well as a border color. So I see the following options:

1. Keep the colors we had before, which might clash with theme colors, but the complete control it gives us allows us to guarantee legibility unless the user explicitly overrides it.

2. Use the icon foreground colors as the background and try to be smart about picking a legible and good looking foreground color. Not yet sure how to actually do this though, but it should be possible to get good results in most cases at least.

And I guess a third option is the current approach of using the icon foreground colors as background and NOT be smart about the foreground color but just use the StatusBar foreground instead, which usually looks neither good nor legible.

Btw, I also removed the Success kind, since it doesn't exist in vscode or Vim, and is not used for anything else either.


bryphe:
Thanks for the quick fix and detailed investigation, @glennsl !

> The notification color keys we've been using have been deprecated since February 2018.

> There's also no direct equivalent in the current set of colors. 

Nice catch. I like the options you presented. My choice would be go with option 1:

> Keep the colors we had before, which might clash with theme colors, but the complete control it gives us allows us to guarantee legibility unless the user explicitly overrides it.

This would mean that, along with some of our Onivim-specific colors (like the mode colors), we'd also have the notification colors be specific to Onivim. Keeping control means we can ensure that the out-of-box experience is exactly as we design.

The other options - "smart" color choices based on theme - seem like it'd be tough to get right in general. So from my POV - option 1 seems like the best choice right now.

> Btw, I also removed the Success kind, since it doesn't exist in vscode or Vim, and is not used for anything else either.

Good call!

glennsl:
> option 1 seems like the best choice right now. And it looks like that's what you implemented here - so we're good to go +1

I implemented option 3 actually, just to test how it'd look (not very good), but I can do another PR to implement option 1 instead shortly.

bryphe:
Ah, sorry I misunderstood @glennsl ! Sounds good, thanks 👍

PR #1594, opened 2020-04-11
Revert Notification Color Defaults To Pre-Revamp Colors

glennsl:
Follow-up to #1591 

This reverts the notification colors to what we had pre-revamp, with one exception. I changed the warning foreground from white to almost black, because white on bright yellow isn't very legible.

Note also that in #1591 I prefixed the keys with `oni.` to distinguish them from the "standard" vscode colors. I'm trying to do that with all the keys, to prevent future collisions, so that users get a better idea of which colors won't be themed by vscode themes and that they therefore might have to tweak themselves, and for us to have a better idea of what we can freely change without breaking vscode compatibility and not have to dig around in the vscode source for.

### DIFF
--- a/src/Feature/Notification/Feature_Notification.re
+++ b/src/Feature/Notification/Feature_Notification.re
@@ -77,35 +77,17 @@ module Colors = {
   let foreground = foreground;
 
   let infoBackground =
-    define(
-      "oni.notification.infoBackground",
-      all(ref(EditorInfo.foreground)),
-    );
+    define("oni.notification.infoBackground", all(hex("#209CEE")));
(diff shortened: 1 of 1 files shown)


bryphe:
> This reverts the notification colors to what we had pre-revamp, with one exception. I changed the warning foreground from white to almost black, because white on bright yellow isn't very legible.

Perfect 👌 

> Note also that in #1591 I prefixed the keys with oni. to distinguish them from the "standard" vscode colors. 

Sounds good. It makes sense to be able to easily differentiate the theme colors we are adding.

PR #1595, opened 2020-04-11
Command Infrastructure refactor

glennsl:
Apart from a better architecture in general, what this refactor gives us is:

- The distributed architecture allows commands to be defined and contributed from feature projects.
- Commands are now defined in one place, rather than in two separate sets with an unclear relationship, as before.
- Integration with the new context keys API, which provides enablement conditions for extension-sourced commands ion addition tot he built in ones (we'll have to add a bunch of context keys for them to actually show up now though).
- Commands are retrieved from extensions on-demand, instead of just once on startup, which means it's ready for enabling and disabling extensions at runtime. (But note that this does not yet include dynamically registered extension commands, only contributed commands)
- The ability to pass an argument to a command, which is needed for some extension integrations.
- Having a single place to look up command definitions is needed for integration with contributed menu items (which is coming up in a near future PR).

TODO (in a future PR):
- [ ] Dynamic command registration from extensions (ie. implement the `vscode.commands.registerCommand` API)
- [ ] Fix id naming inconsistencies
- [ ] Convert oni-specific commands to roughly equivalent vscode commands where available
- [ ] Remove stringly typed `Command` action
- [ ] Make command IDs opaque?


src/Feature/Terminal/Feature_Terminal.re
```diff
@@ -406,3 +401,97 @@ let getLinesAndHighlights = (~colorTheme, ~terminalId) => {
...
+      ...state,
+      id,
+      title,
+    }
+  | Command(NormalMode) => {...state, insertMode: false}
+  | Command(InsertMode) => {...state, insertMode: true}
+
+  | _ => state
+  };
+};
```
glennsl, line 427: BufferRenderer state and reducer has been moved here to untangle command dependencies. And also because this is where it belongs, because of dependency issues like this, but also so it can be abstracted and its implementation details hidden.

glennsl:
> it seems I have broken terminal normal mode by refactoring the buffer renderer slightly, and I'm not sure how. Help would be much appreciated.

Found it! Fixed in https://github.com/onivim/oni2/pull/1595/commits/eccc3d2a51267f02a235f2c998b4d3267bc27eb6. Reviewing the code gave me the clue I needed to figure out I had forgotten to convert one of the scattered and stringly typed `Command` actions. We should eventually move away from the stringly typed `Command` action entirely, so these kinds of issues shouldn't be possible.

src/Core/Command.rei
```diff
@@ -0,0 +1,28 @@
+[@deriving show]
+type t('msg) = {
+  id: string,
+  title: option(string),
+  category: option(string),
+  icon: option([@opaque] IconTheme.IconDefinition.t),
+  isEnabledWhen: WhenExpr.t,
+  msg: [ | `Arg0('msg) | `Arg1(Json.t => 'msg)],
```
glennsl, line 8: This polymorphic variant exists to support commands with contextual arguments. Perhaps not the most elegant approach, but it works well. An alternative might be to distinguish more strongly between extension-sourced commands and built-in commands.

src/Feature/Terminal/Feature_Terminal.re
```diff
@@ -406,3 +401,97 @@ let getLinesAndHighlights = (~colorTheme, ~terminalId) => {
...
+        "terminal.new.current",
+        Command(NewTerminal({cmd: None, splitDirection: Current})),
+      );
+  };
+
+  module Oni = {
+    let normalMode = define("oni.terminal.normalMode", Command(NormalMode));
+    let insertMode = define("oni.terminal.insertMode", Command(InsertMode));
+  };
+};
```
glennsl, line 462: An example of commands defined in and contributed from a feature project.

src/Model/Actions.re
```diff
@@ -35,9 +35,7 @@ type t =
   | BufferSetModified(int, bool)
   | Syntax(Feature_Syntax.msg)
   | Command(string)
-  | CommandsRegister(list(command))
-  // Execute a contribute command, from an extension
-  | CommandExecuteContributed(string)
```
glennsl, line 40: These have been moved, to `Feature_Commands` and `Extensions` respectively.

src/Model/Actions.re
```diff
@@ -61,6 +59,7 @@ type t =
   // Reload keybindings from configuration
   | KeyBindingsReload
   | KeyBindingsParseError(string)
+  | KeybindingInvoked({command: string})
```
glennsl, line 62: This was added to distinguish commands invoked by keybindings from the stringly-typed `Command` action that actually runs the command, which should be replaced by a strongly typed variants and scoped to where the command is defined.

src/Model/ContextKeys.re
```diff
@@ -6,7 +6,13 @@ let menus =
...
       bool("inQuickOpen", model => model != None),
       bool(
-        "quickmenuCursorEnd",
+        "inEditorsPicker",
+        fun
+        | Some({variant: EditorsPicker, _}) => true
+        | _ => false,
+      ),
+      bool(
+        "oni.quickmenuCursorEnd",
```
glennsl, line 15: Oni-specific context keys have been namespaced with `oni.` to distinguish he from the "standard" ones from vscode, fore reasons discussed earlier.

src/Model/ContextKeys.re
```diff
@@ -77,7 +77,19 @@ let other =
...
+      bool("isMax", _state => Revery.Environment.os == Revery.Environment.Mac),
+      bool("isWin", _state =>
+        Revery.Environment.os == Revery.Environment.Windows
+      ),
+      bool("oni.sneakMode", state => Sneak.isActive(state.sneak)),
+      bool("oni.zenMode", state => state.zenMode),
+      bool("oni.keyDisplayerEnabled", state => state.keyDisplayer != None),
+      bool("oni.symLinkExists", _state =>
+        Sys.file_exists("/usr/local/bin/oni2")
+      ),
```
glennsl, line 92: An example of an impure context key, which is still possible, though not ideal.

src/Model/GlobalCommands.re
```diff
@@ -0,0 +1,390 @@
...
+    define("list.selectBackground", ListSelectBackground);
+};
+
+module Oni = {
+  module Explorer = {
+    let toggle =
+      define(
+        ~category="View",
+        ~title="Toggle File Explorer visibility",
+        "oni.explorer.toggle", // use workbench.action.toggleSidebarVisibility instead?
```
glennsl, line 72: There are several instances where we've invented our own commands when there already exists vscode commands that essentially reflects the same intent. I'll do another PR to clean this up later, to try to contain this one as much as possible.

src/Model/GlobalCommands.re
```diff
@@ -0,0 +1,390 @@
...
+
+    let removeFromPath =
+      define(
+        ~category="System",
+        ~title="Remove Oni2 from System PATH",
+        ~isEnabledWhen=WhenExpr.parse("isMac && symLinkExists"),
+        "oni.system.removeFromPath",
+        Command("system.removeFromPath"),
+      );
+  };
```
glennsl, line 133: cc @CrossR. I struggled to understand the logic of the original enablement condition for these, but it seems like this was the intent at least.

src/Model/State.re
```diff
@@ -133,3 +133,11 @@ let initial = (~getUserSettings) => {
...
   textContentProviders: [],
 };
+
+let commands = state =>
+  Command.Lookup.unionMany([
+    Feature_Commands.all(state.commands),
+    Extensions.commands(state.extensions)
+    |> Command.Lookup.fromList
+    |> Command.Lookup.map(msg => Actions.Extension(msg)),
+  ]);
```
glennsl, line 143: After way too many iterations trying to fit this somewhere that's both accessible everywhere needed, has all the dependencies line up with both the new and old architecture, and also makes sense, I finally settled on putting it in `State`. I think this will be a pattern I'll use frequently going forward, until we have a proper composition root where things like these can be defined.

src/Store/CommandStoreConnector.re
```diff
@@ -260,139 +79,15 @@ let start = (getState, contributedCommands) => {
     );
 
   let commands = [
-    ("keyDisplayer.enable", _ => singleActionEffect(EnableKeyDisplayer)),
-    ("keyDisplayer.disable", _ => singleActionEffect(DisableKeyDisplayer)),
```
glennsl, line 264: Most of these are now unnecessary as the mapping is done in the command definition itself.

src/Model/GlobalCommands.re
CrossR: Looks good to me! The original logic was a bit sloppy...

integration_test/KeySequenceJJTest.re
```diff
@@ -2,11 +2,13 @@ open Oni_Model;
 open Oni_IntegrationTestLib;
 
 let keybindings =
-  Some({|
+  Some(
+    {|
 [
-  {"key": "jj", "command": "vim.esc", "when": "insertMode"}
+  {"key": "jj", "command": "oni.vim.esc", "when": "oni.insertMode"}
```
glennsl, line 8: Note these changes in particular, which is probably going to break the config of quite a few people. If we want to namespace them, though, we're going to have to at some point anyway, and earlier seems better.

But we should still try to communicate it in some way, so it doesn't come as so much of a surprise. Should we start to consider some kind of in-app change log for breaking changes? Since we still don't do releases we could try to write the version used last to disk, and when it changes show all the breaking changes since that version on the Welcome screen for example.

src/Model/ContextKeys.re
bryphe: I'm concerned about this - it means we're doing blocking file I/O on potentially every key press or render. Is there another way we could model this, such that it only gets called 'lazily' - ie, when something is using `oni.symLinkExists`? I believe this is only needed when we open the quick menu.

Alternatively - we could store `symLinkExists` in the state somewhere, and check it. It's not perfect, because it could be out of sync, but I think the performance trade-off is worth it.

The concern I have is that synchronous file I/O on our critical rendering / input paths can lead to unpredictable performance characteristics, especially on Windows.

bryphe: (I believe this was less of an issue before, since it only was ran when the command palette was open)

src/Model/Extensions.re
```diff
@@ -14,7 +15,11 @@ type t = {
 [@deriving show({with_path: false})]
 type action =
   | Activated(string /* id */)
-  | Discovered([@opaque] list(ExtensionScanner.t));
+  | Discovered([@opaque] list(ExtensionScanner.t))
+  | ExecuteCommand({
+      command: string,
+      arguments: [@opaque] list(Json.t),
```
bryphe, line 21: Cool! We have arguments now 🎉

src/Model/GlobalCommands.re
bryphe: Thanks @glennsl !

src/Model/State.re
bryphe: Sounds good 👍

src/Store/CommandStoreConnector.re
bryphe: Excellent!

src/Feature/Terminal/Feature_Terminal.re
bryphe: Very cool 👍

src/Core/Command.rei
bryphe: Interesting!

integration_test/KeySequenceJJTest.re
bryphe: I'd prefer us not to break config / backward compatibility where possible - it's something we need to have a story for anyway, so the sooner we exercise this, the better.

> But we should still try to communicate it in some way, so it doesn't come as so much of a surprise. Should we start to consider some kind of in-app change log for breaking changes?

My preference would be to implement an _upgrader_ - when we load keybindings, to correct the incorrect ones, and send a warning that these are deprecated. We've done this once before (when the schema changed): https://github.com/onivim/oni2/blob/d4a60cf7e4b1260bdcb1f47503440b16d96fb2ff/src/Input/Keybindings.re#L235

Is there a case that conflicts that we couldn't unambiguously upgrade today? Or it would be a mechanical mapping of tokens like:
- `vim.esc` -> `oni.vim.esc`
- `insertMode` -> `oni.insertMode`
etc..

src/Model/ContextKeys.re
bryphe: I missed the reason this is required - is it to avoid conflicts?

glennsl: That, and for both us and end-users to get a better understanding of what's oni-specific and what comes from vscode.

src/Model/ContextKeys.re
glennsl: Ah, yeah, good catch. Since context keys can be extended, we could also just add it on-demand when retrieving the command palette. But ideally we'd not have the commands using this defined globally either. The command palette retrieval also changes a bit with the menu contribituons PR, but I'll look into a solution that will work for both.

integration_test/KeySequenceJJTest.re
glennsl: That doesn't actually solve the problem though. It just postpones the breakage until later, when it might be less likely to beak people, but at the same time the tolerance for breaks is much lower than in alpha. And all the while we're still just as liable to get conflicts.

In beta and post 1.0 I think backwards-compatibility is the right approach, but in alpha breaking changes should be expected, as long as we communicate it well (which we currently don't).

glennsl:
Here's an overview of the current breaking changes:

**Commands:**
```
terminal.normalMode -> oni.terminal.normalMode
terminal.insertMode -> oni.terminal.insertMode
explorer.toggle -> oni.explorer.toggle
keyDisplayer.enable -> oni.keyDisplayer.enable
keyDisplayer.disable -> oni.keyDisplayer.disable
sneak.start -> oni.sneak.start
sneak.stop -> oni.sneak.stop
system.addToPath -> oni.system.addToPath
system.removeFromPath -> oni.system.removeFromPath
view.rotateForward -> oni.view.rotateForward
view.rotateBackward -> oni.view.rotateBackward
vim.esc -> oni.vim.esc
workbench.action.reloadSettings -> oni.workbench.action.reloadSettings
workbench.action.enableZenMode -> oni.workbench.action.enableZenMode
workbench.action.disableZenMode -> oni.workbench.action.disableZenMode
```

**Context Keys:**
```
quickmenuCursorEnd -> oni.quickmenuCursorEnd
commandlineFocus -> oni.commandlineFocus
insertmode -> oni.insertMode
normalMode -> oni.normalMode
visualMode -> oni.visualMode
sneakMode -> oni.sneakMode
zenMode -> oni.zenMode
keyDisplayerEnabled -> oni.keyDisplayerEnabled
symLinkExists -> oni.symLinkExists
```

In addition I'd like to replace some of the oni-specific commands one's that use vscode conventions or equivalent vscode commands where those exist:

```
one.explorer.toggle -> workbench.action.toggleSidebarVisibility
oni.workbench.action.enableZenMode -> workbench.action.toggleZenMode
oni.workbench.action.disableZenMode -> workbench.action.toggleZenMode
oni.sneak.start -> oni.workbench.action.toggleSneakMode
oni.sneak.stop -> oni.workbench.action.toggleSneakMode
oni.keyDisplayer.enable -> oni.workbench.action.toggleKeyDisplayer
oni.keyDisplayer.disable -> oni.workbench.action.toggleKeyDisplayer
```

Many of these changes are caused by poor choices made earlier. We therefore ought to take care to make sure we do it right this time to avoid future breaking changes, and if we decide to follow different conventions we could eliminate some of these. We could for example decide to have `vim` be a top-level oni-specific namespace as well, which would eliminate the change from `vim.esc -> oni.vim.esc` and allow changing `oni.insertMode -> vim.insertMode`.

Either way, as long as we have breaking changes we have to decide how to mitigate the breaking, where as I see it we have three choices:

1. Communicate the changes clearly, using a "dynamic" in-app release log that shows all the breaking changes since the last version run once on the Welcome screen, for example. This should be fairly straight-forward and contained to implement, with little added maintenance burden other than having to record the breaking changes. And the infrastructure can support any kind of breaking change as it doesn't rely on automation. It does put more of a burden on the end-user however.

2. Remain backwards compatible by providing transforms from old to new functionality. This is relatively complicated, adds technical debt and the maintenance burden that goes with it, is difficult to adapt and quickly increases in complexity for changes that don't map 1:1, such as some of the proposed future changes. It also doesn't actually solve some of the potential problems the changes here intend to solve, as keeping the old non-namespaced values still leaves us open for conflicts with extension contributions and future changes to vscode, and removing the backwards-compatibility later is still a breaking change. The biggest benefit of this I think is that it doesn't require immediate user intervention.

3. Add a general transformation mechanism for configuration/JSON files. This is much like 2, with most of the same benefits and drawbacks, but can be made more contained as a separate layer between parsing and decoding. The cost is that the generality adds significantly more complexity, and it still won't easily support changes that don't map 1:1. Another benefit is that the files can easily be written back to disk immediately after transformation, which means we should be able to remove old transformations after just one or two releases and it wouldn't require any user intervention at all (although users might still be surprised to see their configuration has changed unexpectedly, especially if they synchronize it across computers, as I do).

I've probably missed something in my descriptions above, but based on the cost/benefit outlined above my preference is still clearly option 1. I think it has the lowest cost, while arguably also having the greatest benefit, mostly because option 2 and 3 might not actually support changes that don't map 1:1, and certainly not easily. I also think it's a very useful mechanism to have even without breaking changes, and even if option 2 or 3 is implemented as well, as awareness of breaking changes is still important even if user intervention isn't required. I'd also rank option 3 far ahead of 2, as I think the generality makes it much more useful despite the extra cost.

bryphe:
> In addition I'd like to replace some of the oni-specific commands one's that use vscode conventions or equivalent vscode commands where those exist:

```
one.explorer.toggle -> workbench.action.toggleSidebarVisibility
oni.workbench.action.enableZenMode -> workbench.action.toggleZenMode
oni.workbench.action.disableZenMode -> workbench.action.toggleZenMode
oni.sneak.start -> oni.workbench.action.toggleSneakMode
oni.sneak.stop -> oni.workbench.action.toggleSneakMode
oni.keyDisplayer.enable -> oni.workbench.action.toggleKeyDisplayer
oni.keyDisplayer.disable -> oni.workbench.action.toggleKeyDisplayer
```

Thanks for going through our commands and cleaning them up, @glennsl ! This mapping looks good to me.

> Many of these changes are caused by poor choices made earlier. We therefore ought to take care to make sure we do it right this time to avoid future breaking changes, and if we decide to follow different conventions we could eliminate some of these.

Definitely. I'm glad we have a better convention proposed here.

> We could for example decide to have vim be a top-level oni-specific namespace as well, which would eliminate the change from vim.esc -> oni.vim.esc and allow changing oni.insertMode -> vim.insertMode.

This would be my preference - whether we go with option 1 or 2. My concern with the `vim.esc`, particularly, is that we have collateral around the 0.4.0 release that shows binding to it - it would be bad experience for someone to see the post / video, try out the binding demo'd in it, and have it not work.

I really appreciate the clear descriptions of the potential approaches 1 through 3. My preference is still to minimize the impact on the end-user via option 2, with the assumption that a conflict is unlikely. This is the approach we'll have to take down the road anyway.

But, I'm OK with option 1, as long as we keep the `vim.esc` command at the top-level. It'd be nice to have the infrastructure to show breaking changes (or just all changes since the last update!) in a nice changelog on the welcome screen. Also opens the door for further enhancements to improve our welcome experience.

glennsl:
I've reverted the namespacing changes, so this should be good to go now. Will follow up soon with a dynamic in-app change log, then redo the namspacing changes as well as the future chanegs outlined above.

src/Model/ContextKeys.re
glennsl: Fixed in https://github.com/onivim/oni2/pull/1595/commits/4590021853321fe5915b2490f882df02fffb8897

PR #1601, opened 2020-04-13
Bugfix: add terminalFocus to quickOpen keybinding

### LINKED ISSUE #1600: Ctrl-Tab doesn't work in a terminal window
Demo: https://i.imgur.com/HcF4WCi.mp4

glennsl:
Fixes #1600


glennsl:
I'm not entirely sure why the quick menu keybindings need to have any condition at all, as they don't in vscode. If there are issues in certain conditions, perhaps we should excluded those conditions instead of including everything else, to avoid future bugs like this.

Issue #1475, opened 2020-03-16
weird font display - Full-width characters overlapping


kmyk:
Full-width characters are rendered with overlapping.

![Screenshot from 2020-03-16 20-32-00](https://user-images.githubusercontent.com/2203128/76754424-73f6e500-67c5-11ea-8fd0-f20f09ea863a.png)

environment: Ubuntu 18.04.4 LTS (Bionic Beaver)

This is the same problem to as described in https://github.com/onivim/oni2/issues/1293#issuecomment-599499581

---

- I tested with Noto Sans, IPA Gothic, and VL Gothic. All fonts have the same problem.
 To test, download [NotoSansCJKjp-hinted.zip](https://noto-website-2.storage.googleapis.com/pkgs/NotoSansCJKjp-hinted.zip) from https://www.google.com/get/noto/help/cjk/ and set `"editor.fontFamily": "/path/to/NotoSansMonoCJKjp-Regular.otf"`. (Setting `"editor.config": "Noto Sans Mono CJK JP"` or `"editor.fontfamily": "/usr/share/fonts/opentype/noto/notosanscjk-regular.ttc"` installed via `apt install fonts-noto-cjk` doesn't work.)
- Test file: [tofu.md](https://github.com/onivim/oni2/files/4337959/tofu.md.txt)
- Setting `"editor.fontSmoothing": "subpixel-antialiased"` (also `"editor.fontSmoothing": "antialiased"` and `"editor.fontSmoothing": "none"`) doesn't solve this. (suggested at https://github.com/onivim/oni2/issues/1293#issuecomment-599505168)

glennsl:
Thanks for reporting, and for the detailed information. I'll try to reproduce this myself when I get the opportunity, as I use Linux myself. Would be great if you could try to reproduce on Windows as well though, to see whether this is Linux-specific.

Could you also post the log from running `oni2 -f --trace --log-filter Window,Font`?

kmyk:
Here is the log on Ubuntu 18.04.4 LTS.
(For Windows, please wait several days...)

```
$ oni2 -f --trace --log-filter Window,Font       
[DEBUG]   +50ms Oni2.Exception : Recording backtraces
[DEBUG]  +121ms Revery.Core.Window : Starting window creation...
[INFO]     +0ms Revery.Core.Window : Creating window Oni2 width: 800 height: 600
[INFO]   +124ms Revery.Core.Window : Window created successfully.
[...]

PR #1608, opened 2020-04-13
Fix #1498 - Accented characters cause cursor to move to beginning of line

### LINKED ISSUE #765: Nordic characters offsets cursor display
If there are any Nordic characters (e.g. æ, ø, and å) on a line before the current cursor position, the cursor is displayed the corresponding number of cells too far right.

Normal mode commands work on the character the cursor is "really" on rather than the one it's displayed at.

Multiple æ's are truncated a few pixels too, so it's likely related to wide character support.

### LINKED ISSUE #1498: Oni2 INSERT MODE accents
## Description
When in INSERT mode you write the portuguese / spanish letter "é", the cursor jumps to the start of the line, but you keep writing on the same position, until the letter is deleted

## Version
commit: d61477c4

## Example
example

bryphe:
__Issue:__ Multibyte UTF-8 characters were being handled incorrectly - we were using the _byte position_ of the cursor (which is what is sent as the cursor column from libvim) as the _character_ position - this would cause problems in any case with a multibyte character, causing the cursor to be offset (and eventually just go to the beginning of the line). For ASCII text, this is OK because it is 1-to-1 (1 byte for 1 character)... but fails with any multibyte character.

__Fix:__
- Update `BufferLine` to handle UTF-8 characters - implement a `toIndex` method that maps from byte position -> character index.
- Use the byte position map to resolve the cursor position as a character position, so that it is rendering correctly and shows the expected position in the status bar.
- Look up character width via a `wcwidth`-like strategy (previously, the width was always hard-coded to '1' character width). 

__Todo:__
- [x] Add integration / regression test for #1498 (get highlighted character, verify position + character under cursor is correct when moving).
- [x] Test / validate syntax highlighting (which is currently based on byte position, rather than character indices)
- [x] Test mouse position - ensuring that the mapping of cursor position -> character position is correct.

__Future work:__
- Test CJK characters (https://github.com/onivim/oni2/issues/1475)
- Instead of using the `wcwidth`/`tty_character_width` heuristic (which is not 100% accurate), we could look at using the actual measured font characters. This would have the benefit of giving us support for non-monospaced fonts.
- Investigate using the type system to enforce the difference between _byte positions_ and _utf-8 character positions_
- Architecture: There is coupling between the `Buffer` and `Editor` - considering moving the `Buffer` to be owned by `Editor`s as the source of truth. They'll be more coupling adding as we implement word wrap.

Fixes #1498 
Fixes #765


src/Core/BufferLine.re
```diff
@@ -17,14 +17,17 @@ type characterCacheInfo = {
   width: int,
 };
 
-let emptyArray = Array.make(0, None);
+let emptyArrayCharacters = Array.make(0, None);
+let emptyArrayIndex = Array.make(0, None);
```
glennsl, line 21: This confuses me. First of all, arrays are mutable, so a simple mistake could suddenly have these be non-empty. Secondly, why are there two of them, if they're not (supposed to be) mutated? Are their physical identity significant somehow? Why not just use the empty array literal `[||]` and compare structurally against that? Or use `Array.length` to check for emptiness.

I could see one reason for this being that the empty array literal will allocate more that doing this, although I would think the compiler should be able to optimize that away. But if it is, this should be documented with a comment. And it's still quite error-prone because of the assumption that it won't be mutated.

src/Core/BufferLine.re
```diff
@@ -17,14 +17,17 @@ type characterCacheInfo = {
...
 
 type t = {
   indentation: IndentationSettings.t,
   // [raw] is the raw string (byte array)
   raw: string,
   // [characters] is a cache of discovered characters we've found in the string so far
   mutable characters: array(option(characterCacheInfo)),
+  // [byteMap] is a cache of byte -> index
+  mutable byteMap: array(option(int)),
```
glennsl, line 30:
```suggestion
  mutable byteIndexMap: array(option(int)),
```
?

src/Core/BufferLine.re
glennsl: Actually, native OCaml doesn't have dynamic arrays. I was thinking in terms of JavaScript/BuckleScript where even empty arrays can be mutated. That shouldn't be a problem here, but this still is very weird code to me, that at least deserves a comment explaining it.

src/Core/BufferLine.re
```diff
@@ -88,8 +98,17 @@ module Internal = {
 let make = (~indentation, raw: string) => {
   // Create a cache the size of the string - this would be the max length
   // of the UTF8 string, if it was all 1-byte unicode characters (ie, an ASCII string).
-  let characters = emptyArray;
-  {indentation, raw, characters, nextByte: 0, nextIndex: 0, nextPosition: 0};
+  let characters = emptyArrayCharacters;
+  let byteMap = emptyArrayIndex;
```
glennsl, line 102: This seems like unnecessary indirection when they can just be assigned to the record fields directly.

src/Core/BufferLine.re
bryphe: I added more description about these empty / sentinel values - hopefully it helps explain how they are used.

glennsl: Thanks. I still think it's weird that the compiler isn't able to optimize away allocation of empty arrays though. Seems like a trivial thing to do, and very low-hanging fruit.

bryphe: I created a benchmark to check - it seems that it does optimize away the allocation, but the structural equality check is still more expensive than a reference equality check:
```
|  Array: structural equality check                                   |  1000000    |  0.0139060020447    |  0         |  0         |  27           |  0          |  0            |
|  Array: reference equality check                                    |  1000000    |  0.00193691253662   |  0         |  0         |  27           |  0          |  0            |
```

glennsl: Yeah, that's probably because it's polymorphic, and why Base/Core and others don't have them in `Pervasives`.

miyalys:
Thank you!

PR #1609, opened 2020-04-14
Fix #1607 - Stack overflow in Diagnostic.explode

### LINKED ISSUE #1607: Stack Overflow in List.map caused by diagnostics
To reproduce:

1. Clone https://github.com/glennsl/oni2/tree/repro/diagnostics/stack-overflow
2. Go to [GlobalColors.re:558](https://github.com/glennsl/oni2/blob/9a4fb3debc36b73401ab986e1bfc7c2b40da0ef0/src/Feature/Theme/GlobalColors.re#L558)
3. Start deleting using backspace from the end of the line

This isn't very recent. It happens at least as far back as https://github.com/onivim/oni2/commit/60adf3c71b5f489b150f0c98438aad0f10c1c4fd, from about two weeks ago.

I've not been able to figure out exactly what causes it, but have narrowed it down to being triggered by incoming diagnostics, as replacing `perFileDiagnostics` at [ExtHostProtocol.re:312](https://github.com/glennsl/oni2/blob/9a4fb3debc36b73401ab986e1bfc7c2b40da0ef0/src/Extensions/ExtHostProtocol.re#L312) with an empty list fixes the issue.

The incoming payload that triggers it is:

```json
[
   [
      {
         "$mid":1,
         "external":"file:///home/glennsl/dev/remote/glennsl/oni2/src/Feature/Theme/GlobalColors.re",
         "path":"/home/glennsl/dev/remote/glennsl/oni2/src/Feature/Theme/GlobalColors.re",
         "scheme":"file"
      },
      [
         {
            "startLineNumber":557,
            "startColumn":64,
            "endLineNumber":424243,
            "endColumn":4,
            "message":"Some record fields are undefined: light hc",
            "severity":8
         },
         {
            "startLineNumber":558,
            "startColumn":11,
            "endLineNumber":558,
            "endColumn":14,
            "message":"This function has type string => Oni_Core__ColorTheme.Defaults.expr\nIt is applied to too many arguments; maybe you forgot a `;'.",
            "severity":8
         },
         {
            "startLineNumber":424242,
            "startColumn":5,
            "endLineNumber":424242,
            "endColumn":7,
            "message":"File \"GlobalColors.re\", line 424242, characters 4-6:\nError: Unclosed \"(\" (opened line 558, column 13)",
            "severity":8
         }
      ]
   ]
]
```

I'm thinking it could have something to do with those line numbers being a tad too big. But after going through a number of `List.map` invocations, logging the length of the list they map over and replacing with `ListEx.safeMap`, I still have not been able to track it down.

bryphe:
__Issue:__ In certain conditions - like a diagnostic covering a very, very large range - we could crash when 'exploding' the diagnostic across lines.

__Defect:__ The `List.map` wasn't stack-safe, so for a very large list range, we'd hit a stackoverflow.

__Fix:__ Create a maximum acceptable length to 'explode', and use the stack-safe variant of `List.map`. Without clamping the list via `List.firstk`, we'd still hit a stack overflow in the consolidation of the Diagnostics in the `_explodeDiagnostics` when we go to flatten.

Fixes #1607

### DIFF
--- a/src/Feature/LanguageSupport/Diagnostic.re
+++ b/src/Feature/LanguageSupport/Diagnostic.re
@@ -18,7 +18,6 @@ let create = (~range, ~message, ()) => {range, message};
 let maxDiagnosticLines = 1000;
 
 let explode = (buffer, diagnostic) => {
-  prerr_endline("Exploding range: " ++ Range.show(diagnostic.range));
   let lineCount = Buffer.getNumberOfLines(buffer);
   let measure = n => {
     Index.toZeroBased(n) < lineCount
--- a/src/Feature/LanguageSupport/Diagnostic.re
+++ b/src/Feature/LanguageSupport/Diagnostic.re
@@ -31,6 +31,6 @@ let explode = (buffer, diagnostic) => {
   };
 
   Range.explode(measure, diagnostic.range)
-  |> ListEx.firstk(maxDiagnosticLines)
+  |> ListEx.firstk(Constants.maxDiagnosticLines)
   |> ListEx.safeMap(range => create(~range, ~message=diagnostic.message, ()));
 };
--- a/src/Feature/LanguageSupport/Diagnostic.re
+++ b/src/Feature/LanguageSupport/Diagnostic.re
@@ -4,6 +4,7 @@
 
 open EditorCoreTypes;
 open Oni_Core;
+open Oni_Core.Utility;
 
 [@deriving show]
 type t = {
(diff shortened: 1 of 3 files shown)


src/Feature/LanguageSupport/Diagnostic.re
```diff
@@ -13,6 +14,9 @@ type t = {
 
 let create = (~range, ~message, ()) => {range, message};
 
+// Clamp the number of lines a diagnostic range can span to a tractable number.
+let maxDiagnosticLines = 1000;
```
glennsl, line 18: Should be in a `Constants` sub-module IMO

PR #1610, opened 2020-04-14
Fix #1599 - Unresponsive terminal with large output

### LINKED ISSUE #1599: Bug: Onivim becomes unresponsive if built-in terminal runs a command with a lot of long lines
I hit this bug when I was trying to run Ruby on Rails + Sidekiq inside Onivim.

The synthetic way of reproducing this is to run a command which generates a lot of long lines.

Steps to reproduce
1. Open terminal using `:term` command (
2. Enter the following command (in bash): `for i in $(seq 1 10000); do printf %1000s |tr " " "-"; done`

Expected result:
Command is being executed in terminal while other parts of editor keep working and respond to user interaction

Actual result:
Editor freezes (almost) instantly and has to be killed.

OS: Arch Linux, Onivim revision 18b0475a24d9cc34a7d9d8e9f55b2a8c1bc33478, built from AUR (`onivim2-git` package)

bryphe:
__Issue:__ The terminal can hang with large amounts of incoming data.

__Defect:__ We were doing several expensive operations during the course of receiving new data from the terminal pty, like:
- Querying for a current working directory
- Dispatching actions excessively (which come with boilerplate: logging, executing subscriptions, etc - which can add up if the actions are too high frequency).

__Fix:__
- Don't compute Sys.getcwd all the time in our subscription code
- Debounce the screen updates, to keep the app responsive

__TODO:__
- [x] Speed up rendering by 'batching' strings (https://github.com/revery-ui/revery-terminal/pull/17)
- [x] GC also came up in profiling - see what we can trim in the 'hot path'

Fixes #1599


src/Service/Terminal/Service_Terminal.re
```diff
@@ -156,6 +159,11 @@ module Sub = {
           ReveryTerminal.resize(~rows, ~columns, state.terminal);
           state.isResizing := false;
           {...state, rows, columns};
+        } else if (state.queuedScreen^ != None) {
+          let screen = Option.get(state.queuedScreen^);
+          dispatch(ScreenUpdated({id: params.id, screen}));
+          state.queuedScreen := None;
+          state;
```
glennsl, line 166: This seems to put a weird dependency between the state being updated and the terminal updating, which can lead to some unintuitive bugs. If we manage to reduce the number of status updates, which is a good thing, the terminal might start to feel unresponsive. And if the number of state updates increase, in the same manner as this would if dispatched directly, it still won't solve the problem but actually exacerbate it.

I'm not sure what downside there is to just using a timer, other than needing a bit more wiring, but it would make the system as a whole significantly simpler I think.

src/Store/StoreThread.re
```diff
@@ -187,7 +188,7 @@ let start =
     let workspaceUri =
       state.workspace
       |> Option.map((ws: Model.Workspace.workspace) => ws.workingDirectory)
-      |> Option.value(~default=Sys.getcwd())
+      |> OptionEx.value_lazy(~default=Sys.getcwd)
       |> Oni_Core.Uri.fromPath;
```
glennsl, line 192: Using a `switch` would make this simpler and clearer than adding a new utility function for lazy evaluation I think.

src/Service/Terminal/Service_Terminal.re
bryphe: Good point - I moved to a timer / debounce strategy, which should be more reliable than relying on the timing of the `update`s.

src/Store/StoreThread.re
bryphe: Hmm, I tried this and have:
```
    let workspaceUri =
      state.workspace
      |> Option.map((ws: Model.Workspace.workspace) => ws.workingDirectory)
      |> (
        fun
        | None => Sys.getcwd()
        | Some(uri) => uri
      )
      |> Oni_Core.Uri.fromPath;
```

But I think this is a bit less readable - did you have something else in mind?

Otherwise, we do have other helper methods that are a similar level of abstraction to `OptionEx.value_lazy` - like [`or_lazy`](https://github.com/onivim/oni2/blob/master/src/Core/Utility/OptionEx.re#L36)

glennsl: I was thinking something more like this:

```suggestion
      switch (state.workspace) {
      | Some(workspace: Model.Workspace.workspace) => workspace.workingDirectory
      [ None => Sys.getcwd()
      }
      |> Oni_Core.Uri.fromPath;
```

Issue #1471, opened 2020-03-14
Oni2 crashes when opening *.rst file


gour:
Hello,

I've the following extensions installed:
```
$ oni2 --list-extensions                                                               
python
restructuredtext
```

and after opening *.rst file, Oni2 crashes with the following log:

[oni2-rst.log](https://github.com/onivim/oni2/files/4332520/oni2-rst.log)

glennsl:
Thanks for the report @gour. From the log I can see that the extension loaded successfully, and I can't see anything else suspicious. Could you try again with `--trace` to get more verbose logging?

Also, is this specific to one `.rst` file, or doers it happen with any?

Is there an `onivim2-crash.log` in the directory you ran it from? If we're able to catch the exception we log the exception information there.

gour:
> Is there an onivim2-crash.log in the directory you ran it from? If we're able to catch the exception we log the exception information there.

```
Yojson.Safe.Util.Type_error("Expected array, got null", 870828711):
```

glennsl:
Oh, that's promising. Seems like it is sending us something we don't expect. Could you try run it again with `OCAMLRUNPARAM=b` set? That should capture the stack trace as well.

```
OCAMLRUNPARAM=b oni2 -f --trace`
```

will log the stack trace and give more detailed logging of what's being sent back and forth.

gour:
> Could you try again with --trace to get more verbose logging?
In the meantime I've installed few more extensions....
```
$ oni2 --list-extensions                                                               
vscode-tailwindcss
context-syntax
fossil
pinegrow-vscode
jinjahtml
jinja
html-css-class-completion
restructuredtext
```
> Also, is this specific to one .rst file, or doers it happen with any?

...and it crashes with other 1.1K ConTeXt  file as well. 

Here is the new crash-log:

```
Yojson.Safe.Util.Type_error("Expected array, got null", 870828711):
Raised at file "util.ml", line 15, characters 20-60
Called from file "util.ml" (inlined), line 19, characters 13-19
Called from file "src/Extensions/ExtHostClient_SCM.re", line 165, characters 31-68
Called from file "util.ml" (inlined), line 19, characters 13-19
Called from file "src/Extensions/ExtHostClient_SCM.re", line 157, characters 14-427
Called from file "src/Extensions/ExtHostClient.re", line 224, characters 6-74
Called from file "src/Extensions/ExtHostTransport.re", line 128, characters 13-53
Called from file "list.ml", line 110, characters 12-15
[...]

PR #1616, opened 2020-04-15
Feature: XML Theme and Grammar support

### LINKED ISSUE #912: Support non-JSON Textmate Grammars and Themes
We currently support JSON textmate grammars and themes, but there are technically XML based ones:

https://github.com/JustusAdam/language-haskell/blob/master/syntaxes/haskell.tmLanguage

This (as far as I can tell) is the only/most used Haskell grammar for code, and its written in XML.

If `reason-textmate` was able to load an XML file and convert it into JSON, then parse that JSON file, we should be okay.

I found a random converter online whilst looking (http://json2plist.sinaapp.com/) and doing that on that file and saving it as JSON did seem to work (though there was some odd highlighting for comments, but that could be unrelated, or a bad conversion), so programmatically doing that conversion should be all we need.

I've raised this here, but its more of a `reason-textmate` issue, and if fixed there I think could be an invisible change on the Oni2 side.

### LINKED ISSUE #1486: Window freezes on Ctrl + Shift + P and searching - Windows 10
Hi, every time I open the editor on my Windows machine, hit Ctrl + Shift + P, and then start to search for *theme*, my whole window crashes and I have to force quit the application and restart it. The odd thing is that when opening from my Git Bash terminal everything works correctly, but when opening from Powershell, Command Prompt, or just the application itself, I can reproduce the issue.

**Attached below is my log file of the most recent crash:**
[log.txt](https://github.com/onivim/oni2/files/4361575/log.txt)
[onivim2-crash.log](https://github.com/onivim/oni2/files/4361576/onivim2-crash.log)

glennsl:
Brings in the changes https://github.com/onivim/reason-textmate/pull/54 and https://github.com/onivim/reason-textmate/pull/55 which added parsing of XML grammars and themes, respectively, to reason-textmate.

Has so far been tested to work with:

- https://github.com/chiragpat/tomorrow-and-tomorrow-night-operator-mono-theme
- https://github.com/glennsl/vscode-duotone-dark-faithful
- https://github.com/JustusAdam/language-haskell

Let me know if there's anything else we should test.

Fixes #912 
Probably fixes #1486 
Addresses #1471, but does not completely fix it. See comment below

### DIFF
--- a/package.json
+++ b/package.json
@@ -252,7 +252,8 @@
     "esy-oniguruma": "onivim/esy-oniguruma#4698ce4",
     "editor-input": "onivim/editor-input#c494950",
     "isolinear": "revery-ui/isolinear#b17ce17",
-    "revery-terminal": "revery-ui/revery-terminal#2ef1c65"
+    "revery-terminal": "revery-ui/revery-terminal#2ef1c65",
+    "reason-textmate": "onivim/reason-textmate#5269e5f"
   },
   "devDependencies": {
     "ocaml": "~4.8",
--- a/package.json
+++ b/package.json
@@ -214,7 +214,7 @@
     "revery": "0.30.0",
     "revery-terminal": "*",
     "reason-tree-sitter": "^1.0.1001",
-    "reason-textmate": "^3.0.0",
+    "reason-textmate": "^3.1.0",
     "esy-sdl2": "*",
     "esy-skia": "*",
     "@opam/fmt": "^0.8.8",
@@ -252,8 +252,7 @@
     "esy-oniguruma": "onivim/esy-oniguruma#4698ce4",
     "editor-input": "onivim/editor-input#c494950",
     "isolinear": "revery-ui/isolinear#b17ce17",
-    "revery-terminal": "revery-ui/revery-terminal#2ef1c65",
-    "reason-textmate": "onivim/reason-textmate#5269e5f"
+    "revery-terminal": "revery-ui/revery-terminal#2ef1c65"
   },
   "devDependencies": {
     "ocaml": "~4.8",
--- a/src/Syntax/GrammarRepository.re
+++ b/src/Syntax/GrammarRepository.re
@@ -25,18 +25,33 @@ let getGrammar = (~scope: string, gr: t) => {
     switch (Ext.LanguageInfo.getGrammarPathFromScope(gr.languageInfo, scope)) {
     | Some(grammarPath) =>
       gr.log("Loading grammar from: " ++ grammarPath);
-      let json = Yojson.Safe.from_file(grammarPath);
-      let resultGrammar = Textmate.Grammar.Json.of_yojson(json);
 
-      switch (resultGrammar) {
-      | Ok(grammar) =>
-        gr.log("Grammar loaded successfully");
-        Hashtbl.add(gr.scopeToGrammar, scope, grammar);
-        Some(grammar);
-      | Error(e) =>
-        gr.log("Grammar loading failed with: " ++ e);
-        None;
+      switch (Yojson.Safe.from_file(grammarPath)) {
+      | json =>
+        switch (Textmate.Grammar.Json.of_yojson(json)) {
+        | Ok(grammar) =>
+          gr.log("JSON Grammar loaded successfully");
+          Hashtbl.add(gr.scopeToGrammar, scope, grammar);
+          Some(grammar);
+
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -251,7 +251,8 @@
     "esy-oniguruma": "onivim/esy-oniguruma#4698ce4",
     "editor-input": "onivim/editor-input#c494950",
     "isolinear": "revery-ui/isolinear#b17ce17",
-    "revery-terminal": "revery-ui/revery-terminal#2ef1c65"
+    "revery-terminal": "revery-ui/revery-terminal#2ef1c65",
+    "reason-textmate": "onivim/reason-textmate#5269e5f"
```
glennsl, line 255: TODO maybe, if we want to do a proper release instead of using a resolution

glennsl:
https://github.com/vscode-restructuredtext/vscode-restructuredtext fails with "invalid backref number/name". Seems like a regex issue.

bryphe:
> https://github.com/vscode-restructuredtext/vscode-restructuredtext fails with "invalid backref number/name". Seems like a regex issue.

Interesting... yes, seems unrelated to the new parsing logic (unless there's some sort of weird string escaping happening).

package.json
bryphe: Sure, I'll do a release. Awesome to have support for plist/xml themes and grammars now!

bryphe: Published as `reason-textmate@3.1.0`

src/Syntax/GrammarRepository.re
```diff
@@ -25,18 +25,33 @@ let getGrammar = (~scope: string, gr: t) => {
...
+          gr.log("JSON Grammar loaded successfully");
+          Hashtbl.add(gr.scopeToGrammar, scope, grammar);
+          Some(grammar);
+
+        | Error(e) =>
+          gr.log("Grammar loading failed with: " ++ e);
+          None;
+        }
+
+      | exception (Yojson.Json_error(_)) =>
```
bryphe, line 42: I wonder if we should determine this based on file extension (or - at the very least - decide which implementation to run first?)

bryphe:
I get to test out the Duotone themes now! 🎉 

image

src/Syntax/GrammarRepository.re
glennsl: I've just assumed they could have any extension, although we could of course still try to infer it from some common extensions. But does it really matter? Do you think there's really a noticeable effect from just trying it blindly?

glennsl:
https://github.com/SuperiorServers/vscode-glua fails because it doesn't have a repository

glennsl:
After making repository optional, that one fails with "invalid backref" too.

src/Syntax/GrammarRepository.re
bryphe: > But does it really matter? Do you think there's really a noticeable effect from just trying it blindly?

The main concern I had was performance related to synchronous file I/O - with the current strategy, for XML grammars, we always have to load them twice from the file system, and some of the grammars can be fairly large (~200Kb or so). 

That's my only concern; if there's no perf impact with this approach, I'm fine with it.

glennsl: They're cached though, so it should at least only happen once per grammar. And the parsing should fail pretty quickly, so unless the JSON parser reads the entire file in before parsing, file size shouldn't matter. The XML parser definitely doesn't, as it's a streaming API, but I also think it's so fault-tolerant that it'll probably just return the entire JSON file as a text node if we switch it around.

I can check what Yojson does though.

glennsl: Yojson uses ocamllex, which I'm pretty sure reads incrementally.

glennsl:
I created a PR for the invalid backref problem at https://github.com/onivim/reason-textmate/pull/57. I see no reason it should block this though.

bryphe:
> I created a PR for the invalid backref problem at onivim/reason-textmate#57. I see no reason it should block this though.

Thanks for looking at this! Agreed,  not blocking 👍

src/Syntax/GrammarRepository.re
bryphe: Thanks for looking into this. If they are both streaming / incremental, should be negligible overhead.

glennsl: The overhead would also only be for XML grammars, which most aren't.

PR #1617, opened 2020-04-15
Dependency: Add libuv + health check

bryphe:
- Bring over the addition of `luv` and the integration of the event loop from #1529 
- Add a health-check to verify packaging and linking of libuv
- Add to version screen


glennsl:
I think we can just merge #1529 instead.

bryphe:
Sounds good, @glennsl !

Issue #1613, opened 2020-04-14
Memory leak in Syntax Highlight Service


CrossR:
There has been 2-3 reports on Discord with people having memory leaks that seem to be coming from the syntax highlight process specifically.

To my knowledge, this has been reported on Linux only so far, and can happen both when the editor is running, and also on close (this one may have changed since #1488, the last report of this was before then).

The process basically uses all available memory on the machine: 

I'll link this issue in Discord and hopefully we can get some logs from someone, as I've been unable to repro on my machine (source build, tar.gz, appimage).

necabo:
So I just compiled 8ba63de2 to ensure my repro still works. Running Arch Linux on 5.6.3-arch1-1.

It's quite simple and 100% reproducible for me. Sending a SIGTERM to Oni2 in order to close it causes the highlight service to go beserk until the oom killer finally gets rid of it. 2-3 minutes after closing Oni2 that way, it looks something like this: 
![oni2](https://user-images.githubusercontent.com/23185845/79273333-209ab400-7ea3-11ea-85a7-17f1a7169fd7.png)

This issue doesn't occur when using `:q` to quit.

CrossR:
With the changes that were made (#1619, #1621) above and the fact there has been no other reporting of this recently, I'll close this out.

If someone is still experiencing this, we can re-open.

PR #1619, opened 2020-04-15
#1613 - Part 1 - Syntax Server: Fail fast on error

bryphe:
__Issue:__ On contributing factor to the syntax server problems in #1613 is that, when something goes wrong (like the channel is closed) - we get stuck spinning in a thread - eating up CPU and memory - rather than failing fast.

__Fix:__ This is the start of investigating #1613, and implements a test case to validate we close successfully on an exception with a read failure. This should partially address the first issue - where, on closing the Onivim process with a signal, the syntax service eats up CPU.

I suspect that the other case - where, during runtime, the syntax server is eating up memory - it's also hitting an exception. 

__TODO:__
- [x] Add test case for process-disconnect

Next steps (separate PRs):
- Improve logging / diagnosability - output syntax server to separate logs, and log out if there is an exception that closes it. Remove the 'forwarding' of logs from syntax server to syntax client, in order to simplify and reduce traffic.
- Add test case for native crash
- Switch to libuv to manage process communication, which is more robust and battle-tested than our read/write thread strategy - this will help us to better manage the connect/disconnect lifecycle, and mitigate classes of errors around the complexity of the threading logic.


src/Syntax_Client/Oni_Syntax_Client.rei
```diff
@@ -41,4 +42,5 @@ let healthCheck: t => unit;
 let close: t => unit;
 
 // Debug methods
-let simulateException: t => unit;
+let simulateReadException: t => unit;
+let simulateMessageException: t => unit;
```
glennsl: I would put these in a `Testing` module. Descriptive code is always better than comments, since comments are easily ignored and quickly get outdated and become misinforming. a sub-module also prevents them from showing up among "normal" completions, where comments don't help at all.

src/Syntax_Server/Oni_Syntax_Server.re
```diff
@@ -151,38 +153,46 @@ let start = (~healthCheck) => {
...
-          // If the messages incurred work, do it!
-          if (State.anyPendingWork(state^)) {
-            log("Running unit of work...");
-            map(State.doPendingWork);
-            log("Unit of work completed.");
-          } else {
-            log("No pending work.");
+          try(
+            {
+              log("Waiting for incoming message...");
```
glennsl, line 156: I don't know if this is part of your plan already, but I'd really like to see this `log` function disappear in favor of using `Timber` with a custom reporter that forwards messages to the main process, as that would give access to all its logging features. Most importantly the severity, but we could also set the log level so we can increase the granularity if needed without affecting performance.

src/Syntax_Client/Oni_Syntax_Client.rei
bryphe: Good idea! 👍

src/Syntax_Server/Oni_Syntax_Server.re
bryphe: Good call! Yes, I plan on using `Timber` and adding a file-logging option. 

I'm considering removing the log forwarding, it's a bit misleading given that the timing is interleaved incorrectly - I think it'd actually be more understandable as a separate file. The important parts of the interaction for most cases can be gleaned from the editor's log - the guts of the syntax server are most important for debugging, IMO. I'll work on a PR and we can discuss more on that.

PR #1620, opened 2020-04-16
feat(workspace): keyboard window navigation

glennsl:
Don't get too excited! This is blocked hard by a bug that likely involves som timing dependency that's really tricky to pin down (at least for me).

To reproduce:
- Split the editor
- Open another editor in the newly created group
- Select the first group, by mouse or by keyboard
- Select the second group again
- Observe that the active editor of the second group moves back to the first editor

It's also possible to get an editor to open from another group. It seems like the underlying issue is that the vim and oni gets out of sync. But I don't understand why, as from what I can tell the refactoring done so far should preserve the order of operations.

Before refactoring, the active editor was calculated before dispatching `WindowSetActive`, which would update the state in `WindowStore` and then synchronize in `VimStore`. After refactoring, `WindowMove*` is dispatched immediately, which will both calculate the active editor and update the state in `WindowStore`, and then synchronize in `VimStore`. That's really the gist of it, and I can't see how that could cause this. And yet it does...

### DIFF
--- a/src/Feature/Layout/Feature_Layout.re
+++ b/src/Feature/Layout/Feature_Layout.re
@@ -1,11 +1,5 @@
 open Utility;
 
-type direction =
-  | Up
-  | Left
-  | Down
-  | Right;
-
 [@deriving show({with_path: false})]
 type size =
   | Weight(float);
@@ -290,26 +284,17 @@ let rec layout = (x, y, width, height, tree) => {
   };
 };
 
-let moveCore = (current, dirX, dirY, tree) => {
+let move = (current, dirX, dirY, tree) => {
   let layout = layout(0, 0, 200, 200, tree);
 
   Internal.move(current, dirX, dirY, layout)
   |> Option.value(~default=current);
 };
 
-let moveLeft = current => moveCore(current, -1, 0);
-let moveRight = current => moveCore(current, 1, 0);
-let moveUp = current => moveCore(current, 0, -1);
-let moveDown = current => moveCore(current, 0, 1);
-
-let move = (direction: direction, current, v) => {
-  switch (direction) {
-  | Up => moveUp(current, v)
-  | Down => moveDown(current, v)
-  | Left => moveLeft(current, v)
-  | Right => moveRight(current, v)
-  };
-};
+let moveLeft = current => move(current, -1, 0);
+let moveRight = current => move(current, 1, 0);
+let moveUp = current => move(current, 0, -1);
+let moveDown = current => move(current, 0, 1);
 
 let rotateForward = (target, tree) => {
   let f =
--- a/src/Feature/Layout/Feature_Layout.rei
+++ b/src/Feature/Layout/Feature_Layout.rei
@@ -1,9 +1,3 @@
-type direction =
-  | Up
-  | Left
-  | Down
-  | Right;
-
 // definition only used for tests
 [@deriving show({with_path: false})]
 type size =
@@ -44,7 +38,7 @@ let removeWindow: ('id, t('id)) => t('id);
 
 let layout: (int, int, int, int, t('id)) => list(sizedWindow('id));
 
-let move: (direction, 'id, t('id)) => 'id;
+let move: ('id, int, int, t('id)) => 'id;
 let moveLeft: ('id, t('id)) => 'id;
 let moveRight: ('id, t('id)) => 'id;
 let moveUp: ('id, t('id)) => 'id;
--- a/src/Model/Actions.re
+++ b/src/Model/Actions.re
@@ -99,6 +99,13 @@ type t =
   | KeyboardInput(string)
   | WindowTitleSet(string)
   | EditorGroupSelected(int)
+  | WindowTreeSetSize(int, int)
+  | WindowMoveLeft
+  | WindowMoveRight
+  | WindowMoveUp
+  | WindowMoveDown
+  | WindowRotateForward
+  | WindowRotateBackward
   | EditorGroupSizeChanged({
       id: int,
       width: int,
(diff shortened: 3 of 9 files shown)


bryphe:
> This is blocked hard by a bug that likely involves som timing dependency that's really tricky to pin down (at least for me).

Hmm, thanks for the details! This must be the bug we discussed previously that was blocking. Glad we have a repro for it.

I suspect that the issue is a multiple-source-of-truth issue - that both the `WindowManager` has a concept of an `activeWindowId`: <https://github.com/onivim/oni2/blob/675a005e41c65b0f8a7facb39e62dd97f92b8188/src/Model/WindowManager.re#L27> _and_ the `EditorGroups` also has a concept of an active id: <https://github.com/onivim/oni2/blob/675a005e41c65b0f8a7facb39e62dd97f92b8188/src/Model/EditorGroups.re#L11>

I bet these are getting out-of-sync, somehow. And the best course of action is for us to move to a single source of truth, either:
- Remove `activeWindowId` from WindowManager, and pass it to the active split when we are creating a split (it needs to know what split the newly created split is relative to).
- Remove `activeEditorId` from EditorGroups - figure out the active editor based on the window split (the window splits already have a reference back to the editor group id: <https://github.com/onivim/oni2/blob/675a005e41c65b0f8a7facb39e62dd97f92b8188/src/Model/WindowTree.re#L19>). 

I think it's unlikely the Vim store is the root of this - the synchronization we do there is for these metrics:
- The window height
- The window wdith
- The top line
- The left colum
<https://github.com/onivim/oni2/blob/675a005e41c65b0f8a7facb39e62dd97f92b8188/src/Store/VimStoreConnector.re#L812>

`libvim` actually doesn't care about windows, aside from forwarding over the window-movement commands, which are handled completely by Onivim 2.

But I think if we can reconcile those multiple sources of truth - either to the window manager or editor groups - it'll make this a lot easier.

BTW, I tried to reproduce this with your steps and I'm not quite able to - here's a gif:

I'm not seeing the result on the last step - the focus moves to the new editor, as expected. But I imagine I'm doing something different.

glennsl:
I'm in the process of refactoring the `WindowManager` into a feature project in preparation for vim-style tabs, to get a better overview of the boundary and what I need to change to add another layer to it. And since moving the `WindowStore` is likely to trigger this issue as well, Having a single source of truth would definitely be beneficial with that as well, so I'll try to untangle that.

I also want to make it independent of the rendering by using ratios instead of absolute coordinates, which should make ti much simpler. Then I'd like to make it a generic container able to hold anything, not just editor groups, so we can extend it to include the sidebar, panel and whatever else we might want as well.

> I'm not seeing the result on the last step - the focus moves to the new editor, as expected. But I imagine I'm doing something different. Let me know if there is something I can do differently to reproduce it.

I'm not sure you're on the right branch. There should be no `EditorSize` module. But other than having to open a different file at the last step, I can recreate the problem using exactly the steps shown in the gif.

glennsl:
> libvim actually doesn't care about windows, aside from forwarding over the window-movement commands, which are handled completely by Onivim 2.

But it does care about the current buffer, and we synchronize that here:

https://github.com/onivim/oni2/blob/675a005e41c65b0f8a7facb39e62dd97f92b8188/src/Store/VimStoreConnector.re#L758-L770

and that can later trigger a `BufferEnter` here, which does set the active editor:

https://github.com/onivim/oni2/blob/675a005e41c65b0f8a7facb39e62dd97f92b8188/src/Store/VimStoreConnector.re#L341-L360

bryphe:
> I'm in the process of refactoring the WindowManager into a feature project in preparation for vim-style tabs, to get a better overview of the boundary and what I need to change to add another layer to it. And since moving the WindowStore is likely to trigger this issue as well, Having a single source of truth would definitely be beneficial with that as well, so I'll try to untangle that.

This sounds great! 

> I also want to make it independent of the rendering by using ratios instead of absolute coordinates, which should make ti much simpler.

Excellent. One thing we'll need to be able to handle is resizing - but we can always convert absolute coordinates (from mouse position) to ratio if we know the total size. Using ratios internally does seem like it'd be more natural - and it would easily handle the case where we resize the entire window.

> Then I'd like to make it a generic container able to hold anything, not just editor groups, so we can extend it to include the sidebar, panel and whatever else we might want as well.

😍 😍 

> I'm not sure you're on the right branch. There should be no EditorSize module.

Oops, you were right... I'm able to reproduce now. Still trying to track down. I did a brain-dump of some ideas I had around removing some of the synchronization in https://github.com/onivim/oni2/pull/1627 (between fixing the multiple sources of truth between windowmanager/editorgroups, and removing the synchronization - maybe we can simplify the problem!)

> But it does care about the current buffer, and we synchronize that here:

Ah, you're right - I would guess that is the source of this bug (getting a rogue `BufferEnter` that's causing up to pop open a new editor...)

glennsl:
> Remove activeWindowId from WindowManager, and pass it to the active split when we are creating a split (it needs to know what split the newly created split is relative to).

Unfortunately this didn't fix the problem. I still think this is a better approach though, so definitely keeping it. Crossing my fingers for #1627 being the fix instead.

bryphe:
I wanted to try and test this again with the latest `master` work to see if it is unblocked now - there's some tricky merges with both https://github.com/onivim/oni2/pull/1668 and the work that I did.

@glennsl - do you think you could merge this up with https://github.com/onivim/oni2/pull/1668, as you're familiar with the changes for `Feature_Layout`, and then I could merge up the remainder with `master`? We can test and see if we're unblocked now.

glennsl:
Seems to work! :tada:

glennsl:
Superseded by #1833

PR #1621, opened 2020-04-16
#1613 - Part 3 - Switch to luv/libuv for syntax highlight process

bryphe:
This is Part 3 of a round of fixes / improvements for the syntax server.

A contributing factor #1613 is the way we are managing the low-level read/writing between processes - when it fails, we end up spinning and eating CPU and RAM. It'd be better to let a library like `libuv` handle this - which has a more resilient and performant implementation of process interop.

Aside from the spin-on-crash issue in #1613, I've also seen intermittent failures around `EBADF` (bad file descriptor) - I'm concerned about the overall resiliency of our current process communication strategy, and have more confidence in `libuv`'s more battle-tested implementation.

This replaces the 'transport' to go from using the OCaml Unix modules pipe/channels to using libuv named pipes to communicate between the editor <-> syntax server.

Once we integrate 'v2' of the exthost - we'll be using libuv for all our cross-process communication.

__TODO:__
- [x] Bring back 'eager' highlighting
- [x] Clean up result handing in client
- [x] Clean up result handling in server
- [x] Remove debug logging
- [x] Get integration tests passing
- [x] Fix noisy error in reason-vscode-exthost
- [x] Windows: Ensure console window doesn't pop up for syntax HL service
- [x] Stability Testing
- [x] LInux: Check on CentOS
- [x] Performance Testing
   - [x] Investigate large-file performance 
     - [x] Implement delta updates (#1664, #1665)
   - [x] Investigate increased GC activity with switch 
   - [x] Reduce syntax traffic around buffer updates
- [x] Validate the transport packet parser handles split payloads
- [x] Add more involved syntax highlight test to skip eager highlighting

__Future work:__
- Finish #1613 fix (requires extension host change too) (test SIGINT, SIGSTP)


src/Store/Features.re
```diff
@@ -71,9 +71,21 @@ let update =
...
+    let effect =
+      switch (out) {
+      | Nothing => Effect.none
+      | ServerError(msg) =>
+        Feature_Notification.Effects.create(
+          ~kind=Error,
+          "Syntax Server error: " ++ msg,
+        )
+        |> Isolinear.Effect.map(msg => Actions.Notification(msg))
+      };
```
glennsl, line 87: We should have a helper function for this, internal to this module, as this is going be repeated many, many times here. Might as well add it now.

src/Core/Utility/FunEx.re
```diff
@@ -2,3 +2,11 @@ let tap = (f, x) => {
...
   x;
 };
+
+let rec repeat = (~count, f) =>
+  if (count <= 0) {
+    ();
+  } else {
+    f();
+    repeat(~count=count - 1, f);
+  };
```
glennsl, line 12:
```suggestion
  if (count > 0) {
    f();
    repeat(~count=count - 1, f);
  };
```

The else branch automatically evaluates to `unit` if omitted.

bryphe: Nice!

src/bin_editor/Oni2_editor.re
```diff
@@ -134,21 +134,14 @@ if (cliOptions.syntaxHighlightService) {
...
-
       // TODO: How many times should we run it?
       // The ideal amount would be just enough to do pending work,
       // but not too much to just spin. Unfortunately, it seems
       // Luv.Loop.run always returns [true] for us, so we don't
       // have a reliable way to know we're done (at the moment).
-      loop(100);
+      Oni_Core.Utility.FunEx.repeat(~count=100, () => {
+        ignore(Luv.Loop.run(~mode=`NOWAIT, ()): bool)
+      });
```
glennsl, line 144: Kind of wondering why you don't just use a for loop though:

```suggestion
      for (_ in 1 to 100) {
        ignore(Luv.Loop.run(~mode=`NOWAIT, ()): bool)
      };
```

It's shorter, and you've used it before.

integration_test/lib/Oni_IntegrationTestLib.re
```diff
@@ -191,7 +191,9 @@ let runTest =
       // Flush any queued calls from `Revery.App.runOnMainThread`
       Revery.App.flushPendingCallbacks();
       Revery.Tick.pump();
-      let _: bool = Luv.Loop.run(~mode=`NOWAIT, ());
+      Oni_Core.Utility.FunEx.repeat(~count=100, () => {
+        ignore(Luv.Loop.run(~mode=`NOWAIT, ()): bool)
+      });
```
glennsl, line 196:
```suggestion
      for (_ in 1 to 100) {
        ignore(Luv.Loop.run(~mode=`NOWAIT, ()): bool)
      };
```

integration_test/dune
```diff
@@ -36,7 +36,6 @@
            RegressionFileModifiedIndicationTest
            RegressionVspEmptyInitialBufferTest
            RegressionVspEmptyExistingBufferTest
-           Regression1097Test
```
bryphe, line 39: See if we can bring this back with the libuv poll fixes

integration_test/lib/Oni_IntegrationTestLib.re
```diff
@@ -99,6 +104,7 @@ let runTest =
       headlessWindow,
       <Oni_UI.Root state />,
     );
+    //Revery.Utility.HeadlessWindow.takeScreenshot(headlessWindow, "screenshot.png");
```
bryphe, line 107: Clean up

PR #1623, opened 2020-04-17
Always load Windows config from %LOCALAPPDATA%

CrossR:
This change means that the Windows config is always loaded from `%LOCALAPPDATA%/Oni2`, regardless of the shell used.

Previously, cmd and powershell would both load correctly but other shells such as git-bash, powershell-core and more would load from a different location, due to `HOME` being set (as far as I can tell).

EDIT: Oh I guess this is technically a breaking change...so we'll need to make sure this is advertised properly, as otherwise people who mainly used Oni2 from say git-bash, will be confused why their config gave up.

### DIFF
--- a/src/Core/Filesystem.re
+++ b/src/Core/Filesystem.re
@@ -213,31 +213,31 @@ let rmdir = path =>
     }
   );
 
-let unsafeFindHome = () =>
+let getUserDataDirectoryExn = () =>
   Revery.(
     switch (Environment.os) {
     | Environment.Windows => Sys.getenv("LOCALAPPDATA")
     | _ =>
       switch (Sys.getenv_opt("HOME")) {
       | Some(dir) => dir
-      | None => failwith("Could not find HOME dir")
+      | None => failwith("Could not find user data directory")
       }
     }
   );
 
-let getOniDirectory = home =>
+let getOniDirectory = dataDirectory =>
   Revery.(
     switch (Environment.os) {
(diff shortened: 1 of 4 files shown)


src/Core/Filesystem.re
```diff
@@ -215,10 +215,13 @@ let rmdir = path =>
 
 let unsafeFindHome = () =>
```
CrossR, line 216: Since I'm in the area, does anyone have a smarter name for this/think I should change this?

Its not really a home folder on Windows, its the config location. Its also only used in the `getOniDirectory`, so could be moved into there and given a smarter name (since its really used to get `XDG_CONFIG_DIR`), which would prevent us using it incorrectly in the future.

glennsl: What about `getUserDataDirectory`?

But also, `unsafe` is incorrect, as it typically means undefined behaviour. The convention for a function that raise exceptions is an `Exn` suffix (or `_exn` in OCaml).

CrossR:
@bryphe, from my PoV this can be merged now, but wanted to hold off since I wasn't sure if you wanted to do anything different with PRs that contain breaking changes.

If not I'll merge tomorrow, or you can merge.

bryphe:
Sorry I missed your message @CrossR ! Just saw it now.

I'm OK with this going on, we'll keep an eye out if anyone 'loses' their config on Windows. @glennsl is working on a 'changes' UI for the welcome screen - this might be our first one to show.

PR #1627, opened 2020-04-17
refactor(vim): remove synchronization effects in VimStoreConnector

bryphe:
Depends on https://github.com/onivim/reason-libvim/pull/115, which is the core of the change.

This was prompted by #1620 (and something @glennsl and I discussed a bit) - removing some of this awkward 'synchronization' logic that happens in the Vim layer, into something more explicit that is easier to reason about.

Instead of 'synchronizing' Vim when certain state changes (which is fragile - we hook up to a grab-bag of messages to figure out when to do this synchronization) - we treat our state as the source of truth, and _push_ it when we send input to Vim.

This is an incremental step that might help us get to a cleaner model for our editor - the next piece of work I'd see for reason-libvim is removing the listeners and using typed effects, like:

```
let input: (~context: Context.t, input: string) => (Context.t, list(Effect.t))
```

Anyway something we can discuss more next week @glennsl !

__TODO:__
- ~~Move `BufferEnter` to an effect in `reason-libvim`~~ 
- ~~Move `BufferUpdate` to an effect in `reason-libvim`~~
- [x] Stability test
- [x] Fix syntax highlighting visible ranges (#1745)


src/Store/VimStoreConnector.re
```diff
@@ -1014,24 +973,6 @@ let start =
...
-    | ViewSetActiveEditor(_) => (state, synchronizeEditorEffect(state))
-    | ViewCloseEditor(_) => (state, synchronizeEditorEffect(state))
-    | Command("workbench.action.nextEditor") => (
-        state,
-        synchronizeEditorEffect(state),
-      )
-    | Command("workbench.action.previousEditor") => (
-        state,
-        synchronizeEditorEffect(state),
-      )
```
bryphe, line 1034: Nice to have this 'synchronization' go away!

src/Store/VimStoreConnector.re
```diff
@@ -847,22 +804,23 @@ let start =
...
                s => {
-                 latestCursors :=
-                   Vim.input(~cursors=[], Zed_utf8.singleton(s));
-                 ();
+                 let (context, _eff) =
+                   Vim.input(~context=latestContext^, Zed_utf8.singleton(s));
+                 latestContext := context;
                },
                text,
              );
```
glennsl, line 818:
```suggestion
             Zed_utf8.fold(
               (s, context) => Vim.input(~context, Zed_utf8.singleton(s)) |> fst,
               contextFromState(state),
               text,
             );
```

src/Store/VimStoreConnector.re
```diff
@@ -725,120 +774,28 @@ let start =
...
+            let (context, _eff) =
+              Vim.input(~context=latestContext^, String.make(1, c));
+            latestContext := context;
             ();
           },
           completion,
         );
-        updateActiveEditorCursors(latestCursors^);
+        let cursors = latestContext^.cursors;
+        updateActiveEditorCursors(cursors);
```
glennsl, line 793:
```suggestion
        updateActiveEditorCursors(latestContext^.cursors);
```

src/Store/VimStoreConnector.re
```diff
@@ -725,120 +774,28 @@ let start =
...
         String.iter(
           c => {
-            latestCursors := Vim.input(~cursors=[], String.make(1, c));
+            let (context, _eff) =
+              Vim.input(~context=latestContext^, String.make(1, c));
+            latestContext := context;
             ();
           },
           completion,
         );
```
glennsl, line 791:
```suggestion
        String.fold(
          (c, context) => Vim.input(~context, String.make(1, c)) |> fst,
          Vim.Context.default(),
          completion,
        );
```

src/Store/VimStoreConnector.re
```diff
@@ -725,120 +774,28 @@ let start =
         isCompleting := true;
         let currentPos = ref(Vim.CommandLine.getPosition());
         while (currentPos^ > position) {
-          let _ = Vim.input(~cursors=[], "<bs>");
+          let _ = Vim.input("<bs>");
```
glennsl, line 777: Accidental partial application in waiting

src/Store/VimStoreConnector.re
```diff
@@ -539,73 +649,12 @@ let start =
...
-          };
+        let context = contextFromState(state);
 
-        let cursors = Vim.input(~autoClosingPairs?, ~cursors, key);
-        let newTopLine = Vim.Window.getTopLine();
-        let newLeftColumn = Vim.Window.getLeftColumn();
+        let ({cursors, topLine, leftColumn, _}: Vim.Context.t, _effects) =
+          Vim.input(~context, key);
+        let newTopLine = topLine;
+        let newLeftColumn = leftColumn;
```
glennsl, line 657:
```suggestion
        let ({cursors, topLine: newTopLine, leftColumn: newLeftColumn, _}: Vim.Context.t, _effects) =
          Vim.input(~context, key);
```

integration_test/EchoNotificationTest.re
```diff
@@ -8,7 +8,7 @@ runTest(~name="InputIgnore test", (_dispatch, wait, runEffects) => {
   );
 
   // Create notification - an echo should trigger onj
-  Vim.command("echo 'hi from test'");
+  Vim.command("echo 'hi from test'") |> ignore;
```
glennsl, line 11: These are all accidental partial application bugs waiting to happen.

```suggestion
  ignore(Vim.command("echo 'hi from test'"): Vim.Context.t);
```

integration_test/QuickOpenEventuallyCompletesTest.re
```diff
@@ -9,11 +9,9 @@ runTest(~name="QuickOpen eventually completes", (dispatch, wait, runEffects) =>
   /* Switch to root directory */
   if (Revery.Environment.os == Revery.Environment.Mac) {
     // CI machines timeout with '/' - so we'll use home (which also reproduces the crash)
-    Vim.command(
-      "cd ~",
-    );
+    Vim.command("cd ~") |> ignore;
   } else {
-    Vim.command("cd /");
+    Vim.command("cd /") |> ignore;
```
glennsl, line 14: You can move these `ignore`s outside the `if`, to avoid the repetition.

src/Store/VimStoreConnector.re
```diff
@@ -425,6 +425,118 @@ let start =
...
+               topLine,
+               width,
+               height,
+               cursors,
+               autoClosingPairs,
+               lineComment,
+               insertSpaces,
+               tabSize: indentation.size,
+             };
+           ret;
```
glennsl, line 535:
```suggestion
             Vim.Context.{
               bufferId,
               leftColumn,
               topLine,
               width,
               height,
               cursors,
               autoClosingPairs,
               lineComment,
               insertSpaces,
               tabSize: indentation.size,
             };
```

src/Store/VimStoreConnector.re
```diff
@@ -425,6 +425,118 @@ let start =
...
 
+  let contextFromState: State.t => Vim.Context.t =
+    state => {
+      state
+      |> Selectors.getActiveEditorGroup
+      |> Selectors.getActiveEditor
+      |> Option.map((editor: Editor.t) => {
+           let {bufferId, cursors, _}: Editor.t = editor;
+           let primaryCursor = Editor.getPrimaryCursor(editor);
+
```
glennsl, line 436: This looks so much better! It's much easier (assuming no bugs on the libvim side of things) to find out exactly what state libvim is in for an operation, and to see exactly what's being "synced" with this.

One possible improvement could be to make it even more granular, like the context keys, where every "piece" is calculated in isolation based on the state. This function is already over one hundred lines long, and so to fully understand a calculation at the bottom, you really have to take every one of those 100+ lines into account to figure out what each component of the calculation actually is.

Also, having a 100+ line anonymous function in a pipe chain isn't great, for much the same reason.

Still a massive step in the right direction though!

integration_test/EchoNotificationTest.re
bryphe: Good point!

I started factoring these to be an action:

```
dispatch(Actions.VimExecuteCommand("echo 'hi from test'"))
```

It's not an ideal action, because it's very function-call-ish. But I realized we would need an effect handler to handle this, because things like `Vim.command("echo 'hi from test'")` will have effects that we'll need to map and react to - so it made sense to centralize that logic.

src/Store/VimStoreConnector.re
bryphe: I used `VimEx.inputString` since it handles UTF8 correctly, too

src/Store/VimStoreConnector.re
bryphe: Used `VimEx.inputString` for this too

src/Store/VimStoreConnector.re
bryphe: FYI, I pulled apart this method and split it up in a `VimContext` module in `Oni_Model`

glennsl: Awesome!

> I pulled apart this method

Very nitpicky, but pet peeve and such, you know... It's a function, not a method. "method" is an OOP-term for a function that's attached to an object.  Outside of OOP there are no methods, just functions. Or procedures, if you're very old and imperatively inclined ;)

bryphe: Sorry, leaked some OOP in, thanks for the correction ;)

bryphe:
The tree-sitter test is failing here because we're not sending the visible ranges correctly - so it's a good time to revisit the way we send that data to the syntax server. I'll implement that in a separate change, though.

bryphe:
I discovered one failure while testing this locally:

1) Open NodeTask.re
2) Close NodeTask.re with mouse
3) Open NodeTask.re again
4) Open EditorSurface.re
5) Close NodeTask.re with mouse
6) Open NodeTask.re

__Expected:__ `NodeTask.re` should re-open
__Actual:__ It doesn't. There's a dependency on getting a `BufferEnter` event to populate the view, so we need to break that dependency

bryphe:
Seems like the failure is fixed after merging in #1772 !

Issue #1423, opened 2020-03-05
Keybinding Suggestions and Contribution Opportunities


glennsl:
This is the place to be if you want to suggest a keybinding, or find a nice and (usually) easy task to start contributing.

### Suggestions

Name | Command | Default | Assigned | PR
-----|----|---|---|---
Save | workbench.action.files.save | <kbd>Mod</kbd>+<kbd>S</kbd> | ?
Focus Tab # | ? | <kbd>Mod</kbd>+<kbd>1</kbd> - <kbd>Mod</kbd>+<kbd>9</kbd> | 
Copy | input.copy? | <kbd>Mod</kbd>+<kbd>C</kbd> | @fanantoxa | #1233
Cut | input.cut? | <kbd>Mod</kbd>+<kbd>X</kbd> | @fanantoxa | #1233
Paste | input.paste? | <kbd>Mod</kbd>+<kbd>V</kbd> | @fanantoxa | #1233
Select All | input.selectAll? | <kbd>Mod</kbd>+<kbd>A</kbd> | @fanantoxa | #1233

Do you have other suggestions? Post a comment below! If the keybinding exists in vscode as well, it would be nice if you could post the name, command and default keybinding used there.

### How to contribute

Keybidnings are defined in `KeyBindingsStoreConnector`. Here's the keybinding for `unfo`, for example:

https://github.com/onivim/oni2/blob/bb2df4cd5d88a4f7e643e1f6b849c0b50ad87874/src/Store/KeyBindingsStoreConnector.re#L162-L166

When invoked this will generate a `Command("undo")` action, which needs to be handled in a Store somewhere. The `undo` comamnd is handled in `VimStoreConnector` here:

https://github.com/onivim/oni2/blob/bb2df4cd5d88a4f7e643e1f6b849c0b50ad87874/src/Store/VimStoreConnector.re#L855

Which invokes the effect defined a bit further up, here:

https://github.com/onivim/oni2/blob/bb2df4cd5d88a4f7e643e1f6b849c0b50ad87874/src/Store/VimStoreConnector.re#L827-L834

A lot of keybindings will be like this, just delegating to libvim to handle it. Some might require a bit more though, but a good place to start is to look into how this is done and vim, and whether it can be delegated in the same manner as this.

If you feel ready to tackle one of the suggestions above, post a comment below to lets us know, then dive in! If you have questions, the best place to ask is in the #dev channel on [our Discord server](https://discord.gg/7maEAxV).

[...]

PR #1629, opened 2020-04-18
Add keybindings for zoom in, zoom out and zoom reset

ArtemDrozdov:
Added keybindings for zoom in, zoom out and zoom reset according #1423 issue.

I checked zooming behavior in VSCode and they have next logic: default value is 0, minimum is -8 and maximum is 9. Zooming step is 1 which equals to 20%.

Short summary what was done in this PR:
1. Removed minimum value of 1.0 for zooming in ConfigurationStoreConnector;
2. Added keybindings to KeyBindingsStoreConnector for <kbd>Cmd</kbd> + <kbd>=</kbd>, <kbd>Ctrl</kbd> + <kbd>=</kbd>, <kbd>Cmd</kbd> + <kbd>-</kbd>, <kbd>Ctrl</kbd> + <kbd>-</kbd>, <kbd>Cmd</kbd> + <kbd>0</kbd>  and <kbd>Ctrl</kbd> + <kbd>0</kbd>;
3. Added Commands and zoom effect handler to VimStoreConnector.

Minimum value for zooming is 0 because on lower values there is no changes in UI.

And probably there should be some logic to process <kbd>Cmd</kbd> + <kbd>=</kbd> on Mac and <kbd>Ctrl</kbd> + <kbd>=</kbd> on Linux/Windows for example. Because with current implementation both of those combinations are working on Mac.

Would love to hear any comments and suggestion!

### DIFF
--- a/src/Store/CommandStoreConnector.re
+++ b/src/Store/CommandStoreConnector.re
@@ -443,7 +443,7 @@ let start = (getState, contributedCommands) => {
     ),
     (
       "workbench.action.zoomReset",
-      state => zoomEffect(state, zoom => Constants.defaultZoomValue),
+      state => zoomEffect(state, _zoom => Constants.defaultZoomValue),
     ),
   ];
 
--- a/src/Store/CommandStoreConnector.re
+++ b/src/Store/CommandStoreConnector.re
@@ -5,6 +5,13 @@ open Oni_Model.Actions;
 
 module KeyDisplayer = Oni_Components.KeyDisplayer;
 
+module Constants = {
+  let zoomStep = 0.2;
+  let defaultZoomValue = 1.0;
+  let minZoomValue = 0.0;
+  let maxZoomValue = 2.8;
+};
+
 let pathSymlinkEnabled = (~addingLink) =>
   (
     Revery.Environment.os == Revery.Environment.Mac
@@ -265,6 +272,33 @@ let start = (getState, contributedCommands) => {
       },
     );
 
+  let zoomEffect = (state: State.t, getCalculatedZoomValue, _) =>
+    Isolinear.Effect.createWithDispatch(~name="window.zoom", dispatch => {
+      let configuration = state.configuration;
+      let currentZoomValue =
+        Configuration.getValue(c => c.uiZoom, configuration);
+
+      let calculatedZoomValue = getCalculatedZoomValue(currentZoomValue);
+      let newZoomValue =
+        Utility.IntEx.clamp(
+          calculatedZoomValue,
+          ~hi=Constants.maxZoomValue,
+          ~lo=Constants.minZoomValue,
+        );
+
+      if (newZoomValue != currentZoomValue) {
+        dispatch(
+          ConfigurationSet({
+            ...configuration,
+            default: {
+              ...configuration.default,
+              uiZoom: newZoomValue,
+            },
+          }),
+        );
+      };
+    });
+
   let commands = [
     ("keyDisplayer.enable", _ => singleActionEffect(EnableKeyDisplayer)),
     ("keyDisplayer.disable", _ => singleActionEffect(DisableKeyDisplayer)),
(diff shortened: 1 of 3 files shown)


src/Store/VimStoreConnector.re
```diff
@@ -975,6 +980,42 @@ let start =
     });
   };
 
+  let zoomStep = 0.2;
+  let defaultZoom = 1.0;
+  let minZoomValue = 0.0;
+  let maxZoomValue = 2.8;
```
glennsl, line 986: Could you put these in a `Constants` sub-module at the top of the file?

src/Store/VimStoreConnector.re
```diff
@@ -975,6 +980,42 @@ let start =
...
+      let configuration = state.configuration;
+      let currentZoomValue =
+        Core.Configuration.getValue(c => c.uiZoom, configuration);
+
+      let calculatedZoomValue =
+        switch (zoomAction) {
+        | ZoomIn => currentZoomValue +. zoomStep
+        | ZoomOut => currentZoomValue -. zoomStep
+        | ZoomReset => defaultZoom
+        };
```
glennsl, line 998: What's the reason you pass in a `zoomAction` instead of just passing in the value directly?

src/Store/VimStoreConnector.re
```diff
@@ -975,6 +980,42 @@ let start =
...
+
+      let newZoomValue =
+        calculatedZoomValue < minZoomValue
+          ? minZoomValue
+          : calculatedZoomValue > maxZoomValue
+              ? maxZoomValue : calculatedZoomValue;
```
glennsl, line 1004: You can use `Utility.IntEx.clamp` for this. It would be significantly more readable I think.

src/Store/VimStoreConnector.re
```diff
@@ -1032,6 +1073,18 @@ let start =
...
+        zoomEffect(state, ZoomIn),
+      )
+    | Command("workbench.action.zoomOut") => (
+        state,
+        zoomEffect(state, ZoomOut),
+      )
+    | Command("workbench.action.zoomReset") => (
+        state,
+        zoomEffect(state, ZoomReset),
+      )
```
glennsl, line 1087: I don't see any reason for these and the effect to need to live in `VimStoreConnector`. Could you move it to `CommandStoreConnector` instead (unless there's a reason I don't see)?

src/Store/VimStoreConnector.re
ArtemDrozdov: Yes, of course.

src/Store/VimStoreConnector.re
ArtemDrozdov: What do you mean by 
> instead of just passing in the value directly

Can't get your point.

glennsl: Looking closer, I guess you can't pass in `currentZoomValue +. zoomStep)` directly instead of `ZoomIn` because you need `currentZoomValue`. But you could do `zoomEffect(state, zoom => zoom +. zoomStep)`.

ArtemDrozdov:
Pushed changes according @glennsl comments:

1. Moved effect logic from `VimStoreConnector` to `CommandStoreConnector`;
2. Added `Constants` submodule in `CommandStoreConnector`;
3. Changed effect logic a little bit.

glennsl:
Oh, CI fails on an unused variable warning. That should be a quick fix though.

src/Store/CommandStoreConnector.re
```diff
@@ -399,6 +433,18 @@ let start = (getState, contributedCommands) => {
...
+      "workbench.action.zoomIn",
+      state => zoomEffect(state, zoom => zoom +. Constants.zoomStep),
+    ),
+    (
+      "workbench.action.zoomOut",
+      state => zoomEffect(state, zoom => zoom -. Constants.zoomStep),
+    ),
+    (
+      "workbench.action.zoomReset",
+      state => zoomEffect(state, zoom => Constants.defaultZoomValue),
```
glennsl, line 446:
```suggestion
      state => zoomEffect(state, _zoom => Constants.defaultZoomValue),
```

PR #1634, opened 2020-04-19
[docs] update keybindings with new zoom additions

glennsl:
Added in https://github.com/onivim/oni2/pull/1629


glennsl:
Guess we don't need to wait for CI just for docs...

PR #1635, opened 2020-04-19
Refactor: Use more (correct) theme colors

glennsl:
This adds some new theme colors in place of hard-coded colors or incorrect theme colors being used, as well as correcting some uses of existing theme colors.


src/Components/Input.re
```diff
@@ -75,27 +76,40 @@ module Constants = {
...
+      |> Option.value(~default=Colors.transparentBlack);
+  };
 
-  let default =
+  let foreground = color("input.foreground");
+  let background = color("input.background");
+  let border = color("input.border");
+  let placeholderForeground = color("input.placeholderForeground");
+  let selection = color("selection.background");
+};
```
glennsl, line 93: Am example of a non-app-specific component using theme colors. This wasn't possible before because colors were hard-coded into an app-specific type, and so had to be passed in one-by one. This makes the input component much nicer to use. Of course there's still plenty of that around with Revery components. Perhaps we should move the `ColorTheme` infrastructure to Revery though.

src/Feature/Theme/Feature_Theme.re
```diff
@@ -30,11 +30,20 @@ let defaults =
...
+    Colors.MinimapSlider.defaults,
+    Colors.MinimapGutter.defaults,
     Colors.Oni.defaults,
+    Colors.Oni.Modal.defaults,
     Colors.Oni.Sneak.defaults,
+    Colors.Panel.defaults,
+    Colors.PanelTitle.defaults,
+    Colors.PanelInput.defaults,
     Colors.ScrollbarSlider.defaults,
+    Colors.Selection.defaults,
```
glennsl, line 46: The new color sections that have been added.

PR #1646, opened 2020-04-20
Refactor: Feature_Layout - Part 1: WindowManager, WindowTree and WindowTreeLayout

glennsl:
This is the first part of a refactor of the WindowTree functionality into a separate Feature_Layout project, intended to lay the groundwork for several coming features: Generalized keyboard window navigation (not just for editor groups), window resizing and vim-style tabs. This PR moves, merges and simplifies WindowManager, WIndowTree and WindowTreeLayout into the top-level Feature_Layout module. It also adds the following features over the previous implementation:

- Generalizes it to work with any kind of window content.
- Removes `activeWindowId` to keep `EditorGroup.activeId` as the single source of truth
- removes absolute sizing in favor of a weight-based approach which simplifies the interaction between the UI and the underlying model.

Unfortunately the remaining work is blocked by a nasty and mysterious timing dependency that is being partially addressed by https://github.com/onivim/reason-libvim/pull/115. Fingers crossed that that will pan out.

Once that's been solved, the next steps are:
- [ ] Replace `WindowsStore` wqith an update function in `Feature_Layout`.
- [ ] Move layout views to `Feature_Layout`.
- [ ] Move window movement logic from `VimStore` and CommandStore` into `Feature_Layout`.
- [ ] Implement "fixed" size windows, that remains the same size as its container is resized.
- [ ] Implement "locked" windows that will remain in place no matter what.
- [ ] Implements tests using exposed APIs so internals can be hidden.


src/Feature/Layout/Feature_Layout.re
```diff
@@ -0,0 +1,284 @@
...
+  | Down
+  | Right;
+
+type t('content) =
+  | Split([ | `Horizontal | `Vertical], list(t('content)))
+  | Window({
+      weight: float,
+      content: 'content,
+    })
+  | Empty;
```
glennsl, line 15: The window is now called a window instead of a split, and the split is called a split instead of "parent". Otherwise this is pretty much the same tree structure as before.

src/Feature/Layout/Feature_Layout.re
```diff
@@ -0,0 +1,284 @@
...
+
+[@deriving show({with_path: false})]
+type window('content) = {
+  content: 'content,
+  x: int,
+  y: int,
+  width: int,
+  height: int,
+};
```
glennsl, line 24: This name isn't great, as it's specifically a window as a product of `layout`, so more like `laidOutWindow`? But preferably using a less awkward term.

src/Feature/Layout/Feature_Layout.rei
```diff
@@ -0,0 +1,53 @@
...
+};
+
+let initial: t('content);
+
+let windows: t('content) => list('content);
+let addWindow:
+  (
+    ~target: option('content)=?,
+    ~position: [ | `Before | `After],
+    [ | `Horizontal | `Vertical],
```
glennsl, line 37: I wanted to unify all these arguments into one that accurately models the problem instead of ignoring parts of the input in certain conditions. It could be a single argument with ``[`Above('content) | `Below('content) | `LeftOf('content) | 'RIghtOf(content) | `Leftmost | `Rightmost]`` but the way it's currently used makes that a bit awkward in some places.

src/Store/CommandStoreConnector.re
```diff
@@ -230,29 +230,24 @@ let start = (getState, contributedCommands) => {
...
-        WindowTree.createSplit(
-          ~editorGroupId=newEditorGroup.editorGroupId,
-          (),
-        );
+      dispatch(AddSplit(direction, newEditorGroup.editorGroupId));
 
-      dispatch(AddSplit(direction, split));
+      // This needs to be dispatched after the split, since this will set the
+      // active editor group, which is then used as the target for the split.
+      dispatch(EditorGroupAdd(newEditorGroup));
```
glennsl, line 237: Another timing dependency caused by function-actions. This one was easily solved though, but the fix had to be applied in two places.

src/Store/VimStoreConnector.re
```diff
@@ -625,12 +626,12 @@ let start =
...
-        dispatch(Actions.EditorGroupAdd(eg));
 
-        let split =
-          WindowTree.createSplit(~editorGroupId=eg.editorGroupId, ());
+        dispatch(Actions.AddSplit(direction, eg.editorGroupId));
 
-        dispatch(Actions.AddSplit(direction, split));
+        // This needs to be dispatched after the split, since this will set the
+        // active editor group, which is then used as the target for the split.
+        dispatch(Actions.EditorGroupAdd(eg));
```
glennsl, line 634: The other place

src/Store/WindowsStoreConnector.re
```diff
@@ -133,8 +107,7 @@ let start = () => {
       | Init => initializeDefaultViewEffect(state)
       // When opening a file, ensure that the active editor is getting focus
       | ViewCloseEditor(_) =>
-        if (List.length(WindowTree.getSplits(state.windowManager.windowTree))
-            == 0) {
+        if (List.length(Feature_Layout.windows(state.layout)) == 0) {
```
glennsl, line 110:
```suggestion
        if (Feature_Layout.windows(state.layout) == []) {
```

src/UI/EditorLayoutView.re
```diff
@@ -45,40 +35,30 @@ let renderTree = (state, theme, tree) => {
...
-
-  let children = renderTree(state, theme, windowTree);
+let%component make = (~state: State.t, ~theme, ()) => {
+  let%hook (maybeDimensions, setDimensions) = Hooks.state(None);
+  let children =
+    switch (maybeDimensions) {
+    | Some((width, height)) =>
+      renderTree(~width, ~height, state, theme, state.layout)
+    | None => React.empty
+    };
```
glennsl, line 55: Instead of sending the dimensions all the way to the `WindowManager` state and back, it's now contained here with a state hook.

integration_test/dune
```diff
@@ -34,6 +34,7 @@
            RegressionCommandLineNoCompletionTest
            RegressionDeleteAllLinesTest
            RegressionFileModifiedIndicationTest
+           RegressionNonExistentDirectory
```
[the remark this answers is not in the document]
bryphe: Nice catch!

src/Feature/Layout/Feature_Layout.re
bryphe: Sounds good 👍

src/Feature/Layout/Feature_Layout.re
bryphe: Maybe `measuredWindow` or `sizedWindow`?

src/Feature/Layout/Feature_Layout.re
```diff
@@ -0,0 +1,284 @@
...
+  | Up
+  | Left
+  | Down
+  | Right;
+
+type t('content) =
+  | Split([ | `Horizontal | `Vertical], list(t('content)))
+  | Window({
+      weight: float,
+      content: 'content,
```
bryphe, line 13: Great to see this beginning generic now - thanks!

src/Feature/Layout/Feature_Layout.re
```diff
@@ -0,0 +1,284 @@
...
+          let newMinX = cur.x < minX ? cur.x : minX;
+          let newMinY = cur.y < minY ? cur.y : minY;
+          let newMaxX = cur.x + cur.width > maxX ? cur.x + cur.width : maxX;
+          let newMaxY = cur.y + cur.height > maxY ? cur.y + cur.height : maxY;
+          let newDeltaX = cur.width / 2 < deltaX ? cur.width / 2 : deltaX;
+          let newDeltaY = cur.height / 2 < deltaY ? cur.height / 2 : deltaY;
+
+          (newMinX, newMinY, newMaxX, newMaxY, newDeltaX, newDeltaY);
+        },
+        (0, 0, 1, 1, 100, 100),
```
bryphe, line 51: Where do these `100` values come from?

src/Feature/Layout/Feature_Layout.re
```diff
@@ -0,0 +1,284 @@
...
+    let findSplit = children => {
+      let predicate =
+        fun
+        | Window({content, _}) => content == target
+        | _ => false;
+
+      List.exists(predicate, children);
+    };
+
+    switch (currenTree) {
```
bryphe, line 104: Minor typo: `currenTree` -> `currentTree`

src/Feature/Layout/Feature_Layout.re
```diff
@@ -0,0 +1,284 @@
+open Utility;
+
+type direction =
+  | Up
+  | Left
+  | Down
+  | Right;
+
+type t('content) =
+  | Split([ | `Horizontal | `Vertical], list(t('content)))
```
bryphe, line 10: I wonder if we will need to add a `weight` to the Split as well? For example:

```
| A | B |
--------
| C | D |
```
(ie, `Split(Horizontal, [`Vertical([splitA, splitB]), `Vertical([splitC, splitD])])`)

- if we wanted to resize the _height_ of A/B or C/D - it seems like that would need to live on the `Split`. Just something to think about as we get into resizing

src/Feature/Layout/Feature_Layout.re
bryphe: It seems like these are the delta - what if I'm moving from a very large window (ie, `800`) past a smaller window (ie, `50`) - it seems like this `move` could overshoot small windows. Might've been a mistake I made in the logic

src/UI/EditorLayoutView.re
bryphe: Cool!

src/Store/CommandStoreConnector.re
bryphe: Interesting. It'd be nice to remove this - I would imagine we could push one or both out of the effects and directly into the state update. Some untangling though needed.

src/Store/CommandStoreConnector.re
```diff
@@ -43,29 +43,24 @@ let start = () => {
...
-      let split =
-        WindowTree.createSplit(
-          ~editorGroupId=newEditorGroup.editorGroupId,
-          (),
-        );
+      dispatch(AddSplit(direction, newEditorGroup.editorGroupId));
 
-      dispatch(AddSplit(direction, split));
+      // This needs to be dispatched after the split, since this will set the
+      // active editor group, which is then used as the target for the split.
```
bryphe, line 49: Thanks for adding the comment!

src/Feature/Layout/Feature_Layout.re
glennsl: I like `sizedWindow`, so going with that. Thanks!

src/Store/CommandStoreConnector.re
glennsl: Yeah, something like that. I wanted to try to untangle it when replacing `WindowsStore` with `Feature_Layout.update`, but that's of course blocked by the other timing bug.

src/Feature/Layout/Feature_Layout.re
glennsl: I haven't touched this part of the code, though I really, really wanted to because it's so imperative. But it seems to work fine, so "don't fix what ain't broke" and such. I'll probably do a pass over this when implementing the resizing though, because it'll need to account for the weights and locked windows and such then.

src/Feature/Layout/Feature_Layout.re
glennsl: Good catch!

src/Feature/Layout/Feature_Layout.re
glennsl: Yeah, that's a good point. The weight should probably be associated with the "child", not the window, so the list of children would be more like `list((float, t('content)))`. I'll leave it as-is as it's not actually used yet, and will fix it when I do the resizing shortly.

PR #1654, opened 2020-04-22
Bugfix: Hover window doesn't fit diagnostic messages

### LINKED ISSUE #1183: Intellisense Hover Popup Distorted [Linux/Gnome]
I had to do the below (resize window) just to make sure it wasn't a glitch with the text coming through:
image

Here's the initial way I saw it (app window 1/2 screen):
image

```console
(✓) 13:39:09 [~/code/gb/api::master]
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Pop!_OS 18.04 LTS"
(✓) 13:39:16 [~/code/gb/api::master]
$ echo $XDG_SESSION_TYPE
wayland
(✓) 13:40:00 [~/code/gb/api::master]
$ uname -a
Linux p5520 5.3.0-7625-generic #27~1576774585~18.04~c7868f8-Ubuntu SMP Thu Dec 19 20:37:47 UTC  x86_64 x86_64 x86_64 GNU/Linux
```

glennsl:
**Issue:** When diagnostic messages spanned multiple lines, the hover window wouldn't fit the message. It wasn't tall enough, and compensated for this by being absurdly long.

**Defect:** Line breaks weren't taken into account when measuring the height and width of a piece of text. Instead it was just assumed that it would be rendered as a single line, with the height being the height of the font and the width being the number of characters multiplied by the font width.

**Fix:** Messages are now split into lines, which are measured individually and added up.

Additionally, this adds removal of extraneous indentation, i.e. indentation that occurs on every line of the message.

Fixes #1183

### DIFF
--- a/src/Core/Utility/StringEx.re
+++ b/src/Core/Utility/StringEx.re
@@ -66,6 +66,19 @@ let trimRight = str => {
   aux(length - 1);
 };
 
+let indentation = str => {
+  let rec loop = i =>
+    if (i >= String.length(str)) {
+      i;
+    } else if (isSpace(str.[i])) {
+      loop(i + 1);
+    } else {
+      i;
+    };
+
+  loop(0);
+};
+
 let extractSnippet = (~maxLength, ~charStart, ~charEnd, text) => {
   let originalLength = String.length(text);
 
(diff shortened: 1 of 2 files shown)


bryphe:
Such a huge improvement!

image

Thank you @glennsl !

PR #1655, opened 2020-04-22
Syntax Highlighting - Perf: Fix flashing character at end of line

bryphe:
While investigating performance for #1621 - I saw some issues around the way we send the colorized tokens from the textmate syntax highlighting service up to the editor.

One issue is that we are sending up `ColorizedToken.t` for pure whitespace tokens - this isn't actually useful, and in fact, causes problems at the end of the line - it causes the trailing character to temporarily be colored as whitespace instead of being colorized the same as the previous token. This can cause a jarring 'flash' when a new character is entered, as by default, it should match the previous token until syntax highlighting is completed. However, with the whitespace tokens being sent up - it temporarily has the color of the whitespace / new line.

Another way to fix this - which would be desirable - is to have more granularity in terms of buffer updates (which are currently only line-based). Knowing a range within the line of the update can also help us handle this, by shifting the tokens based on the delta range.

In addition, there were a couple smaller issues fixed:
- Don't allocate in `StringEx.startWith` - just check if the characters `0` through `prefixLength` match.
- Cache hex-parsing results, since this happens a lot over the course of syntax highlighting (and there tends not to be much variety).
- Use `String.concat(stringList, " ")` as opposed to `fold_left` and `++`, because it is faster.

__Future work:__
- Implement minimized delta updates, in addition to the work here.


src/Core/Utility/StringEx.re
```diff
@@ -21,14 +21,27 @@ let contains = (query, str) => {
...
+        for (idx in 0 to prefixLength - 1) {
+          if (prefix.[idx] != str.[idx]) {
+            raise(NoMatchException);
+          };
+        };
+        true;
+      }
+    ) {
+    | NoMatchException => false
+    };
```
glennsl, line 44: Slightly shorter:

```suggestion
    let rec match = i =>
      i == prefixLength || (prefix.[i] == str.[i]) && match(i + 1));
    match(0);
```

Or if that's a bit too cryptic:

```suggestion
    let rec match = i =>
      if (i == prefixLength) {
        true;
      } else {
        prefix.[i] == str.[i] && match(i + 1);
      };
    match(0);
```

src/Syntax/TextmateTokenizerJob.re
```diff
@@ -100,6 +101,20 @@ let onBufferUpdate = (bufferUpdate: BufferUpdate.t, lines, v: t) => {
   Job.map(f, v);
 };
 
+let _hexToColor = Hashtbl.create(128);
```
glennsl, line 104: A leading underscore is still not a good idea as it has a different compiler-enforced meaning than you want it to have. I'd suggest using an `Internal` module instead, or if it's just used for a single function, scope it to that. This could even be generalized into a `cached` utility function I think:

```
let cached = (size, f) => {
  let cache = Hashtbl.create(size);

  input =>
    switch (Hashtbl.find_opt(cache, input)) {
    | Some(output) => output
    | None =>
      let output = f(input);
      Hashtbl.add(cache, input, output);
      output;
    };
};
```

src/Syntax/TextmateTokenizerJob.re
```diff
@@ -100,6 +101,20 @@ let onBufferUpdate = (bufferUpdate: BufferUpdate.t, lines, v: t) => {
...
+  switch (Hashtbl.find_opt(_hexToColor, hex)) {
+  | None =>
+    let color = Revery.Color.hex(hex);
+    Hashtbl.add(_hexToColor, hex, color);
+    color;
+  | Some(color) => color
+  };
+};
+
+exception NoWhitespaceException;
```
glennsl, line 116: You can define local exceptions inside functions now (since 4.08 maybe?)

src/Syntax/TextmateTokenizerJob.re
```diff
@@ -115,38 +130,52 @@ let doWork = (pending: pendingWork, completed: completedWork) => {
...
+          for (idx in startIndex to endIndex - 1) {
+            if (!StringEx.isSpace(line.[idx])) {
+              raise(NoWhitespaceException);
+            };
+          };
+          true;
+        }
+      ) {
+      | NoWhitespaceException => false
+      };
```
glennsl, line 157:
```suggestion
    let rec isWhitespaceOnly = (startIndex, endIndex) =>
      if (startIndex >= endIndex) {
        true;
      } else {
        !StringEx.isSpace(line.[startIndex] && isWhitespaceOnly(startIndex + 1, endIndex);
      };
```

There's a pattern here. Opportunity for `StringEx.forAll` with an optional start and end index?

src/Syntax/TextmateTokenizerJob.re
bryphe: Oh, cool! Didn't know that. I factored this out to use `StringEx.forAll` instead, but good to know for the future.

src/Syntax/TextmateTokenizerJob.re
bryphe: Added a `StringEx.forAll` to help with this

src/Core/Utility/StringEx.re
```diff
@@ -23,6 +23,36 @@ let explode = str =>
...
+  [forAll(~start, ~stop, ~f, str)] validates a predicate [f] for each character, from [start] (inclusive) to [stop] (exclusive)
+
+  Returns [true] if predicate [f(c)] returns [true] from all characters from [start] to [stop], [false] otherwise.
+*/
+let forAll = (~start=?, ~stop=?, ~f, str) => {
+  let start =
+    switch (start) {
+    | None => 0
+    | Some(v) => v
+    };
```
glennsl, line 36:
```suggestion
let forAll = (~start=0, ~stop=?, ~f, str) => {
```

src/Core/Utility/StringEx.re
```diff
@@ -23,6 +23,36 @@ let explode = str =>
...
+
+  let stop =
+    switch (stop) {
+    | None => String.length(str)
+    | Some(v) => v
+    };
```
glennsl, line 42:
```suggestion
  let stop = stop |> Option.value(~default=String.length(str));
```

src/Core/Utility/Cache.rei
```diff
@@ -0,0 +1 @@
+let create: (~initialSize: int, 'a => 'b, 'a) => 'b;
```
glennsl, line 1: `create` suggests that it's returning a cache object of some kind rather than a function. I think something like `embellish` would be more suggestive of what it actually does.

Or perhaps `Utility.Cache.cached` wouldn't be so bad, if a bit repetitive.

bryphe: `memoize` seems pretty descriptive for this - I switched it to that. Let me know if you prefer `cached` or `embellish` over `memoize` though.

glennsl: Oh yeah, `memoize` is a great name!

PR #1668, opened 2020-04-25
feat(workspace): window/split resizing, part 1 - basic functionality and commands

glennsl:
This refactors `Feature_Layout` to support window resizing and adds a number of commands to do so. Default keybindings are unfortunately blocked by multiple levels of bugs. It will still work with custom keybindings however, if carefully selected, and the PR should still be mergeable. It's just limited in functionality.
Peek 2020-04-25 22-34

TODO:
- [ ] Keybindings (blocked by several bugs)
- [ ] Mouse resizing
- [ ] implement `workbench.action.toggleEditorWidths`
- [ ] implement `<C-W>_` and `<C-W>|` (maximum height/width)
- [ ] add minimum window size constraint?
- [ ] wire up `:resize` and `:vertical resize`

### DIFF
--- a/src/Feature/Layout/Feature_Layout.re
+++ b/src/Feature/Layout/Feature_Layout.re
@@ -6,17 +6,19 @@ type direction =
   | Down
   | Right;
 
-type t('content) =
-  | Split([ | `Horizontal | `Vertical], list(t('content)))
-  | Window({
-      weight: float,
-      content: 'content,
-    })
-  | Empty;
+[@deriving show({with_path: false})]
+type t('id) =
+  | Split([ | `Horizontal | `Vertical], list(container('id)))
+  | Window('id)
+  | Empty
+and container('id) = {
+  weight: float,
+  content: t('id),
+};
 
 [@deriving show({with_path: false})]
-type sizedWindow('content) = {
-  content: 'content,
+type sizedWindow('id) = {
+  id: 'id,
   x: int,
   y: int,
   width: int,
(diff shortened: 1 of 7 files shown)


bryphe:
So cool @glennsl - awesome progress!

bryphe:
>  add minimum window size constraint?

A minimum size sounds good - we could use the editor font character width / line height to start.

> Keybindings (blocked by several bugs)

Are these the two we talked about: the `<C-w>` blocking other window bindings, and the issue with the virtual keys (ie, we're unable to bind to `<C-w>+` today - or are there additional ones, too?

Really excited for this functionality! 🎉

glennsl:
> A minimum size sounds good - we could use the editor font character width / line height to start.

That would require implementing it in the view though, which means you'll get a "dead zone" where the underlying weight is increased or decreased but not have a detectable effect and therefore seem to not work.

> Are these the two we talked about: the `<C-w>` blocking other window bindings, and the issue with the virtual keys (ie, we're unable to bind to `<C-w>+` today - or are there additional ones, too?

<C-w><S-=> also doesn't work, so it doesn't seem possible to distinguish between `+` and `=` for example, which is needed to implement several of these bindings.

And working around `<C-w>` blocking other window bindings by implementing window movement as ordinary keybindings also isn't possible because of the synchronization/timing dependency bug.

glennsl:
This one really doesn't want to go in. I think every CI box has failed now, each in turn.

glennsl:
I'm sure this will pass CI at some point in the next century or so...

glennsl:
2-in-1! Lucky me!

glennsl:
.azp run

glennsl:
It works! :tada:

gusbemacbe:
Hei @glennsl !

Are you still working on the sidebar resizing?

glennsl:
In a way. The goal is to (mostly) treat it as just another window, so the ability to resize windows would enable resizing the sidebar as well. Eventually.

PR #1673, opened 2020-04-28
Refactor: Integrate vscode-exthost with scanner / manifest

bryphe:
Paired with: https://github.com/onivim/reason-vscode-exthost/pull/22 - this refactors to use `vscode-exthost` to manage extension scanning and extension manifest, along with all the types that are used by the manifest.

Functionally, there should be no change, and this does not yet use `vscode-exthost` to actually run the extension host - but it hooks up some of the extension pieces, so that we have less plumbing to do when we actually go to make the switch to replace the `ExtHostTransport`/`ExtensionClient` with the new extension host.


bryphe:
I'll close this out, and migrate `reason-vscode-exthost` over instead

PR #1676, opened 2020-04-28
Fix #1578 - Terminal command + args configuration settings

### LINKED ISSUE #1578: Is it possible to customize the default shell for builtin terminal?
The behavior that I observe right now (on Windows) is as follows:

- `:term` opens cmd.exe shell
- so does `:terminal`
- "Open terminal" commands in command palette open PowerShell

I'd like all of these to open `wsl.exe` shell instead, Looking at #1403 I discovered that $SHELL environment variable affects the choice. After setting this globally, `:term` and `:terminal` open wsl.exe but "Open terminal" continues to open PowerShell.

It would be nice to be able to specify a consistent default across these different options via editor settings since I'm not sure what the implications of overriding SHELL globally are on Windows.

bryphe:
This implements some new configuration settings for the terminal, along with arguments to pass:

Shell command (per-platform):
- `terminal.integrated.shell.linux`
- `terminal.integrated.shell.windows`
- `terminal.integrated.shell.osx`

...as well as arguments to pass to the shell command:
- `terminal.integrated.shellArgs.linux`
- `terminal.integrated.shellArgs.windows`
- `terminal.integrated.shellArgs.osx`

__TODO:__
- [x] Wire up `shellArgs` to the shell launch config
- [x] Test on Windows
- [x] Test on Linux
- [x] Test on Mac
- [x] Integration Test
- [x] Add documentation

Fixes #1578


src/Feature/Terminal/Feature_Terminal.re
```diff
@@ -71,6 +72,35 @@ type outmsg =
...
+      ~default=[],
+    );
+  let shellArgsLinux =
+    setting(
+      "terminal.integrated.shellArgs.linux",
+      list(string),
+      ~default=[],
+    );
+  let shellArgsOSX =
+    setting("terminal.integrated.shellArgs.osx", list(string), ~default=[]);
```
glennsl, line 101: The convention I've used elsewhere is to make every dot-separated segment except the last a module, as that makes the translation to and from the keys really easy and predictable. Assuming it would be implied that everything here is prefixed with `terminal.integrated` you could for example have `Configuration.Shell.windows` and `Configuration.ShellArgs.osx`.

bryphe: Sounds good!

PR #1687, opened 2020-04-29
Fix #1686 - Sneak color unreadable in Monokai theme

### LINKED ISSUE #1686: Sneak isn't usable on 29/4/2020-5:43:44 Build
As you can see in the image, the letters are unreadable.
Oni-SneakBroken

My system:
Linux dlag-pc 5.6.7-arch1-1 #1 SMP PREEMPT Thu, 23 Apr 2020 09:13:56 +0000 x86_64 GNU/Linux

bryphe:
__`master`__:

(Note several themes are unreadable, like `Monokai` - called out in #1686)

__PR__:

(Some themes do not have an ideal contrast, but better than being totally unreadable...)

Fixes #1686

### DIFF
--- a/src/Feature/Theme/GlobalColors.re
+++ b/src/Feature/Theme/GlobalColors.re
@@ -587,7 +587,8 @@ module Oni = {
 
   module Sneak = {
     let background =
-      define("oni.sneak.background", all(ref(Selection.background)));
+      define("oni.sneak.background", all(ref(Menu.selectionBackground)
+      ));
     let foreground =
       define("oni.sneak.foreground", all(ref(Menu.foreground)));
     let highlight =
--- a/src/Feature/Theme/GlobalColors.re
+++ b/src/Feature/Theme/GlobalColors.re
@@ -587,8 +587,7 @@ module Oni = {
 
   module Sneak = {
     let background =
-      define("oni.sneak.background", all(ref(Menu.selectionBackground)
-      ));
+      define("oni.sneak.background", all(ref(Menu.selectionBackground)));
     let foreground =
       define("oni.sneak.foreground", all(ref(Menu.foreground)));
     let highlight =
--- a/src/UI/SneakView.re
+++ b/src/UI/SneakView.re
@@ -31,6 +31,13 @@ module Styles = {
   let item = (x, y, theme) => [
     backgroundColor(Colors.Oni.Sneak.background.from(theme)),
     position(`Absolute),
+      boxShadow(
+        ~xOffset=3.,
+        ~yOffset=3.,
+        ~blurRadius=5.,
+        ~spreadRadius=0.,
+        ~color=Revery.Color.rgba(0., 0., 0., 0.2),
+      ),
     top(y),
     left(x + Constants.size / 2),
     Style.height(Constants.size),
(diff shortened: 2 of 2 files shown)


src/Feature/Theme/GlobalColors.re
```diff
@@ -587,7 +587,8 @@ module Oni = {
 
   module Sneak = {
     let background =
-      define("oni.sneak.background", all(ref(Selection.background)));
+      define("oni.sneak.background", all(ref(Menu.selectionBackground)
+      ));
```
bryphe, line 591: Looks like this changed in https://github.com/onivim/oni2/pull/1635 - I moved it back to what it was prior. Is there something I should look out in changing this back, @glennsl ?

One thing we could consider is setting just a legible default, and customizing our built-in themes with oni.sneak colors.

src/UI/SneakView.re
```diff
@@ -31,6 +31,13 @@ module Styles = {
   let item = (x, y, theme) => [
     backgroundColor(Colors.Oni.Sneak.background.from(theme)),
     position(`Absolute),
+      boxShadow(
+        ~xOffset=3.,
+        ~yOffset=3.,
+        ~blurRadius=5.,
+        ~spreadRadius=0.,
+        ~color=Revery.Color.rgba(0., 0., 0., 0.2),
+      ),
```
bryphe, line 40: Figured I'd add a box shadow to the sneak squares while I was here...

src/Feature/Theme/GlobalColors.re
glennsl: Hmm, no I don't think so. When I changed this I think my assumption was that `selection.background` was more general and would be used as fallback for the more specific selection colors, so that lazy theme authors could get away with just defining that and not bother with the more specific ones. But that assumption didn't hold, and I knew that when I merged the PR, but I missed this one it seems.

> One thing we could consider is setting just a legible default

Or use a single theme color and derive the other by computing the opposite, for example, as both we and vscode do for the cursor.

bryphe: Cool, thanks for the details.

> Or use a single theme color and derive the other by computing the opposite, for example, as both we and vscode do for the cursor.

Hmm, didn't think about that, that's a good idea!

I tried it with each color - unfortunately there were still a few themes that were less than optimal, 

ie: `Selection.background` + opposite for foreground with gruvbox:
image

`Menu.selectionBackground` + opposite with laserwave:
image
(not as bad, but a bit hard to see on my lower contrast monitor..)

But it's good to know we have that in our toolbox.

glennsl: Hmm, yeah it would be nice to have a smarter one that maximizes contrast. `opposite`, which think I made emulate vscode's behaviour, just does `1 - intensity` for each of the color components, which obviously won't work very well for values close to 0.5.

bryphe: That makes sense - I wonder if another color space would work better for finding the opposite (like [HSL](https://github.com/onivim/oni2/pull/1687). Some interesting ideas here too: https://gamedev.stackexchange.com/questions/38536/given-a-rgb-color-x-how-to-find-the-most-contrasting-color-y

glennsl: If the color is converted to HSL it would be pretty easy I think. The "opposite" is just `hue + 0.5 mod 1.0`. The saturation can probably remain untouched, and the luminance is mostly what determines the contrast, so just adjust that up or down to until the difference is sufficient (but not maximized, as that would just be black or white)

glennsl: > I wonder if another color space would work better for finding the opposite (like HSL

Ah, great minds think alike!

PR #1689, opened 2020-04-29
perf(terminal): input regression

bryphe:
__Issue:__ Terminal input latency regressed in https://github.com/onivim/oni2/pull/1610 - the debouncing added some frames of input latency in the terminal. 

__Fix:__ Allow single terminal data / cursor move event to go through, before debouncing.


src/Service/Terminal/Service_Terminal.re
```diff
@@ -67,22 +67,30 @@ module Sub = {
         let makeDebouncedDispatch = () => {
           let scheduled = ref(false);
           let latestAction = ref(None);
+          let debounceCount = ref(0);
```
glennsl, line 70: This seems like a very imperative and error-prone way of doing this. I'm pretty sure a single `state` variant would be sufficient instead of multiple mutable variables that can be in any number of invalid states.

src/Service/Terminal/Service_Terminal.re
```diff
@@ -67,22 +67,30 @@ module Sub = {
...
+                      debounceCount := 0;
+                      latestAction^ |> Option.iter(dispatch);
+                      scheduled := false;
+                      latestAction := None;
+                    },
+                    Revery.Time.zero,
+                  );
+                ();
+              };
+            }
```
glennsl, line 93: I also don;t think this does what you want it to do. Since there's no timer to decrement the debounceCount, you've just moved the latency to a second invocation which might occur at any time in the future.

src/Service/Terminal/Service_Terminal.re
```diff
@@ -67,22 +67,30 @@ module Sub = {
         let makeDebouncedDispatch = () => {
```
glennsl, line 67: It would also be nice if this was generic. I see no reason it needs to be hard-coded for `dispatch`.

src/Service/Terminal/Service_Terminal.re
bryphe: Good point - working on a refactor

src/Service/Terminal/Service_Terminal.re
bryphe: Ya, it's totally busted! Appreciate you catching this. I've factored it out and adding some test cases to better cover this.

src/Service/Terminal/Service_Terminal.re
bryphe: True. I refactored to a `FunEx.debounce1` method

glennsl: Very nice. I'm not sure about just calling it `debounce1` though, it has a subtly different behaviour than a "normal" debounce. Perhaps `debounceLater1`? Or have an optional  `immediate` argument like [underscore](http://underscorejs.org/#debounce) does.

bryphe: Ah, you're right... Looking at the debounce docs, it's more akin to `throttle` because we don't reset the timer on each invocation (like `debounce` does). I switched the name, and added arguments for `~leading` and `~trailing` behavior.

PR #1694, opened 2020-04-30
Fix #660 - Clipboard: Specify Character mode when not multiple lines

### LINKED ISSUE #660: + register in normal mode pastes in the new line
whenever i do "+p it pastes the result in the new line. Eg. if i have 'bla' in the clipboard and have file:
```
line[1]
line2
```
where `[x]` indicates cursor position in normal mode and i do "+p then i get:
```
line1
[b]la
line2
```
instead of:
```
line1bl[a]
line2
```
It works fine in the insert mode with Ctrl+v.

bryphe:
Pull in https://github.com/onivim/reason-libvim/pull/116 , modify clipboard provider to specify `character` mode when not multiple lines

Fixes #660 (and also `xp` behavior, when use-clipboard-on-paste is enabled)

### DIFF
--- a/src/Store/VimStoreConnector.re
+++ b/src/Store/VimStoreConnector.re
@@ -79,7 +79,7 @@ let start =
     let isMultipleLines =
       fun
       | [s] => String.contains(s, '\n')
-      | [..._] => true
+      | [_, ..._] => true
       | [] => false;
 
     let splitNewLines = s => String.split_on_char('\n', s) |> Array.of_list;
--- a/src/Store/VimStoreConnector.re
+++ b/src/Store/VimStoreConnector.re
@@ -79,7 +79,7 @@ let start =
     let isMultipleLines =
       fun
       | [s] => String.contains(s, '\n')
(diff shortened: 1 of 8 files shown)


src/Store/VimStoreConnector.re
```diff
@@ -76,6 +76,12 @@ let start =
       |> List.map(c => String.make(1, c))
       |> String.concat("");
 
+    let isMultipleLines =
+      fun
+      | [s] => String.contains(s, '\n')
+      | [..._] => true
```
glennsl, line 82: What? Apparently this is valid. It doesn't do what you think it does though. It translates to just `_`. Fortunately you should get a warning that the last case is unused. It should be:

```suggestion
      | [_, ..._] => true
```
Seems like you might have discovered a new bug in Reason though. I can even kind of imagine the nasty syntactic hack that would cause this.

bryphe: Ouch, you're right! Nice catch!

PR #1695, opened 2020-04-30
CI: Fix OSX libintl link error

bryphe:


bryphe:
This will be included in #1694

PR #1697, opened 2020-04-30
Exthost - DocumentContentProvider, Decorations, Diagnostics

bryphe:
Implement several other API notifications for the main thread:

- DocumentContentProvider
- Decorations
- Diagnostics

And add a test case for diagnostics.

### DIFF
--- a/src/Core/Utility/ListEx.re
+++ b/src/Core/Utility/ListEx.re
@@ -11,6 +11,23 @@ let rec last =
   | [x] => Some(x)
   | [_, ...t] => last(t);
 
+let mapResult: list(result('a, 'b)) => result(list('a), 'b) =
+  items => {
+    List.fold_left(
+      (acc, curr) => {
+        acc
+        |> ResultEx.flatMap(list => {
+             switch (curr) {
+             | Ok(v) => Ok([v, ...list])
+             | Error(_) as err => err
+             }
+           })
+      },
+      Ok([]),
+      items,
+    );
+  };
+
 /**
  * Return all but the last element in a list.
  */
--- a/src/Core/Utility/ListEx.re
+++ b/src/Core/Utility/ListEx.re
@@ -11,23 +11,6 @@ let rec last =
   | [x] => Some(x)
   | [_, ...t] => last(t);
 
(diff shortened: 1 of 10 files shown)


src/Core/Utility/ListEx.re
```diff
@@ -11,6 +11,23 @@ let rec last =
...
+             | Ok(v) => Ok([v, ...list])
+             | Error(_) as err => err
+             }
+           })
+      },
+      Ok([]),
+      items,
+    );
+  };
+
```
glennsl, line 30: This is a pretty bad name, as it doesn't actually do any mapping from he user POV. It also belongs in `ResultEx`, not `ListEx`.  It's a tricky one to name though. In Haskell and Scalaz it's called `sequence`, in Elm it's called `combine`, in ocaml-containers it's called `flatten` and in Base it's called `all`.

That last one should serve as a hint that you don't even have to implement and name it yourself, but can just use `Base.Result.all`.

src/Exthost/Diagnostic.re
```diff
@@ -0,0 +1,52 @@
...
+      );
+
+    Ok({
+      range,
+      message: ret.message,
+      severity: ret.severity,
+      // TODO:
+      //source: ret.source,
+      //code: ret.code,
+    });
```
glennsl, line 44: Why not use `ocaml-decoders` and avoid all this unnecessary conversion and duplicate types? IMO `deriving yojson` should be outright banned.

src/Core/Utility/ListEx.re
bryphe: Ah, cool - I figured it must exist somewhere.  I keep forgetting that we have `Base` around... would be nice if there was something like `hoogle` for all our dependencies, to search by function signature.

src/Exthost/Diagnostic.re
bryphe: Ah, I ported this over from existing code... switched to use the decoder strategy now, though.

glennsl: Ah, sorry, from the description I thought it was new. But good to get it converted anyway!

PR #1700, opened 2020-05-01
Dependency: esy-skia -> d60e5fe

glennsl:
Actually pick up the clang 10 build fixes. Not sure why the revery update PR worked with the stale resolution. And not sure why this resolution is needed now, but without it it falls back to 0.6.0. So there you go.

### DIFF
--- a/package.json
+++ b/package.json
@@ -238,7 +238,7 @@
     "revery": "revery-ui/revery#9f725e0",
     "editor-core-types": "onivim/editor-core-types#6a8afaf",
     "esy-cmake": "prometheansacrifice/esy-cmake#2a47392def755",
-    "esy-skia": "revery-ui/esy-skia#07c13a9",
+    "esy-skia": "revery-ui/esy-skia#d60e5fe",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -238,7 +238,7 @@
     "revery": "revery-ui/revery#9f725e0",
     "editor-core-types": "onivim/editor-core-types#6a8afaf",
     "esy-cmake": "prometheansacrifice/esy-cmake#2a47392def755",
-    "esy-skia": "revery-ui/esy-skia#07c13a9",
+    "esy-skia": "revery-ui/esy-skia#d60e5fe",
```
bryphe, line 241: It might be worth just removing both of these resolutions, which shouldn't be necessary (and cause extra confusion when we upgrade..)

glennsl: Tried that, but then it falls back to 0.6.0 for some reason. From the OP:

> not sure why this resolution is needed now, but without it it falls back to 0.6.0. So there you go.

bryphe: Got it, sorry I missed that.

PR #1701, opened 2020-05-01
Refactor - Exthost: Move handlers to Msg module

bryphe:


bryphe:
It's not too different from the `Exthost_SCM`/`Exthost_Terminal` model we have - I just pivoted it based on `Request/Reply/Response` to document the protocol. The modules are still standalone and could be split out. The challenge is the `Request` depends on `Client`, which depends on `Msg`, so I pivoted it this way to avoid the dependency cycle.

glennsl:
Hmm, yeah I don't really understand the overall context of this, but as you say it's easy to split them out later anyway. I just find a `Msg` layer very odd architecturally as it describes a technical concept rather than a domain concept.

bryphe:
> I just find a Msg layer very odd architecturally as it describes a technical concept rather than a domain concept.

Right, `Msg` probably isn't the right name anymore - might get shuffled around as I implement sending responses back. I'm sure there's a better, more descriptive name. 🤔  But I'll bring this in for now and continue to think about it.

PR #1702, opened 2020-05-01
Exthost - API - ExtHostDocuments: Document edit / update

bryphe:


src/Exthost/Eol.re
```diff
@@ -0,0 +1,12 @@
+[@deriving (show({with_path: false}), yojson({strict: false}))]
```
bryphe, line 1: This - along with the `ModelContentChange`/`ModelChangeEvent` modules were ported from the existing protocol. In the interest of saving time, so I can get us wired up to the v2, I'm going to keep as-is and not convert them to the decoder approach.

Issue #1659, opened 2020-04-23
Add window size to configuration


janvanveldhuizen:
As you cannot easily change the window size (In Windows the Alt+Space shortcut doesn't work) I would suggest to add a "workbench.windowSize" option in config file, in order to be able to always launch at your prefered size.

glennsl:
Don't you think it would be better to have workspace persistence, as proposed in #640? That's how most apps solve this.

janvanveldhuizen:
Yes, that makes more sense.
Basically, what I miss is the standard Windows alt+space menu, where I can quickly maximize the windows with Alt-Space-X.

glennsl:
Right, yeah you might want to make an issue in https://github.com/revery-ui/revery for that. Doesn't sound like it'd be too hard to add, but it does require some native integration.

glennsl:
Should this be considered fixed by #1703? Enabling <kbd>Alt</kbd>+<kbd>Space</kbd> is a Revery-issue entirely I think.

janvanveldhuizen:
I agree

Issue #639, opened 2019-08-11
Feature: Hot exit (persist unsaved changes on quit)


bryphe:
This is related to the following issues:

__Prerequisities:__
https://github.com/onivim/oni2/issues/610
https://github.com/onivim/oni2/issues/600

Once our 'quit' lifecycle is refined (ie, we're handling unsaved changes correctly...) - I'd like to add another layer in terms of "hot exit". 

Hot exit would work as follows:
- Persist the swapfile for active, unsaved buffers
- Persist the undofile for all open buffers for persistent undo

On quit, we'd save all the relevant swap files in the Onivim folder, and reference them via a configuration file (`workspace.json`). When we load, we'd check and see if there was an active `workspace.json`, activate it (ie, populate the splits as before), and then load the swapfiles and undofiles for buffers as we load them.

Essentially - it'd just act as a convenient, automatic way to reuse the swapfiles & undofiles that Vim already has for this sort of persistence.

Some open questions:
- What new APIs would we need for `libvim`? Would we want an explicit API to save / load swapfiles and undo files, or just point to a per-directory workspace?

PR #1703, opened 2020-05-01
feat(workspace/#640): persistence infrastructure + persist workspace and window state

### LINKED ISSUE #619: OSX: Onivim 2 requesting access to Calendar / Contacts
Reported on Twitter: https://twitter.com/brecke/status/1159200070077681665

I'm assuming this is only OSX, but trying to check.

Likely, the cause is the fuzzy-finder traversing into some directories - Atom has a good write-up on this here: https://discuss.atom.io/t/why-does-macos-say-that-atom-wants-to-access-my-calendar-contacts-photos-etc/59291 (In addition, VSCode has the same issue here: https://github.com/Microsoft/vscode/issues/55236)

An option would be to, by default, filter out these paths from the fuzzy finder.

It would be helpful to confirm is this _only_ occurs when running the fuzzy finder, or if this manifesting in other ways, too.

### LINKED ISSUE #640: Feature: Workspace Persistence
I'd like to add the concept of a _session_ to Onivim 2, which would include the following:

- The working directory the user was in
- Open splits / window configuration (which buffers are in which editor group)
- Window size & location
- Opt-in hooks for various features to persist (ie, like Zen mode)

We'd save this information to a specific location in the user's folder, likely as JSON.

On load, we'd pick up this persisted info and restore the user's workspace / session.

__Related:__

- Once we have these basics configured, we should extend the format to handle hot exit: https://github.com/onivim/oni2/issues/639
- Should integrate with https://github.com/onivim/oni2/issues/440 (when changing workspaces, we persist the current state of the 'old' workspace, and load the persisted state of the 'new' workspace)

### LINKED ISSUE #906: Use a better default working directory
Currently it seems to default to the directory it is launched from interactively, which the user might not have permission to write to, and which probably isn't where the user wants to edit files anyway.

Better defaults might be:
- The OS-specific Documents directory
- The user/home directory
- A dedicated workspace directory created initially and in whatever OS-specific directory is appropriate.
- Nothing
- To ask

glennsl:
This adds the necessary persistence infrastructure, as well as global and workspace stores for persisting the following items:

- Oni2 version
- Workspace path
- Window position
- Window size

It can be merged as-is, but there's a couple minor issues I'd like to iron out that requires upstream changes:
- [x] Fix "jumping" when restoring position
- [x] Store and restore maximized state

Edit: Actually, this shouldn't be merged until I've converted the write to use Libuv, as it's currently blocking.

Fixes #640
Fixes #619 - We now fall back to Documents instead
Fixes #906 - We now fall back to Documents, and then ultimately to $HOME
Addresses #639, #1659


src/Store/StoreThread.re
```diff
@@ -80,17 +82,15 @@ let start =
       ~setTitle,
       ~setVsync,
       ~window: option(Revery.Window.t),
-      ~cliOptions: option(Oni_Core.Cli.t),
+      ~filesToOpen=[],
+      ~overriddenExtensionsDir=?,
+      ~shouldLoadExtensions=true,
+      ~shouldSyntaxHighlight=true,
+      ~shouldLoadConfiguration=true,
```
glennsl, line 89: Switched form passing around the whole `cliOptions` record in favor of explicit arguments. This gives more control over the options, a better udnerstanding of the dependencies, and allowed the `Cli` module to be moved to `bin_editor` where CLI parsing and such belongs.

src/bin_editor/Cli.re
```diff
@@ -176,25 +155,14 @@ let parse =
...
-    | ([], [hd, ..._], _) => Rench.Path.dirname(hd)
-    | ([], [], "/") => homeOrWorkingDirectory
-    | ([], [], workingDirectory) => workingDirectory
+    switch (directories) {
+    | [first, ..._] => Some(first)
+    | [] =>
+      switch (filesToOpen) {
+      | [first, ..._] => Some(Rench.Path.dirname(first))
+      | [] => None
+      }
```
glennsl, line 165: The CLI parsing now only deals with CLI parsing instead of making fallback decisions. If there's no valid file or folder arguments passed on the command line, `folder` will be `None` and the decision on what to do delayed so other sources (like the persistence store in this case) can be checked independently as well.

src/Model/Workspace.re
```diff
@@ -8,11 +8,12 @@
...
   rootName: string,
 };
 
-type t = option(workspace);
-
-let initial: t = None;
+let initial = workingDirectory => {
+  workingDirectory,
+  rootName: Filename.basename(workingDirectory),
+};
```
glennsl, line 19: Workspace is no longer an `option`. I'm not sure why it ever was, as it's not necessary and was a pain in the neck to deal with everywhere. It was actually just `None` until the `FileExplorer` set it to the current working directory, which was a really odd place to make that decision and too late to properly deal with persistence, so now it's set when the state is created initially instead.

src/Store/Persistence.re
```diff
@@ -0,0 +1,247 @@
...
+    let path =
+      Filesystem.getStoreFolder()
+      |> Result.map(storeFolder => Filename.concat(storeFolder, store.hash))
+      |> ResultEx.flatMap(Filesystem.getOrCreateConfigFolder)
+      |> Result.map(folder => Filename.concat(folder, "store.json"))
+      |> Result.get_ok;
+
+    let outChannel = open_out(path);
+    Printf.fprintf(outChannel, "%s", str);
+    close_out(outChannel);
```
glennsl, line 153: TODO: This should use Luv to avoid blocking I/O

src/bin_editor/Oni2_editor.re
```diff
@@ -190,25 +251,38 @@ if (cliOptions.syntaxHighlightService) {
...
       );
     let _: Window.unsubscribe =
-      Window.onFocusLost(w, () => dispatch(Model.Actions.WindowFocusLost));
+      Window.onFocusLost(window, () =>
+        dispatch(Model.Actions.WindowFocusLost)
+      );
+    let _: Window.unsubscribe =
+      Window.onSizeChanged(window, _ => persistWorkspace());
+    let _: Window.unsubscribe =
+      Window.onMoved(window, _ => persistWorkspace());
```
glennsl, line 285: These trigger workspace persistence.

src/bin_editor/Oni2_editor.re
```diff
@@ -156,21 +216,22 @@ if (cliOptions.syntaxHighlightService) {
       if (isDirty^) {
         update(<Root state=currentState^ />);
         isDirty := false;
+        persistGlobal();
```
glennsl, line 219: The global store is persisted only when the state changes.

src/bin_editor/Oni2_editor.re
```diff
@@ -74,6 +76,70 @@ let cliOptions =
...
+        let store = Workspace.storeFor(workingDirectory);
+
+        (
+          Store.get(Workspace.windowX, store),
+          Store.get(Workspace.windowY, store),
+          Store.get(Workspace.windowWidth, store),
+          Store.get(Workspace.windowHeight, store),
+        );
+      }
+    );
```
glennsl, line 116: This is where the persisted window state is retrieved.

src/Model/Workspace.re
bryphe: Unfortunately, there are cases where we can't necessarily make a good decision on a 'default' workspace directory - for example, this bug on OSX: https://github.com/onivim/oni2/issues/619

By default, when you launch Onivim 2 on OSX, it uses `~` as the working directory... which has some nasty side effects as documented in https://github.com/onivim/oni2/issues/619. It's a pretty horrible first run experience on Mac, at the moment.

In that case - we're better off not setting a workspace, and giving the user an option to choose one (similar to how VSCode behaves). So we're going to have to represent the no-active-workspace case in some way (and in that case - we shouldn't pick up any persistence).

src/Store/Persistence.re
```diff
@@ -0,0 +1,247 @@
...
+
+  let create = () => ref(None);
+
+  let send = (inPipe, outPipe, data) => {
+    inPipe := Some(data); // send
+    let received = outPipe^; // receive
+    inPipe := None; // reset
+    received; // return
+  };
+};
```
bryphe, line 17: Nice 😉

src/Store/Persistence.re
bryphe: We should look at this across all our File I/O (like the extension scanning, too!)

src/Store/Persistence.re
```diff
@@ -0,0 +1,247 @@
...
+  type state = (Oni_Model.State.t, Revery.Window.t);
+
+  let windowX =
+    define("windowX", int, 0, ((_state, window)) =>
+      Revery.Window.getPosition(window) |> fst
+    );
+  let windowY =
+    define("windowY", int, 0, ((_state, window)) =>
+      Revery.Window.getPosition(window) |> snd
+    );
```
bryphe, line 215: Neat. It looks like we can define persistence per-feature, too.

src/bin_editor/Cli.re
bryphe: Sounds good!

src/bin_editor/Oni2_editor.re
```diff
@@ -74,6 +76,70 @@ let cliOptions =
...
+let initWorkingDirectory = () => {
+  let path =
+    switch (cliOptions.folder) {
+    | Some(folder) => folder
+    | None =>
+      switch (Store.Persistence.(Store.get(Global.workspace, Global.store))) {
+      | Some(path) => path
+      | None =>
+        switch (Revery.Environment.getWorkingDirectory()) {
+        | "/" => Sys.getenv_opt("HOME") |> Option.value(~default="/")
```
bryphe, line 89: We may want to use this method: https://github.com/onivim/oni2/blob/5d4c99a9409429358e171abf7c9e67b16bfb1bd2/src/Core/Filesystem.re#L216

src/bin_editor/Oni2_editor.re
```diff
@@ -74,6 +76,70 @@ let cliOptions =
...
+          // ~y,
+          ~width,
+          ~height,
+          (),
+        ),
+      app,
+      "Oni2",
+    );
+
+  Window.setPosition(window, x, y);
```
bryphe, line 137: Ah, so you're work on the SDL2 / Revery API will let us pass this directly in to avoid jank on window creation 👍

src/Store/Persistence.re
bryphe: I guess with the dependency structure, we may not be able to at this time, as `Persistence` is part of store. Does it make sense to move it to a Feature project, like for Themes or Configuration? (And provide an effect for persistence?)

src/Store/VimStoreConnector.re
```diff
@@ -1122,20 +1122,19 @@ let start =
         copyActiveFilepathToClipboardEffect,
       )
 
-    | VimDirectoryChanged(directory) =>
+    | VimDirectoryChanged(workingDirectory) =>
```
bryphe, line 1125: When we change workspaces - do we load persisted settings for the new workspace?

bryphe:
So nice to have the window size and position persisted! 🎉

src/Store/VimStoreConnector.re
glennsl: It doesn't, but it could. It would be kind of weird to have the window position and size suddenly change just because you `:cd` somewhere. But if we want other settings to apply at that point, we certainly could. The problem is more how to detect that the workspace changed perhaps. Could do it on `DirectoryChanged` I guess.

src/bin_editor/Oni2_editor.re
glennsl: Indeed, hoping so!

src/bin_editor/Oni2_editor.re
glennsl: Good idea!

src/Store/Persistence.re
glennsl: The Persistence infrastructure could be moved to `Core`, but the stores themselves need to have access to `State.t` I think. It shouldn't be a problem to add a `map` function and add feature-defined items to those stores though.

glennsl: I had that in mind, but forgot at some point. Will add to the todo list

src/Model/Workspace.re
glennsl: > By default, when you launch Onivim 2 on OSX, it uses ~ as the working directory.

Is that because it is explicitly passed in as a CLI argument, or just because that's our fallback? If the latter, why not just use a different fallback? It would make our life a lot easier.

Either way, it was de facto always set before though, because of the file explorer.So to revert it we'll have to decide what to do with that when there's no workspace.

bryphe: > Is that because it is explicitly passed in as a CLI argument, or just because that's our fallback?

It's not expliclity passed, but I guess it's the default working directory OSX uses when opened via Finder.

> If the latter, why not just use a different fallback? It would make our life a lot easier.

We've tried to find a good fallback but have been unsuccessful - I think the best thing for us to do is to create an experience around 'no-workspace-yet' (ie, an 'empty' experience for the file explorer that prompts the user to open a workspace).

As you mentioned, we already had a 'default' workspace set anyway, so this isn't a step backward - but it is something we'll have to figure out a solution in order to improve the first-run experience on OSX and fix https://github.com/onivim/oni2/issues/619

src/Store/VimStoreConnector.re
bryphe: Good point. I guess there are two classes of settings - ones that you would only want applied initially, and ones that you'd want to be applied anytime... Since we only have the one 'class' today, we can defer this until we have persisted settings that have a need for this.

src/Model/Workspace.re
glennsl: > We've tried to find a good fallback but have been unsuccessful 

`Documents` doesn't work? Or a dedicated oni2 workspace directory? There's some more discussion about this in #906 btw.

> I think the best thing for us to do is to create an experience around 'no-workspace-yet' 

That means we have to deal with the possibility of there not being a workspace everywhere, and therefore make a ton of decisions about what to do instead. What should we do with search if there's no workspace? What should we do with `:cd`? What should we do with `:e`? With `!`? Is no-oping or special casing everything that doesn't work without a workspace a better first-run experience?

> As you mentioned, we already had a 'default' workspace set anyway

Not just in the state though.There's always a working directory, and plenty of functionality uses that instead of the workspace. Being able to use `:e` and `:cd` without the UI giving you a clue about where you are certainly doesn't sound like a great experience.

> so this isn't a step backward

Buit it is, because we now persist the workspace. So the logic for that needs to be convoluted even more to account for the possibility of there being no workspace. Not a major step backwards, but it does make this and everything else much more complicated.

I really can;t see how this would end up being a better experience for anyone.

bryphe: At the high level, all I want is a fix for https://github.com/onivim/oni2/issues/619 - whatever that entails. If it means that we have to handle a `None` case, then that's what we have to do - it's simply not an acceptable user experience as it stands today. 

On the other hand, if we can find a solution that solves https://github.com/onivim/oni2/issues/619 with a good default, I'm all for that and of course I'd prefer to keep things simple 😄 

> `Documents` doesn't work?

That's the best suggestion so far, IMO! We can give that a shot, I believe it should fix https://github.com/onivim/oni2/issues/619 , without adding the complexity of handling the no-workspace case.

glennsl: Cool! Should we try that just for macOS then, or for all platforms? it does seem more natural for Windows as well, but on Linux I think there's a possibility that there is no `Documents` folder so we'll have to fall back to `$HOME`. Which is arguably a more natural choice on Linux anyway.

bryphe: I like the idea of using the documents folder, and then falling back to home if its not available (when starting via Finder / UI - when starting from the terminal, we can use that working directory).

Might be a good use case for the reason-native `dir` library: https://github.com/facebookexperimental/reason-native/tree/master/src/dir#diruser - They have a `Dir.User.document` API, that, if it is `None`, we could fall-back to `Dir.User.home`.

Sounds like we have a good plan, and shouldn't block this PR 👍  Thanks for brainstorming this!

glennsl: > when starting via Finder / UI - when starting from the terminal, we can use that working directory

How do we distinguish between being started from Finder/UI vs the terminal?

src/Store/Persistence.re
glennsl: This is actually a bit more involved than I first thought, as it uses `Filesystem` just above here to get the right path. And that logic is so convoluted that I'd really rather not touch it. I guess I'll just duplicate the parts that are blocking here.

glennsl: Or just do it once during instantiation and cache it. The read needs to be blocking anyway.

src/Model/Workspace.re
bryphe: I think that's the big issue for us to figure out... 

On OSX, there is a weird `-psn` argument (that we filter out):
https://github.com/onivim/oni2/blob/b3be9160e3a617c32a574f53588e53af0bc578c3/src/bin_launcher/Oni2.re#L97
 I'm not sure if this only happens on first launch (from gatekeeper), or if it happens anytime launched through Finder, but if it is the latter, that would be a reliable way on OSX.

Another option might be to check if the `SHELL` env variable is set - wouldn't be the case for launching from finder, but would be when running from the terminal...

src/Store/Persistence.re
bryphe: Seems reasonable - I need to profile startup perf and see what the bottlenecks are. Might be some targeted places we could make async (like extension scanning / loading), that would be more impactful.

bryphe: > Or just do it once during instantiation and cache it. The read needs to be blocking anyway.

Yeah, it seems like reading persistence has to be sync on startup, so we know the parameters to pass to the window.

src/Model/Workspace.re
glennsl: Another alternative might be to do like `vscode`, require a `.` argument to open the working directory from the terminal.

bryphe: That sounds good!

glennsl: > Might be a good use case for the reason-native dir library: https://github.com/facebookexperimental/reason-native/tree/master/src/dir#diruser 

I'm not sure how to use this. It doesn't seem to be released. And since it uses FFI I can't easily just copy it into out project either, which might also have licensing issues.

glennsl: Oh there's an `opam` file in the root I can probably use.

bryphe: Ah, ya - looks like we could add a dependency like this with the json file, as well:
```
    "@reason-native/dir": "facebookexperimental/reason-native:dir.json#bb80cbc"
```

glennsl: Oh, that works as well? That doesn't seem to be documented. Worked fine with the opam file though, but I had to add `Fp` as well. Which we were going to do anyway so...

src/bin_editor/Oni2_editor.re
glennsl: Used `Dir.User.document` and `Dir.home` instead, as discussed above.

src/Store/Persistence.re
glennsl: Done.

src/Model/Workspace.re
glennsl: Done in https://github.com/onivim/oni2/pull/1703/commits/0eaedb31569816addebbdb3963706347d0eb4328. I'll mark this as fixing #619 too then.

bryphe:
I think the CI is hitting the same issue I saw in #1627 - I worked around it by adding this resolution:
```
"@opam/conf-pkg-config": "1.1"
```

glennsl:
Yeah, I saw that and adopted it in https://github.com/revery-ui/revery/pull/826 as well. Will bring it over here once that's merged, unless you've merged #1627 by then.

glennsl:
Hmm, this seems to fail everywhere except on my computer. Currently building on my Macbook to try there, but could you try this on Windows as well @bryphe ?

bryphe:
> Hmm, this seems to fail everywhere except on my computer. Currently building on my Macbook to try there, but could you try this on Windows as well @bryphe ?

Sure, taking a look - my guess would be that the syntax server is failing to start for some reason - might've been impacted by changes to the cli or logging in `Oni2_editor.re` 🤔

glennsl:
One of them definitely has an issue with the syntax highlighting, but the rest seem to all fail for different reasons. I'll restart to see if perhaps we can get even more flavors!

bryphe:
Fun 😄 It seems on OSX, it's failing when running this command

 `esy x Oni2 -f --version`

which outputs: 
```
##[error]Bash exited with code '2'.
```

exit code 2 is usually an OCaml exception - so we could try adding `--debug` to that invocation and see if we can get a stacktrace.

bryphe:
Same sort of failure on Windows. I think adding that `--debug` flag might give us a clue. I'll try it out on Windows too.

bryphe:
I think it's related to not having a `~/.config/oni2` folder on the CI machines - I moved mine and am getting an exception now (`esy x Oni2_editor -f --version` isn't outputting anything)

...but it's great that it caught this!

bryphe:
It seems like in the case where `~/.config/oni2` isn't created, it's crashing here: 
```
  let store =
    instantiate("global", () => Schema.[entry(version), entry(workspace)]);
```

glennsl:
Hmm, shouldn't that be created automatically? And shouldn't some integration test try to exercise user configuration or keybindings or some such?

bryphe:
It gets created here in `Filesystem.re`:
```
let getStoreFolder = () =>
  getUserDataDirectory()
  >>= getOniDirectory
  >>= (dir => getPath(dir, "store"))
  >>= getOrCreateConfigFolder;
```
I think the problem is the `getOrCreateConfigFolder` (where the creation happens) is happening after trying to get the store path

bryphe:
> And shouldn't some integration test try to exercise user configuration or keybindings or some such?

The integration tests do exercise user config and keybindings, but they create separate folders for it (mainly to prevent it from paving a developer's configuration when they run it 😬 )

glennsl:
There's really not much happening in `instantiate` before that gets called:

```
  let instantiate = (name, entries) => {
    let hash = Internal.hash(name);
    let filePath =
      Filesystem.getStoreFolder()
      |> Result.map(storeFolder => Filename.concat(storeFolder, hash))
      |> ResultEx.flatMap(Filesystem.getOrCreateConfigFolder)
      |> Result.map(folder => Filename.concat(folder, "store.json"))
      |> Result.get_ok;
    let store = {name, hash, filePath, entries: entries()};
```

glennsl:
> The integration tests do exercise user config and keybindings, but they create separate folders for it 

Right, that's the complicated override code.  It would be better to override the config folder more generally I think, and maybe that'd solve this issue as well.

glennsl:
In the meantime I can make sure to handle this more gracefully in the Persistence code. Just log an error and just continue without actually reading and writing anything.

bryphe:
It looks like the bug is in the `mkdir` function in `Filesystem` - which we call in `getOrCreateConfigFolder`:
```
    | None => 
      /*
       No folder was present, lets make one.
       */
      mkdir(configDir, ()) >>= (() => return(configDir))
```

The function is: 
```
let mkdir = (path, ~perm=userReadWriteExecute, ()) => {
  Unix.(
    try(mkdir(path, perm) |> return) {
    | Unix_error(err, _, _) =>
      error(
        "can't create directory '%s' because '%s",
        path,
        error_message(err),
      )
    }
  );
};
```

AFAIK, mkdir doesn't let you make a path of nested subdirectories... We need a better `mkdir` that handle a 'path' of subdirectories.

Before, we'd `mkdir("~/.config/oni2")`, but now we're using `mkdir("~/.config/oni2/store")`... and I guess that extra path is causing it to fail. A quick fix could be to ensure that we're creating the `~/.config/oni2` folder first, before anything else. A better long-term fix would be to set up mkdir to recursively create subdirectories, so that it can handle arbitrary paths.

And it seems like this could also fail in the case where a user had no `~/.config/` directory at all...

glennsl:
@reason-native/fs has a `mkDirP`:

https://github.com/facebookexperimental/reason-native/blob/bb80cbca60210bd1832c280c1bbcfb2bf3add6d5/src/fs/Fs.rei#L209-L221

bryphe:
I tried the latest, and I'm still seeing a crash on WIndows unfortunately.

src/bin_editor/Oni2_editor.re
```diff
@@ -58,14 +58,16 @@ let printVersion = _cli => {
   0;
 };
 
+Log.debug("Startup: Parsing CLI options");
```
bryphe, line 61: One thing I've seen cause crashes on Windows is logging (writing to `stdout`/`stderr`) before we get a chance to init a console instance (which happens in `Cli.parse`, if `-f` is specified). But I removed this line, and it didn't fix the crash

bryphe:
Ok, identified the crash!
```
(Invalid_argument "First token in path C:\\Users\\bryphe is not absolute.")
Raised at file "src/fp/Fp.re", line 220, characters 6-98
Called from file "src/dir/Dir.re", line 117, characters 16-35
Called from file "src/dir/Dir.re", line 187, characters 13-48
Called from file "src/dir/Dir.re", line 341, characters 8-20
```

Looks like `Fp` doesn't support windows style back-slashes. Looking at a place we can normalize these paths to work around it for now.

bryphe:
Ok, took some time but got it working now - turns out there were a couple blockers... I documented them more thoroughly in a PR to `dir`: https://github.com/facebookexperimental/reason-native/pull/251

glennsl:
Oh, I didn't notice that this is ready to be merged. I was waiting for reason-native to be updated! Thanks @bryphe!

PR #1706, opened 2020-05-02
Exthost v2 - API - Completion

bryphe:
This is the initial work to wire up language features to the 'v2' extension host, adding a test case for completion.

This adds a `request` API to the client which returns an `Lwt.t` promise with the response payload, and like the previous API, allows specifying a parser/transformer to convert the raw JSON to a strongly typed domain object.

### DIFF
--- a/src/Core/Utility/ResultEx.re
+++ b/src/Core/Utility/ResultEx.re
@@ -8,6 +8,14 @@ let flatMap = f =>
   | Ok(x) => f(x)
   | Error(_) as err => err;
 
+let map2 = (f, a, b) => {
+  switch (a, b) {
+  | (Ok(vA), Ok(vB)) => Ok(f(vA, vB))
+  | (Error(_) as errA, _) => errA
+  | (_, Error(_) as errB) => errB
+  };
+};
+
 let tapError = f =>
   fun
   | Ok(_) as ok => ok
--- a/src/Core/Utility/ResultEx.re
+++ b/src/Core/Utility/ResultEx.re
@@ -8,14 +8,6 @@ let flatMap = f =>
   | Ok(x) => f(x)
   | Error(_) as err => err;
 
-let map2 = (f, a, b) => {
-  switch (a, b) {
-  | (Ok(vA), Ok(vB)) => Ok(f(vA, vB))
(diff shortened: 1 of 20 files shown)


src/Exthost/Request.re
```diff
@@ -80,6 +80,37 @@ module ExtensionService = {
...
+module LanguageFeatures = {
+  let provideCompletionItems =
+      (
+        ~handle: int,
+        ~resource: Uri.t,
+        ~position: OneBasedPosition.t,
+        ~context: CompletionContext.t,
+        client,
+      ) => {
+    let parser = _ => {
```
bryphe, line 92: TODO: Implement the parser. The data model for completion items has actually changed significantly since the previous iteration - it uses a 'condensed' JSON model to reduce bytes-over-wire and to parse (which is nice for perf!)

test/Exthost/LanguageFeaturesTest.re
```diff
@@ -0,0 +1,54 @@
...
+      |> Test.waitForReady
+      |> Test.waitForExtensionActivation("oni-language-features")
+      |> Test.withClient(
+           Request.DocumentsAndEditors.acceptDocumentsAndEditorsDelta(
+             ~delta=addedDelta,
+           ),
+         )
+      |> Test.withClientRequest(
+           ~name="Get completion items",
+           ~validate=_ => true,
```
bryphe, line 46: TODO: Implement actual validation

src/Core/Utility/ResultEx.re
```diff
@@ -8,6 +8,14 @@ let flatMap = f =>
...
   | Error(_) as err => err;
 
+let map2 = (f, a, b) => {
+  switch (a, b) {
+  | (Ok(vA), Ok(vB)) => Ok(f(vA, vB))
+  | (Error(_) as errA, _) => errA
+  | (_, Error(_) as errB) => errB
+  };
+};
+
```
glennsl, line 18: When it gets so complicated that you need to use `map2` it might be better to use the let ppx?

bryphe: Good idea, I forgot about that - I'll give it a try

PR #1708, opened 2020-05-02
refactor(exthost): Upgrade from 1.33.5 -> 1.45.3

bryphe:
This is updating to use the latest extension host - there were several changes that were made:

Our previous 'extension host' integration had a custom transport layer that was hacked into the VSCode codebase - so that it could expose a `stdin`/`stdout` interface for communicating with Onivim 2. This was problematic, primarly because it was difficult to keep our forked extension host up-to-date as the VSCode codebase changed.

In this new model, we use libuv and talk to it as a node process would - this means that we no longer need a custom transport layer, and only a couple of lines need to be modified (mostly to make it run via node instead of electron) - so much easier to keep up-to-date and pull in new features!

We had several extensions blocked because they require a newer version of the Code extension host than we had bundled - so this will help unblock those issues and get us to the next round of blockers.

This PR wires up this new extension host to be used by Onivim 2, rather than the old one. Once we've fully converted the API, we can remove the old strategy.

There are several APIs that are still needed to be implemented for parity with the old version:

- [x] Workspace: acceptWorkspaceData
- [x] Configuration: acceptConfigurationChagned
- [x] provideTextDocumentContext
- [x] provideDecorations
- [x] SCM
   - [x] `provideOriginalUri` effect
   - [x] `spliceResourceStates`
- [x] Diagnostics
- [x] Terminal
- [x] Language Features
  - [x] Completion registration / request items
  - [x] Definition registration / request
  - [x] Document highlights registration / request
  - [x] Find all references
  - [x] Document symbol
- [x] Fix initialization timing
- [x] Upgrade bootstrapping of exthost to pick up new version
- [x] Reason fix `esy` path issue
- [x] `oni-dev` extension activation error: `Activating extension 'undefined' failed: The \"path\" argument must be of type string. Received type undefined.`
- [x] Reason - fix completion. Why is it not returning?
- [x] CSS - why is color completion not showing up?
- [x] TS / JS - Investigate 'invalid language tag' error (try upgrading extension)
- [x] TS / JS - why aren't completions showing up, even though they are returned?
- [x] Git - investigate activation error: `Activating extension 'undefined' failed: Cannot read property 'machineId' of undefined.`
- [x] Git - investigate `$spliceResourceState` parse failure
- [x] Remove old ExthostTransport (and the flaky test...)
- [x] Remove old ExthostClient
- [x] Remove old ExthostProtocol
- [x] Get unit tests green
- [x] Fix spawn process arguments
   - [x] Use `hide*` flags for windows
- [x] Get integration tests green

__Split to separate PR:__
- Log `stdout` / `stderr` from extension host
- Upgrade vendored node dependency (looks like this isn't blocking the core upgrade, but we can move up to node 12.x now)


src/Exthost/DocumentFilter.re
```diff
@@ -18,3 +18,10 @@ let decode = {
     )
   );
 };
+
+let matches = (~filetype: string, selector) => {
+  switch (selector.language) {
+  | None => false
+  | Some(language) => String.equal(language, filetype)
+  };
+};
```
glennsl, line 27: Could be reduced to:

```suggestion
let matches = (~filetype: string, selector) =>
  selector.language == Some(filetype);
```
Might not even be worth it as a stand-alone function

src/Store/ExtensionClient.re
```diff
@@ -13,132 +13,164 @@ module LanguageFeatures = Feature_LanguageSupport.LanguageFeatures;
...
+      () => {
+        let uri = Buffer.getUri(buffer);
+        let position = Exthost.OneBasedPosition.ofPosition(location);
+
+        Exthost.Request.LanguageFeatures.provideDocumentHighlights(
+          ~handle=id,
+          ~resource=uri,
+          ~position,
+          client,
+        )
```
glennsl: Very minor, but since this patern is repeated may times in this module. These variables seem to just add unnecessary indirection:

```suggestion
        Exthost.Request.LanguageFeatures.provideDocumentHighlights(
          ~handle=id,
          ~resource=Buffer.getUri(buffer),
          ~position=Exthost.OneBasedPosition.ofPosition(location),
          client,
        )
```

bryphe: Went through and cleaned these up 👍

PR #1709, opened 2020-05-02
feat(Extension Host): Workspace APIs

bryphe:


src/Core/Uri.re
```diff
@@ -42,7 +42,33 @@ module Scheme = {
...
+           | _ => fail("No scheme"),
+         );
+
+    let decodeString =
+      string
+      |> and_then(scheme => {
+           Internal.isKnownScheme(scheme)
+             ? succeed(scheme |> Internal.ofString)
+             : succeed(Custom(scheme))
+         });
```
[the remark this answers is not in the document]
bryphe: ah, good catch

src/Core/Uri.re
```diff
@@ -49,20 +51,13 @@ module Scheme = {
       list(string)
       |> and_then(
            fun
-           | [scheme, ..._] =>
-             Internal.isKnownScheme(scheme)
-               ? succeed(scheme |> Internal.ofString)
-               : succeed(Custom(scheme))
+           | [scheme, ..._] => scheme |> ofString |> succeed
            | _ => fail("No scheme"),
```
glennsl: It's a good idea to be explicit and avoid the wildcard when at all possible so you get the benefit of exhaustiveness checking. And here you certainly can:

```suggestion
           | [scheme, ..._] => scheme |> ofString |> succeed
           | []=> fail("No scheme"),
```

test/Exthost/WorkspaceTest.re
```diff
@@ -0,0 +1,92 @@
...
+       )
+    |> waitForWorkspaceEvent(~name="Workspace showActive", evt => {
+         evt.eventType == "workspace.show" && evt.workspacePath == None
+       })
+    |> Test.withClient(
+         Request.Workspace.acceptWorkspaceData(~workspace=Some(workspace1)),
+       )
+    |> waitForWorkspaceEvent(~name="Workspace: changed to workspace 1", evt => {
+         evt.eventType == "workspace.changed"
+         && evt.workspacePath == Some("/test/workspace1")
```
bryphe, line 80: This verification needs to be fixed on Windows (gets coerced to windows-style paths)

PR #1716, opened 2020-05-03
fix(i18n): handle unicode characters properly in Input

fanantoxa:
Updates input component to handle Unicode characters properly
- [x] Ctrl + p
- [x] Ctral + Shft + p
- [x] `:`, `/`
- [x] Ctrl + SHft + f
- [x] SCM
- [x] Add tests

### DIFF
--- a/src/Feature/InputText/Model.re
+++ b/src/Feature/InputText/Model.re
@@ -1,5 +1,4 @@
 open Oni_Core;
-open Utility;
 
 [@deriving show]
 type msg =
@@ -280,7 +279,6 @@ let%test_module "Model" =
      };
 
      let uTestString = "😊↪Вім is Cool";
-     let uTestStringLength = Zed_utf8.length(uTestString);
 
      let%test_module "paste" =
        (module
--- a/src/Feature/InputText/Model.re
+++ b/src/Feature/InputText/Model.re
@@ -35,11 +35,11 @@ module Internal = {
   let wordSeparators = " ./\\()\"'-:,.;<>~!@#$%^&*|+=[]{}`~?";
 
   let separatorOnIndexExn = (index, text) => {
-    String.contains(wordSeparators, text.[index]);
+    Zed_utf8.contains(wordSeparators, Zed_utf8.sub(text, index, 1));
   };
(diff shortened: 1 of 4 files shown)


fanantoxa:
@glennsl I have one bug with this implementation. Not sure if this should be fixed here or somewhere else:
So the problem is with `:` and `/` Menus.
When I start put Unicode (Ukrainian in this case) characters in at it add them properly and updates cursor properly.
If I press `<BS>` at the end it removes last character properly.
But If I press `Left` It doesn't update it to proper cursor position
(I have feeling that it still count 8byte chars in this case)
Could you help me with this please?

comand_bug

bryphe:
This is awesome @fanantoxa , thank you for improving this!

> But If I press Left It doesn't update it to proper cursor position
(I have feeling that it still count 8byte chars in this case)
Could you help me with this please?

Ah, interesting... I think this is similar to https://github.com/onivim/oni2/pull/1608 - where the 'cursor position' we're getting from Vim is being reported as the _byte position_ instead of a _character index_ - so we have to convert it from byte -> character to get the proper 'cursor caret position'. I'll take a look a bit more tomorrow

fanantoxa:
@glennsl Hey It's ready for review :wink:

fanantoxa:
@bryphe I was told to ask you for review)

src/Feature/InputText/Model.re
```diff
@@ -288,11 +278,13 @@ let%test_module "Model" =
...
 
+     let uTestString = "😊↪Вім is Cool";
+
      let%test_module "paste" =
        (module
         {
-          let pasteText = "hello from clipboard";
-          let pasteTextLength = pasteText |> String.length;
+          let pasteText = "😊↪Вім hello from clipboard";
+          let pasteTextLength = pasteText |> Zed_utf8.length;
```
bryphe, line 287: Nice! Thanks for adding all these test cases @fanantoxa 😄

bryphe:
Thanks for the ping @fanantoxa ! Changes look great to me. Really appreciate all the test cases you've updated & added.

PR #1717, opened 2020-05-04
Correct double click behavior on Mac (maximize)

zbaylin:
- Added `window` to `state` (this might be controversial)
- Added `window` param to `Titlebar`
- Execute the maximize function on double click

Issue #1631, opened 2020-04-19
Double clicking macOS title bar doesn't resize the window


opeik:
Very minor, but would be a nice quality of life feature.

Expected behaviour:

- When you double click the title bar, the window resizes to use all available space 

Actual behaviour:

- Nothing

glennsl:
I'm pretty sure you mean title bar, soI corrected that for you. Otherwise I do not know what you mean.

ArtemDrozdov:
Created issue in Revery repository to add double click event on Node so this feature can be implemented.

PR #1719, opened 2020-05-05
feat(other): fix macOS titlebar double-click behavior to reflect system default

zbaylin:
This reads the output from `defaults read 'Apple Global Domain' AppleActionOnDoubleClick` and determines what to do based on that: minimize, maximize, or nothing.

Extends #1717 and #1631


zbaylin:
This passed CI now! @bryphe I'll merge this on your approval.

PR #1726, opened 2020-05-06
chore: remove pkg-config workaround

glennsl:
Root cause fixed in https://github.com/esy-ocaml/esy-opam-override/pull/62


bryphe:
Awesome! Thanks @glennsl

PR #1728, opened 2020-05-06
chore(CI): update Azure Pipelines to use stages

CrossR:
Split up CI steps into stages, to aid re-running.
This should mean that we can more easily run individual stages of the CI, rather than the whole thing on a failure.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,284 +6,297 @@
 name: $(Build.SourceVersion)
 
 # Master build triggers
-trigger: 
+trigger:
 - master
 
 # PR Triggers
 pr:
   autoCancel: true
   branches:
-    include: 
+    include:
     - master
   paths:
-    exclude: 
+    exclude:
     - README.md
 
-jobs:
-- job: Kickoff
-  displayName: "Kickoff"
-  pool:
-    vmImage: 'ubuntu-latest'
+# Multistage pipeline, but no stage depends on another.
+# Split by Linux, Mac, Windows, Docker
+stages:
+- stage: LinuxStages
+  dependsOn: []
+  jobs:
+  - job: Kickoff
+    displayName: "Kickoff"
+    pool:
+      vmImage: 'ubuntu-latest'
 
-  steps:
-  - script: ./scripts/set-short-commit-variable.sh
-    name: shortCommitVariableStep
+    steps:
+    - script: ./scripts/set-short-commit-variable.sh
+      name: shortCommitVariableStep
 
-- job: Hygiene_Checks
-  displayName: 'Hygiene Checks'
-  timeoutInMinutes: 0
-  pool:
-    vmImage: 'macOS 10.14'
+  - job: Linux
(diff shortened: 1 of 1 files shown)


CrossR:
I've labelled this WIP rather than as a draft, since I think (could be very wrong) that draft PRs don't run the CI, which isn't great for this.

The idea here is that we have 3 build and 3 validate stages. So if the windows validate step fails, we just re-run that (using the existing artifact!), rather than the full thing. I've not looked if/how the in issue system works with it, but worst case, you have to click re-run in the UI.

Stable CI would be nicer, but this should let us get stuff in quicker regardless. Critically, this also doesn't actually change the CI at all, it just alters the ordering of stuff, so there should be no difference at all (which is good when we are so close to certain limitation like space).

CrossR:
Oh, and an example on my fork, to see how it works.

https://dev.azure.com/CrossR-1/oni2/_build/results?buildId=503&view=results

If you click though to the jobs you can see the windows bit failed and I re-ran it. Additionally I re-ran the mac test and it picks up and uses the existing build artifact, making re-running that very quick and not a 20 min thing.

CrossR:
Of course for this PR, everything would pass first time so I can't show re-running here, where its been failing repeatedly in actual code changes in other PRs.

At least I can see the "Rerun stage" button, and it looks like the permissions are being picked up correctly, since I'm logged in via GitHub on the Azure website.

CrossR:
Scrap that, I tried to rerun some test stages and it said I didn't have permissions.

Think this is unfortunately one of the places you'll need to sort it out @bryphe.
I'm also going to split it up some more, so the docker step is also separate, just since the docker build occasionally fails by itself.

bryphe:
Thanks for fixing up our build, @CrossR !

> Think this is unfortunately one of the places you'll need to sort it out @bryphe.

Hmm, I looked into this, and I can't see where to grant 're-run' permission - I just see this setting, which is currently enabled:

image

Was hoping there was a way to use `/azp run` like `/azp re-run` or `/azp runfailed` - that would be very convenient... but didn't find anything there either.

CrossR:
This 

https://docs.microsoft.com/en-us/azure/devops/pipelines/policies/permissions?view=azure-devops

seems to imply that you don't have control over the permissions directly, rather you add users to existing groups and then they inherit permissions from there. Or perhaps I'm understanding this wrong.

>Was hoping there was a way to use /azp run like /azp re-run or /azp runfailed - that would be very convenient... but didn't find anything there either.

Yeah, it looks like there is only the existing command which is unfortunate. I'll have a look around and see if there is any where they say they are updating this. The feature of using variables from previous stages is from last week or so, so it looks like stages support is fairly new anyways.

glennsl:
You can use `/azp run` to trigger a specific _pipeline_ by name, but I think _stages_ are something else.

CrossR:
I'm following a ticket on this, but I don't think its worth carrying on until there is a better way of running specific stages from here . I'll close this out and hopefully in the future we can move to something like this is they add a `/azp run failed` syntax.

PR #1718, opened 2020-05-04
docs: add changelog conventions to contribution guidelines

glennsl:

PR #1730, opened 2020-05-06
feat(other): add changelog

glennsl:
This adds a nifty changelog, as well as an "update" screen that opens after updating, showing all changes since the last version used and, more importantly, any breaking changes users should be aware of.

The changelog is based on the conventions outlined in #1718, and is automatically generated from the commit log and PRs by a script intended to be run during the nightly build, to be packaged together with the releases. The script takes two arguments, the path to the changelog file and an optional authentication token for the GitHub API to increase the rate limit.

### Update screen
2020-05-06-223720_811x694_scrot

### Full changelog
2020-05-06-223737_823x697_scrot

### DIFF
--- a/assets/changelog.xml
+++ b/assets/changelog.xml
@@ -1,5 +1,62 @@
 <changelog>
-  <commit hash="5d4c99a9" pr=1623 time=588257227>
+  <commit type="feat" scope="other/#1631" hash="92bab00d" pr=1717 time=588630193>
+    correct double click behavior on macOS (maximize)
+  </commit>
+  <commit type="feat" scope="Extension Host" hash="4edce141" pr=1709 time=588627097>
+    Workspace APIs
+  </commit>
+  <commit type="refactor" scope="exthost" hash="89236cac" pr=1711 time=588617053>
+    v2 - API - provideTextDocumentContent
+  </commit>
+  <commit type="chore" scope="dependency" hash="ea560d1d" pr=1714 time=588519632>
+    add fs, fp, dir
+  </commit>
+  <commit type="fix" scope="search/#1693" hash="4b76ab42" pr=1713 time=588516594>
+    closing the search pane doesn't "defocus it"
+  </commit>
+  <commit type="chore" scope="dependency" hash="996f6b48" pr=1712 time=588503999>
+    revery -> 7853ddb
+  </commit>
+  <commit type="refactor" scope="exthost" hash="33790f4e" pr=1706 time=588448480>
+    v2 - API - completion
+  </commit>
+  <commit type="fix" scope="file-explorer/#873" hash="cf6af66d" pr=1705 time=588433491>
+    exit zen mode when toggling explorer
+  </commit>
+  <commit type="fix" scope="vim" hash="3890823d" pr=1704 time=588433444>
+    linewise clipboard paste (yyp)
+  </commit>
+  <commit type="refactor" scope="exthost" hash="b3be9160" pr=1702 time=588373057>
+    API - ExtHostDocuments: Document edit / update
+  </commit>
+  <commit type="refactor" scope="exthost" hash="ef4c6b33" pr=1701 time=588357323>
+    move handlers to Msg module
+  </commit>
+  <commit type="chore" scope="dependency" hash="670657c4" pr=1700 time=588345462>
+    esy-skia -> d60e5fe
+  </commit>
+  <commit type="fix" scope="workspace" hash="ccb322ff" pr=1699 time=588343021>
+    broken save command
+  </commit>
+  <commit type="refactor" scope="exthost" hash="9838721d" pr=1697 time=588341199>
+    DocumentContentProvider, Decorations, Diagnostics
+  </commit>
+  <commit type="fix" scope="vim/#1671" hash="7c6f4a78" pr=1685 time=588298351>
+    - buffer loses unsaved changes when switching
+  </commit>
+  <commit type="chore" scope="ci" hash="bc2ac4d2" pr=1698 time=588298337>
+    Linux (CentOS) - remove cleanup prior to tests
+  </commit>
+  <commit type="chore" scope="dependency" hash="25539c48" pr=1696 time=588294522>
+    revery -> 9f725e0
+  </commit>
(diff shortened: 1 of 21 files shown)


src/Feature/Changelog/Feature_Changelog.re
```diff
@@ -0,0 +1,402 @@
...
+               fun
+               | {typ: Some("fix" | "perf"), _} => true
+               | _ => false,
+             );
+
+        <ScrollView style=Styles.scrollContainer>
+          <View style=Styles.content>
+            <Parts.Breaking items=breaking uiFont theme />
+            <Parts.Features items=features uiFont theme />
+            <Parts.Fixes items=fixes uiFont theme />
```
glennsl, line 394: Trying out a new convention. The `Parts` module is meant for internal sub-components. I was hoping it would be possible to do `Parts.breaking`, but unfortunately Reason won't accept that.

src/Store/VimStoreConnector.re
```diff
@@ -525,15 +525,36 @@ let start =
...
+          Actions.BufferRenderer(
+            BufferRenderer.RendererAvailable(
+              bufferId,
+              BufferRenderer.UpdateChangelog({
+                since: Persistence.Global.version(),
+              }),
+            ),
+          ),
+        );
+      };
```
glennsl, line 557: Unfortunately I had to replace the Welcome screen with the Update screen as opening both would have it fo into some kind of infinite `BufferEnter` loop.

src/UI/EditorGroupView.re
```diff
@@ -1,40 +1,19 @@
-/*
- * Editor.re
- *
- * Editor component - an 'Editor' encapsulates the following:
- * - a 'tabbar'
- * - an editor surface - usually a textual buffer view
- */
-
```
glennsl, line 8: Decided to clean this up as well, as there was a lot of convoluted, deceptive and unused stuff that made it hard to debug an issue I had with ``position(`Absolute)`` overflowing the actual editor area and overlapping the tabs.

assets/changelog.xml
```diff
@@ -0,0 +1,104 @@
+<changelog>
```
bryphe, line 1: Cool!

src/Feature/Changelog/Feature_Changelog.re
```diff
@@ -0,0 +1,397 @@
...
+      | None => "unknown"
+      };
+
+    let color =
+      switch (commit.typ) {
+      | Some("feat") => Colors.Oni.visualModeBackground.from(theme)
+      | Some("fix") => Colors.Oni.operatorModeBackground.from(theme)
+      | Some("perf") => Colors.Oni.normalModeBackground.from(theme)
+      | _ => Colors.foreground.from(theme)
+      };
```
bryphe, line 199: I like the detail and thought you put into the formatting - the screenshots look great!

src/Feature/Changelog/Feature_Changelog.re
```diff
@@ -0,0 +1,397 @@
...
+
+  switch (parse(simpleXml)) {
+  | commits => Ok(commits)
+  | exception (Failure(message)) => Error(message)
+  };
+};
+
+// VIEW
+
+let model = Lazy.from_fun(read);
```
bryphe, line 102: Nice use of `Lazy` to defer the reading until it's needed

src/Feature/Changelog/Feature_Changelog.re
bryphe: Looks nice! We could add this guidance to our coding standards. Seems reasonable to have a `Parts` module for internal sub components, and the `Styles` module for our internal styles.

src/UI/EditorGroupView.re
bryphe: Thanks for cleaning this up!

glennsl:
What do you think about the convention of having the pr number as supplementary information on the `scope` @bryphe? And do you think it adds any value to show it in the changelog, or should I cut out whatever come after a `/`?

bryphe:
> What do you think about the convention of having the pr number as supplementary information on the scope @bryphe? And do you think it adds any value to show it in the changelog, or should I cut out whatever come after a /?

I like the idea of showing the PR in the changelog. Would be really cool if we had it set up so the PR links were sneakable/clickable - and could pop open up a link to the issue.

It seems like it might be awkward, though, to have to include it in the naming - in the case where you're opening a PR, you don't know the PR number until after you open (likewise for creating a git branch)

glennsl:
These aren't the PRs though, but associated issues, i.e. the issue the commit fixes.

The PR numbers are also stored in the log however, but not displayed. The idea was to implement clickable links to open the PR at some point later, as I think that would require some native integration.

bryphe:
> __These aren't the PRs though__, but associated issues, i.e. the issue the commit fixes.

Ah, sorry I was confused, I thought you were asking about the __pr number__.

But I think I understand what you're asking now - for a format like below that contains the _issue number_:
```
fix(vim/#123): get_op_type error when using gcc
```

I do like having the issue number referenced in the commit, and I think it does make sense to parse it, and store it, like we do for PR number.

IMO It would be nice to show both the PR number and issue number if available, ie, eventually, something like:

```
feature | workspace | #640 | Fixes bug with workspace (PR #641)
```

Right now, if we don't do anything, it would show up as `workspace/#640`, and I'm not a fan of that notation with the `/` - it looks unpolished to me. Seems like we could at least add the parsing now to extract the issue number, if available, store it in our XML along with the PR, and implement a nicer display / link for it later.

glennsl:
I fully agree. I'll do that then.

glennsl:
Wow. CI completed in under an hour. I was starting to wonder if it actually ran! Is that just from #1738 @bryphe?

CrossR:
Would be even faster if the mac CI would reliably download its cache... with the cache the mac builds take around 15mins, not 50+. Though its not the end of the world when both linux ones take 40~ mins, and don't have a cache.

bryphe:
Looks like it did @CrossR !

image

 I don't think #1738 helped too much - but if there isn't a backlog (no time spent waiting for a machine) - all the individual jobs were < 60 min.

But ya, unfortunately it looks like the cache restore isn't working as expected - seems like every platform has a unique problem with it. It would really speed up the builds if the cache was working reliably!

PR #1734, opened 2020-05-06
refactor(exthost) - V2 - Decorations API

bryphe:


src/Exthost/Request.re
```diff
@@ -27,6 +28,93 @@ module Configuration = {
...
+                      ThemeColor.decode
+                      >>=:: (
+                        color =>
+                          succeed({priority, bubble, title, letter, color})
+                      )
+                  )
+              )
+          )
+      )
+    );
```
glennsl, line 80: Ouch.. This is why we use `obj` 😆

src/Exthost/Request.re
```diff
@@ -27,6 +28,93 @@ module Configuration = {
...
+               List.fold_left(
+                 (acc, (id, decoration)) => {
+                   let id = int_of_string(id);
+                   IntMap.add(id, decoration, acc);
+                 },
+                 IntMap.empty,
+                 items,
+               ),
+             )
+           });
```
glennsl, line 99: Seems like this should use `map`?

src/Exthost/Request.re
glennsl: Ah, I see this is an array. In that case it might be better to use `index` with the pipeline API perhaps. Something like:

```suggestion
    Json.Decode.(
      Pipeline.(
        decode((priority, bubble, title, letter, color) => {priority, bubble, title, letter, color})
        |> custom(index(0, int))
        |> custom(index(1, bool))
        |> custom(index(2, string))
        |> custom(index(3, string))
        |> custom(index(4, ThemeColor.decode))
      )
    );
```

This should format better. Would be nice to have an `obj`-like decoder for arrays too though.

bryphe: Ouch, ya, this was brutal. I didn't know about the Pipeline API - that's a tremendous improvement - thanks @glennsl !

src/Exthost/Request.re
bryphe: Ah, right

PR #1736, opened 2020-05-07
refactor(exthost): V2 - SCM - Port over SCM types / messages / requests

bryphe:


src/Exthost/Msg.re
```diff
@@ -502,6 +502,144 @@ module Telemetry = {
...
+                  | _ => Error("spliceResourceStates: Unexpected json")
+                  },
+                splices,
+              )
+            | _ => Error("spliceResourceStates: Unexpected json"),
+            groupSplices,
+          )
+
+        | _ => Error("Unexpected arguments for $spliceResourceStates")
+      };
```
bryphe, line 629: Need to flatten / adjust these types so that we can have a single `Ok(...)` result from it - the previous implementation relied on making multiple dispatches with a nested `List.iter`

src/Exthost/Exthost.rei
```diff
@@ -698,6 +788,14 @@ module Request: {
       Lwt.t(list(Location.t));
   };
 
+  module SCM: {
+    let provideOriginalResource:
+      (~handle: int, ~uri: Uri.t, Client.t) => Lwt.t(option(Uri.t));
+
+    let onInputBoxValueChange:
+      (~handle: int, ~value: string, Client.t) => unit;
+  };
```
bryphe, line 797: TODO: Implement these

src/Exthost/Exthost.rei
```diff
@@ -122,45 +122,49 @@ module SCM: {
...
     [@deriving show({with_path: false})]
-    type t = {
+    type resource = {
       handle: int,
-      uri: Uri.t,
-      icons: list(string),
-      tooltip: string,
+      resourceUri: Uri.t,
+      icon: Icon.t,
+      toolTip: string,
```
glennsl: What's the reason for these changes (other than icons -> icon)? "Tooltip" is widely considered to be a single word, and as far as I can tell is considered so in vscode as well. This casing is also inconsistent with the `tooltip` field in `command` above. `resourceUri` seems redundant, although ti does seem to have some grounds for consistency with vscode. And the conventional name for the main type of a module is still `t`.

bryphe: > What's the reason for these changes (other than icons -> icon)?

I was going for consistency with the `vscode` types: https://github.com/onivim/vscode-exthost/blob/50bef147f7bbd250015361a4e3cad3305f65bc27/src/vs/workbench/api/common/extHost.protocol.ts#L794

But you're right, `toolTip` and `icons` wasn't consistent - I'll fix those.

> And the conventional name for the main type of a module is still t.

I used this originally, but it caused problems, because `Splice` couldn't refer to a list of resources as `list(t)`. I tried various approaches like aliasing the type (ie, `type t = { ... }; type resource = t;` so that `Splice` could refer back to resource without thinking that `t` was referring to itself  - but I couldn't get the interface to match up.

glennsl: > I was going for consistency with the vscode types

I don't think it's all that important to be consistent with vscode, especially for types liek these that are aliased (in Feature_SCM) and predominantly used outside the context of the extension host. Not the end of the world either way though.

> I used this originally, but it caused problems, because Splice couldn't refer to a list of resources as list(t). I tried various approaches like aliasing the type (ie, type t = { ... }; type resource = t; so that Splice could refer back to resource without thinking that t was referring to itself - but I couldn't get the interface to match up.

There are a number of other ways of fixing this that won't break the API as well. Specifying `type resource = t;` should cause no more problems than renaming it. You can specify `type nonrec t = ...` on `Splice.t` to avoid `t` referencing itself. You can use local destructive substitution to give `Resource.t` a different name wihtout exporting it as a separate type: `type resource := t;`. And you could also move `Splice` up a level so it can reference `Resource.t`.

An ordinary alias should also work though, so I'd be curious to see the error.

bryphe: > I don't think it's all that important to be consistent with vscode,

I'd like to keep the `Exthost` layer close to the domain model presented here: https://github.com/onivim/vscode-exthost/blob/master/src/vs/workbench/api/common/extHost.protocol.ts

- that will make it easier for me when we update the exthost, and the domain model changes, to keep things in sync. From that vscode domain model, I'd see us transforming them to our domain models in the service / feature projects (ie, `Feature_SCM` could massage the exthost types into types that make more sense for our domain).

Compressing that transformation - like we did here - https://github.com/onivim/oni2/blob/4b1479e04c8a891e7b95e7948dba7f7191e01637/src/Extensions/ExtHostClient_SCM.re#L195 - makes it much more difficult for me to reason about the previous mapping of the old exthost data model to our data models, and on top of that, more difficult to convert to the new vscode exthost model. It's a lot more layers to try and get right (especially w/o any test coverage)

So the consistency is really to make my life easier as we continue to upgrade the extension host.

> You can specify type nonrec t = ... on Splice.t to avoid t referencing itself

Good call! I forgot about `nonrec` - that solves it 👍 (I can't use `[@deriving show]` with `nonrec`, but I don't think that's a big deal for this type..)

bryphe: I'll bring this in to unblock #1708 - but will be interesting to continue to refine the `Exthost` / feature boundary as we wire things up to the v2 exthost.

PR #1742, opened 2020-05-11
chore(osx): Add code-signing for watchdog and spdlog

bryphe:


bryphe:
This only impacts OSX and the OSX build passed - so I'll override and merge in now.

PR #1752, opened 2020-05-12
fix(technical/#1749): ballooning window size

glennsl:
**Issue:** Every time Oni2 was opened in a previous workspace the window size would grow from what was persisted before is some form of scaling was used.

**Defect:** The window size was persisted using pixel size, but restored using screen coordinates, which would have it grow by the scaling factor each time.

**Fix:** Add a function upstream in Revery to get the window size in screen coordinates and persist that instead.

### DIFF
--- a/bench.esy.lock/index.json
+++ b/bench.esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "1507c79509bff28587cb149a9636414c",
+  "checksum": "b12dcb2545299645884955b7d7c2b635",
   "root": "Oni2@link-dev:./package.json",
   "node": {
     "yarn-pkg-config@github:esy-ocaml/yarn-pkg-config#db3a0b63883606dd57c54a7158d560d6cba8cd79@d41d8cd9": {
@@ -1188,7 +1188,7 @@
         "@opam/dune-configurator@opam:2.5.1@aeb9d8d5",
         "@opam/dune@opam:2.5.1@a0c1e658",
         "@opam/dir@github:bryphe/reason-native:dir.opam#fd0225@d41d8cd9",
-        "@opam/decoders-yojson@opam:0.4.0@cb39dea6",
+        "@opam/decoders-yojson@opam:0.3.0@b9081413",
         "@opam/charInfo_width@opam:1.1.0@9d8d61b2",
         "@glennsl/timber@1.0.0@d41d8cd9", "@esy-ocaml/reason@3.5.2@d41d8cd9",
         "@esy-ocaml/libffi@github:onivim/libffi#590b041@d41d8cd9"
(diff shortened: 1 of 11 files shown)


CrossR:
I'm still getting a growing window on Windows 10 with the current build from the portal (https://github.com/onivim/oni2/commit/fa0d3d691938bcc8c6dba79f44f5559634b7cd8c). Perhaps the fix was only Linux specific?

CrossR:
As far as I can see, the the "raw" size and the size used in this PR are the same?

The raw size uses `w.metrics.size.width` which is set in `_getMetricsFromGlfwWindow`, which in turn calls `Sdl2.Window.getSize`, which is what the new method uses.

The docs for that SDL2 call implies we need to use something else: https://wiki.libsdl.org/SDL_GetWindowSize, so do we need to hook up and use `GetRendererOutputSize`?

bryphe:
Thanks for looking into this, @CrossR !

I suspect there's a case with scaling not handled properly - either via `GDK_SCALE` or on a high-dpi windows display. Does your Windows 10 machine have display scaling?

CrossR:
Yeah, its a 4K monitor with 150% scaling, so could be scaling or high-DPI issues.

bryphe:
Cool, thanks @CrossR - I'm able to reproduce this on a Windows 10 device with scaling, and also on Linux with `GDK_SCALE` set

glennsl:
I don't have a high-DPI monitor, so I'm pretty much just stabbing in the dark.

I can't see how something like `GetRendererOutputSize` would work though, as that returns pixel size. Assuming this still follows the conventional definition of pixel, as a representation of physical coordinates, there should be more pixels per "screen coordinate" with a high-DPI and/or a scale factor above 1.0. And since the problem is that the window grows, not shrinks, the problem seems to be that we're getting pixels and/or unscaled coordinates.

Could you run some tests with the reason-sdl2 example @CrossR so we can really figure out the relation between these coordinates?

bryphe:
I can help with this @glennsl - I have setups to test on both Linux/Windows.

I think the main problem is inconsistency in the Revery `Window` API in talking about sizes - there are two window sizes that SDL gives us - the 'screen space' size, and the 'framebuffer' size. On top of that, we also account for display scaling (like frameworks like Electron handle). So that makes three 'spaces' for window sizes:
- 1) Framebuffer pixel size
- 2) Screen space size
- 3) Screen space size, with display scaling

The problem was, the API surface wasn't consistent in handling this - `create`, `getSize`, and `setSize` should all be consistent in the space they use - which they weren't. `create` would use coordinates in the 'display scaled' space (3), but getSize/getRawSize would get the dimensions in (2) - likewise with `setSize`.

I opened https://github.com/revery-ui/revery/pull/840 to try and clean up that API and clarify this... hopefully it helps.

PR #1754, opened 2020-05-13
fix(changelog): Correct invalid XML attributes

zbaylin:
Before the XML tags for commits had integer literals, which XML does not support. This PR simply changes the attributes to be wrapped with double quotes.

It also adds a note about using the script to generate changelogs locally.

### DIFF
--- a/assets/changelog.xml
+++ b/assets/changelog.xml
@@ -1,82 +1,85 @@
 <changelog>
-  <commit type="chore" scope="dependency" hash="fa0d3d69" pr="1753" time="1589302850">
+  <commit hash="10ba05e2" time="1589378010">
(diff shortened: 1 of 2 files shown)


zbaylin:
@glennsl also noted that `pr` was null for PRs that have single commits. I fixed that too.

zbaylin:
No problem! Do you mind if I skip CI for this? I don't think there are any tests in there that are relevant to this script. This will help me add links for the PRs in the changelog!

bryphe:
Do you mind if I skip CI for this? I don't think there are any tests in there that are relevant to this script.

@zbaylin - sure, as long as you confirm this and sure it won't break the build 👍 

> This will help me add links for the PRs in the changelog!

💯

Issue #803, opened 2019-10-02
Unable to open file if all tabs have been closed


AndreasNasman:
After closing all open tabs, opening files (through clicking, `:e` or `quickOpen.open`) stops working.

bryphe:
Was able to reproduce this:

![image](https://user-images.githubusercontent.com/13532591/75503053-749c1880-5989-11ea-8c0a-f268e45812b1.png)

bryphe:
This is fixed now with #1776 👍

Issue #733, opened 2019-09-04
Closing window with unsaved changes results in no other window being created for new buffers


avimmer:
Related to #610 

Version: 2.0.2.0
Date: 8/31/2019, 5:18:41 AM
Commit: 237e14b

* Start onivim2
* Observe there's a new untitled buffer open
* Enter insert mode
* Type "test"
* Observe window close button change to a modified buffer indicator
* Press the modified buffer indicator
* Window closes and you're presented with an empty pane
* Try opening any other buffer using file explorer or command
* Observe new buffers being opened as indicated by the status line, but no actual window appears until restart. It feels like you can modify those buffers without any visual feedback.

bryphe:
Nice catch @avimmer ! Definitely a bug - we end up in a broken state where buffers aren't being opened.

leavengood:
This can still happen with the Welcome pane too. Steps now:

1. Start onivim2
2. Use `:e some_new_file.txt` to get a new file.
3. Close the Welcome tab.
4. Make some changes in the new file.
5. Click the modified circle to close the new file.
6. Bad state now entered.

I believe an empty window should be a valid state in general, because we probably need a checkbox on the welcome screen to be able to turn it off, like in VSCode. In fact, now that I test VSCode, their empty window has something sort of like the current onivim2 welcome tab:

<img width="693" alt="Screen Shot 2020-03-11 at 5 11 39 PM" src="https://user-images.githubusercontent.com/637/76464254-66072580-63bb-11ea-8477-01dc74a6b285.png">


So obviously we need a bit more work there, but it is probably not too hard. I may investigate some of this.

PR #1759, opened 2020-05-13
feat(workspace): add drag and drop support to EditorView

zbaylin:
If you drag multiple files onto the editor view, it will check
all the paths to make sure they're files (not dir, socket, link, etc.)
If they are files, it will dispatch an OpenFileByPath action, which
will/should open the file in the editor!

oni2-drag-drop1

__Todo:__
- [x] Investigate multi-file drop (@bryphe )

__Test Cases:__
- [ ] Single file drop on Windows (dev build)
- [x] Single file drop on Windows (packaged build)
- [x] Single file drop on OSX (dev build)
- [x] Single file drop on OSX (packaged build)
- [x] Single drop on Linux


src/UI/EditorView.re
```diff
@@ -22,16 +22,28 @@ module Styles = {
...
+let onFileDropped = (event: NodeEvents.fileDropEventParams) =>
+  List.iter(
+    path => {
+      let stats = Unix.stat(path);
+      // This tests if the path dropped is a file. We can't handle opening directories, etc. yet
+      if (stats.st_kind == S_REG) {
+        GlobalContext.current().dispatch(
+          Model.Actions.OpenFileByPath(path, None, None),
+        );
+      };
```
glennsl, line 34: side-effects should not be triggered in the UI, as that makes it difficult to track and control what's happening,a nd to evolve the logic when we need to update the state as well, for example. In short, it elads to spaghetti code. The rule of thumb is that the UI should only dispatch messages to be handled by stores/update functions, which then update the model and emit side-effects through `Isolinear.Effect`.

But even more generally, even in imperative-land, doing blocking I/O on the UI thread is considered a big no-no as it can blocks the UI from rendering and in some cases can really hurt responsiveness.

zbaylin: Yeah that's my bad -- I kinda just quickly worked on this to get it to somewhat work. I should have added a WIP tag on the PR generally.

Thanks for the info!

zbaylin:
@glennsl I re-architected this to what (I think) is more in line with the Isolinear/Oni architecture as a whole. Let me know if I messed anything up.

And thanks again to @bryphe for walking me through the architecture last night!

zbaylin:
As a side note, I'm getting weird behavior with dropping multiple files that I obviously would like to fix before this gets merged. If I drop two files and then try to close one of them, it just swaps positions with the other in the tabs.

Probably easier to just watch what I mean:
oni2-multi-drag-drop-error

You can't really tell from the framerate of the GIF but the log in the back is also going crazy -- also not sure why.

All I'm doing is dispatching, in this case, 2 `OpenFileByPath` Actions, so I'm not exactly sure why that would cause this behavior.

src/Feature/FileDrop/dune
```diff
@@ -0,0 +1,10 @@
+(library
+    (name Feature_FileDrop)
```
bryphe, line 2: Thanks for putting this scaffolding together, Zach!

src/Feature/FileDrop/Feature_FileDrop.re
```diff
@@ -0,0 +1,3 @@
+[@deriving show({with_path: false})]
```
bryphe, line 1: This might seem small for now, but this is the perfect place for us to add a model once we incorporate dragging 👍

src/Store/Features.re
```diff
@@ -186,6 +186,25 @@ let update =
...
 
+  | FileDrop(msg) =>
+    let eff =
+      switch (msg) {
+      | FilesDroppedOnEditor({paths}) =>
+        Isolinear.Effect.createWithDispatch(
+          ~name="feature.filedrop.openFiles", dispatch => {
+          List.iter(
+            path => {
+              let stats = Unix.stat(path);
```
bryphe, line 197: Thanks for moving the file I/O to an effect!

bryphe:
Looking good to me, @zbaylin ! Awesome work.

> All I'm doing is dispatching, in this case, 2 OpenFileByPath Actions, so I'm not exactly sure why that would cause this behavior.

I'm doing some work in this area today, so I'll investigate this in a bit. Your code seems fine, so I think it's exposing / hitting a bug.

One thing we should think about are the set of test cases we should try to make sure this is ready-to-go - some tests I can think of:
- [ ] Single file drop on Windows
- [ ] Single file drop on OSX
- [ ] Single drop on Linux

And then a couple other thoughts:
- How does this work with _Zen mode_? Does it switch to the last file, or should it leave zen mode? 🤔 
- For now, we just ignore directories, but maybe as a next step we could look at changing directories / workspaces, if just a single folder is dragged.

zbaylin:
> I'm doing some work in this area today, so I'll investigate this in a bit. Your code seems fine, so I think it's exposing / hitting a bug.

Sounds good. Thanks for that.

> One thing we should think about are the set of test cases we should try to make sure this is ready-to-go - some tests I can think of:

I can test on Linux. Unfortunately I can't get Oni to build on my Windows VM for various reasons. 

> For now, we just ignore directories, but maybe as a next step we could look at changing directories / workspaces, if just a single folder is dragged.

VSCode switches workspaces on a single folder being dragged. I'm not sure what the behavior is for multiple folders, or a folder mixed in with files. I can look into that today.

bryphe:
> I can test on Linux. Unfortunately I can't get Oni to build on my Windows VM for various reasons.

Ok - I added a list of test cases at the top so we get a nice progress bar on the PR to track whats left:
image

I'll test Windows for you - and seems like OSX is already covered. Are there any other test cases you can think of that we need to check?

> VSCode switches workspaces on a single folder being dragged. I'm not sure what the behavior is for multiple folders, or a folder mixed in with files. I can look into that today.

Awesome, that'd be great! We can implement this in a separate PR, so it's not blocking this work. I'm thinking we may want another `msg` to model this, like `FolderDropped`

zbaylin:
Not for this PR's scope -- but eventually we'll want to expand this to the file tree. I can't think of any more specific tests for this PR. If and when you figure out that bug with multiple files we can test that, but other than that I can't think of anything.

bryphe:
Sounds great @zbaylin !

bryphe:
One other test case I think of is verifying on the packaged application (just in case there's something busted with the `Info.plist` metadata - I added it to the list). Let me know if you need any help creating a packaged build via `esy create-release`

zbaylin:
@bryphe Linux works!

GIF:
oni2-single-drag-drop-linux

zbaylin:
Same with macOS packaged!

GIF:
oni2-single-drag-drop-mac-packaged

glennsl:
> You can't really tell from the framerate of the GIF but the log in the back is also going crazy -- also not sure why.

That sounds a lot like the issue I had with opening both the update screen and welcome screen initially. It would go into some kind of infinite BufferEnter loop. Solved it by just replacing the welcome screen instead, which is less than ideal, so it'd be cool if fixing this bug solved that as well.

bryphe:
> That sounds a lot like the issue I had with opening both the update screen and welcome screen initially

Definitely. I'm working through cleaning up that codepath today, to reduce the awkward timing dependencies:
- https://github.com/onivim/oni2/pull/1762
- https://github.com/onivim/oni2/pull/1763
- https://github.com/onivim/oni2/pull/1764

and then:
- Remove some imperative `BufferEnter` dependencies
- Remove the awkward `RecalculateEditorView` action
- Merge in https://github.com/onivim/oni2/pull/1627

Hoping to fix the multiple-buffer-open issue, along with these issues today:
- https://github.com/onivim/oni2/issues/803
- https://github.com/onivim/oni2/issues/733

bryphe:
It might take a bit of time to work through it - sorry that you're both blocked on this!

In the meantime, @zbaylin - a couple options:
- We could restrict this PR to just a single file, and merge as-is, and add in the multiple-file behavior once I've worked those open file issues
- You could switch gears to another piece of work (like the changelog links?), and we could merge this to `master` once I've finished that set of work and the multiple-file-open is unblocked

zbaylin:
@bryphe I'm fine leaving it until the multiple files bug is fixed. I can help @glennsl with the changelog in the meantime!

src/Feature/FileDrop/Feature_FileDrop.re
glennsl: I have to admit I'm a bit more concerned. Not so much because of the granularity, although that is a minor concern too, but more because this seems to be a cross-cutting concern. It's not clear to me how this will integrate with other features that need to support drag and drop. Why would dropping a file on the file explorer, for example, even need to involve a separate file drop feature project? It seems like very little code would live here, but would instead have to be implemented in the consuming feature projects (file explorer/workspace) or in the glue code, as is largely the case here.

I can see that there's a missing feature project here for this specific feature to live in however, but I think that missing project is for the workspace as a whole.

src/Store/Features.re
```diff
@@ -186,6 +186,25 @@ let update =
...
+              if (stats.st_kind == S_REG) {
+                dispatch(OpenFileByPath(path, None, None));
+              };
+            },
+            paths,
+          )
+        })
```
glennsl, line 204: Ir would be nice to have this effect moved out of the update function, to avoid it turning into a blob. I also think ti does too much. Ideally this would be two separate effects, one for the `stat`, which ought to be non-blocking and dispatch a new user-supplied message when done, and then another to open the file. That would also enable the stat effect to be moved into the feature project, while the open file effect would need to be an `outmsg` because it's hard-coded to `Actions.t` for historic reasons.

This area is still a bit murky with a lot of historical baggage and few good examples of how to do these things properly, so don't worry too much about it, but the general lesson here is that supplying the message to be dispatched when it is invoked allows effects to be generalized.

glennsl: Here's one example of such an effect:

https://github.com/onivim/oni2/blob/d971dae8cb82807a310744bfca2b995ecdfad929/src/Store/FileExplorerStore.re#L11-L25

Used here:

https://github.com/onivim/oni2/blob/d971dae8cb82807a310744bfca2b995ecdfad929/src/Store/FileExplorerStore.re#L52-L59

And another:

https://github.com/onivim/oni2/blob/d971dae8cb82807a310744bfca2b995ecdfad929/src/Feature/SCM/Feature_SCM.re#L65-L74

Used here:

https://github.com/onivim/oni2/blob/d971dae8cb82807a310744bfca2b995ecdfad929/src/Store/ExtensionClientStoreConnector.re#L246-L249

zbaylin: @bryphe and I worked this out last night and I think the consensus we came to was that eventually the drag and drop feature will have it's own state once the _drag_ part is implemented. I.e. if you held a dragged file over a certain area for x amount of time, something happens, etc. I'm obviously not as familiar with the underlying architecture as the two of you are, so it's possible @bryphe is able to explain better than I am.

I understand what you mean with regards to separating the actions -- I can do that. Thanks for the examples!

bryphe:
> I can see that there's a missing feature project here for this specific feature to live in however, but I think that missing project is for the workspace as a whole.

@glennsl - do you think we should move this to a feature project like `Feature_Workspace`? In discussing we figured the state that we managed would potentially be a _drag state_ for the gesture (ie, the drag events Zach implemented here: https://github.com/revery-ui/esy-sdl2/pull/13

>  Ideally this would be two separate effects, one for the stat, which ought to be non-blocking and dispatch a new user-supplied message when done, and then another to open the file

So something like `Feature_Workspace.Effects.stat: (~path: string, ~mapResult: StatResult => 'msg) => Isolinear.Effect.t('msg)`?

glennsl:
> do you think we should move this to a feature project like Feature_Workspace?

Yeah, I think that would be more natural. You'd have a hierarchy of features, akin to a hierarchy of components, with the File Explorer as a sub-feature. With ownership of a piece of state at the highest level necessary in the hierarchy.

Having state "on-the-side" leads to unclear ownership I think.

> So something like `Feature_Workspace.Effects.stat: (~path: string, ~mapResult: StatResult => 'msg) => Isolinear.Effect.t('msg)`?

I would say more like:

```reason
Service_Filesystem.stat: (~path: string, ~onResult: StatResult => 'msg) => Isolinear.Effect.t('msg)
```

That is, moved to a service project, or even an external library, and using `~onResult` instead of `~mapResult` as that suggests more strongly that the `'msg` will be dispatched.

src/Feature/FileDrop/Feature_FileDrop.re
bryphe: > Not so much because of the granularity, although that is a minor concern too, but more because this seems to be a cross-cutting concern. It's not clear to me how this will integrate with other features that need to support drag and drop. Why would dropping a file on the file explorer, for example, even need to involve a separate file drop feature project?

Right, I think this is a fair concern! It seems like - at least the way this is implemented right now - `FilesDroppedOnEditor` - makes sense to be part of the `Feature_Editor` (since it's handling the use-case where a file is dropped on the editors, and not on the explorer or other parts of the shell, at the moment).

The main issue is that `Editor` is one of our least-factored features at the moment - we don't have a central `msg` or `updater` yet.

We could start a `Feature_Editor.msg` with `FilesDropped`, and stub out an updater - and then incrementally move the remaining functionality towards that (ie, there are lots of `Editor*` actions in `Actions.t` that could be migrated to this `msg` later).

src/Service/OS/Service_OS.rei
```diff
@@ -1 +1,6 @@
-module Effect: {let openURL: string => Isolinear.Effect.t(_);};
+module Effect: {
+  let openURL: string => Isolinear.Effect.t(_);
+  let stat: (string, Unix.stats => 'msg) => Isolinear.Effect.t('msg);
```
bryphe, line 3: Thanks for the refactoring here, Zach!

Issue #1766, opened 2020-05-14
Oni crashing with Yojson.Json_error("Blank input data"):


izik1:
os: Arch Linux (i3)
commit: #fa0d3d69 (presumably, 2020-05-14's nightly otherwise), AppImage
steps to repro, as a bash script:
```bash
# backup / remove global store file:
mv ~/.config/oni2/store/global/store.json ~/.config/oni2/store/global/_store.json

# control case (nothing bad should happen here)
oni2 --debug --trace -f 
# equivalently:
# oni2 --debug --trace -f .

# avoid contaminating future testcase
rm ~/.config/oni2/store/global/store.json

# exhibit a:
mkdir -p /tmp/onitest/example-a

 oni2 --debug --trace -f /tmp/onitest/example-a
# there should currently be a window that contains the framebuffer that the terminal last had when the window was created. (see ^1)

# in order to stop oni:
# navigate to terminal that opened oni
<ctrl-C>
killall -SIGKILL Oni2_editor

cat /tmp/onitest/example-a/onivim2-crash.log 
# observe that `/tmp/onitest/example-a/onivim2-crash.log` should contain the following: 
# Yojson.Json_error("Blank input data"):

cat ~/.config/oni2/store/global/store.json
# and that the global store file is indeed empty.

# surely, it will work now, right?
 oni2 --debug --trace -f /tmp/onitest/example-a
# oni should crash without opening any window or logging anything to stdout

cat onivim2-crash.log
# this should also contain: Yojson.Json_error("Blank input data"):
# note that _this_ error log is in `.` whereas the last one was in the opened directory.

# we need to clear the global store again
rm ~/.config/oni2/store/global/store.json

# but this time we should remove `onivim2-crash.log` too
rm onivim2-crash.log

# exhibit b:
mkdir -p /tmp/onitest/example-b/1

# this is required otherwise it'll complain that this directory doesn't exist later
mkdir -p ~/.config/oni2/store/21cdfff0b


echo '{"windowX":2,"windowY":40,"windowWidth":800,"windowHeight":600,"windowMazimized":false}' > ~/.config/oni2/store/21cdfff0b/store.json

echo '{ "workspace": "/tmp/onitest/example-b/1" }' > ~/.config/oni2/store/global/store.json

# oni should open normally. But it doesn't.
oni2 --debug --trace -f /tmp/onitest/example-b/1
# kill it as above.

# note that there shouldn't be any crashlog.
# observe one of the first logs should be:
[...]

PR #1768, opened 2020-05-15
fix(persistence/#1766): Blank input data

bryphe:
__Issue:__ If the `store.json` is corrupt, we crash on launch - described by #1766 

__Defect:__ There are some additional error cases - like a corrupted `store.json` file - that could cause us to crash.

__Fix:__ Log out errors, but revert to defaults if we can't load the persistence file.

### DIFF
--- a/test/Core/PersistenceTests.re
+++ b/test/Core/PersistenceTests.re
@@ -1,0 +1,45 @@
+open Oni_Core;
+open TestFramework;
+
+module Store = Persistence.Store;
+open Persistence.Schema;
+
+describe("Persistence", ({test, _}) => {
+  let setup = () => {
+    let storeFolder =
+      Filename.get_temp_dir_name()
+      ++ "store-testXXXXXX"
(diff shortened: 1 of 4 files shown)


test/Core/PersistenceTests.re
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  test("get defaults", ({expect, _}) => {
+    let { store, testBool, _} = setup();
+    expect.bool(Store.get(testBool, store)).toBe(false);
+  });
+
+  test("empty file (regression test for #1766)", ({expect, _}) => {
+    let { storeFolder, testBool, _ } = setup();
+
```
glennsl, line 37:
```suggestion
  test("get defaults", ({expect, _}) => {
    let {store, testBool, _} = setup();
    expect.bool(Store.get(testBool, store)).toBe(false);
  });

  test("empty file (regression test for #1766)", ({expect, _}) => {
    let {storeFolder, testBool, _} = setup();

```

PR #1769, opened 2020-05-15
fix(exthost): marshalled URIs not revived correctly

glennsl:
This adds an `$mid` field to the encoded JSON of URIs, so that the exthost marshaller can properly detect it.

### DIFF
--- a/test/Exthost/UriTest.re
+++ b/test/Exthost/UriTest.re
@@ -8,7 +8,7 @@ let waitForMessage = (~name, f, context) => {
   |> Test.waitForMessage(
        ~name,
        fun
-       | Msg.MessageService(ShowMessage({message, severity})) =>
+       | Msg.MessageService(ShowMessage({message, severity, _})) =>
          f(severity, message)
        | _ => false,
      );
--- a/test/Exthost/UriTest.re
+++ b/test/Exthost/UriTest.re
@@ -1,0 +1,37 @@
+open TestFramework;
+
+open Oni_Core;
+open Exthost;
+
+let waitForMessage = (~name, f, context) => {
+  context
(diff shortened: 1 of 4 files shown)


test/Exthost/UriTest.re
```diff
@@ -0,0 +1,37 @@
...
+       | _ => false,
+     );
+};
+
+describe("UriTest", ({test, _}) => {
+  test("revive uri", _ => {
+    Test.startWithExtensions(["oni-uri"])
+    |> Test.waitForExtensionActivation("oni-uri")
+    |> Test.withClient(
+         Request.Commands.executeContributedCommand(
```
bryphe, line 22: Cool, thanks for adding the test! I wasn't sure where we were hitting this, but it is clear now.

glennsl: If you want more details of the code path:

https://github.com/onivim/vscode-exthost/blob/923c38b016c87a205957456e13c62f8dfd3bdc62/src/vs/workbench/api/common/extHostCommands.ts#L189

https://github.com/onivim/vscode-exthost/blob/923c38b016c87a205957456e13c62f8dfd3bdc62/src/vs/workbench/api/common/extHostCommands.ts#L53

https://github.com/onivim/vscode-exthost/blob/923c38b016c87a205957456e13c62f8dfd3bdc62/src/vs/base/common/marshalling.ts#L42-L45

https://github.com/onivim/vscode-exthost/blob/923c38b016c87a205957456e13c62f8dfd3bdc62/src/vs/base/common/uri.ts#L384-L395

So you see it's the marshaller that uses `$mid` for disambiguation. `URI.revive` doesn't even check it, so if something just assumes it's a URI it would have worked fine.

PR #1770, opened 2020-05-15
chore(exthost): redirect exthost I/O streams to stdin/stderr

glennsl:
Useful mostly for debugging purposes.

### DIFF
--- a/src/Store/ExtensionClient.re
+++ b/src/Store/ExtensionClient.re
@@ -410,6 +410,23 @@ let create = (~config, ~extensions, ~setup: Setup.t) => {
       ~windows_hide=true,
       ~windows_hide_console=true,
       ~windows_hide_gui=true,
+      ~redirect=[
+        Luv.Process.inherit_fd(
+          ~fd=Luv.Process.stdin,
+          ~from_parent_fd=Luv.Process.stdin,
+          (),
+        ),
+        Luv.Process.inherit_fd(
+          ~fd=Luv.Process.stdout,
+          ~from_parent_fd=Luv.Process.stderr,
+          (),
+        ),
+        Luv.Process.inherit_fd(
+          ~fd=Luv.Process.stderr,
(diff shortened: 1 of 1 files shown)


src/Store/ExtensionClient.re
```diff
@@ -410,6 +410,23 @@ let create = (~config, ~extensions, ~setup: Setup.t) => {
...
+        Luv.Process.inherit_fd(
+          ~fd=Luv.Process.stdin,
+          ~from_parent_fd=Luv.Process.stdin,
+          (),
+        ),
+        Luv.Process.inherit_fd(
+          ~fd=Luv.Process.stdout,
+          ~from_parent_fd=Luv.Process.stderr,
+          (),
+        ),
```
bryphe, line 423: This will break windows, in the case where a console isn't allocated...

What we need to do to get this logging is actually to create a new pipe that to redirect the output, and then read from that - like here: https://github.com/onivim/oni2/blob/995227ad05af144d85215c89a6cbce11d1bfc298/src/Extensions/NodeTask.re#L36

glennsl: And we need separate pipes for each stream? That's a lot fo hassle just for this. Could we instead check if it's attached and only redirect if it is?

bryphe: > And we need separate pipes for each stream?

Maybe you can use the same one? I'm not actually sure, I've never tested that.

> Could we instead check if it's attached and only redirect if it is?

That seems reasonable - checking `Timber.App.isEnabled` might be sufficient:

https://github.com/onivim/oni2/blob/4d413cfaa48d5c9b2692f9625677a24fc53e5771/src/bin_editor/Cli.re#L114

glennsl:
No worries, just forgot about it for a bit :)

Did you test this on Windows to make sure it doesn't crash?

bryphe:
@glennsl -  tested it on Windows - works great, no crash when running w/o a console 👍

PR #1772, opened 2020-05-15
refactor(editor) - Remove imperative Recalculate actions

bryphe:
We have an awkward set of `Recalculate` actions, dispatched here: https://github.com/onivim/oni2/blob/4d413cfaa48d5c9b2692f9625677a24fc53e5771/src/Store/StoreThread.re#L341 - between this and the `synchronization` effects addressed in https://github.com/onivim/oni2/pull/1627, there are a lot of tricky timing dependencies that occur along the editor pipeline.

This is an incremental refactor, to help remove that timing dependency - with the observation that we will need the `Editor` to know more about the buffer contents (for features like word wrap, etc). This replaces the 'recalculation' (which really was just calculating some metrics like max line length and number of lines for the view) with the actual event that requires it (ie, `BufferUpdate`).

To model the buffer-stuff-that-the-editor cares about, I added an `EditorBuffer.t` - this is just a subset of `Oni_Core.Buffer.t` that exposes the properties we sync (there's extra information on the core buffer, like metadata or filetype, that we don't synchronize right now - so we don't expose that.

In addition, there is a `BufferEnter` event that is overloaded (we also dispatch it for changes to the filename) - I'd really like to get rid of this event in general, as it also involves a timing dependency, but requires some further work. In the meantime, I split it out so that there is a separate action to model the filename-changed occurrence.

There's still further refactoring needed, but this is chipping away at the timing dependencies that are still blocking #1627 

__TODO:__ 
- [x] Fix splitting via a command when there is no 'path' associated with the buffer (blocking `AddRemoveSplitTest.exe`
- [x] Stability test

__Test cases:__
- [x] Initial welcome buffer shows up
- [x] Changelog can be opened via command palette
- [x] `:vsp`
- [x] `:sp` with file
- [ ] startup with single file (ie, `oni2 package.json`)
- [x] rename file with `:sav`
- [x] use `enew` to create a buffer without path, and split
- [x] use `gd` to navigate in a buffer

### DIFF
--- a/src/Model/Actions.re
+++ b/src/Model/Actions.re
@@ -24,9 +24,14 @@ type t =
   | BufferHighlights(BufferHighlights.action)
   | BufferDisableSyntaxHighlighting(int)
   | BufferEnter({
-      metadata: [@opaque] Vim.BufferMetadata.t,
+      id: int,
       fileType: option(string),
       lineEndings: [@opaque] option(Vim.lineEnding),
+      filePath: option(string),
+      modified: bool,
+      version: int,
+      // TODO: This duplication-of-truth is really awkward,
+      // but I want to remove it shortly
       buffer: [@opaque] Buffer.t,
     })
   | BufferFilenameChanged({
--- a/src/Model/Actions.re
+++ b/src/Model/Actions.re
@@ -27,6 +27,14 @@ type t =
       metadata: [@opaque] Vim.BufferMetadata.t,
       fileType: option(string),
       lineEndings: [@opaque] option(Vim.lineEnding),
+      buffer: [@opaque] Buffer.t,
+    })
+  | BufferFilenameChanged({
+      id: int,
+      newFilePath: option(string),
+      newFileType: option(string),
+      version: int,
+      modified: bool,
     })
   | BufferUpdate({
       update: [@opaque] BufferUpdate.t,
--- a/src/Model/Actions.re
+++ b/src/Model/Actions.re
@@ -28,7 +28,7 @@ type t =
       fileType: option(string),
       lineEndings: [@opaque] option(Vim.lineEnding),
       filePath: option(string),
-      modified: bool,
+      isModified: bool,
       version: int,
       // TODO: This duplication-of-truth is really awkward,
       // but I want to remove it shortly
@@ -39,7 +39,7 @@ type t =
       newFilePath: option(string),
       newFileType: option(string),
       version: int,
-      modified: bool,
+      isModified: bool,
     })
   | BufferUpdate({
       update: [@opaque] BufferUpdate.t,
--- a/src/Model/Actions.re
+++ b/src/Model/Actions.re
@@ -81,7 +81,6 @@ type t =
   | DiagnosticsSet(Uri.t, string, [@opaque] list(Diagnostic.t))
   | DiagnosticsClear(string)
   | SelectionChanged([@opaque] VisualRange.t)
-  | RecalculateEditorView([@opaque] option(Buffer.t))
   | DisableKeyDisplayer
   | EnableKeyDisplayer
   | KeyboardInput(string)
--- a/src/Store/VimStoreConnector.re
+++ b/src/Store/VimStoreConnector.re
@@ -59,7 +59,7 @@ let start =
       setClipboardText,
     ) => {
   let (stream, dispatch) = Isolinear.Stream.create();
-  let hasInitialized = ref(false);
+  let libvimHasInitialized = ref(false);
 
   let languageConfigLoader =
     Ext.LanguageConfigurationLoader.create(languageInfo);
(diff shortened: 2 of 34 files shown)


src/Model/Actions.re
```diff
@@ -27,6 +27,14 @@ type t =
...
       fileType: option(string),
       lineEndings: [@opaque] option(Vim.lineEnding),
+      buffer: [@opaque] Buffer.t,
+    })
+  | BufferFilenameChanged({
+      id: int,
+      newFilePath: option(string),
+      newFileType: option(string),
+      version: int,
+      modified: bool,
```
glennsl, line 37:
```suggestion
      isModified: bool,
```

src/Store/VimStoreConnector.re
```diff
@@ -59,6 +59,7 @@ let start =
       setClipboardText,
     ) => {
   let (stream, dispatch) = Isolinear.Stream.create();
+  let hasInitialized = ref(false);
```
glennsl, line 62:
```suggestion
  let libvimHasInitialized = ref(false);
```

It's not immediately clear what exactly this refers to. I first thought it might refer to `Actions.Init` having been processed, which would be really hacky, but still... You really can't be too specific with global mutable variables.

bryphe:
> I'd really like to see the Editor using a first-class module instead to abstract away the underlying data structures. 

I agree, I think we're on the same page here - towards the goal of generalizing the editor, I'd like to functorize it, and have the consumer provide 'concrete' implementations of the buffer, input strategy, etc. I believe this a similar abstraction to what you have in mind.

> I'll leave it up to you whether you think it would be worth looking into now. EditorBuffer is a super-awkward concept, but it's definitely not the only awkward concept in that general area, so I'm sure we can live with it temporarily.

I'd prefer to handle this later, as I'm trying to stay scoped in terms of removing the timing dependency and unblocking the multi-dispatch and window navigation. But it is certainly something we should think about as we work in this space.

glennsl:
> I'd like to functorize it, and have the consumer provide 'concrete' implementations of the buffer, input strategy, etc. I believe this a similar abstraction to what you have in mind.

Yeah, that would work much the same, but functors are a lot more invasive and boilerplatey. You'd have to turn every top-level module that needs access to these into a functor, and pass those around as well to make sure there's only one module instance everyone refers to. A first-class module on the other hand could even be introduced incrementally. It could start of as just an interface to `Buffer`, replacing `EditorBuffer` here pretty much 1-to-1. That's why I suggested it here.

Functors are useful, and sometimes necessary, to deal with the value restriction in certain scenarios involving polymorphism, because they allow turning type variables into concrete types. I don't think we need any polymorphism at all here though.

bryphe:
> Yeah, that would work much the same, but functors are a lot more invasive and boilerplatey

Ah, maybe we are thinking about different ideas - here's sort of what I was thinking in terms of 'generalizing' the editor:

```

module type Editor = {
   type t;
   ... editor methods
};

let view: (~editor: Editor.t) => React.element;

module type Buffer = {
   type t;
   let maxLineLength: t => int;
   let numberOfLines: t => int;
   let getLine: (t, int) => option(BufferLine.t);
};

type effects =
| BufferUpdate
| NavigateToBuffer
| WindowLeft
...
| NewSplit
| Close;

module type EditorImplementation = {
     module BufferImpl: Buffer;
      
    let input: (t, BufferImpl.t) => (t, list(effects)); 
};
module Make: (EditorImpl: EditorImplementation) => Editor;
```

with the idea being we could have:

```
module OnivimEditor = Editor.Make({
   module BufferImpl = Vim.Buffer; // or some wrapper

   let input = // wrapper around Vim.input to map effects
});
```

or a:

```
module GeneralEditor = Editor.Make({
   module BufferImpl = // a thin wrapper around an array or rope

  let input = (inputString) => BufferImpl.writeString(inputString);

});
```

It's kind of hand-wavy, but could share the rendering implementation, word wrap, even features like completion between editors - with it set up so that you bring-your-own input function and buffer. But sounds like you might have something different in mind, @glennsl

bryphe:
I'll bring this in so I can merge with #1627 and look at the next round of blockers - but we should keep discussing this in terms of evolution for the `Editor`. We can chat more in our meeting Mon

glennsl:
> But sounds like you might have something different in mind, @glennsl

I think it would look almost exactly like this. Just remove `Make` and pass in `EditorImplementation` as a first-class module instead. I think you'll find there's going to be a lot of boilerplate if you try to implement it as a functor, but otherwise the idea is good. And a functor is strictly more powerful than a first-class module I think, but it's going to become very hairy when you get dependencies between functor instances instead of plain modules because you need `EditorImplementation` to be accessible in those other modules. Unless you manage to factor the code to avoid it, but that would also be a major refactor.

bryphe:
> I think it would look almost exactly like this. Just remove Make and pass in EditorImplementation as a first-class module instead.

Ah, I see!

> And a functor is strictly more powerful than a first-class module I think, but it's going to become very hairy when you get dependencies between functor instances instead of plain modules because you need EditorImplementation to be accessible in those other modules. 

This is very true. Hit this with `Camomile` - we used a functorized module with our configuration, but libraries would use `CamomileLibrary.Default`, and not pick up our config...

PR #1774, opened 2020-05-15
feat(changelog): Rework and improve the changelog UI/UX

zbaylin:
Now the changelog provides much more information

image

- Commits with breaking changes are prefixed with ⚠️ 
- Commits can be expanded to view more information
  - Commit hashes are listed/clickable (this can be thought of the version where the change was added)
  - PR numbers are listed/clickable (if applicable)
  - Actual breaking changes are listed
  - If the commit message has multiple lines, the rest of the lines will be shown


glennsl:
Some initial impressions:
- Seems like only the arrow is clickable. It would be nice if the summary text was clickable as well.
- The commit hash disappears when hovered
- The PR number and arrow don't have hover states
- The view state isn't preserved if switching to another tab and back
- Mutliple breaking changes aren't distinguished very well. Could use some more spacing, or bullet points or something.
- Clicking the links causes the epiphany/Gnome Web background web process to crash continuously with the message:

```
"** (epiphany:270034): WARNING **: 22:00:50.239: Web process crashed
bwrap: Can't create file at /.flatpak-info: Read-only file system
```

Looks great though!

glennsl:
> Clicking the links causes the epiphany/Gnome Web background web process to crash continuously with the message

Seems like this is only if it's not already open. So I guess it's launched with the Oni2 executables directory as its CWD perhaps?

Not sure this is really a bug with the Revery implementation, but might be that we can fix it there anyway.

zbaylin:
> Seems like only the arrow is clickable. It would be nice if the summary text was clickable as well.
Sounds good

> The commit hash disappears when hovered
Do you mind checking this with different themes for me? I'm just using your theme's `textLink.activeForeground` color, so if that's the same as your background then it will look like it disappears.

> The view state isn't preserved if switching to another tab and back
That's true -- I guess I didnt consider that when I used local state.

> Mutliple breaking changes aren't distinguished very well. Could use some more spacing, or bullet points or something.
I can do that too.

Thanks for the feedback! 😃

src/Feature/Changelog/Feature_Changelog.re
```diff
@@ -189,17 +199,79 @@ module View = {
...
+    let breakingChangeIcon = [width(breakingChangeIconWidth)];
+
+    let caret = [width(caretWidth), marginTop(2)];
+
+    module MoreInfo = {
+      let main = [
+        marginTop(4),
+        marginLeft(
+          typWidth + scopeWidth + breakingChangeIconWidth + caretWidth,
+        ),
```
glennsl, line 212: It might be better to just put the whole expanded block in the summary column and let the layout engine do this instead of adding up all the column widths manually.

src/Store/Features.re
```diff
@@ -186,6 +186,25 @@ let update =
...
+        })
+
+      | CommitHashClicked(hash) =>
+        Isolinear.Effect.create(~name="feature.changelog.openCommit", () => {
+          let url =
+            Printf.sprintf("https://github.com/onivim/oni2/commit/%s", hash);
+          Revery.Native.Shell.openURL(url) |> ignore;
+        })
+      };
+    (state, eff);
```
glennsl, line 206: These really shouldn't live here, but in an update function in `Feature_Changelog` instead. And the effect ought to be defined in some library so it can be reused.

zbaylin: Thanks for the feedback! I can move it to an update function. The reason I didn't do that was because I thought update functions were strictly for getting new state, which this doesn't do, but I guess I was wrong about that. 🤷‍♂️ 

I thought about adding a `OpenURLExternally` `Action`, but from my discussions with you and @bryphe I thought that sounded too imperative/"active". Is this something you think we would want?

glennsl: > I thought update functions were strictly for getting new state

Ah, yeah not really. It's for processing `msg`s. But it's very rare you'd have a feature window any state, so it's certainly not a bad approximation.

> I thought about adding a OpenURLExternally Action, but from my discussions with you and @bryphe I thought that sounded too imperative/"active". Is this something you think we would want?

You don't need an action, but your intuition is indeed also correct. That would be very imperative, which is what effects are for:

```
let openUri = uri =>
        Isolinear.Effect.create(~name="openUri", () =>
          Revery.Native.Shell.openURL(uri) |> ignore
        )
```

glennsl: Also, you should never `use `ignore` without specifying the type you ignore. That's just as bad as `let _ = ...`. It's functionally equivalent.

A nice way to use ignore with a type is:

```
ignore(Revery.Native.Shell.openURL(uri): whatever);
```

glennsl: imperative APIs that return somrthing other than `unit` is also not great. In fact a nice feature of APIs that return `unit` or take =unit` as its only argument is that it makes it obvious that it produces side-effects, as if it didn't it would be pointless.

src/Feature/Changelog/Feature_Changelog.re
```diff
@@ -128,6 +133,11 @@ module View = {
   module Styles = {
     open Style;
 
+    let typWidth = 60;
```
bryphe, line 136: Minor: Should this be `typeWidth` instead of `typWidth`?

src/Feature/Changelog/Feature_Changelog.re
```diff
@@ -233,36 +305,202 @@ module View = {
...
+        | None => commit.summary
+        };
+      let onCaretClick = _e => setExpanded(exp => !exp);
+
+      <View>
+        <View style=Styles.commit>
+          <breakingChangeIcon commit theme />
+          <typ commit uiFont theme />
+          <scope commit uiFont theme />
+          <Clickable onClick=onCaretClick style=Styles.caret>
```
bryphe, line 445: I like the way you structured the UI / components! 💯

src/Feature/Changelog/Feature_Changelog.re
```diff
@@ -293,10 +539,12 @@ module View = {
   module Update = {
     module Parts = {
       let line = (~commit, ~uiFont, ~theme, ()) => {
+        let splitSummary = commit.summary |> String.split_on_char('\n');
+
         <View style=Styles.commit>
           <typ commit uiFont theme />
           <scope commit uiFont theme />
-          <summary commit uiFont theme />
+          <title text={splitSummary |> List.hd} uiFont theme />
```
bryphe, line 547: Is there any chance `List.hd` could throw here? If so - we might want to protect against it (otherwise it could manifest as a crash-on-launch)

src/Feature/Theme/GlobalColors.re
```diff
@@ -34,6 +34,20 @@ let activeContrastBorder =
...
+      {dark: hex("#3794FF"), light: hex("#006AB1"), hc: hex("#3794FF")},
+    );
+
+  let activeForeground =
+    define(
+      "textLink.activeForeground",
+      {dark: hex("#3794FF"), light: hex("#006AB1"), hc: hex("#3794FF")},
+    );
+};
+
```
bryphe, line 50: Thanks for tracking these down and adding them, Zach!

src/Feature/Changelog/Feature_Changelog.re
zbaylin: The name of the component is `typ` (I assume because `type` is a reserved keyword, this was there when I started working), so I thought I would be consistent with that. I can change it if you think it's clearer the other way!

src/Feature/Changelog/Feature_Changelog.re
zbaylin: Weird, I thought I pushed a change to fix that. One sec!

zbaylin: Oh I forgot that that appeared twice -- I fixed it the first time but not the second.

src/Feature/Changelog/Feature_Changelog.re
bryphe: Ah, got it! That makes sense to be consistent with the component 👍

src/Feature/Changelog/Feature_Changelog.re
bryphe: Awesome, thanks Zach!

zbaylin:
@glennsl I moved to a global state model, which, as you can see below, now keeps the changes open between tabs!

oni2-changelog-persistent

src/Feature/Changelog/Feature_Changelog.rei
```diff
@@ -1,19 +1,36 @@
...
   module Full: {
     let make:
       (
+        ~state: model,
         ~theme: ColorTheme.Colors.t,
         ~uiFont: UiFont.t,
         ~onPullRequestClicked: int => unit,
         ~onCommitHashClicked: string => unit,
+        ~onChangeExpanded: commit => unit,
+        ~onChangeContracted: commit => unit,
```
glennsl: You don't really need separate handlers for internal messages. Instead you can create a "scoped" dispatch function, like is done here:

https://github.com/onivim/oni2/blob/599e880bda42adeb9b69318c73ef30c77036de48/src/Store/Features.re#L237-L239

That also allows you to make the `msg` type abstract, leaking fewer internal details.

src/Feature/Changelog/Feature_Changelog.re
```diff
@@ -32,6 +43,25 @@ let isSameDate = (a, b) => {
...
 
+let update = (model, msg) =>
+  switch (msg) {
+  | PullRequestClicked(pr) =>
+    let url = Printf.sprintf("https://github.com/onivim/oni2/pull/%d", pr);
+    (model, URL(url));
+  | CommitHashClicked(hash) =>
+    let url =
+      Printf.sprintf("https://github.com/onivim/oni2/commit/%s", hash);
+    (model, URL(url));
```
glennsl, line 54: These should really just emit effects directly, but that requires moving the `openURL` effect to some place that's accessible here. Perhaps you could move both this and `stat` to a `Service_OS` project? Or maybe we should have a `Service_Revery` project for this, and a `Service_Filesystem` for `stat`?

glennsl: Also, `URL` is not a great name for a `msg`, outgoing or otherwise, as it says nothing about either what has happened or what's expected to happen. If you later have a URL that you need a different behavior for, then now you have a problem (well, ambiguous code is already a problem of course, but this demonstrates why it's ambiguous).

src/Feature/Changelog/Feature_Changelog.rei
zbaylin: I consolidated it in the latest commit. Let me know if that works!

glennsl: Yeah, that looks good! I'm wondering why you decided to cal it `dispatchMsg` instead of just `dispatch` as we use everywhere else though.

zbaylin: You're right, I just didnt want to confuse it with a more general `dispatch` that takes `Action`s rather than the `msg`s. Sorry about that!

glennsl: Right. `Action` isn't available here though, so there shouldn't be any confusion. It's an intentional part of the architecture that these components should be self-contained and have a clear boundary so that you can easily compose them. And because each component should have roughly the same structure (model/msg/update/view) the architecture is in a sense fractal. And so you want it to have the same name because it suggests its role within this component. And also, if you're going to have separate names for every level of the fractal you're going to have to become very creative ;)

zbaylin: > And also, if you're going to have separate names for every level of the fractal you're going to have to become very creative ;)

Haha very true! Thanks for the input. I changed it in the latest commit.

src/Feature/Changelog/Feature_Changelog.re
zbaylin: I changed this around in the latest commit to what I _think_ you mean. Let me know if I still need to change anything!

glennsl: Perfect. Thanks!

zbaylin:
Apart from the unrelated CI failure, does this look good to you both @bryphe @glennsl? I think I addressed all the concerns, but let me know if I missed anything!

glennsl:
Code looks great! There's a couple things I miss in terms of functionality though:
- There's no hover state on the summary, other than the cursor changing. It would be nice if the color of the text changed as well.
- The issue number isn't shown as an expanded field, when available. You can always get to that through the PR of course, but it would be nice to be able to see and go to it directly.

zbaylin:
@glennsl Not sure how I missed the issue number part! Both of those are really simple additions, I'll work on them right now. Thanks for bringing them to my attention!

zbaylin:
@glennsl I added those two improvements. Thanks again for the input. As long as the code I added is good, I think this is ready to be merged!

src/Feature/Changelog/Feature_Changelog.re
```diff
@@ -275,20 +281,36 @@ module View = {
...
+          Colors.foreground.from(theme) |> Revery.Color.multiplyAlpha(0.75),
+        ),
+      ];
+
+      let titleInactiveStyle = (font: UiFont.t, ~theme) => [
+        fontFamily(font.fontFile),
+        fontSize(12.),
+        color(Colors.foreground.from(theme)),
+      ];
+    };
```
glennsl: The `Style` suffixes here are redundant, as they're already in a module called `Styles`.

zbaylin: Very true, sorry about that! I had the tag name for the `ClickableText` elements on the brain which is why I messed that up. Thanks!

zbaylin: Fixed in the latest commit!

PR #1776, opened 2020-05-16
fix(editor/#733): Closing last window can result in no buffers being able to opened

### LINKED ISSUE #733: Closing window with unsaved changes results in no other window being created for new buffers

bryphe:
__Issue:__ There are a few ways, like #733 , where it was possible to get to a state where all buffers are closed and there is a blank screen. If that wasn't bad enough... you can't even reopen any buffers, you're just stuck on the blank screen!

__Defect:__ We'd clear out all window splits and editor groups in this case, but once they're cleared, there wasn't a way to get them back.

__Fix:__ Consolidate the 'close-editor' logic - the `editorGroups` and window layout are tightly coupled, so bring those together - we may want to think about merging these closer together. But this factors out some hard-to-follow nested reducer logic. 

Fixes #733

### DIFF
--- a/src/Store/StoreThread.re
+++ b/src/Store/StoreThread.re
@@ -66,6 +66,7 @@ let registerCommands = (~dispatch, commands) => {
 
 let start =
     (
+      ~showUpdateChangelog=true,
       ~getUserSettings,
       ~configurationFilePath=None,
       ~keybindingsFilePath=None,
@@ -114,6 +115,7 @@ let start =
   let commandUpdater = CommandStoreConnector.start();
   let (vimUpdater, vimStream) =
     VimStoreConnector.start(
+      ~showUpdateChangelog,
       languageInfo,
       getState,
       getClipboardText,
--- a/integration_test/Regression733Test.re
+++ b/integration_test/Regression733Test.re
@@ -1,14 +1,10 @@
-open Oni_Core;
 open Oni_Model;
 open Oni_IntegrationTestLib;
-
-open Feature_Editor;
 
 // Regression test case covering #733:
 // Closing all active splits causes no other splits to be opened
 runTestWithInput(
-  ~name="Regression733Test",
-  (input, dispatch, wait, runEffects) => {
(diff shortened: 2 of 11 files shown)


src/Store/StoreThread.re
```diff
@@ -66,6 +66,7 @@ let registerCommands = (~dispatch, commands) => {
 
 let start =
     (
+      ~showUpdateChangelog=true,
```
glennsl, line 69: I really don't like these dirty hacks being scattered all over the place just for the sake of testing, with very little indication of their actual purpose. Making the code more convoluted and prone to error really undermines the point of writing tests in the first place. IMO, doing it properly could be mocking out the entire store folder, or the persistence mechanism, for example.

bryphe: > IMO, doing it properly could be mocking out the entire store folder, or the persistence mechanism, for example.

@glennsl - That would be great! It would've be nice to have this in the initial persistence implementation (along with tests). Would you be up for adding this?

bryphe: Ideally we could tackle it holistically, and address the `configurationFilePath` and `keybindingsFilePath` too: https://github.com/onivim/oni2/blob/599e880bda42adeb9b69318c73ef30c77036de48/src/Store/StoreThread.re#L70 just having a way to specify a root config folder. That would help significantly.

bryphe: In the meantime, I'm going to merge this so I can unblock the next round of fixes.

PR #1782, opened 2020-05-18
chore(CI): Fix CentOS 7 installation of git2

bryphe:
CentOS 7 defaults to having `git` v1, but we need v2 for our dependencies - this updates the script to pick up git v2+


bryphe:
CentOS 7 build is passing, so I'll bring this in -

PR #1784, opened 2020-05-18
fix(cli/#1783): add back custom extension dir option

### LINKED ISSUE #1783: oni2 --extensions-dir not working / improperly documented
The [documentation](https://onivim.github.io/docs/configuration/extensions) states that

> This can be overridden with the --extensions-dir command line flag however.

I tried both

    oni2 --extensions-dir ~/.vscode/extensions

and 

    oni2 --extensions-dir=~/.vscode/extensions

None seem to work

CrossR:
Fix #1783.

Add back the plumbing for the CLI flag to specify a custom extension folder.

### DIFF
--- a/src/Store/StoreThread.re
+++ b/src/Store/StoreThread.re
@@ -86,7 +86,7 @@ let start =
       ~minimize,
       ~window: option(Revery.Window.t),
       ~filesToOpen=[],
-      ~overriddenExtensionsDir=?,
+      ~overriddenExtensionsDir=None,
       ~shouldLoadExtensions=true,
       ~shouldSyntaxHighlight=true,
       ~shouldLoadConfiguration=true,
--- a/src/bin_launcher/Oni2.re
+++ b/src/bin_launcher/Oni2.re
@@ -40,13 +40,13 @@ let spec =
     (
       "--disable-syntax-highlighting",
       passthrough,
-      "Turn off syntax highlighting.",
+      " Turn off syntax highlighting.",
     ),
-    ("--disable-extensions", passthrough, "Turn off extension loading."),
+    ("--disable-extensions", passthrough, " Turn off extension loading."),
     (
       "--disable-configuration",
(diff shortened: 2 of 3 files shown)


src/bin_launcher/Oni2.re
```diff
@@ -40,13 +40,13 @@ let spec =
     (
       "--disable-syntax-highlighting",
       passthrough,
-      "Turn off syntax highlighting.",
+      " Turn off syntax highlighting.",
```
CrossR, line 43: Whilst I was in the CI, I fixed this.

If you specify the description with a space at the start, it will be properly aligned with all the rest of the flags.

src/Store/StoreThread.re
```diff
@@ -86,7 +86,7 @@ let start =
       ~minimize,
       ~window: option(Revery.Window.t),
       ~filesToOpen=[],
-      ~overriddenExtensionsDir=?,
+      ~overriddenExtensionsDir=None,
```
CrossR, line 89: Maybe I'm being an idiot...but I thought that in this context `?` and `None` are equivalent?

However when doing `esy` I get that `cliOptions.overriddenExtensionsDir` is an `option(string)` and it is expecting a `string`....which I don't really understand. The docs here https://reasonml.github.io/docs/en/function#optional-labeled-arguments seem to suggest that it should work.

bryphe: Ah, ya, `None` vs `?` is very confusing... looks good 👍

src/bin_launcher/Oni2.re
bryphe: Thanks for cleaning this up, too 👍

bryphe:
Thanks for fixing this @CrossR , much appreciated. We're missing test coverage on our CLI parameters... so I'll add a CLI test suite so we can make sure this doesn't break in the future 👍

PR #1787, opened 2020-05-19
fix(macos/#1462): fix macOS titlebar conditional rendering

zbaylin:
Before the titlebar was hidden when the window was maximized. Now it's hidden when the window is fullscreen!

image

### DIFF
--- a/src/Store/Reducer.re
+++ b/src/Store/Reducer.re
@@ -53,8 +53,13 @@ let reduce: (State.t, Actions.t) => State.t =
       | WindowFocusGained => {...s, windowIsFocused: true}
       | WindowFocusLost => {...s, windowIsFocused: false}
       | WindowMaximized => {...s, windowIsMaximized: true}
+      | WindowFullscreen => {...s, windowIsFullscreen: true}
       | WindowRestored
-      | WindowMinimized => {...s, windowIsMaximized: false}
+      | WindowMinimized => {
+          ...s,
+          windowIsFullscreen: false,
+          windowIsMaximized: false,
+        }
       | _ => s
       };
     };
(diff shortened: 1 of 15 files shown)


src/Store/Reducer.re
```diff
@@ -53,8 +53,13 @@ let reduce: (State.t, Actions.t) => State.t =
       | WindowFocusGained => {...s, windowIsFocused: true}
       | WindowFocusLost => {...s, windowIsFocused: false}
       | WindowMaximized => {...s, windowIsMaximized: true}
```
bryphe, line 55: Is it possible to go from `fullscreen` to `maximized`? If so, the `windowIsFullScreen` value might stay `true` when the `WindowMaximized` event comes in. I couldn't make this happen, though. Maybe we should set `windowIsFullScreen: false` in the `WindowMaximized` case, though?

zbaylin: That's a great question -- I didn't consider that! I don't think it's possible, as I couldn't get it to do that either. The second point is correct though -- I'll add that now!

zbaylin: This is fixed in the latest commit! Thanks again, @bryphe!

src/Store/Reducer.re
```diff
@@ -52,14 +52,10 @@ let reduce: (State.t, Actions.t) => State.t =
       | ReallyQuitting => {...s, isQuitting: true}
       | WindowFocusGained => {...s, windowIsFocused: true}
       | WindowFocusLost => {...s, windowIsFocused: false}
-      | WindowMaximized => {...s, windowIsMaximized: true}
-      | WindowFullscreen => {...s, windowIsFullscreen: true}
+      | WindowMaximized => {...s, windowDisplayMode: Maximized}
```
bryphe, line 55: Thanks for moving from the correlated-bools to a variant, this seems much easier to work with than trying to understand the relationship between `isMaximized` and `isFullscreen` - now those states are mutually exclusive 👍

Issue #1085, opened 2019-12-19
MacOS: Using the OS's mechanism for opening a file


raguay:
I'm trying to set Oni2 as my main editor for other programs to launch as well based on extension. For example, opening a file from Alfred using Onivim2. Onivim2 will open, but without the file.

On the command line, it would be the same as `open -a /Applications/Onivim2.app ~/.zshrc`. The OS opens the program and tells it to open the file. But, Onivim2 is ignoring the request for opening the file. This is particular to macOS.

I can create Alfred actions to do it properly on the command line for me, but in the long run Onivim2 should work with the OS and it's methods of launching the file as well.

leematos:
I attempted this today, and used --args to do so. The behavior varied:

- With a file:

 open -a /Applications/Onivim2.app --args /Users/lbot/.bash_profile

it renders the file but no sidebar:

<img width="838" alt="image" src="https://user-images.githubusercontent.com/359140/75122804-730fdf00-566f-11ea-8956-358d799f9721.png">

If I pass in a folder, it renders the full app (ignore the messed up onilogo, was testing the logos on my branch:

<img width="828" alt="image" src="https://user-images.githubusercontent.com/359140/75122823-963a8e80-566f-11ea-91f0-0280ff644a5f.png">

raguay:
Try to use "Open with..." from Finder (or most other Finder type programs).
It doesn't open the file, just launches Oni2. Also, set Oni2 as your
default editor for a file extension and try to open that file by double
click. Again, it just opens the program. Oni2 should support these methods
of opening a file if it is to be a main stream editor.

raguay:
Just a note of thanks. It is working great for me now. Thank you so much.

Issue #552, opened 2019-08-02
Packaging - OSX: First launch of packaged app fails, requiring a second open


bryphe:
Even after the gatekeeper issues are addressed in #510 - there is one other issue that hurts the first run experience. After clicking 'open', you can see the app icon bounce in the dock... but the app doesn't open. It requires double-clicking or opening again to actually launch.

bryphe:
Adding some logging for this case:

```
LOG: Executable name: /private/var/folders/6g/bw9ys08145dgtk_cbzf8pwgw0000gn/T/AppTranslocation/BD8871E4-878E-447B-A8C7-B28746330460/d/Onivim2.App/Contents/MacOS/Oni2
LOG:  arg: /private/var/folders/6g/bw9ys08145dgtk_cbzf8pwgw0000gn/T/AppTranslocation/BD8871E4-878E-447B-A8C7-B28746330460/d/Onivim2.App/Contents/MacOS/Oni2
LOG:  arg: -psn_0_2282029
```

The first run adds this extra argument `-psn_X_XXXXXXX`, which causes us to fail:
```
/Users/bryphe/oni2/_esy/default/store/i/oni2-e8573589/bin/Oni2: unknown option '-psn_0_2253350'.
Usage: 
  -help  Display this list of options
  --help  Display this list of options
```

PR #1791, opened 2020-05-20
fix(macOS/#1085): allow opening of files through system protocol

zbaylin:
As of right now, if you try to open a file in Oni on macOS either through "Open With" in Finder or with the `open` command in Terminal.app, Oni activates but doesn't open the file passed in.

This does two main things:
- adds file associations to the `Info.plist` so that we're shown when trying to "Open With" a compatible file:
image

- binds to Revery's new `App.onFileOpen` event system to trigger opens within Oni2 itself

It also changes the executable in the `Info.plist` from `Oni2` to `Oni2_editor` -- this is necessary because otherwise the file information gets lost between the processes. From my conversations with @bryphe, this seems to be okay, I just have to make sure that Gatekeeper doesn't interfere with CLI flags.

Addresses #1085. Thank you @raguay for bringing this to our attention!

### DIFF
--- a/package.json
+++ b/package.json
@@ -238,7 +238,7 @@
   "resolutions": {
     "@esy-ocaml/libffi": "onivim/libffi#590b041",
     "@opam/yojson": "onivim/yojson:yojson.opam#f480aef",
-    "revery": "revery-ui/revery#4f0f851",
+    "revery": "revery-ui/revery#326f2df",
     "editor-core-types": "onivim/editor-core-types#6a8afaf",
     "esy-cmake": "prometheansacrifice/esy-cmake#2a47392def755",
     "esy-skia": "revery-ui/esy-skia#d60e5fe",
--- a/package.json
+++ b/package.json
@@ -238,7 +238,7 @@
   "resolutions": {
     "@esy-ocaml/libffi": "onivim/libffi#590b041",
     "@opam/yojson": "onivim/yojson:yojson.opam#f480aef",
-    "revery": "revery-ui/revery#f9fedcbe",
+    "revery": "revery-ui/revery#4f0f851",
     "editor-core-types": "onivim/editor-core-types#6a8afaf",
     "esy-cmake": "prometheansacrifice/esy-cmake#2a47392def755",
     "esy-skia": "revery-ui/esy-skia#d60e5fe",
--- a/package.json
+++ b/package.json
@@ -19,6 +19,70 @@
     "productName": "Onivim2",
     "fileAssociations": [
       {
+        "name": "Plain text",
+        "role": "Editor",
+        "ext": [
+          ".txt",
+          ".README",
+          ".LICENSE"
+        ]
+      },
+      {
+        "name": "YAML document",
+        "role": "Editor",
+        "ext": [
+          ".yml",
+          ".yaml"
+        ]
+      },
+      {
+        "name": "Git configuration",
+        "role": "Editor",
+        "ext": [
+          ".gitignore",
+          ".gitconfig",
+          ".gitattributes"
+        ]
+      },
+      {
+        "name": "Docker configuration",
+        "role": "Editor",
+        "ext": [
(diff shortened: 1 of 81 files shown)


package.json
```diff
@@ -18,6 +18,70 @@
...
+        ]
+      },
+      {
+        "name": "ReasonML source",
+        "role": "Editor",
+        "ext": [
+          ".re",
+          ".rei"
+        ]
+      },
```
bryphe, line 76: Thanks for adding these additional file associations!

scripts/release.js
```diff
@@ -123,8 +123,15 @@ if (process.platform == "linux") {
...
-      CFBundleExecutable: "Oni2",
+      CFBundleExecutable: "Oni2_editor",
       NSHighResolutionCapable: true,
+      CFBundleDocumentTypes: package.build.fileAssociations.map(fileAssoc => {
+            return {
+                CFBundleTypeExtensions: fileAssoc.ext.map(ext => ext.substr(1)),
+                CFBundleTypeName: fileAssoc.name,
+                CFBundleTypeRole: fileAssoc.role
+            }
+        })
```
bryphe, line 134: nice! 👍

scripts/release.js
```diff
@@ -123,8 +123,15 @@ if (process.platform == "linux") {
       CFBundleVersion: `${package.version}`,
       CFBundlePackageType: "APPL",
       CFBundleSignature: "????",
-      CFBundleExecutable: "Oni2",
+      CFBundleExecutable: "Oni2_editor",
```
bryphe, line 126: This looks good, but we'll need to port over this CLI logic:
https://github.com/onivim/oni2/blob/8c600e062746eb647dd8270cbc770de2365f51dc/src/bin_launcher/Oni2.re#L97

zbaylin: I noted that in the original PR comment 👍. I was looking at `Cli.re`, but it seems to me that we dont pass in `Sys.argv` like we do in `Oni2`. Not sure what the best solution for that is, unfortunately.

bryphe:
Regarding the command-line-flag that needs to be ported over, it can be a bit tricky to test! That's because, as far as I know, there isn't a very reliable way to set the 'gatekeeper' quarantine flag, reliably.

So the steps I use to test this are:

Setup: Install `npm install -g http-server` to have it around

1) Create the `_release/Onivim2.app` folder via `esy create-release`
2) Run `./scripts/osx/package.sh` to create a `tar.gz` in the `publish` folder
3) `cd _publish`
4) `http-server .` <-- start a local http server in the `publish` folder
5) Go to a web browser, and navigate to `http://127.0.0.1:8080/Onivim2-<commitid>-tar.gz`
6) The tar.gz should be downloaded
7) Extract the `tar.gz`, and run Onivim 2. When you download a file - even if its locally - the gatekeeper quarantine bit gets set!
8) Because the app isn't code signed, you'll need to go to `Security & Privacy` and allow it to open
9) Try opening again - this time, you'll get a warning, but you'll be able to Open

__Expected:__ Application should start successfully
__Actual:__ You can see icon bounce in the dock, but it fails to start. Starting it again works!

The reason it's failing to start is because its being passed that extra `-psn` parameter that isn't handled. Some details on when we hit this before: https://github.com/onivim/oni2/issues/552

scripts/release.js
bryphe: Ah, ya, we need to use a different call - `Arg.argv_parse` - that lets us pass in a list of arguments (`Argv.parse` is like `Arg.argv_parse` with `Sys.argv` passed in). 

I made the change here: https://github.com/onivim/oni2/pull/1795 (and wanted to add some test coverage on our CLI anyway, so this was a good opportunity)

bryphe:
Looks like this needs an `esy format`

zbaylin:
@bryphe I somehow missed that! Thanks! Will fix now.

Issue #441, opened 2019-07-15
Scrollbar/minimap: click and drag doesn't work (Linux: X11/Xwayland)


szbergeron:
Platform: linux, oni running in xwayland on Sway with libinput

Action: click on minimap or scrollbar handle, move cursor up or down with long file
Expected result: viewport scrolls
Actual result: dragging with mouse held down produces no result, sometimes handle sticks to mouse on cursor release and moving cursor without clicking drags viewport around wildly and unpredictably

bryphe:
This should be fixed now in latest `master` - feel free to open a new issue if you're still seeing issues around this 👍

PR #1792, opened 2020-05-20
fix(ux): mouse capture issues

### LINKED ISSUE #817: Display gets crazy after left clicking preview and moving mouse.
preview

When I left click on preview and move mouse weird thing happens. With each mouse move the document is scrolling, even if I release left mouse button.
It does not stop if I left click in some different area.
Ok, it stops when I left click in preview again. Is it desirable behavior?

glennsl:
This brings in a new and improved hook-based mouse capture API which should resolve several cases where the mouse capture gets "stuck". It was also supposed to fix the wonky minimap mouse dragging (#441), but no luck figuring that out yet. This is on par with the current behaviour though, so can still be merged as-is to unblock other tasks.

Should fix #817. ~~Might address #250. These depend a bit on interpretation though.~~

### DIFF
--- a/bench.esy.lock/index.json
+++ b/bench.esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "48e3daa64897c89dc48413a6c57c0eb3",
+  "checksum": "3fd3e130b7d6f99d979244b0ee0f4722",
   "root": "Oni2@link-dev:./package.json",
   "node": {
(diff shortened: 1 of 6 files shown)


bryphe:
> This is on par with the current behaviour though, so can still be merged as-is to unblock other tasks.

Sure, sounds good 👍

glennsl:
Could you also verify if you think this fixes #817 and #250 @bryphe?

bryphe:
#817 looks good to me; I'll keep #441/#250 around though to track the remaining wonkiness (it's not clear if they are the same issue, or if #441 is tracking something Linux specific - so would be good to keep them both just so I remember to test the linux paths called out)

PR #1800, opened 2020-05-22
feat(workspace/#463): split resizing by mouse

### LINKED ISSUE #463: No way to resize splits

glennsl:
mouse-resizing

No need to look too closely at the code, at least the `Feature_Layout` implementation, as I'm planning on refactoring this to use pixel coordinates instead of weight since I've eventually realized that this is an inherent part of the problem domain.

There's also a bit of wonkiness with the handle not properly tracking the mouse once it hits a limit. This is because the mouse coordinates from the events are relative to the window, while the handle and splits are positioned relative to their positional container, so the resizing is done through relative change, which works fine until it is constrained and the handle remains in place while the mouse continues to move. The behaviour is predictable though, and might be easier (or just different) to fix after refactoring, so I think it's fine for now.

Fixes #463

### DIFF
--- a/src/UI/EditorLayoutView.re
+++ b/src/UI/EditorLayoutView.re
@@ -17,12 +17,6 @@ module Styles = {
 
   let container = [flexGrow(1), flexDirection(`Row)];
 
-  let split = direction => [
-    flexDirection(direction == `Vertical ? `Row : `Column),
-    backgroundColor(direction == `Vertical ? Colors.red : Colors.blue),
-    flexGrow(1),
-  ];
-
   let verticalHandle = (node: Feature_Layout.sized(_)) => [
     cursor(MouseCursors.horizontalResize),
     position(`Absolute),
--- a/src/UI/EditorLayoutView.re
+++ b/src/UI/EditorLayoutView.re
@@ -8,6 +8,10 @@ open Revery;
 open UI;
 open Oni_Model;
 
+module Constants = {
+  let handleSize = 10;
+};
+
 module Styles = {
   open Style;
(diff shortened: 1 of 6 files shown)


bryphe:
> No need to look too closely at the code, at least the Feature_Layout implementation, as I'm planning on refactoring this to use pixel coordinates instead of weight since I've eventually realized that this is an inherent part of the problem domain.

Seems reasonable to me. With pixel coordinates, we'll have to adjust on resize when the window size changes (which happens for 'free' with weights) - but it makes some parts of the problem easier (like, minimum sizes). 

> The behaviour is predictable though, and might be easier (or just different) to fix after refactoring, so I think it's fine for now.

Thanks for calling this out. I agree, this is already a huge step forward compared to what we have today. Appreciate all your work to implement this!

src/UI/EditorLayoutView.re
```diff
@@ -8,63 +8,147 @@ open Revery;
...
+    left(node.x),
+    top(node.y + node.height - Constants.handleSize / 2),
+    width(node.width),
+    height(Constants.handleSize),
+  ];
+};
+
+let%component handleView =
+              (~direction, ~node: Feature_Layout.sized(_), ~onDrag, ()) => {
+  let%hook (captureMouse, _state) =
```
bryphe, line 41: Nice use of the new `mouseCapture` hook 😄

PR #1801, opened 2020-05-22
feat(Windows): custom titlebar

zbaylin:
This adds a custom titlebar on Windows, similar to the one on macOS.

image


src/UI/Titlebar.re
```diff
@@ -7,57 +7,252 @@ open Revery.UI;
...
+          GlobalContext.current().dispatch(Model.Actions.WindowCloseClicked);
+
+        <View
+          style={
+            isHovered
+              ? Styles.Windows.Button.hoverClose(theme)
+              : Styles.Windows.Button.container
+          }
+          onMouseUp
+          onMouseEnter={_ => setHovered(_ => true)}
```
bryphe, line 148: I like that you added the hover styles, awesome!

src/UI/WindowResizers.re
```diff
@@ -0,0 +1,98 @@
...
+  Since we have the frame disabled, we have to write our own
+  behavior for this.
+*/
+open Revery;
+open Revery.UI;
+
+module Styles = {
+  open Style;
+
+  let size = 5;
```
bryphe, line 12: Thanks for factoring this to a constant 👍

src/UI/WindowResizers.re
```diff
@@ -0,0 +1,98 @@
...
+    backgroundColor(Revery.Colors.blue),
+  ];
+
+  let topRight = [
+    position(`Absolute),
+    Style.right(0),
+    Style.top(0),
+    width(size),
+    height(size),
+    backgroundColor(Revery.Colors.pink),
```
bryphe, line 74: We'll have to remove these colors prior to merging 😄

src/UI/WindowResizers.re
```diff
@@ -0,0 +1,98 @@
...
+  <View style=Styles.left mouseBehavior=Revery.UI.ResizeLeft />,
+  <View style=Styles.bottomLeft mouseBehavior=Revery.UI.ResizeBottomLeft />,
+  <View style=Styles.bottom mouseBehavior=Revery.UI.ResizeBottom />,
+  <View style=Styles.bottomRight mouseBehavior=Revery.UI.ResizeBottomRight />,
+  <View style=Styles.right mouseBehavior=Revery.UI.ResizeRight />,
+  <View style=Styles.topRight mouseBehavior=Revery.UI.ResizeTopRight />,
+  <View style=Styles.top mouseBehavior=Revery.UI.ResizeTop />,
+];
+
+let make = () => resizers |> React.listToElement;
```
bryphe, line 98: Thanks for moving this to a component!

src/UI/Titlebar.re
zbaylin: Is this default behavior on your system? On stock apps on my VM the cursor is just the standard pointer. My VM has a bunch of weird quirks though so it's entirely possible it's just mine.

bryphe: You're right, I double-checked on my non-VM-Windows machine, and actually this would differ from the default behavior - thanks for checking! So we don't need to add cursor styles.

CrossR:
Just two comments: It looks like the logo could do with a teeny bit of padding when maximised. 

Also the image is pretty low res, though I don't think there is much you can do about that @zbaylin, so maybe a thing to just note down for now since there is a few places we need to bump the resolution of the logos, which I think @bryphe has mentioned before, but its pretty low priority for now.

image

bryphe:
Looks like the integration tests might need to be updated with the new methods:
```
File "integration_test/lib/Oni_IntegrationTestLib.re", lines 155-177, characters 4-5:
155 | ....Store.StoreThread.start(
156 |       ~showUpdateChangelog=false,
157 |       ~getUserSettings,
158 |       ~setup,
159 |       ~onAfterDispatch,
...
174 |       ~window=None,
175 |       ~filesToOpen,
176 |       (),
177 |     ).
Error: This expression has type
         close:(unit -> unit) ->
         restore:(unit -> unit) ->
         (Store.StoreThread.Model.Actions.t -> unit) * (unit -> unit)
       but an expression was expected of type 'a * 'b
```

(You can test this locally with `esy '@integrationtest' install` & `esy '@integrationtest' build`)

zbaylin:
Awesome, will fix that as soon as my Windows VM finishes building the new dependencies haha 🕐. I also think I saw 1 or 2 `esy b dune build @check` errors that I should resolve too. Overall though Im super happy with how this turned out!

zbaylin:
I think I fixed all the `esy b dune build @check` errors (which include the integration test errors) -- CI will let me know if I messed anything up haha.

bryphe:
Ah, looks like it needs an `esy format` too to pass the hygiene check - but other builds look good!

bryphe:
> Also the image is pretty low res, though I don't think there is much you can do about that @zbaylin, so maybe a thing to just note down for now since there is a few places we need to bump the resolution of the logos, which I think @bryphe has mentioned before, but its pretty low priority for now.

Yes, good point! I need to get the icon tweaked a bit to have more solid outline for low-res - it doesn't look right scaled down

CrossR:
I've noticed a few regression after this unfortunately (I apparently didn't test this as thoroughly as I thought I had).

 - The Window dimensions are off for me, I expect as part of the many scaling issues we've had.
 - The maximise behaviour seems to go too far, and fill the whole screen, rather than just up to the toolbar.
 - There is a slight dead zone in the tool bar where I can't drag, but that could be related to the window scaling issue.

The scaling/window size issue is probably causing #1817 too.

CrossR:
Also, here is an video of what I'm seeing: https://drive.google.com/open?id=1vMrR31pi7HJ-TJKXREOjDctLAZ0JMYGw

(Window handles in the wrong place, then not being able to move the window, then the taskbar being covered on maximise)

PR #1807, opened 2020-05-22
feat(macOS): custom document icons

zbaylin:
This is a small yet (I think) visually pleasing/enhancing addition: custom document icons for macOS.

They all follow a similar look to this one:
reasonml

The main change (obviously) is the name of the language/filetype at the bottom.

Here is what it looks like in Finder:
image


bryphe:
Very cool @zbaylin - didn't even think about this! I really like the icons you created.

These only show up when Onivim 2 is set as the 'default' editor for a file, right? I don't want to take over a user's Finder without them consenting - but I'm a big fan of these 😄

zbaylin:
Yes, it only ever shows when the user has Oni2 as their default editor (or when Oni2 is the only app which can open the file, but that's also technically "default"). You can try it yourself too if you want to verify that, but it's at least the case for me 😃. I'm glad you like the icons -- to me they're simple but effective 👍

Issue #1166, opened 2020-01-04
macOS cmd-tab icon looks non-retina


jaredly:
<img width="444" alt="Screen Shot 2020-01-04 at 4 06 16 PM" src="https://user-images.githubusercontent.com/112170/71772684-63480a00-2f0c-11ea-9d22-2dfc744fb29b.png">

this is on master, running with `esy run`

leematos:
I dug into this one, and it looks like the logo is being rendered based on the logo.png:

`
[DEBUG]    +0ms Revery.Core.Window : Loading icon from: /Users/lbot/Development/onivim/oni2/_release/./Onivim2.app/Contents/MacOS/logo.png`

This is where it happens in revery: https://github.com/revery-ui/revery/blob/2f6f985927dcdf362c2ac1d915c7f752278485dd/src/Core/Window.re#L428

The logo.png file is currently 64x64, so that explains that.

The outstanding questions are:

- Can/should we be using the icon set here on Mac?
- If not, is there commonplace wisdom on cross-platform app icon sizing?

We can make logo.png 1024x1024, but I don't know what kind of effects that would have on windows/linux.

I think this classifies as a revery bug, as the framework should handle that abstraction imo.

mr-sour:
![image](https://user-images.githubusercontent.com/2181180/82620800-08533e80-9b9f-11ea-9e7e-6d196511acd5.png)
I assume this is the same bug but it looks low res if you have large dock icons to. I vote to just up it to 256x256 and let the operating system scale it down if need be. It looks like linux and mac-os can support pgn icons upto 512x512 but windows becomes has the lowest resolution at 256x256. A more ideal solution would be assets for each build using icns ico and pgn with max support size but for a cross platform pgn it looks like 256 is the best bet.

PR #1813, opened 2020-05-23
fix(macOS): change app icon to be more inline with system design

zbaylin:
@bryphe and I were talking about how the current Oni2 logo on macOS doesn't really fit the current design language of the OS, specifically with the filled in background.

This PR adds a couple things:
- an SVG of the raw Oni2 icon, which can be used to generate any resolution bitmap
- an SVG of the Oni2 icon with an outer glow to be used as the icon on macOS
- an ICNS of the aforementioned macOS SVG
  - this was generated from the SVG using [Image2Icon](http://www.img2icnsapp.com/), which generates all the required resolutions in the macOS spec for icons

This is purely just a shot at what I think improves the look a little bit, but I am 100% open to feedback/input. I'm certainly no graphic designer!

Here's what the new icon looks like:
Onivim2

The outer glow color is the same as the dark-blue/violet background in the full logo.

Here's what it looks like in Finder/other contexts:
Dark mode:
Screen Shot 2020-05-23 at 11 38 56 AM
Light mode:
Screen Shot 2020-05-23 at 11 38 39 AM

For comparison, here's how the old icon looks relative to other macOS apps:
image

Please don't hold back your critiques, haha!


bryphe:
Looks great, @zbaylin ! Thank you so much for improving this.

I know one piece of UI we had particular issues with was in the `cmd`+`tab` UI: https://github.com/onivim/oni2/issues/1166

Does this fix that piece of UI, too? I think I'm still seeing the old icon, but not sure if it's maybe picking up a different version of the app.

zbaylin:
That part is set by SDL2. I can bump up the resolution of the icon.png using that SVG and see if that fixes it!

bryphe:
Oh, interesting - didn't realize that! I guess that `SDL_SetWindowIcon` method?

>  I can bump up the resolution of the icon.png using that SVG and see if that fixes it!

Sounds great!

zbaylin:
@bryphe I actually came up with a better idea. If SDL doesn't set the app icon, it just uses the one of the bundle (`Onivim2.icns`). This is definitely preferable because ICNS files have multiple correctly scaled versions of the icon for each resolution. So I added a flag to `Info.plist` to set the environment variable `ONI2_BUNDLED` to `1` when running in .app mode. In `Oni2_editor`, if this value is 1 it doesn't set the icon through SDL. Kinda hard to show in a screenshot, but it's pretty obvious when you check it out yourself!

src/bin_editor/Oni2_editor.re
```diff
@@ -120,14 +120,28 @@ let createWindow = (~forceScaleFactor, ~workingDirectory, app) => {
     | _ => true
     };
 
+  let icon =
+    switch (Revery.Environment.os) {
+    | Mac =>
+      switch (Sys.getenv_opt("ONI2_BUNDLED")) {
```
bryphe, line 126: Cool idea! I didn't realize you could set environment variables via the `plist`

src/bin_editor/Oni2_editor.re
```diff
@@ -120,14 +120,28 @@ let createWindow = (~forceScaleFactor, ~workingDirectory, app) => {
...
+      | Some(bundled) =>
+        switch (bundled) {
+        | "1" => None
+        | _ => Some("logo.png")
+        }
```
bryphe, line 131: I think we could simplify this - `Some(_) => None` (whenever the `ONI2_BUNDLED` env variable is set at all, just use `None`)

scripts/release.js
```diff
@@ -133,7 +133,10 @@ if (process.platform == "linux") {
                 CFBundleTypeRole: fileAssoc.role,
                 CFBundleTypeIconFile: "macDocumentIcons/" + fileAssoc.icon.mac
             }
-        })
+        }),
+      LSEnvironment: {
```
bryphe, line 137: Neat! Didn't know this was an option

bryphe:
> @bryphe I actually came up with a better idea. If SDL doesn't set the app icon, it just uses the one of the bundle (Onivim2.icns). This is definitely preferable because ICNS files have multiple correctly scaled versions of the icon for each resolution.

That makes sense - great idea 👍 This will be better at handling things like retina displays for sure.

bryphe:
Looks so much better! Fixes #1166 👍

src/bin_editor/Oni2_editor.re
zbaylin: Awesome I just changed that in the most recent commit. Thanks!

PR #1822, opened 2020-05-26
test: add ppx_inline_test

glennsl:
This adds `ppx_inline_test` as a dependency, wires it up to be run on CI and during development, and adds a few inline tests to exercise it in order to see if it fails on any of our CI boxes.

The developer experience with this is actually amazing. Try running `esy @test watch:inline` and add a few tests. That feedback loop is _tight_!.


package.json
```diff
@@ -368,44 +337,41 @@
...
+    "@opam/fs": "*",
+    "@opam/logs": "^0.7.0",
+    "@opam/luv": "~0.5.1",
     "@opam/lwt": "*",
+    "@opam/markup": "^0.8.2",
     "@opam/ocamlbuild": "*",
     "@opam/ppx_deriving": "*",
     "@opam/ppx_deriving_yojson": "*",
+    "@opam/ppx_inline_test": "v0.13.1",
     "@opam/ppx_let": "v0.13.0",
```
glennsl, line 351: Sorry about the mess. I used `esy add` for this, which went to town on the formatting. I actually had no idea that sub-command existed, but felt it should so I tried, and voila. Much more convenient than adding dependencies manually, so I guess this is the formatting we ought to try to maintain.

bryphe:
> The developer experience with this is actually amazing. Try running esy @test watch:inline and add a few tests. That feedback loop is tight!.

😎 very cool! It's a great developer experience

package.json
bryphe: Neat, I forgot about that command too. Maintain this formatting sounds good.

test.json
```diff
@@ -1,13 +1,15 @@
...
-    "run-ci": "esy '@test' x OniUnitTestRunnerCI",
+    "run": "esy '@test' x OniUnitTestRunner && esy @test inline",
+    "run-ci": "esy '@test' x OniUnitTestRunnerCI && esy @test inline",
     "core": "esy '@test' x OniUnitTestRunnerCore",
     "cli": "esy '@test' x OniUnitTestRunnerCLI",
     "model": "esy '@test' x OniUnitTestRunnerModel",
     "exthost": "esy '@test' x OniUnitTestRunnerExtHost",
-    "watch": "esy '@test' dune build --watch --root . -j4"
+    "watch": "esy '@test' dune build --watch --root . -j4",
+    "inline": "esy dune runtest",
```
bryphe, line 11: Just tested on Windows - unfortunately this results in an 'assembler error'.

We can fix it by running `esy b dune runtest` (running in the build environment - on Windows, the MingW toolchain, which is needed for linking by dune, is only available in the build environment. Otherwise, dune fails trying to link to run the test). Seems to still work on OSX too.

test.json
```diff
@@ -1,13 +1,15 @@
 {
   "source": "./package.json",
   "scripts": {
-    "run": "esy '@test' x OniUnitTestRunner",
-    "run-ci": "esy '@test' x OniUnitTestRunnerCI",
+    "run": "esy '@test' x OniUnitTestRunner && esy @test inline",
```
bryphe, line 4: Unfortunately the `&&` doesn't seem to be handled the same on Windows - it doesn't actually execute `esy @test inline` (so the `esy '@test' run` call passes, when it should fail due to the failing test case).

For CI - we could add an `esy @test inline` to our build steps, to make sure they get run on Windows: https://github.com/onivim/oni2/blob/c5849ea07e56625445f0151b483f58fd556b2c6e/.ci/esy-build-steps.yml#L26

test.json
glennsl: Thanks for the heads up! Fixed in https://github.com/onivim/oni2/pull/1822/commits/f8f37945749b97cdaf10395d72bbb42900401dce

test.json
glennsl: Ah, sorry, seems this is a shortcoming of esy in general actually. For the CI this was just temporary to quickly test it, so no big deal there, but it would have been nice to have a single command to run tests. Not sure how to do that now...

bryphe: > but it would have been nice to have a single command to run tests. 

No worries, maybe there is a way to create an executable to run it (or run it as part of `OniUnitTestRunnerCI`: https://github.com/janestreet/ppx_inline_test#building-and-running-the-tests-outside-of-jane-street) - but not sure if it's worth going to that level.

glennsl: Yeah, if you're cool with just doing `esy @test inline && esy @test run` locally. Should be easier to remember once we get more inline tests, which I expect we will.

glennsl: @Et7f3 claims https://github.com/onivim/oni2/pull/1822/commits/9861370cac07525cb17b22481a6ed11c34e6e620 will work on Windows because of esy-bash. Can you test @bryphe?

Et7f3: That's what is done in revery esy.install

bryphe: Unfortunately doesn't seem to work - I just get: 
```
error: project is not installed, run `esy install`
```
on Windows.

I think the way `esy` runs scripts in the `package.json` is unfortunately different than the way `esy` runs the `install`/`build` commands.

Et7f3: You have to first install the test sandbox and the default sandbox I think. It is the same than `esy b cmd` it fail if sandbox isn't installed.

bryphe: Hmm, good point @Et7f3 - seems I didn't have the default sandbox installed.

However, after running `esy install`, I'm getting this now with `esy @test run`:
```
/cygdrive/c/Users/bryphe/AppData/Roaming/npm/esy: line 1: /node_modules/esy/_build/default/bin/esy.exe: No such file or directory
```

glennsl: `esy @test inline` runs `esy b dune runtest`, so could just be that too. Runnning `esy '@test ' b dune runtest` instead shouldn't be a problem though.

Et7f3: Esy doesn't seem to add itself in the path...
Try:
"bash -c \"dune runtest && dune exec ./test/OniUnitTestRunner\""

Et7f3: GitHub removed my backslash 🤨

glennsl: `dune runtest` needs to be run in the build environment though, so maybe `"esy b bash -c \"dune runtest && dune exec ./test/OniUnitTestRunner\""` then?

Et7f3: Yup it seem better.

glennsl: Hmm, the inline tests work fine then, but the rely tests fails:

```
Running 1 test suite   
 RUNS  CLIRunning executable: /home/glennsl/dev/remote/glennsl/oni2/Oni2
--version
Got output: ||
Status: WEXITED: 127
Expected output to contain: Onivim 2
But received: 
Running executable: /home/glennsl/dev/remote/glennsl/oni2/Oni2
--list-extensions
--extensions-dir=/home/glennsl/dev/remote/glennsl/oni2/development_extensions
Got output: ||
Status: WEXITED: 127
Expected output to contain: dev-extension
But received: 
Running executable: /home/glennsl/dev/remote/glennsl/oni2/Oni2_editor
-psn_0_989382
--list-extensions
Got output: ||
Status: WEXITED: 127
Expected status to be: WEXITED: 0
But received: WEXITED: 127
 FAIL  CLI
  • CLI › launcher › --version should show output

    Exception (Failure "Output incorrect")

  • CLI › launcher › --list-extensions should show extensions

    Exception (Failure "Output incorrect")

  • CLI › editor › rogue -psn argument shouldn't cause a failure

    Exception (Failure "Exit code incorrect")


Test Suites: 1 failed, 0 passed, 1 total
Tests:       3 failed, 0 passed, 3 total
Time:        0.007s
```

bryphe: Hmm, looks like something strange with the working directory in the `dune exec` case - this path can't be right: `/home/glennsl/dev/remote/glennsl/oni2/Oni2`. 

`dune exec` must be changing how we find the executing directory for the test cases here: https://github.com/onivim/oni2/blob/b537fac3f612f621378900d53df3d45335f04a72/test/Cli/TestRunner.re#L16

Path of least resistance might be to just have the two separate commands for now

glennsl: Yeah, did that for now. They really need to be run in different environments it seems.

Issue #1817, opened 2020-05-24
On Windows, window edges extend outside of the screen when maximised


Phaen:
When maximising the Oni2 window on Windows, all four edges of the window extend outside of the screen and are not rendered. The following images demonstrate the issue:

This is the top-left of a regular floating window. This is expected behaviour, nothing out of the usual.
![Top-left of regular floating window](https://user-images.githubusercontent.com/5429189/82751317-9c4b2300-9db6-11ea-9065-04873275a8ff.png)

When maximising the window, you can see that both edges are slightly cut off. Besides the lack of spacing, this is apparent from the cut logo and the disappearing indicators for the file explorer and search.
![Maximised window (both edges are cut off)](https://user-images.githubusercontent.com/5429189/82751340-c866a400-9db6-11ea-90fe-2a4070f8158e.png)

While lazy resizing to a regular window, it becomes more apparent that the edges were not rendered. (Window isn't re-rendered during a lazy resize, filling the unrendered edges with content of the old render.)
![While resizing to regular window](https://user-images.githubusercontent.com/5429189/82751390-17acd480-9db7-11ea-8442-73f081c6f438.png)

CrossR:
I've stuck a comment on #1801 with some issues I've seen, I think that is related.

zbaylin:
Thanks @Phaen for bringing this to our attention! I think this happens because we render _over_ the default frame, and when the window is maximized, the frameshape changes, which we (well, more accurately SDL2) doesn't adjust for. I'll mess around with this either today or early this week to come up with a fix!

bryphe:
This is fixed now with #1823 - thanks @zbaylin !

PR #1823, opened 2020-05-26
fix(Windows): remove overdraw when maximized

zbaylin:
This is WIP as of now, because it fixes the overdraw for me, but I'm not sure it will be fixed for everyone. This addresses #1817 and the concerns brought up by @CrossR in #1801.

@CrossR and @bryphe, can you test this on your machines to make sure it's a universal fix? I want to make sure it works before it gets merged.

### DIFF
--- a/bench.esy.lock/index.json
+++ b/bench.esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "696484669c273e4ff36e8f859e0e7350",
+  "checksum": "a0f6964dd4e6f5632edba7d438578e4d",
   "root": "Oni2@link-dev:./package.json",
   "node": {
     "yarn-pkg-config@github:esy-ocaml/yarn-pkg-config#db3a0b63883606dd57c54a7158d560d6cba8cd79@d41d8cd9": {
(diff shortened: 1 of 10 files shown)


zbaylin:
I forgot to mention I tested it with different levels of scaling and it seems to work, but since we had scaling issues over the weekend I want to be sure.

zbaylin:
I also made revery-ui/revery#862, which will fix the taskbar disappearing when maximized 😃

CrossR:
This seems to work for me! I tested this with revery-ui/revery#72ee949 and they both work together nicely. Has fixed the overdraw issue for me, as well as my taskbar issue (3840x2160 at 150% scaling, and tested some of the niche things like moving taskbar size etc)

Its also a tiny thing, but perhaps the drag handles shouldn't be visible when maximised? I don't think its actually an issue, its just one of those tiny things. I've no idea why it is the case you can't resize when maximised...but I know its how every windows app works.

Other than that, great work!

CrossR:
Actually, just in case this is relevant to the revery PR (not sure where the bug lies) but something between the saving and loading of window sizes is going a bit strange. If I maximise oni2 and then close, when I reopen, it says its maximised (from the icon) but its not quite maximised.

zbaylin:
That's a great point @CrossR! That should be an easy fix. Thanks!

zbaylin:
@CrossR I dont think that's related to this/the revery PR -- I was having that issue off and on before as well. Definitely something I can investigate though!

zbaylin:
@CrossR I just pushed a commit that doesn't "render" the `WindowResizers` when the Window is maximized. Thanks again for pointing that out!

vhanla:
Actually this fix is just shifting the margin ignoring the real issue with Revery which I mentioned here revery-ui/revery#1001 because the actual maximized window size is bigger than the monitor resolution.



It invades space of other monitors and taskbar area too.  
  
I also provided a temporary workaround using SetWindowsHookEx to WH\_CALLWNDPROCRECT that almost fixes the maximization under work area rect vhanla/revery@d19b9d1  
  
However, using hooks would be the last thing to do, but I couldn't find the right place where I can manipulated main window WNDPROC for handling windows messages.  
  
Anyhow, you can check if that workaround works in your Windows installation too, by just replacing in `package.json` **revery** resolution to my commit above mentioned.  
 

package.json  
`"resolutions" : {`  
 `"revery": "vhanla/revery#d19b9d1",`

In the following screen capture I set the UI/Root.re 's \[margin(**0**) otherwise that margin shifting is noticeable.



As I said (in the issue reported at revery's repository page), it doesn't completely fixes that issue, since now there is a 1 pixel gap in the bottom of the window, even on a different DPI scaling, fortunately it doesn't invade space of other monitors neither taskbar area now.  
  
BTW: I also included WS\_SYSMENU and WS\_MINIMIZEBOX to make it responsive to minimize and restore when clicking the taskbar icon.  
  
Hopefully you will fix that issue. Kind regards.

PR #1824, opened 2020-05-26
refactor(libvim): Bring in reason-libvim

bryphe:
This brings in `reason-libvim` into the main source tree here (ie, as a monorepo), to reduce the friction when making cross-cutting changes between `reason-libvim` and Onivim 2.


glennsl:
After this I keep having issues with `test/reason-libvim/test.crlf` being modified and marked dirty by git, despite it being listed in `.gitatrtibutes`. I've tried working around it by running `git update-index --assume-unchanged` on it, but that still gives me issues if I try to check out another branch and such.

CrossR:
I'm having the same issue on both of my machines....I can't really see why, the file says it should have `crlf` in the `.gitattributes` and it does have that. I don't fully know what is being auto converted, but its a pain that it throws up issues when swapping branches.

I downloaded a zip of the repo and checked that to see if that would rule out my git doing something odd on checkout...but that was the exact same.

bryphe:
Strange.. I'm seeing this too. I'll take a look tomorrow and see if I can narrow down what's causing it. Worst case - we already have a CRLF file checked into the repo, and I can adjust the tests to use the one we already have.

PR #1827, opened 2020-05-27
fix(configuration/#1826): initial value of editor.insertSpaces deviates from default

### LINKED ISSUE #1826: Wrong default value for editor.insertSpaces
According to the [documentation](https://github.com/onivim/oni2/blob/1f1ceb227d26c5656cb0ce8b375d17d6d116a847/docs/docs/configuration/settings.md#L55) the default value for `editor.insertSpaces` is `true`. However calling `Preferences: Open Configuration File` for the first time will show a configuration file with `editor.insertSpaces` set to `false`.

glennsl:
Fixes #1826. The default value itself was fixed in #1580


glennsl:
We really ought to generate the initial configuration file from the defaults, to avoid issues like this.

bryphe:
Thanks for catching and fixing this, @glennsl !

PR #1833, opened 2020-05-27
refactor: layout feature - part 1

glennsl:
This is a major overhaul of the Layout feature to make it nicely encapsulated and clean up the internals to be ready for upcoming features such as global window keyboard navigation and vim-style "layout tabs".

Some highlights:
- Moved EditorLayoutView into the feature
- Moved layout commands into the feature as contributions
- Replaced the libvim `onMovement` hook with proper keybindings
- Consolidated the movement logic
- Moved tests inline to avoid exposing internals in the public API (and also simplified them quite a lot, so they're easier to deal with when refactoring)
- Separated the generic data structures from the specific feature model

Future refactoring work:
- Bring EditorGroup, Tabs etc. into the feature
- There's still a bit of nasty logic with `ViewCloseEditor` in `WindowsStore` [here](https://github.com/onivim/oni2/blob/1567361d1126b1818eec9dc5feb1c1af17edbeaf/src/Store/WindowsStoreConnector.re#L43-L67), but cleaning that up requires bringing `EditorGroup` into `Feature_Layout` I think.


src/Feature/Layout/Layout.re
```diff
@@ -0,0 +1,407 @@
...
+       actual == window(1);
+     };
+
+     let%test "parent split changes direction if needed" = {
+       let actual =
+         hsplit([window(2), window(1)]) |> addWindow(`Vertical, 3);
+
+       // TODO: Bug
+       actual == hsplit([window(3), window(2), window(1)]);
+       //hsplit([window(1), vsplit([window(3), window(2)])]);
```
glennsl, line 79: This existing bug was uncovered when splitting the tests for `addWindow` and `insertWindow` (which used to be combined in `addWindow` with the `~target` argument) so the same test cases are covered for both. Very unlikely to hit this, but should be fixed regardless.

src/Feature/Layout/Layout.re
```diff
@@ -0,0 +1,407 @@
...
+          children,
+        ),
+      )
+    | `Window(_) => node
+    }
+  };
+};
+
+/**
+ * resetWeight
```
glennsl, line 405:
```suggestion
 * resetWeights
```

src/Store/Features.re
```diff
@@ -143,6 +143,29 @@ let update =
...
+          editorGroups:
+            EditorGroups.setActiveEditorGroup(
+              editorGroupId,
+              state.editorGroups,
+            ),
+        }
+      | Nothing => state
+      };
+    (state, Effect.none);
+
```
glennsl, line 168: The active editor group is now managed by the layout feature, but still owned at root level and just passed into the update and view. We could probably move it into the `Focus/FocusManager` mechanism too.

src/Store/KeyBindingsStoreConnector.re
```diff
@@ -338,6 +338,37 @@ let start = maybeKeyBindingsFilePath => {
...
+      {
+        key: "<C-W>R",
+        command: Feature_Layout.Commands.rotateForward.id,
+        condition: WhenExpr.Value(True),
+      },
+      {
+        key: "<C-W><S-R>",
+        command: Feature_Layout.Commands.rotateBackward.id,
+        condition: WhenExpr.Value(True),
+      },
```
glennsl, line 371: This keybinding doesn't work, because of one of the bugs uncovered with the `editor-input` I think.

src/Feature/Layout/Layout.re
glennsl: Fixed in https://github.com/onivim/oni2/pull/1833/commits/1563e39cee15dfe503c7637e3d5ab2ea203a2bbb

src/Store/KeyBindingsStoreConnector.re
```diff
@@ -338,6 +338,62 @@ let start = maybeKeyBindingsFilePath => {
...
+      {
+        key: "<C-W><S-,>", // TODO: Does not work and should be `<`, but blocked by bugs in editor-input,
+        command: Feature_Layout.Commands.increaseHorizontalSize.id,
+        condition: WhenExpr.Value(True),
+      },
+      {
+        key: "<C-W><S-.>", // TODO: Does not work and should be `>`, but blocked by bugs in editor-input
+        command: Feature_Layout.Commands.decreaseHorizontalSize.id,
+        condition: WhenExpr.Value(True),
+      },
```
glennsl, line 391: Several of these keybindings do not work due to bugs in `editor-input`, but have still been added and marked as such for future reference.

bryphe:
> This is a major overhaul of the Layout feature to make it nicely encapsulated and clean up the internals to be ready for upcoming features such as global window keyboard navigation and vim-style "layout tabs".

Exciting! Thanks for this work, @glennsl

src/Store/KeyBindingsStoreConnector.re
bryphe: Ah, thanks for calling them out and marking them here 👍

src/Feature/Layout/AbstractTree.re
```diff
@@ -0,0 +1,119 @@
...
+      };
+
+    {...node, kind: `Split((dir, children))};
+
+  | `Window(_) => node
+  };
+};
+
+let%test_module "rotate" =
+  (module
```
bryphe, line 74: Thanks for the lengths you went to be able to add these cases - they look great!

src/Feature/Layout/Feature_Layout.re
```diff
@@ -1,458 +1,438 @@
...
+module Commands = {
+  open Feature_Commands.Schema;
+
+  let rotateForward =
+    define(
+      ~category="View",
+      ~title="Rotate Windows (Forwards)",
+      "view.rotateForward",
+      Command(RotateForward),
+    );
```
bryphe, line 322: Thanks for adding these commands 👍

src/Store/CommandStoreConnector.re
```diff
@@ -48,22 +48,6 @@ let start = () => {
...
-             Feature_Layout.move(direction, group.editorGroupId, state.layout)
-           );
-
-      switch (maybeEditorGroupId) {
-      | Some(editorGroupId) =>
-        dispatch(Actions.EditorGroupSelected(editorGroupId))
-      | None => ()
-      };
-    });
-  };
```
bryphe, line 65: Thanks for the clean-up you did here!

bryphe:
> There's still a bit of nasty logic with ViewCloseEditor in WindowsStore here, but cleaning that up requires bringing EditorGroup into Feature_Layout I think.

Yes, this makes sense... they are tightly coupled right now (ie, removing a group requires removing / cleaning up splits). I wasn't sure if this should be a separate feature (ie, a `Feature_EditorGroups` that depends on `Feature_Layout`, but parameterizes it), or if it should be part of `Feature_Layout`. 

This PR makes it clearer, though, since Feature_Layout itself is less general (with `AbstractTree` being the general part) - so seems like we'd want to bring the `EditorGroup` concept in here

PR #1848, opened 2020-05-29
chore(windows): Investigate syntax health failure

bryphe:
The healthcheck for windows, for the syntax server, has been failing - an exit code of 2 intermittently comes from the syntax service, but just for the windows builds.

An exit code of 2 is usually indicative of an ocaml exception, so we should be dropping a crash log in that case - this updates the health check to look for a crash log and output the results to give us a clue on where it is crashing.


bryphe:
Of course... the syntax service doesn't fail this time. But this should give us more info when it does fail the healthcheck.

PR #1853, opened 2020-05-30
feat(macOS): add DMG icon

zbaylin:
Another small visual detail. These are fun to create!

This just adds a custom DMG icon for the packaged release of Onivim2 on macOS. The icon looks like this:
dmg-icon

This is what it looks like mounted:
Screen Shot 2020-05-29 at 9 54 42 PM


bryphe:
Woah, very cool - great touch @zbaylin ! Thank you!

PR #1859, opened 2020-05-30
fix(textmate/#1825): Crash when parsing Nim syntax

### LINKED ISSUE #1825: Nim VsCode extension crashing oni2
I installed the [Nim language extension](https://github.com/pragmagic/vscode-nim) by navigating to the oni2 extensions directory and git cloning the repo. Oni starts fine, but if I try to open a .nim file, it crashes immediately. 

The crash log is very short: `Yojson.Safe.Util.Type_error("Can't get member 'name' of non-object type string", _):`

Please let me know if more details are needed. Thanks!

bryphe:
__Issue:__ The nim syntax extension causes Onivim 2 to crash

__Defect:__ There are a couple issues:
- First, failing to parse the syntax file shouldn't cause a crash
- Second, we're not handling a capture group construction:

We expect capture groups to look like:
```
{"1":{"name":"meta.preprocessor.pragma.nim"}}
```
(be nested in a JSON object with a `name` property).

However, there are some capture groups that have a simplified syntax:
```
{"0":"punctuation.definition.string.end.nim"}
```

which our JSON parser wasn't handling.

__Fix:__ Update our capture group parsing to handle the streamlined form.

Fixes #1825

### DIFF
--- a/src/Core/dune
+++ b/src/Core/dune
@@ -16,7 +16,6 @@
         yojson
         ppx_deriving.runtime
         ppx_deriving_yojson.runtime
-        textmate
         editor-core-types
         timber
         ReasonFuzz
--- a/src/reason-textmate/Pattern.re
+++ b/src/reason-textmate/Pattern.re
@@ -77,6 +77,22 @@ module Json = {
       };
     };
 
+  module Decode = {
+    open Oni_Core;
+    open Json.Decode;
+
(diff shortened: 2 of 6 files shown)


bryphe:
Did a simple check here:

image

At least, we have some basic highlighting now

PR #1875, opened 2020-06-05
refactor(explorer): new arrow icon, v alignments

lessp:
This is a minor one, but:
- Switches to use the foreground-color `SideBar` provided by the theme
- Switches icon to the `angle`-one
- Adjusts some margins in order to align things vertically (see `CODE_OF_CONDUCT.md`)
- Increases the font-size by 1pt 

**Before:**
Screenshot 2020-06-05 at 14 35 01
**After:**
Screenshot 2020-06-05 at 14 32 33

### DIFF
--- a/src/UI/FileTreeView.re
+++ b/src/UI/FileTreeView.re
@@ -29,6 +29,9 @@ module Styles = {
     height(Constants.tabHeight),
   ];
 
+  // Minor adjustment to align with text
+  let folder = [marginTop(4)];
+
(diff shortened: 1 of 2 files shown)


lessp:
> Nice! Does this make it look closer to vscode with the same themes? I can't build and check myself right now unfortunately.

Yep, at least a little:

image

bryphe:
Thanks for the improvements here, @lessp !

PR #1876, opened 2020-06-05
chore: fix gcc 10 build issues

glennsl:
This upgrades revery to the latest commit in order to make it build with GCC 10. And as part of that also upgrades to the new font family API.


src/Core/UiFont.re
```diff
@@ -1,37 +1,17 @@
...
-  fontFileSemiBoldItalic,
-  fontSize,
+let default = {
+  normal: Revery.Font.Family.fromFile("Inter-UI-Regular.ttf"),
+  bold: Revery.Font.Family.fromFile("Inter-UI-Bold.ttf"),
+  italic: Revery.Font.Family.fromFile("Inter-UI-Italic.ttf"),
+  semiBold: Revery.Font.Family.fromFile("Inter-UI-Medium.ttf"),
+  semiBoldItalic: Revery.Font.Family.fromFile("Inter-UI-MediumItalic.ttf"),
+  size: 12.,
 };
```
glennsl, line 17: Just loading all the variants separately seem to work fine. Not entirely sure how to combine these into a single family.

src/Service/Font/Service_Font.re
```diff
@@ -6,6 +6,7 @@ module Log = (val Log.withNamespace("Oni2.Service.Font"));
 [@deriving show({with_path: false})]
 type font = {
   fontFile: string,
+  fontFamily: [@opaque] Revery.Font.Family.t,
   fontSize: float,
   measuredWidth: float,
   measuredHeight: float,
```
glennsl, line 12: Added `fontFamily` here because this is also used with `<Text>` in some places.

bench/lib/Helpers.re
```diff
@@ -25,6 +25,10 @@ let simpleState = {
 
 let defaultFont: Service_Font.font = {
   fontFile: Revery.Environment.executingDirectory ++ "FiraCode-Regular.ttf",
+  fontFamily:
+    Revery.Font.Family.fromFile(
+      Revery.Environment.executingDirectory ++ "FiraCode-Regular.ttf",
+    ),
```
zbaylin, line 31: I actually don't think you need to prepend the executing directory. I originally had that function named `Family.asset` but @bryphe (rightfully) recommended I change it to something a little more obvious that it could be from any file.

The file string eventually gets passed here:
https://github.com/revery-ui/revery/blob/c4c6c23f97a0278e1877d37306bdb89e45bb439e/src/Font/FontCache.re#L66

Which, as you can see, calculates the path as an asset anyway. I can check if you want, but it might be easier for you since you already have it built. Let me know!

src/Components/Input.re
```diff
@@ -107,6 +107,8 @@ let%component make =
               (
                 ~theme,
                 ~style=Styles.default(~theme),
+                ~fontSize=18.,
+                ~fontFamily=Revery.Font.Family.fromFile("Roboto.Regular.ttf"),
```
zbaylin, line 111: Is `Roboto.Regular.ttf` a typo? I see `Roboto-Regular.ttf` elsewhere.

glennsl:
The only problem I've encountered so far is a bit of weirdness with the completions UI:

2020-06-05-225002_237x86_scrot

Haven't been able to figure out the cause unfortunately.

src/Components/Input.re
```diff
@@ -183,8 +181,9 @@ let%component make =
     let dimensions =
       Revery_Draw.Text.measure(
         ~smoothing=Revery.Font.Smoothing.default,
-        ~fontFamily=Styles.fontFamily,
-        ~fontSize=Styles.fontSize,
+        ~fontFamily=
+          Revery.Font.Family.toPath(fontFamily, Normal, false, false),
```
zbaylin, line 185: Unrelated to this but I should change this function to take the `FontFamily.t` with the weight, italicized, and monospaced as separate arguments.

src/Core/UiFont.re
zbaylin: The `fromFile` function takes an optional variant resolver. You can do something like:

```ocaml
let family = Font.Family.fromFile(~variant=(~weight, ~italicized, ~monospaced as _) =>
  switch ((weight, italicized)) {
  | (Weight.Bold, false) => "Inter-UI-Bold.ttf"
  ...
  | _ => "Inter-UI-Regular.ttf"
  }, "Inter-UI-Regular.ttf")
```

bench/lib/Helpers.re
glennsl: Yeah I wondered about that too, as it's not done anywhere else. Hard to verify here though since this is in the benchmark, although I suppose it doesn't really matter as long as some font is loaded.

src/Components/Input.re
glennsl: Indeed it is. Good catch!

src/Components/Input.re
glennsl: Yeah, that's almost the only complaint I have with the API. That and the unlabelled boolean arguments, which isn't very intuitive.

src/Core/UiFont.re
zbaylin: I have a suspicion changing this may also fix that weird line bug you were seeing

bryphe:
Thanks for all the work merging this up, @glennsl !

Hmm, ya, looking to the `Completions` issue - I see this:

image

Whereas before this change, it was like this:
image

It seems like somehow we're picking up two different fonts after this change for the `<HighlightText />` used by the completions view... and it's not obvious to me why.

src/Components/HighlightText.re
```diff
@@ -5,6 +5,8 @@ let make =
       ~text: string,
       ~highlights: list((int, int)),
       ~style,
+      ~fontFamily=?,
+      ~fontSize=?,
       ~highlightStyle,
       (),
     ) => {
```
bryphe, line 12: There's a case here that's missed - for the last element - we need to also set `?fontFamily` and `?fontSize` on this text node, too

bryphe:
Ah, found the issue with the completions - a case is missing where we need to add `?fontFamily` and `?fontSize`:
```
  let highlighted = {
    let textLength = String.length(text);

    // Assumes ranges are sorted low to high
    let rec highlighter = last =>
      fun
      | [] => [
          <Text style text={String.sub(text, last, textLength - last)} />,
        ]

      | [(low, high), ...rest] => [
          <Text
            style
            ?fontFamily
            ?fontSize
            text={String.sub(text, last, low - last)}
          />,
          <Text
            style=highlightStyle
            ?fontFamily
            ?fontSize
            text={String.sub(text, low, high + 1 - low)}
          />,
          ...highlighter(high + 1, rest),
        ];

    highlighter(0, highlights) |> React.listToElement;
  };
```

This case, for the last text run, is missing the fontFamily setting - so it falls back to the default font:
```
      | [] => [
          <Text style text={String.sub(text, last, textLength - last)} />,
        ]
```

bryphe:
I noticed one other place where the fontFamily wasn't being set - the Quickmenu. Fixed both the HighlightText and Quickmenu issues in https://github.com/onivim/oni2/pull/1876/commits/6e5c9457a527fbbbbae3a415c6473b7571bf9ac2

bryphe:
Unfortunately ASAN is also picking up a false memory leak from `libfontconfig.so`, from the change to bring in a default font family:
```
=================================================================
==36825==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 256 byte(s) in 1 object(s) allocated from:
    #0 0x7f32f5e03602 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x98602)
    #1 0x7f32f5b450b9  (/usr/lib/x86_64-linux-gnu/libfontconfig.so.1+0x1d0b9)

Indirect leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7f32f5e0379a in __interceptor_calloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x9879a)
    #1 0x7f32f5b457c8  (/usr/lib/x86_64-linux-gnu/libfontconfig.so.1+0x1d7c8)

SUMMARY: AddressSanitizer: 288 byte(s) leaked in 2 allocation(s).
```
These look like process-global memory that is allocated by `libfontconfig` that only gets cleanewd up when the process is closed. I'm looking into a way to suppress libfontconfig leaks, and still run ASAN elsewhere.

src/Core/UiFont.re
glennsl: Oh, neat! That API is a bit confusing though, with the default case specified outside, and one of the solver cases therefore ignored. A separate `fromFiles` function would be better I think. I might try to tackle that and the few other minor API issues I've noticed before I do this.

Issue #1397, opened 2020-03-02
Just bought onivim... macOS 10.15, app does not boot.


safaiyeh:
<img width="588" alt="Screen Shot 2020-03-02 at 8 36 37 AM" src="https://user-images.githubusercontent.com/8675043/75696749-ef528580-5c60-11ea-88b7-ab2d95f304a1.png">

bryphe:
Hi @safaiyeh ,

Sorry about the crash on launch! 

Some questions:
- I'm assuming you installed with the `dmg`?
- How are you launching the app? Via Finder?
- Do you see a window pop-up at all and disappear? Or just nothing at all? 

We could try getting some log output - if you run this command and share the output:

`/Applications/Onivim2.app/Contents/MacOS/Oni2 -f --trace --debug` 

might give us clues about where things are going wrong.

safaiyeh:
• Yes I installed with the `dmg`.
• I tried via Finder & Spotlight Search.
• Nothing at all, usually the icon bounces on the dock indicating it is opening. No bounce.

Logs:
```
/Applications/Onivim2.app/Contents/MacOS/Oni2 -f --trace --debug
dyld: Library not loaded: @executable_path/../Frameworks/libffi.6.dylib
  Referenced from: /Applications/Onivim2.app/Contents/MacOS/Oni2_editor
  Reason: no suitable image found.  Did find:
	/Applications/Onivim2.app/Contents/MacOS/../Frameworks/libffi.6.dylib: open() failed with errno=13
	/Applications/Onivim2.app/Contents/MacOS/../Frameworks/libffi.6.dylib: stat() failed with errno=1
	/Applications/Onivim2.app/Contents/MacOS/../Frameworks/libffi.6.dylib: open() failed with errno=13
	/Applications/Onivim2.app/Contents/MacOS/../Frameworks/libffi.6.dylib: stat() failed with errno=13
	file system relative paths not allowed in hardened programs
```

bryphe:
Thanks for the details and the log output, @safaiyeh ! Really appreciate it.

Doing some investigation... seems like a similar issue to these:
- https://forums.developer.apple.com/thread/115451
- https://forums.developer.apple.com/message/365393#365393

And in particular... the discussion in this issue looks like RStudio hit something similar: https://github.com/rstudio/rstudio/issues/4597

Seems like the new hardened-runtime is causing us to fail when loading our `libffi.6.dylib`, because we're using an `@executable_path`-relative path to load it (like RStudio).

[...]

PR #1898, opened 2020-06-07
refactor(ux): combine ui fonts into a single family

glennsl:
Also updates to the latest version of revery, with a breaking change for the `Image` component to support remote images.

### DIFF
--- a/bench.esy.lock/index.json
+++ b/bench.esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "50523377bb7d08b6c3315ab45f94de88",
+  "checksum": "57c776521ed53d05646cd96a4bcb9ee8",
   "root": "Oni2@link-dev:./package.json",
   "node": {
     "yarn-pkg-config@github:esy-ocaml/yarn-pkg-config#db3a0b63883606dd57c54a7158d560d6cba8cd79@d41d8cd9": {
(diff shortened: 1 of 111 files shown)


glennsl:
`create-release` fails on mac with:

```
Found a dynamic library: ["libcrypto.1.1.dylib","libssl.1.1.dylib"]
There should be only static libraries to successfully package.
```

glennsl:
I've tried to build `esy-openssl` using `-static` on macOS (see https://github.com/reason-native-web/vendor/compare/master...glennsl:esy-openssl/build/static), but unfortunately get this error:

    ld: library not found for -lcrt0.o

Googling that error led me to [this SO post](https://stackoverflow.com/a/3801032) which claims this is not possible on macOS.

Not sure how to proceed here other than to revise the strategy of trying to statically link everything on macOS.

bryphe:
Thanks for making this change, @glennsl ! Much more streamlined 🎉 

> Not sure how to proceed here other than to revise the strategy of trying to statically link everything on macOS.

I'll investigate this today. The main concern is hitting issues like https://github.com/onivim/oni2/issues/1397 - crashes on boot for _some_ users.

 I think we have a good understanding now (for some reason, `libffi` was being flagged and quarantined by AV) - so we can potentially bring in `libcrypto` and `libssl` as dependencies. I'd like to at least test and verify we don't hit the same issue, though - ideally run the AV software on CI so we can ensure that our packaged builds won't be falsely flagged or hit a quarantine issue.

bryphe:
FYI - I split out the upgrade / codesigning changes here: https://github.com/onivim/oni2/pull/1900

bryphe:
I've hooked up the code-signing of those libraries, and validated that it at least isn't hitting the issue in #1397 - so should be unblocked now.

PR #1901, opened 2020-06-08
feat(editor): add Hover support w/ Feature project

zbaylin:
⚠️ Very WIP ⚠️ 

This PR will add hover support when it's done. As of right now, it's still a draft. I'll log my progress as I go

- [x] Create `Feature_Hover`
- [x] Add basic commands
- [x] Add `Markdown` component
- [x] Register commands with command palette
- [x] Integrate with extension host
- [ ] Figure out correct display model 🤔 
- To be continued...


src/Feature/Hover/Feature_Hover.re
```diff
@@ -0,0 +1,28 @@
...
+  | Hide;
+
+module Commands = {
+  open Feature_Commands.Schema;
+
+  let show =
+    define(
+      ~category="Hover",
+      ~title="Show hover panel",
+      "hover.show",
```
bryphe, line 17: For the command IDs, we should try and keep these in sync with VSCode - it's important for an extension compatibility, because some extensions might try and show hover. 

It looks like the equivalent is `"editor.action.showHover"`: https://github.com/microsoft/vscode/blob/eff2fb75151dac67b6835584b6e91808ec1379ae/src/vs/editor/contrib/hover/hover.ts#L234

src/Feature/Hover/Feature_Hover.re
```diff
@@ -0,0 +1,28 @@
+/* Feature_Hover.re
+     This feature project contains logic related to Hover
+   */
+
+[@deriving show({with_path: false})]
+type command =
```
bryphe, line 6: Looks like a good start!

src/Feature/Hover/Feature_Hover.re
zbaylin: Oh okay, I didn't even realize those were based in VSCode command ID's 😳. Thanks for the tip!

zbaylin: Fixed in the latest commit! It also looks like they dont have a hide. From what I can tell they hide when a non-modifier key is pressed. I got rid of the `Hide` command too.

zbaylin:
Well I got something working:

oni2_hoverproto1

Not sure if you want to call it "hover" because it doesn't hover over the code nor is it triggered by a mouse hover, but the "hover" information from VSCode shows up! The rest is just UX, which is exciting

bryphe:
Nice - I'm even getting hovers for Reason/OCaml with this!
image

zbaylin:
@bryphe That's _awesome_! If you try it now you should see it render underneath the highlighted token 😎

bryphe:
Nice @zbaylin ! This will be very useful for me personally - I often what to interogate the types that are inferred, and didn't have the ability to do that without hover 😄 

For example:
image

Looks like we even render proper markdown:
image

bryphe:
Some notes / ideas:

1) If you open a hover, move, and then open a second hover - it briefly shows the first hover, at the first position, prior to opening. Maybe something we could adjust in the state - ie, not show until we have the latest?

2) A subtle box shadow on the hover might look nice, and help decouple it from the UI.

3) In the case where there is non-code text, like:
image
The non-code text looks a bit large and out of sync with the UI text. It also is spilling out of the hover - maybe an overflow issue.

4) In cases where it's very large, perhaps we should have a max-height and add a scrollbar?

5) If the editor is scrolled, and we try and get hover on a bottom element, it won't be visible - perhaps we could put it above the cursor / highlight in that case?

Minor feedback though, this is looking awesome. I really like the rounded corners and the detail you've put into it so far! It's going to be so nice to have this (finally!). Great work!

zbaylin:
Thanks @bryphe! I'm working on getting the styling to look a bit better. I'm getting some weird overflow/wrap issues but I think that they can all be fixed by messing around with the flex params. revery-ui/revery#907 should also help things look a little prettier 😆. I'm very excited for this too for the exact reason you mentioned -- being able to see the type info is a game changer.

Glad I could make your workflow any easier 👍.

zbaylin:
>     1. If you open a hover, move, and then open a second hover - it briefly shows the first hover, at the first position, prior to opening. Maybe something we could adjust in the state - ie, not show until we have the latest?

I can actually fix this by clearing the contents of the hover on hide -- I don't think that would mess with anything negatively.

src/Feature/Hover/Feature_Hover.re
```diff
@@ -125,11 +125,11 @@ module Styles = {
   open Style;
   module Colors = Feature_Theme.Colors;
 
-  let outer = (~x, ~y) => [
+  let outer = (~x, ~y, ~theme) => [
     position(`Absolute),
     left(x),
     top(y),
-    border(~width=1, ~color=Revery.Colors.black),
+    border(~width=1, ~color=Colors.EditorHoverWidget.border.from(theme)),
```
bryphe, line 132: Nice, thanks for picking this up!

bryphe:
So nice to have the mouse hover hooked up now, too! Nice work @zbaylin 

I tested it out and noticed a few issues:

The first hover I tried was the `Styles.container` in `EditorSurface.re` - it showed up incorrectly (note the `=` overlapping the `list`), and it seems like there should be a scrollbar, but there wasn't one:

image

Then, I added a text with an error - and it showed up the previous hover (instead of the error message, as I expected), but seemingly properly formatted:

image

zbaylin:
Hmm that first one almost looks like it would be a bug in the Markdown system in Revery. I hadn't seen that, so thanks for bringing it to my attention.

That second one can be fixed by just clearing the contents/range buffer on close. Sorry about that.

src/Components/Markdown.re
```diff
@@ -0,0 +1,130 @@
...
+       let syntaxHighlighter =
+           (
+             ~tokenTheme,
+             ~colorTheme,
+             ~languageInfo,
+             ~grammars,
+             ~language,
+             lines,
+           ) => {
+         let grammarRepository =
```
bryphe, line 23: Ideally, we shouldn't need to recreate the `grammarRepository` every time - this means we'll have to reload grammars every time we render. We have a common instance of `grammarRepository` right now - can we use that here?

src/Feature/Editor/Editor.re
```diff
@@ -250,6 +250,26 @@ let scrollDeltaPixelY = (~pixelY, view) => {
   scrollToPixelY(~pixelY, view);
 };
 
+let getCursorOffset = (~buffer, ~cursorPosition: EditorCoreTypes.Location.t) => {
```
bryphe, line 253: Thanks for moving this to the editor!

src/Components/Markdown.re
zbaylin: Oof that was my bad! I moved it to the state but didn't change it in the component :facepalm. Luckily it's an easy fix!

src/Feature/Editor/SurfaceView.re
```diff
@@ -73,24 +81,74 @@ let%component make =
...
+       });
+    hoverTimerActive := true;
+    lastMousePosition := Some((evt.mouseX, evt.mouseY));
+    resetHoverTimer();
+  };
+
+  let%hook () =
+    Hooks.effect(
+      If(
+        (t, _) => Revery.Time.toFloatSeconds(t) > Constants.hoverTime,
```
bryphe, line 122: This seems a bit unclear to me - why are we ignoring the second argument for it?

src/Feature/Editor/SurfaceView.re
```diff
@@ -73,24 +81,74 @@ let%component make =
...
+        hoverTimer,
+      ),
+      () => {
+        lastMousePosition^
+        |> Option.iter(((mouseX, mouseY)) => {
+             getMaybeLocationFromMousePosition(mouseX, mouseY)
+             |> Option.iter(((line, col)) => {
```
bryphe, line 129: This seems like it could be simplified:
```
lastMousePosition^
|> OptionEx.flatMap(((mouseX, mouseY)) => getMaybeLocationFromMousePosition(mouseX, mouseY))
|> Option.iter(((line, col)) => ...)
```
We can remove these nested iters by using `flatMap`/`bind`.

src/Feature/Editor/SurfaceView.re
zbaylin: Ah yeah I should probably add a comment about that. Usually the `If` hook compares `x` to a previous version of itself. However I want to compare it to a constant value. Theoretically I can just move it to an isolated if statement, but I think the hook provides clarity that it's an effect and also runtime benefits, if I'm not mistaken.

src/Store/InputStoreConnector.re
```diff
@@ -46,7 +46,10 @@ let start = (window: option(Revery.Window.t), runEffects) => {
   let handleTextEffect = (state: State.t, k: string) => {
     switch (Model.FocusManager.current(state)) {
     | Editor
-    | Wildmenu => [Actions.KeyboardInput(k)]
+    | Wildmenu => [
+        Actions.KeyboardInput(k),
+        Actions.Hover(Feature_Hover.KeyPressed(k)),
```
bryphe, line 51: Just curious - why do we have to add this specifically for the wildmenu?

zbaylin: Oh that's interesting. I guess my formatting was bad and it lumped it in there when I ran `esy format`. You're right that it shouldn't be 👍

src/Components/Markdown.re
zbaylin: Actually I'm realizing that here it's creating a Textmate grammar repository from an Oni_Syntax one. We carry around the Oni_Syntax one which gets passed here, should I also add a Textmate one to state? I can also make it stateful, so it doesnt re-create it every render.

src/Store/InputStoreConnector.re
zbaylin: Ah it's not specifically for the wild menu. I just added it in where the editor case was. I can split those if you think it makes sense.

src/Feature/Editor/SurfaceView.re
bryphe: Makes sense to keep it an effect, thanks for the clarification! A comment would be helpful 👍

src/Store/InputStoreConnector.re
bryphe: Ah, I see, I missed that the `Editor` was in the same handler

src/Components/Markdown.re
bryphe: Ah, got it - so it's using our common grammar repository (taking advantage of the cache) - just wrapping it. That should be fine!

bryphe:
Just tested out the latest - looks great to me!

I only noticed one more minor thing - when moving the mouse across a reason file, several hovers would get queued up, and they would be displayed in series (you'd see some 'flash' briefly). We could fix this by keeping track of the latest hover request, and ignoring previous ones.

Otherwise, this is awesome - excited to have this in here! Thanks for working through the scrolling and the wrapping issues, @zbaylin !

PR #1908, opened 2020-06-10
fix(editor/#1475): Wide characters disappearing on scroll

bryphe:
__Issue:__ This is a fix for the 3rd issue called out in #1475 - the fact that some of the characters were disappearing on scroll. It turns out that this doesn't just impact wide characters (like the double-width CJK characters in the screenshot), but also could affect other cases where the size of the character (like, a `/t` with tab size of 4) was greater than 1.

__Defect:__ For rendering, we grab a substring of the text that is in the visible viewport. When we talk about substrings - there are 3 'string spaces' that are important:
1) Byte space - the individual bytes in a string. This is the fastest to work with, but often isn't what we need - in UTF-8, characters may be represent by multiple bytes.
2) Character indices - individual UTF-8 characters. For ASCII, this would be 1-to-1 with the byte indices, but many characters require multiple bytes.
3) Position - the actual positioning of the character, given the rendered width (for example, a `tab` with tab size 4 would take up 4 cells, so the string `/ta` with those settings would have `a` at position 4).

__Fix:__ The issue is that we were conflating these spaces in coming up with that string to render - we were using the _character indices_ when we actually needed to use the _position_. So the fix is to figure out where the string starts and ends based on the _position_, convert those to _bytes_ (which the colorizer operates on), and then render from there.

__TODO:__
- [ ] Add a test case
- [ ] Profile - the `positionToByte` conversion isn't very fast. Need to see if it is a hot path for rendering.
- [ ] Validate scroll functionality after merge

__Future:__
- Is there a way we can use the type system to help differentiate / enforce those 'string spaces'?
- This code could definitely be refactored, streamlined, and optimized


bryphe:
This is fixed in a more robust way (based on pixel positions) thanks to @zbaylin 's work in #2171 🎉

PR #1913, opened 2020-06-10
feat(layout): minimum window sizes and cascading

glennsl:
cascading

Implements minimum sizes and "cascading", fixes the tracking of the handle relative to the mouse pointer, and remembers the window state before dragging a handle so that moving the handle back reverts it. It also lays the groundwork for implementing maximization and directional resizing.

### DIFF
--- a/src/Feature/Layout/Feature_Layout.re
+++ b/src/Feature/Layout/Feature_Layout.re
@@ -1,13 +1,20 @@
 // MODEL
 
-type model = Layout.t(int);
+type model = {
+  tree: Layout.t(int),
+  uncommittedTree: option(Layout.t(int)),
+};
 
-let initial = id => Layout.singleton(id);
+let initial = id => {tree: Layout.singleton(id), uncommittedTree: None};
 
-let windows = Layout.windows;
-let addWindow = Layout.addWindow;
-let insertWindow = Layout.insertWindow;
-let removeWindow = Layout.removeWindow;
+let updateTree = (f, model) => {...model, tree: f(model.tree)};
+
+let windows = model => Layout.windows(model.tree);
+let addWindow = (direction, focus) =>
+  updateTree(Layout.addWindow(direction, focus));
+let insertWindow = (target, direction, focus) =>
+  updateTree(Layout.insertWindow(target, direction, focus));
+let removeWindow = target => updateTree(Layout.removeWindow(target));
 
 let move = (focus, dirX, dirY, layout) => {
   let positioned = Positioned.fromLayout(0, 0, 200, 200, layout);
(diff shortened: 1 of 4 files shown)


src/Feature/Layout/Layout.re
```diff
@@ -26,35 +28,419 @@ let empty = vsplit([]);
 
 let singleton = id => vsplit([window(id)]);
 
+// INTERNAL
+open {
```
bryphe, line 32: What does this `open {` do?

src/Feature/Layout/Layout.re
```diff
@@ -26,35 +28,419 @@ let empty = vsplit([]);
...
+           ~stopWhen=i => i < 0,
+           nodes,
+         );
+
+       let%test_module "reclaimLeft" =
+         (module
+          {
+            // Because floats are tricky to compare, this inflates them by a
+            // magnitude of two, converts them to int and then compares,
+            // effectively comparing each float value with a tolerance of 0.01..
```
bryphe, line 80: Thanks for the details here!

src/Feature/Layout/Layout.re
```diff
@@ -26,35 +28,419 @@ let empty = vsplit([]);
...
+            };
+
+            let%test "available < limit, target == 1" = {
+              let nodes = [|window(1), window(2), window(3)|];
+
+              let reclaimed = reclaimLeft(~limit=10., 1, nodes);
+
+              (reclaimed, nodes) == (0.7, [|0.3, 1., 1.|]);
+            };
+
```
bryphe, line 103: Nice set of test cases 👍

bryphe:
The cascading is really cool! In combination with the fixed mouse-handle-tracking, it's really impressive looking. Thanks for your work on this, @glennsl !

src/Feature/Layout/Feature_Layout.re
```diff
@@ -1,13 +1,20 @@
 // MODEL
 
-type model = Layout.t(int);
+type model = {
+  tree: Layout.t(int),
+  uncommittedTree: option(Layout.t(int)),
```
bryphe, line 5: Cool idea to have a 'working' tree, and then committing it once the resize has completed

src/Feature/Layout/Layout.re
glennsl: Learned this trick from @Et7f3 not long ago. It defines an empty module and then opens it, effectively defining top-level functions that won't be exported. Roughly equivalent to:

```
module Internal = {
  ...
};
open Internal;
```

The reason I'm trying this out instead of using an explicit `Internal` module is that you sometimes have an internal function depend on an exported function which in turn depends on an internal function, meaning you need several separate blocks of internal code, which is a bit awkward with explicit `Internal` modules.

What do you think about it?

PR #1924, opened 2020-06-12
feat(layout): keyboard navigation between sidebar, pane and editors

glennsl:
This adds keyboard navigation between editors, the sidebar and the bottom pane as an extension of the existing window movement keybindings. We lack good visual indication of where the keyboard focus actually is, but otherwise it works great!

Adding proper indication of focus will require some design work and is too big to in scope for this PR I think, but let me know if you think it should be added before merging.

### DIFF
--- a/src/Feature/Layout/Feature_Layout.re
+++ b/src/Feature/Layout/Feature_Layout.re
@@ -1,5 +1,10 @@
 // MODEL
 
+type panel =
+  | Left
+  | Center(int)
+  | Bottom;
+
 type model = {
   tree: Layout.t(int),
   uncommittedTree: [
@@ -77,7 +82,7 @@ type msg =
 
 type outmsg =
   | Nothing
-  | Focus(int);
+  | Focus(panel);
 
 let rotate = (direction, focus, model) => {
   ...model,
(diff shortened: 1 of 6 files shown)


src/Feature/Layout/Feature_Layout.re
```diff
@@ -1,5 +1,10 @@
 // MODEL
 
+type panel =
+  | Left
+  | Center(int)
+  | Bottom;
```
glennsl, line 6: This can easily be extended with `Top` and `Right` as well, but I didn't want to take that into account everywhere as long as we don't use it.

src/Feature/Layout/Feature_Layout.re
```diff
@@ -134,143 +139,227 @@ let update = (~focus, model, msg) => {
 
   | Command(MoveLeft) =>
     switch (focus) {
-    | Some(focus) => (model, Focus(moveLeft(focus, activeTree(model))))
+    | Some(Center(focus)) =>
+      let newFocus = model |> activeTree |> moveLeft(focus);
+      if (newFocus == focus) {
+        (model, Focus(Left));
+      } else {
+        (model, Focus(Center(focus)));
```
bryphe, line 147: I'm able to move right to left with this change - should this be `Focus(Center(newFocus))`?

bryphe:
I noticed one bug - mentioned it in a code comment.

Are there keybindings that work in the file explorer for moving the selection up/down? I thought we did have a concept of focus - and the arrow keys worked - but perhaps I'm mistaken.

src/Feature/Layout/Feature_Layout.re
glennsl: Indeed it should. Sorry, I don't know how I missed that. Fixed in https://github.com/onivim/oni2/pull/1924/commits/f9ddab0125d272f8b27acdf34022a2ae3e8704db

glennsl:
> Are there keybindings that work in the file explorer for moving the selection up/down? I thought we did have a concept of focus - and the arrow keys worked - but perhaps I'm mistaken.

Yeah, up, down and enter are the only keybindings implemented so far. You need to have an existing focus though.

Now that all the infrastructure is in place, we ought to move the file explorer to a feature project, add commands for navigation and keybindings using those commands. Should be pretty straight-forward at this point I think.

PR #1925, opened 2020-06-12
docs: update keybinding reference with layout additions

glennsl:
Updates the keybindings reference with the latest additions to layout. Keybindings that don't work because of bugs in `editor-input` have been formatted with strike-through.


bryphe:
Thanks for documenting these, @glennsl !

Issue #1122, opened 2019-12-28
UX: Missing bold and italic options in themes (and elsewhere)


bryphe:
One of the goals of the project is a __beautiful, modern experience__ - but one way we're falling short of this, at the moment, is the lack of bold or italics in themes (and in our text rendering in general).

Example:
![image](https://user-images.githubusercontent.com/13532591/71550623-dd671480-2989-11ea-9ac6-93a48a686649.png)

vs

![image](https://user-images.githubusercontent.com/13532591/71550618-d6d89d00-2989-11ea-9db7-8750f88cd851.png)

Theme used is here: https://github.com/PierreCapo/voir-vscode-theme

bryphe:
@zbaylin implemented this in #1930 🎉

PR #1930, opened 2020-06-13
feat(editor): hook up bold and italic props from Textmate

zbaylin:
I think I have all the scaffolding in place, but themes that appear bold/italic in VSCode for me don't here. I wonder if there is something broken with the Textmate parsing with regards to the bold/italic modes? Logging the values also shows it's always false on parse.


src/Core/Oni_Core.re
```diff
@@ -9,7 +9,7 @@ module BufferLine = BufferLine;
 module BufferPath = BufferPath;
 module BufferUpdate = BufferUpdate;
 module BuildInfo = BuildInfo;
-module ColorizedToken = ColorizedToken;
```
bryphe, line 12: Thanks for renaming this! More general than just a color now 👍

src/Feature/Editor/BufferLineColorizer.rei
```diff
@@ -7,19 +7,24 @@ open Revery;
 open Oni_Core;
 
 /*
- * Type [tokenColor] is a tuple of [(backgroundColor, foregroundColor)]
+ * Type [tokenTheme] is a record containing theme information
  */
-type tokenColor = (Color.t, Color.t);
+type tokenTheme = {
```
bryphe, line 12: `tokenTheme` sounds like a theme to me - maybe just `token` (ie, `BufferLineColorizer.token`) or `themedToken` - a token that is themed?

src/Feature/Editor/Draw.re
```diff
@@ -94,13 +89,19 @@ let drawShapedText = {
...
-    let text =
-      Revery.Font.(shape(context.font, text) |> ShapeResult.getGlyphString);
+  (~context, ~x, ~y, ~color, ~bold, ~italic, text) => {
+    let font =
+      Revery.Font.Family.resolve(
+        ~italic,
+        bold ? Revery.Font.Weight.Bold : Revery.Font.Weight.Normal,
+        context.fontFamily,
+      )
+      |> Stdlib.Result.get_ok;
```
bryphe, line 99: We need to be careful with the `Stdlib.Result.get_ok` method - in the case where, for some reason, the `resolve returns an `Error` - this would manifest as a crash.

Seems like this could happen a few ways:
1) The user switches a theme, the application crashes because it couldn't resolve a now-bold or now-italic token
2) The user switches a fonts, and the application crashes
3) ...even worse - the user types a token, it gets resolved as bold / italic, causing a crash and the user losing work.

3 would be the worst case! We should avoid `Stdlib.Result.get_ok` during runtime, and handle the `result`. I can see two possibilities:
- Possibility 1: This doesn't actually need to be `result` - it always returns _something_ (ie, a normal font), even if couldn't resolve the bold/italic version. Or, we could provide an alternative method that is like `resolveOrFallback` - that tries to return `bold/italic`, but gives the normal variant if it can't find it. That wouldn't need to return a `result`.
- Possibility 2 - there isn't a good a way to handle when there isn't an italic / bold variant, and in that case, the best thing we could do is just avoid rendering (ie, use `Result.iter` to do the rendering).

I think we should be able to have a non-result API that falls back to the 'normal' variant if it can't resolve the `bold`/`italic`/`both` variant, but either way we shouldn't have a chance of crashing here.

src/Service/Font/FontLoader.re
```diff
@@ -57,7 +58,7 @@ let loadAndValidateEditorFont =
               requestId,
               {
                 fontFile: fullPath,
-                fontFamily: Revery.Font.Family.fromFile(fullPath),
+                fontFamily: Revery.Font.Family.system(familyName),
```
bryphe, line 61: It looks like this logic is unfortunately broken in a couple cases:
- When no `editor.fontFamily` is specified (ie, the default case) - we should make sure we're loading our default font.
- When a path is specified, ie `/path/to/some-font.ttf` - we should handle that case, and just have a custom resolver that always resolves to the font file.

src/reason-textmate/TokenTheme.re
```diff
@@ -115,11 +115,24 @@ let of_yojson = (~defaultBackground, ~defaultForeground, json: Yojson.Safe.t) =>
         | _ => None
         };
 
+      print_endline(json |> Yojson.Safe.to_string);
+
+      let fontStyle = fs =>
+        Yojson.Safe.Util.member("fontStyle", json)
+        |> str
```
bryphe, line 122: Thanks for catching this!

src/reason-textmate/TokenTheme.re
```diff
@@ -115,11 +115,24 @@ let of_yojson = (~defaultBackground, ~defaultForeground, json: Yojson.Safe.t) =>
         | _ => None
         };
 
+      print_endline(json |> Yojson.Safe.to_string);
```
bryphe, line 118: Will need to remove debug logging prior to merging (can crash windows when there is no `stdout`/`stderr` attached)

bryphe:
So excited to finally have this in, @zbaylin ! Thanks for exploring it

Sharing the screenshot you showed me:

image

😍

bryphe:
This will fix https://github.com/onivim/oni2/issues/1122 🎉

src/Service/Font/FontLoader.re
zbaylin: I think I fixed this in the latest commit. Thanks for pointing it out! Let me know if anything is still broken that you can find. I wasn't able to.

src/Feature/Editor/Draw.re
zbaylin: That sounds great to me! Would that API be housed in `Service_Font`, as you see it?

src/reason-textmate/TokenTheme.re
zbaylin: Ah thanks for catching that. It's gone now!

src/Feature/Editor/BufferLineColorizer.rei
zbaylin: Good point. I changed it to `themedToken` in the latest commit!

src/Feature/Editor/Draw.re
bryphe: > Would that API be housed in Service_Font, as you see it?

Good question. I was actually thinking it could make sense for [`FontFamily`](https://github.com/revery-ui/revery/blob/master/src/Font/FontFamily.rei) - a `resolve` that returns a result (a possibility of failure), and a `resolveWithFallback` which always returns a `FontCache.t` - just, in the case where bold/italic/variant isn't available, we'd use the default cache.

I'd be fine putting it in `Service_Font` for now, too, as an intermediate to try out - and when we feel good about the API (or see another use case in Revery), we could push it back at that time.

zbaylin: Awesome, for the time being I put it in `ServiceFont` as `resolveWithFallback` that resolves to the default font in the event of an error. Let me know if you think this is satisfactory!

PR #1931, opened 2020-06-13
fix(file-explorer): focus issues

glennsl:
**Issue:** The file explorer would not initially set the focus to a file opened from outside of the file explorer. Additionally, the focus would be set to invalid paths, or paths that do not exist in the tree, in certain conditions such as initially when the welcome screen is shown and the focus would be set to `oni://Welcome`

**Defect:** When setting the focus, the path was not checked against the tree. Additionally, the focus was set before the nodes in the path had been loaded.

**Fix:** Validate that the path exists on the tree when settings focus, and set the focus only after loading and revealing the path.


bryphe:
Thanks for fixing this, @glennsl ! Works great 👍

PR #1939, opened 2020-06-15
feat(extensions): Add LaserWave Italic extension back

bryphe:
Bring `LaserWave Italic` back in, since we now support the italics:

image


zbaylin:
This will make for an awesome demo screenshot 💯

PR #1942, opened 2020-06-15
feat(editor): signature help

zbaylin:
⚠️ Again, very WIP ⚠️ 

This creates a `Feature_SignatureHelp` module that will be the integration of the information we get from the extension host


zbaylin:
I think I'm getting a race condition bug where, when the trigger key is hit, the buffer isn't updated in the extension host so the request returns `null`. @bryphe have you dealt with any similar bugs in the past that might help me figure this one out?

src/Core/Json.re
```diff
@@ -18,10 +18,12 @@ module Decode = {
         string => result(value, string)
     );
 
-    let of_file = file =>
+    let of_file = file => {
+      Console.log(file);
```
bryphe, line 22: We'll need to remove the debug logging

src/Exthost/SignatureHelp.re
```diff
@@ -65,15 +65,30 @@ module RequestContext = {
 module ParameterInformation = {
   [@deriving show]
   type t = {
-    label: string,
+    label: [ | `String(string) | `Range(int, int)],
     documentation: option(MarkdownString.t),
   };
 
+  let label =
```
bryphe, line 72: Thanks for fixing this up!

src/Feature/Hover/Feature_Hover.re
```diff
@@ -34,15 +34,8 @@ let initial = {
...
-  let current = ref(0);
-
-  let get = () => {
-    let id = current^;
-    current := id + 1;
-    id;
-  };
-};
+module IDGenerator =
+  Oni_Core.Utility.IDGenerator.Make({});
```
bryphe, line 38: We actually have a similar concept in Revery, too (`Revery.UniqueId.Make()`)

src/Components/HoverView.re
```diff
@@ -0,0 +1,152 @@
+open Revery;
```
bryphe, line 1: Nice! Thanks for refactoring this out to a common place - good chunk of code to share between hover and signature help 👍

src/Feature/Hover/Feature_Hover.re
zbaylin: Oh cool, didn't notice that!

src/Feature/SignatureHelp/Feature_SignatureHelp.rei
```diff
@@ -0,0 +1,77 @@
...
+  | InfoReceived({
+      signatures: list(Exthost.SignatureHelp.Signature.t),
+      activeSignature: int,
+      activeParameter: int,
+      requestID: int,
+    })
+  | EmptyInfoReceived(int)
+  | RequestFailed(string)
+  | SignatureIncremented
+  | SignatureDecremented;
```
bryphe, line 34: Might be worth renaming this such that they describe the event that occurs - like `SignatureIncrementClicked`/`SignatureDecrementClicked`. That way, if there ends up being some subtle difference in behavior between clicking and the command - it's obvious where to handle it.

src/Model/ContextKeys.re
```diff
@@ -25,48 +25,53 @@ let menus =
...
+        }
+      ),
+      bool("visualMode", state =>
+        switch (ModeManager.current(state)) {
+        | TerminalVisual
+        | Visual => true
+        | _ => false
+        }
+      ),
+      bool("parameterHintsVisible", state =>
```
bryphe, line 70: Thanks for adding this key!

src/Service/Exthost/Service_Exthost.re
```diff
@@ -108,6 +108,27 @@ module Effects = {
         );
       });
     };
+
+    let provideSignatureHelp =
```
bryphe, line 112: Nice work wiring this up! 🎉

src/Store/KeyBindingsStoreConnector.re
```diff
@@ -459,6 +459,18 @@ let start = maybeKeyBindingsFilePath => {
         command: Feature_Layout.Commands.toggleMaximize.id,
         condition: windowCommandCondition,
       },
+      {
+        key: "<A-DOWN>",
+        command: Feature_SignatureHelp.Commands.incrementSignature.id,
+        condition:
+          "editorTextFocus && parameterHintsVisible" |> WhenExpr.parse,
+      },
```
bryphe, line 467: Thanks for adding these!

src/Core/Json.re
zbaylin: Oops! I thought I had removed all of them. Thanks for pointing this one out! Fixed now 👍

src/Feature/SignatureHelp/Feature_SignatureHelp.rei
zbaylin: Ah you're right. I changed it in the latest commit. Thanks!

PR #1945, opened 2020-06-16
fix(exthost): Signature Help - hook up documentation of parameters / signature

bryphe:
This change wires up the `documentation` properties for the `ParameterInformation` and `Signature` information coming from the extension host, and updates the test case so that we're actually sending some signature info and validating it comes through.

### DIFF
--- a/src/Exthost/Exthost.rei
+++ b/src/Exthost/Exthost.rei
@@ -83,7 +83,11 @@ module Location: {
 };
 
 module MarkdownString: {
-  type t = string;
+  [@deriving show]
+  type t;
+
+  let fromString: string => t;
+  let toString: t => string;
 
   let decode: Json.decoder(t);
 };
@@ -298,8 +302,7 @@ module SignatureHelp: {
     [@deriving show]
     type t = {
       label: string,
-      // TODO
-      //documentation: option(string),
+      documentation: option(MarkdownString.t),
     };
     let decode: Json.decoder(t);
   };
@@ -308,8 +311,7 @@ module SignatureHelp: {
     [@deriving show]
     type t = {
       label: string,
-      // TODO:
-      //documentation: options(MarkdownString.t),
+      documentation: option(MarkdownString.t),
       parameters: list(ParameterInformation.t),
     };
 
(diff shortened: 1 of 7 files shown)


src/Exthost/Exthost.rei
```diff
@@ -83,7 +83,11 @@ module Location: {
 };
 
 module MarkdownString: {
-  type t = string;
+  [@deriving show]
+  type t;
+
+  let fromString: string => t;
+  let toString: t => string;
```
bryphe, line 90: This might seem a bit convoluted, since the underlying type in `MarkdownString.re` is.... `string`.

But eventually there is going to be more we need to handle, like:
```
// https://github.com/onivim/vscode-exthost/blob/c7df89c1cf0087ca5decaf8f6d4c0fd0257a8b7a/src/vs/base/common/htmlContent.ts#L10
//type t = {
//	stringValue: string, // value
//	isTrusted: bool,
//	supportThemeIcons: bool,
// uris
//}
```

So if we make it abstract - it gives us more flexibility to rev the underlying implementation, without breaking any consumers.

src/Feature/Hover/Feature_Hover.re
```diff
@@ -18,7 +18,7 @@ type provider = {
 type model = {
   shown: bool,
   providers: list(provider),
-  contents: list(string),
+  contents: list(Exthost.MarkdownString.t),
```
bryphe, line 21: This is just a minor change to pick up the `MarkdownString.t` abstract type

test/collateral/extensions/oni-language-features/extension.js
```diff
@@ -62,9 +62,9 @@ function activate(context) {
 
 	const signatureHelpProvider = {
 		provideSignatureHelp: (_document, _position, _token, _context) => {
-			const signature1 = new vscode.SignatureInformation("signature 1", null);
+			const signature1 = new vscode.SignatureInformation("signature 1", "signature 1 documentation");
 			signature1.parameters = [
-				new vscode.ParameterInformation("parameter 1", null)
+				new vscode.ParameterInformation("parameter 1", "parameter 1 documentation")
 			];
```
bryphe, line 68: @zbaylin - if you decide you want to test signature help via the `oni-dev-extension`, this will be a good sample to start with 👍

bryphe:
Sorry this was late coming in, @zbaylin !

Issue #1947, opened 2020-06-17
Add syntax highlighting to hover popup


yunti:
Thanks for the recent addition of hover info and the `gh` shortcut too which is great, and makes a huge difference to the experience in onivim2

Perhaps it's already on your next steps, but it would be great to have syntax highlighting in the hover info.

Below is a comparison versus vscode. 
Which shows what a difference highlighting makes.
(Additionally onivim seem to have a larger font used for the file location which appears in the popup, giving it the emphasis. Rather than the type info.
Vscode doesn't show this (or do I have it turned off?) which makes the hover info more focused and the current file used is probably obvious so doesn't seem to add much here).

vscode:
<img width="388" alt="image" src="https://user-images.githubusercontent.com/1826271/84901667-b7523f80-b0a3-11ea-9573-c5f07ea695c7.png">
Onivim:
<img width="349" alt="image" src="https://user-images.githubusercontent.com/1826271/84902052-347db480-b0a4-11ea-99c2-5cd7e1e8b7ce.png">

bryphe:
Hi @yunti ,

> Thanks for the recent addition of hover info and the gh shortcut too which is great, and makes a huge difference to the experience in onivim2

Great to hear!

Thanks for logging the issue. Seems like there are two issues:

1) We're showing the file location, which probably doesn't need to be there
2) No syntax highlighting

For the first issue - I need to check, but we may be handling the setting for `show_module_path_on_hover` incorrectly:
https://github.com/jaredly/reason-language-server/blob/f523b11e63c19dce25b8ceae895ad28ffcb085b6/src/lsp/NotificationHandlers.re#L111

For the second issue - it seems that, when there is an anonymous code-block in the markdown, like:
```
     ["'''\n(\n  ~children: 'children,\n  ~key: option(string),\n  unit\n) => Js.t(< children : 'children >)\n''';''
       ];
```

We aren't handling it quite correctly - we should be falling back to the filetype of the active editor, but I think we're just falling back to plaintext now.

bryphe:
#1954 fixes the first issue

@zbaylin , is there a convenient to set a 'default' filetype for the markdown renderer? If so - we should set the default (no language specified) to the filetype of the active buffer. I believe that would address the second issue, that there is no syntax highlighting applied.

zbaylin:
That's a good question. Right now there isn't but it's definitely something I can add in the Markdown component here. It shouldn't be too hard, I can look into it tonight!

bryphe:
[...]

PR #1955, opened 2020-06-18
fix(markdown): add default language parameter to Markdown component

zbaylin:
This is to address the second half of #1947. If there isn't/an invalid language, we fallback to this default.

In use with the RLS
image

### DIFF
--- a/src/Components/Markdown.re
+++ b/src/Components/Markdown.re
@@ -18,6 +18,7 @@ open {
              ~languageInfo,
              ~grammars,
              ~language,
+             ~defaultLanguage,
              lines,
            ) => {
          let grammarRepository =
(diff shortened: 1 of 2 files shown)


src/Components/Markdown.re
```diff
@@ -26,10 +27,19 @@ open {
...
-             language,
-           );
+           switch (
+             Oni_Extensions.LanguageInfo.getScopeFromLanguage(
+               languageInfo,
+               language,
+             )
+           ) {
+           | Some(scope) as s => s
+           | None =>
```
bryphe, line 37: Minor tweak - you could use `OptionEx.or_` to remove the need for the switch statement.

src/Components/Markdown.re
```diff
@@ -26,10 +27,19 @@ open {
...
+             Oni_Extensions.LanguageInfo.getScopeFromLanguage(
+               languageInfo,
+               defaultLanguage,
```
bryphe, line 40: Awesome! Thanks for adding this fallback

PR #1966, opened 2020-06-20
fix(app): raise window on unsaved-buffer close

zbaylin:
This brings the typical behavior that any window with unsaved work is raised on an attempt to quit. This uses Revery's new `Window.raise` to achieve that.


zbaylin:
I was originally going to include a SIGINT handler in this PR as well, but that requires some more investigation and I think this is important/significant enough to stand on it's own

src/Store/StoreThread.re
```diff
@@ -85,6 +85,7 @@ let start =
       ~minimize,
       ~close,
       ~restore,
+      ~raise,
```
glennsl, line 88: `raise` is very ambiguous here. My first instinct, even knowing that you've been working on raising the window, is that refers to raising an exception. And it will in fact also shadow the global `raise` function that does raise exceptions (hence why it's highlighted as a keyword). I would recommend renaming this to `raiseWindow`.

src/Store/StoreThread.re
```diff
@@ -368,6 +369,8 @@ let start =
     window,
   );
 
+  let signalHandle = Luv.Signal.init() |> Result.get_ok;
+
```
glennsl, line 373: This doesn't seem to be used anywhere. Assuming it's left over form the `SIGINT` handler you were investigating?

src/Store/StoreThread.re
bryphe: Agreed! Not the first time we've had a naming collision with Sdl2 - the `show` function to show a window was easy to confuse for the `show` function to convert to a string. `raiseWindow` sounds good to clarify and avoid the shadowing issue 👍

src/Store/StoreThread.re
zbaylin: Ah yeah thanks! I thought I had run `esy b dune build @check`, but I guess I hadn't.

src/Store/StoreThread.re
zbaylin: Ah thanks, I totally forgot about the `raise` function shadowing. I'll change that now!

bryphe:
> I was originally going to include a SIGINT handler in this PR as well, but that requires some more investigation and I think this is important/significant enough to stand on it's own

That sounds like a good plan 👍

PR #1972, opened 2020-06-20
fix(cli): OSX - 'oni2' command is not forking by default

bryphe:
__Issue:__ When running `oni2` with the latest builds, on OSX, it is staying attached to the terminal process. The expected behavior is it should detach, unless `-f` or `--nofork` are passed - in which case it should stay attached.

__Fix:__ Run `Oni2` instead of `Oni2_editor`.

Looks like this is a regression from https://github.com/onivim/oni2/pull/1927 - I still see the `ONI2_BUNDLED` environment variable get passed, so I think this is OK. @zbaylin - is there anything else I should check here?

### DIFF
--- a/scripts/osx/run.sh
+++ b/scripts/osx/run.sh
@@ -13,4 +13,4 @@ DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
 
 export ONI2_BUNDLED=1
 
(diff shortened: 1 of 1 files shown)


zbaylin:
As long as the icon looks okay it should be good. Sorry I overlooked this!

bryphe:
> As long as the icon looks okay it should be good. Sorry I overlooked this!

Cool, no problem! We're still getting the new icon with this change 👍

Issue #1973, opened 2020-06-20
Hover: Hover shows on all visible editors


bryphe:
If you open a `hover` via `gh` - it will show on all visible editors, ie:

![image](https://user-images.githubusercontent.com/13532591/85211469-5f634500-b2fe-11ea-8e6a-cd56724e9561.png)

__REPRO__:
- Open Onivim2 in the `oni2` directory
- Open a reason file
- Run `:vsp README.md`
- Show a hover in the reason file via `gh`

Expected: The hover should only be shown in the active editor window
Actual: The hover is shown in all windows

zbaylin:
Oh interesting, I'll fix this now. Sorry about that!

bryphe:
Fixed now with #1978 - thanks @zbaylin !

PR #1978, opened 2020-06-21
fix(hover & signature help): check what editor the request was sent from

zbaylin:
This fixes the multiple hovers issue @bryphe noted in #1973. The request now takes an editor, and the view checks that it is the same when displaying.

image

### DIFF
--- a/src/Feature/Hover/Feature_Hover.re
+++ b/src/Feature/Hover/Feature_Hover.re
@@ -23,7 +23,7 @@ type model = {
   triggeredFrom:
     option([ | `CommandPalette | `Mouse(EditorCoreTypes.Location.t)]),
   lastRequestID: option(int),
-  editor: option(Feature_Editor.Editor.t),
+  editorID: option(int),
 };
 
 let initial = {
@@ -33,7 +33,7 @@ let initial = {
   range: None,
   triggeredFrom: None,
   lastRequestID: None,
-  editor: None,
+  editorID: None,
 };
 
 module IDGenerator =
@@ -52,7 +52,7 @@ type msg =
       contents: list(Exthost.MarkdownString.t),
       range: option(EditorCoreTypes.Range.t),
       requestID: int,
-      editor: Feature_Editor.Editor.t,
+      editorID: int,
     })
   | HoverRequestFailed(string)
   | MouseHovered(EditorCoreTypes.Location.t)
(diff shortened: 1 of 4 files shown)


src/Feature/Hover/Feature_Hover.re
```diff
@@ -276,15 +289,18 @@ module View = {
...
-    <Oni_Components.HoverView x y theme=colorTheme>
-      {List.map(markdown => <hoverMarkdown markdown />, model.contents)
-       |> React.listToElement}
-      {model.contents != [] && diagnostic != []
-         ? <horizontalRule theme=colorTheme /> : React.empty}
-      {List.map(diag => <hoverDiagnostic diagnostic=diag />, diagnostic)
-       |> React.listToElement}
-    </Oni_Components.HoverView>;
+    switch (model.editor) {
+    | Some(editor') when editor' === editor =>
```
glennsl, line 293: You should generally avoid using physical equality (`===`) as it can easily lead to bugs, especially in functional programming where data structured tend to be replaced rather than mutated. Instead just compare the ids.

src/Feature/Hover/Feature_Hover.re
```diff
@@ -23,6 +23,7 @@ type model = {
   triggeredFrom:
     option([ | `CommandPalette | `Mouse(EditorCoreTypes.Location.t)]),
   lastRequestID: option(int),
+  editor: option(Feature_Editor.Editor.t),
```
glennsl, line 26: I don't think you need to store the entire editor record here. Just the `id` should be sufficient. Storing the entire editor violates single source of truth as it can easily get stale/out-of-sync, which leads to bugs.

src/Feature/Hover/Feature_Hover.re
zbaylin: Thanks, that's a much better idea. I was getting the `compare functional values` crash and lazily decided to use `===`. Thanks for correcting me!

src/Feature/Hover/Feature_Hover.re
zbaylin: This is also changed in my latest commit. Thanks again!

zbaylin:
Just realized that the same thing would happen for signature help, so I tested it and it does. I'll fix that in this PR too

PR #1981, opened 2020-06-22
fix(signature help/#1980): close on buffer updates

### LINKED ISSUE #1980: Signature Help: Remains open after backspacing
It looks like there is a bug on the closing condition of signature help - if I backspace past where the signature help was first opened, it remains opened.

We'll need to figure out the right condition to close in this case - an option would be to save the cursor position when we first trigger it, and if the cursor ends up before that position, we'd cancel the gesture.

Another option would be to save the cursor position of the last successful request, and if the cursor ends up prior to that position, we'd request again (and possibly cancel if null/none is returned for the signature help provider).

zbaylin:
This should fix #1980. I have tested it in a bunch of scenarios and it seems to work, but my implementation is slightly hacky so I welcome anyone else who wants to try it 😅.


src/Feature/SignatureHelp/Feature_SignatureHelp.re
```diff
@@ -348,6 +361,41 @@ let update = (~maybeBuffer, ~maybeEditor, ~extHostClient, model, msg) =>
       },
       Nothing,
     )
+  | CursorMoved(editorID) =>
```
bryphe, line 364: Since we have to keep an eye on different events to decide when to show / retrigger signature help, that suggests maybe a subscription could streamline and simplify things (by automatically reacting to state changes). 

We don't have a good example to point to for the language services, but something to think about..

src/Store/Features.re
```diff
@@ -448,6 +448,26 @@ let update =
       };
     let effect = [shEffect] |> Effect.batch;
     ({...state, signatureHelp}, effect);
+
+  | EditorCursorMove(editorID, _) =>
+    let maybeBuffer = Selectors.getActiveBuffer(state);
+    let maybeEditor =
+      state |> Selectors.getActiveEditorGroup |> Selectors.getActiveEditor;
```
bryphe, line 455: This might need to be merged up with the new `Feature_Layout` changes

src/Feature/SignatureHelp/Feature_SignatureHelp.re
```diff
@@ -348,6 +361,41 @@ let update = (~maybeBuffer, ~maybeEditor, ~extHostClient, model, msg) =>
...
+            ~model,
+            ~context,
+            ~requestID,
+          );
+        (
+          {...model, shown: true, lastRequestID: Some(requestID)},
+          Effect(effects),
+        );
+      | _ => (model, Nothing)
+      };
```
bryphe, line 396: We could potentially streamline these nested switches with the `ppx_let`, ie:

```
let%bind editorId = model.editorID;
...
let%bind location = model.location;
```

zbaylin:
@bryphe I just merged this up with `master`. Let me know it if doesn't fit into the new API well!

src/Store/Features.re
zbaylin: Just did this in e953d67! Let me know if it doesn't fit particularly well.

src/Feature/SignatureHelp/Feature_SignatureHelp.re
zbaylin: Thats an interesting idea -- I'm very unfamiliar with Subscriptions in general, but nothing I can't learn! Let me know if you want me to redo this PR to use one.

PR #1982, opened 2020-06-22
fix(completion): Positioning regression

bryphe:
There was a regression in the completion-overlay positioning from https://github.com/onivim/oni2/pull/1953 - we were incorrectly adding the `gutterWidth` to the `y` (vertical) position as opposed to the `x` (horizontal) position.

### DIFF
--- a/src/Feature/Editor/OverlaysView.re
+++ b/src/Feature/Editor/OverlaysView.re
@@ -69,8 +69,8 @@ let make =
       editor,
     );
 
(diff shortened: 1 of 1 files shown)


zbaylin:
I was worried this was something I had caused with the other overlays 😅

bryphe:
> I was worried this was something I had caused with the other overlays 😅

Sorry for scaring you! Not your fault, was on me 😬 

I need to hook up some image verification tests at some point to help protect against this sort of regression...

PR #1984, opened 2020-06-22
fix(hover & signature help/#1947): various text display issues

### LINKED ISSUE #1947: Add syntax highlighting to hover popup

zbaylin:
This fixes the following issues with Signature Help and Hover:
- no syntax highlighting
- incorrect font color
- text appearing too small

Fixes #1947

### DIFF
--- a/bench.esy.lock/index.json
+++ b/bench.esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "9b7851c9f2d94a0c2546161ea1a5be7e",
+  "checksum": "1d9e69635ff238f82479b61d1430b4d3",
   "root": "Oni2@link-dev:./package.json",
   "node": {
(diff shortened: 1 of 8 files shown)


bryphe:
New highlighting looks great!

PR #1989, opened 2020-06-23
fix(layout): Fix various regressions in zen mode

bryphe:
This fixes several regressions in zen mode from #1977, like:
- Zen mode shows all splits
- Zen mode does not hide the tab bar
- Zen mode does not de-activate with `:vsp` or `:sp`

Some future work:
- It might make sense to bring the state management for zen mode into `Feature_Layout`, since it mainly has an impact on layout (and some view operations change the state, like adding a split).
- Zen mode configuration should be moved to the new configuration system


src/UI/EditorView.re
```diff
@@ -248,11 +248,16 @@ let make =
...
+  let hideZenModeTabs =
+    state.configuration |> Configuration.getValue(c => c.zenModeHideTabs);
+
   <View onFileDropped style={Styles.container(theme)}>
     <View style={Styles.container(theme)}>
       <Feature_Layout.View
         uiFont={state.uiFont}
         theme
+        isZenMode={state.zenMode}
+        shouldShowTabsInZenMode={!hideZenModeTabs}
```
glennsl, line 260: Do we need layout to know about zen mode? Would it not be better to just have `showTabs`?

There's also the `workbenchEditorShowTabs` configuration setting.

bryphe: Good catch @glennsl ! You're right, would be better just to have `showTabs`. Switched to this in https://github.com/onivim/oni2/pull/1989/commits/e78992fa43bac1791be678b3240cee1f0a806b14

PR #1992, opened 2020-06-24
fix(editor): Completion showing up in multiple windows

bryphe:
__Issue:__ Our completion UI would show up in all visible editor windows, like:

image

__Defect:__ This looks like a regression from #1977 - the `isActive` field is hardcoded to `true`, and the completion rendering uses that flag to decide which split to render on. 

__Fix:__ Wire up the `isActive` property - which is being set elsewhere, but just not plumbed through.

Future work:
- The completion should be its own feature project, and keep track of the editor we make the request (like we're doing for hover and signature help).


bryphe:
> Ah, sorry. Should have marked that with a TODO or something!

No worries! Easy to miss in a big refactoring.

PR #2005, opened 2020-06-25
fix(vim/#434): Auto-complete in command line does not handle whitespace

### LINKED ISSUE #434: Handle (white)space in command mode autocompletion
Autocompletition in command mode doesn't handle etries with spaces correctly.

1.  When dropping a filesystem based suggestion (in favour of another one: ie, choosing another) for command (`e`, `chdir`) parameters, only the last word (or Word?) gets deleted.
command_mode_space_bug
The command mode's state machine should delete the currently selected completition entry when a discard transition occours.

2. When selecting an entry with spaces, it doesn't get escaped (like it did in Oni), and argument will be invalid for the command - for example chdir fails to folders with spaces in them (on Windows at least)

bryphe:
Fixes #434 (well, once the test cases are green)

The command-line still needs some love - the completion behavior is not ideal (there's a discrepancy between the highlighted item and the text entered, which can be confusing and un-intuitive). But this improves the specific case of white-space-in-paths.


Samonitari:
On first skimming, it seems fine!
I plan to spend some more time with oni2 in the weekend though, so I'll report if I've found some missed parts!

PR #2008, opened 2020-06-25
chore(ci): Add Ubuntu 20.04 preview image to CI

bryphe:


bryphe:
I'll hold off on this change for now - there's some tweaks that would need to be made to the CI caching to make this work.

Issue #592, opened 2019-08-05
Support for traditional Vim buffer/tab behavior


mhanberg:
Oni2 currently implements tabs similar to other GUI text editors.

This is _not_ behavior that I find valuable and would find it beneficial to return to the traditional Vim behavior. (I am fine with this being opt in.)

yunti:
Likewise intellij/webstorm etc... has a great similar feature where you can turn the tabs off just let the editor manage opening closing files.

bryphe:
This is (one of) our top feature requests - the concept of Vim tab-pages to manage splits. Some details here: https://vim.fandom.com/wiki/Using_tab_pages

@glennsl , do you think you could take a look at this? This would involve handling the vim `tabpage` commands, like:
- `tabnew`
- `tabedit`.

We actually get these commands (at least the 'tabedit' / 'tabnew') commands here: https://github.com/onivim/oni2/blob/401bc02467d3a3d3aa27acc6bd7f70342a4492ef/src/Store/VimStoreConnector.re#L266

But we don't do anything with them; we treat it as if the user is just editing a file.

Vim 'tab pages' group together window splits; this is really nice to jump between window layouts. With our current `State`, we only have a root-level set of window splits, defined here: https://github.com/onivim/oni2/blob/401bc02467d3a3d3aa27acc6bd7f70342a4492ef/src/Model/State.re#L42

We'd basically want a new concept of `tabPages`, that could either reference a window manager, or contain window managers, something like: 
```
module TabPage {
   type t = { 
        tabName: option(string); 
        windowSplits: WindowManager.t 
    }
}

module TabPages = {
    type t = {
        activeTabId: int,
        idToPage: IntMap.t(TabPage.t)
     }
};
```

When there are multiple tabpages, we'd show these as an _additional_ tab bar, above the editor group tab bars. If a `TabPage` does not have a name set - we'd use the name and icon of the current buffer. The dirty icon would be shown based on all the buffers in the editor group (if any is dirty - we will show the dirty icon). 

When there is only a single tabpage, we could show nothing - this would be the default case for 'non-tab' users. A boolean option like `workbench.editorShowVimTabPages`  would be helpful for users who have a tab-first workflow - this could show the tabpages even when there is a single tabpage.

For a power-Vim-tab-user, they'd want to set both:
- `workbench.editorShowTabs` to `false`
- `workbench.editorShowVimTabPages` to `true`

[...]

PR #2013, opened 2020-06-25
feat(layout/#592): layout tabs - part 1

glennsl:
layout tabs

This adds a new concept of "layout tabs", which behave more like the tabs in vim. By default these tabs are positioned at the bottom to avoid two levels of tabs at the top, but they can be positioned at the top using the `oni.layout.layoutTabPosition` configuration setting. They can also be hidden, or made to  always show, using `oni.layout.showLayoutTabs`.

To mimic the layout of vim, use:

```
"oni.layout.layoutTabPosition": "top",
"editor.workbench.showTabs": false,
```

There's still a bit of work to get faithful vim behaviour though. These are the next steps I plan to work on:

- [ ] Wire up the `:tab*` commands (`:tabnew` and `:tabedit` already work, although not entirely correctly)
- [ ] Wire up `gt` and `gT`. This must be done from `libvim` for the moment due to bugs in `editor-input`, but shouldn't be too hard.
- [ ] Add a configuration setting for a "single tab mode" that will reuse the first tab instead fo adding hidden tabs when `showTabs` is off.
- [ ] For technical reasons it is currently not possible to open a new buffer directly in a new split or layout tab. Instead what will happen is that the new split or layout tab is opened with he currently active editor in addition to the new one. This is because opening a buffer will automatically open an editor (via `BufferEnter`), and it's not possible to open an empty split or layout tab before opening the buffer. Therefore the only two options are to either open a new split or layout tab with the currently active editor and then open the buffer, or to open the buffer in the old split/layout tab before opening a new one. Either way there'll be an extra editor that shouldn't be there.

Addresses #592


bryphe:
Wow, this is looking great @glennsl ! Thanks for the work on this.

bryphe:
>  Wire up gt and gT. This must be done from libvim for the moment due to bugs in editor-input, but shouldn't be too hard.

I can help with this piece, if you'd like, and haven't started it -

bryphe:
One thing to think about that isn't called out on your list is the `quit` behavior, too - if `:q` is used in the layout tab, it should be closed, moving to the previous tab. If it's the last layout tab, the editor should be closed.

src/Components/Tabs.re
```diff
@@ -49,7 +49,7 @@ let schedulePostRender = f => postRenderQueue := [f, ...postRenderQueue^];
 let component = React.Expert.component("Tabs");
 let make =
     (
-      ~children as render: 'a => element,
+      ~children as render: (~isSelected: bool, int, 'a) => element,
```
bryphe, line 52: I wonder if this `int` should be a named parameter as well, like `~index:int`?

src/Feature/Layout/Model.re
```diff
@@ -132,57 +132,80 @@ type model = {
...
+    }
+  );
+
+let windows = model => Layout.windows(model |> activeLayout |> activeTree);
 
 let visibleEditors = model =>
   model
   |> windows
-  |> List.filter_map(id => groupById(id, model))
+  |> List.filter_map(id => model |> activeLayout |> groupById(id))
```
bryphe, line 204: Nice! Seems like adding the layout tabs was a pretty straightforward addition to the model, with the groundwork you laid out in #1977

src/Feature/Layout/View.re
```diff
@@ -0,0 +1,616 @@
...
+
+  let id: t => int;
+  let title: t => string;
+  let icon: t => option(Oni_Core.IconTheme.IconDefinition.t);
+  let isModified: t => bool;
+
+  let render: (~isActive: bool, t) => Revery.UI.element;
+};
+
+module Tab = {
```
bryphe, line 26: Thanks for generalizing this

glennsl:
> I can help with this piece, if you'd like, and haven't started it -

Already working on it. Implemented it twice even! First as an extension of the goto callback, but then after discovering it was tightly (and weirdly) tied to multiple `:tab` commands I decided to go with a separate tabpage callback.

> One thing to think about that isn't called out on your list is the quit behavior, too - if :q is used in the layout tab, it should be closed, moving to the previous tab. If it's the last layout tab, the editor should be closed.
>
> Related to this - if I close (press the x) of the active tab, when there are tabs available, the editor closes. I would expect it to switch to the next available tab.

Ah, yeah, sorry. That behaviour was intended only for when the last layout tab was closed. I forgot to test and handle these other cases, so it just falls back to that. Will fix!

glennsl:
The Ubuntu 16 CI seems to fail consistently on `esy bootstrap` now. I think this is the relevant part of the log:

```
    autoconf.texi:8017: misplaced }
    autoconf.texi:8018: must be after `@defmac' to use `@defmacx'
    autoconf.texi:8019: misplaced }
    autoconf.texi:8206: must be after `@defmac' to use `@defmacx'
    autoconf.texi:8271: misplaced }
    autoconf.texi:8290: misplaced }
    autoconf.texi:8317: misplaced }
    autoconf.texi:8380: must be after `@defmac' to use `@defmacx'
    conftest.c:4597: must be after `@defmac' to use `@defmacx'
    conftest.c:15929: must be after `@defmac' to use `@defmacx'
    Makefile:251: recipe for target 'autoconf.info' failed
```

Makes no sense to me though...

bryphe:
> Already working on it. Implemented it twice even! First as an extension of the goto callback, but then after discovering it was tightly (and weirdly) tied to multiple :tab commands I decided to go with a separate tabpage callback.

Cool, sounds good. Thanks for looking into it!

> The Ubuntu 16 CI seems to fail consistently on esy bootstrap now. I think this is the relevant part of the log:

Ouch, ya, I'm looking into it - it's unrelated to your changes. I also see this in the log:
```
    /home/vsts/.esy/3/b/autoconf-dafb3f25/build-aux/missing: line 51: aclocal-1.11: command not found
    WARNING: 'aclocal-1.11' is missing on your system.  You should only need it if
             you modified 'acinclude.m4' or 'configure.ac'.  You might want
             to install the Automake and Perl packages.  Grab them from
             any GNU archive site.
```

It shouldn't need to run `aclocal` at all, so it's curious why it's trying... In the meantime, if this is the only check that's failing - feel free to ignore / override. I'll work on getting that check back up.

src/Components/Tabs.re
glennsl: Fixed in https://github.com/onivim/oni2/pull/2013/commits/f43e24dfdc0e5346d08bdf825c2c8143fa3979e3

glennsl:
The closing behaviour should be fixed now. But please do check thoroughly, as it's surprisingly complex with all the different levels and ways of closing.

PR #2016, opened 2020-06-25
fix(editor/#2012): Use nth_opt rather than nth in SignatureHelp

### LINKED ISSUE #2012: Crash: Feature_SignatureHelp (Failure nth)
From

Log file: https://github.com/onivim/oni2/files/4832232/onivim2-crash.log

```
(Failure nth):
Raised at file "stdlib.ml", line 29, characters 17-33
Called from file "src/Feature/SignatureHelp/Feature_SignatureHelp.re", line 429, characters 6-52
Called from file "src/Feature/SignatureHelp/Feature_SignatureHelp.re", line 583, characters 6-251
Called from file "src/UI/EditorView.re", line 107, characters 10-432
Called from file "src/Feature/Editor/EditorSurface.re", line 314, characters 4-34
Called from file "lib/Brisk_reconciler_internal.re", line 831, characters 12-205
Called from file "lib/Brisk_reconciler_internal.re", line 735, characters 12-235
Called from file "lib/Brisk_reconciler_internal.re", line 1346, characters 14-401
Called from file "lib/Brisk_reconciler_internal.re", line 1090, characters 16-497
Called from file "lib/ListTR.re", line 32, characters 22-40
```

zbaylin:
Fixes #2012. This is a strange bug because it means that the extension host is providing invalid indices, but this should fix it. Sorry I made a faulty assumption!

### DIFF
--- a/src/Feature/SignatureHelp/Feature_SignatureHelp.re
+++ b/src/Feature/SignatureHelp/Feature_SignatureHelp.re
@@ -424,71 +424,77 @@ module View = {
         ~baseFontSize=uiFont.size,
         ~codeBlockFontSize=editorFont.fontSize,
       );
-    let signature: Signature.t = List.nth(model.signatures, signatureIndex);
-    let parameter: ParameterInformation.t =
-      List.nth(signature.parameters, parameterIndex);
-    let renderLabel = () => {
-      switch (parameter.label) {
-      | `Range(start, end_) =>
-        let s1 = String.sub(signature.label, 0, start);
-        let s2 = String.sub(signature.label, start, end_ - start);
-        let s3 =
-          String.sub(
-            signature.label,
-            end_,
-            String.length(signature.label) - end_,
-          );
-        [
-          <Text
-            text=s1
-            fontFamily={editorFont.fontFamily}
-            fontSize={editorFont.fontSize}
-            style={Styles.parameterText(~theme=colorTheme)}
-          />,
-          <View style={Styles.activeParameter(~theme=colorTheme)}>
+    let maybeSignature: option(Signature.t) =
+      List.nth_opt(model.signatures, signatureIndex);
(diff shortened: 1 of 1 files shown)


src/Feature/SignatureHelp/Feature_SignatureHelp.re
```diff
@@ -523,17 +529,19 @@ module View = {
...
+        {switch (maybeParameter) {
+         | Some({documentation: Some(docs), _})
+             when Exthost.MarkdownString.toString(docs) != "" =>
            [
              <horizontalRule theme=colorTheme />,
              <signatureHelpMarkdown markdown=docs />,
            ]
            |> React.listToElement
          | _ => React.empty
          }}
```
bryphe, line 541: You could perhaps decrease nesting by using monadic bind / flatmap, ie:
```
maybeParameter
|> OptionEx.flatMap(({documentation, _}: Parameter.t) => documentation)
|> Option.map(docs => {
    if (Exthost.MarkdownString.toString(docs)) == 0 ? React.empty : 
           [
              <horizontalRule theme=colorTheme />,
             <signatureHelpMarkdown markdown=docs />,
           ]
           |> React.listToElement;
}
|> Option.value(~default=React.empty);
```
Not a big deal here, though, the delta is pretty small.. but a good tool to keep in mind to flatten nested switches / pattern matches.

bryphe:
> This is a strange bug because it means that the extension host is providing invalid indices, but this should fix it.

We should be robust even in cases where we get bad data from the extension host, ie, an extension with bugs shouldn't crash the editor. Our best protection is to use the APIs that don't throw (ie, `List.nth_opt` vs `List.nth`), so that we can model these cases in the type system.

zbaylin:
@bryphe Sounds good, it was an oversight on my part. I'll know better in the future!

Issue #1869, opened 2020-06-03
Cannot run `esy install` - error unable to fetch @opam/ocaml-lsp-server


OlegBrony:
when I run it I get this error: 
```
 $ esy install
info install 0.6.2 (using package.json)
error unable to fetch @opam/ocaml-lsp-server@github:ocaml/ocaml-lsp:ocaml-lsp-server.opam#04733ed:     
      
        source: github:ocaml/ocaml-lsp:ocaml-lsp-server.opam#04733ed
        error: error: command failed: 'git' '-C' '/home/user/programms/oni2/_esy/default/tmp/esy-86f879' 'submodule' 'update' '--init' '--recursive'
               stderr:
                        Cloning into 'vendor/cmdliner'...
                        Cloning into 'vendor/merlin'...
                        fatal: reference is not a tree: 05e32270b74e2650cef7240188dd74632feaf6d9
                        Cloning into 'vendor/ocaml-syntax-shims'...
                        Unable to checkout '05e32270b74e2650cef7240188dd74632feaf6d9' in submodule path 'vendor/merlin'
                        
               stdout:
                        Submodule 'vendor/cmdliner' (https://github.com/dbuenzli/cmdliner.git) registered for path 'vendor/cmdliner'
                        Submodule 'vendor/merlin' (https://github.com/rgrinberg/merlin) registered for path 'vendor/merlin'
                        Submodule 'ocaml-syntax-shims' (https://github.com/ocaml-ppx/ocaml-syntax-shims) registered for path 'vendor/ocaml-syntax-shims'
                        Submodule path 'vendor/cmdliner': checked out 'db4d02a9eb47b5c43127a67cb121004b03ea3719'
                        Submodule path 'vendor/ocaml-syntax-shims': checked out '38a6dd23543d9c4c3678fc6741962832fd539877'
                        
                 fetching dist: github:ocaml/ocaml-lsp:ocaml-lsp-server.opam#04733ed
error: installation error  
  fetching @opam/ocaml-lsp-server@github:ocaml/ocaml-lsp:ocaml-lsp-server.opam#04733ed
esy: exiting due to errors above
```
System: Linux Mint
how can I fix it?

bryphe:
Hi @OlegBrony - this should be fixed now with https://github.com/onivim/oni2/pull/1912 (updating our `ocaml-lsp` dependency). Worth trying the latest to see!

Some other notes:
[...]

PR #2021, opened 2020-06-26
docs(installation): Update esy version in docs

CrossR:
Tiny change to recommend the latest version of `esy` in the docs, rather than specifically `0.6.2` since there was some issues that were fixed in the later version and aligns the docs with the CI (#1869, #1866).

This seems like the best way to prevent us having to change every update, or alternatively we could stick a link in to the CI config to recommend using what we use in the CI.

### DIFF
--- a/docs/docs/for-developers/building.md
+++ b/docs/docs/for-developers/building.md
@@ -10,7 +10,7 @@ sidebar_label: Building from Source
 
 - Install [Git](https://git-scm.com/)
 - Install [Node](https://nodejs.org/en)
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for updating this, @CrossR !

> Tiny change to recommend the latest version of esy in the docs, rather than specifically 0.6.2 since there was some issues that were fixed in the later version and aligns the docs with the CI (#1869, #1866).

Sounds good. Yeah, I wish there was a macro or something where we could change the esy version in one place and have our docs and CI pick it up! But this seems like the best option at the moment.

bryphe:
...and reminded me that we should try the latest `esy`: https://github.com/onivim/oni2/pull/2025

PR #2029, opened 2020-06-27
fix(welcome): Linux / win32 shorcut key info

### LINKED ISSUE #2028: Welcome view shows wrong keybinding for linux / win32
image
Cmd shuld be Ctrl on windows / linux

amiralies:
Closes #2028

### DIFF
--- a/src/UI/WelcomeView.re
+++ b/src/UI/WelcomeView.re
@@ -105,6 +105,12 @@ let animation =
     |> delay(Revery.Time.milliseconds(150))
   );
 
+let cmdOrCtrl =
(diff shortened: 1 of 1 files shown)


amiralies:
Note that i'm not sure if it's the best place to handle this. may be abstracted at some point

bryphe:
Awesome, thanks for fixing this @amiralies !

> Note that i'm not sure if it's the best place to handle this. may be abstracted at some point

Definitely - I'd like to have a keybind-lookup for this, eventually - that way the welcome screen will show the correct keys if they are rebound. But this is great in the meantime.

PR #2032, opened 2020-06-27
feat(layout/#592): layout tabs - part 3 - single tab mode

glennsl:
This adds a "single tab mode" that hides the tab bar and restricts each group to a single tab. The effect of this is that `:q` will always close the split/group even if several buffers have been opened in it.

There is one caveat, however. If single tab mode is enabled when there are multiple tabs in a group, the extra tabs won't disappear immediately. Only when another buffer is opened will they be replaced. Any fix I can think of for this would be excessively complex for something that is designed to always be on or off, not toggled mid-session, so I doubt it's really worth it to fix it.

Addresses #592 - This should be the final piece of the puzzle for this, depending on how complete the support needs to be at this point.

### DIFF
--- a/src/Feature/Layout/Configuration.re
+++ b/src/Feature/Layout/Configuration.re
@@ -58,4 +58,4 @@ let layoutTabPosition =
     ~default=`bottom,
   );
 
-let singleTabMode = setting("oni.layout.singleTabMode", bool, ~default=true);
+let singleTabMode = setting("oni.layout.singleTabMode", bool, ~default=false);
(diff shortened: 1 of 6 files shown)


bryphe:
> This adds a "single tab mode" that hides the tab bar and restricts each group to a single tab. The effect of this is that :q will always close the split/group even if several buffers have been opened in it.

Awesome!

> Any fix I can think of for this would be excessively complex for something that is designed to always be on or off, not toggled mid-session, so I doubt it's really worth it to fix it.

Assuming that's it's always on-and-off seems reasonable. We could potentially improve the UX or enforce this, like "this setting will be changed the next time you open Onivim" - but out of scope for this PR. Thanks for your work on this!

PR #2092, opened 2020-07-12
fix(clipboard): Support pasting into text controls

bryphe:


CrossR:
I think this PR has caused visual block mode to be unusable as the `C-v` binding is now always active (rather than just insert/command mode) so you can't use it in normal mode. Was that intended?

bryphe:
Ah, thanks for catching that @CrossR - definitely not intended. Reverting that part of the change now!

PR #2098, opened 2020-07-13
fix(exthost): Successful install and load extension (flutter/dart) (#2097)

### LINKED ISSUE #2097: Installing extension results in error by parsing package.json (Expected a string, but got [ "workspace" ])
Hey Oni2 Team.

I found an issue according to installing the flutter and dart extensions. The following text describes the scenario and possible solution.

## Scenario:

Install extension using:
`oni2 --install-extension ~/Downloads/Dart-Code.dart-code-3.12.2.vsix
Successfully installed extension: ~/Downloads/Dart-Code.dart-code-3.12.2.vsix`
which is then available at
`ls ~/.config/oni2/extensions 
Dart-Code.dart-code-3.12.2.vsix`

The  extension will not appear in the extension tab in oni2.
While running `oni2 with -f` the terminal shows
`~/.config/oni2/extensions/Dart-Code.flutter-3.12.2.vsix/package.json:
    Expected a string, but got [ "workspace" ]`

## Possible solution
In the package.json of the extension the "extensionKind" key has a list of string as a value. (As described at the visual studio documentation: https://code.visualstudio.com/api/advanced-topics/remote-extensions#:~:text=%22extensionKind%22%3A%20%5B%22workspace,extensions%20fall%20into%20this%20category.)
only 
```"extensionKind": ["workspace"]``` 
```"extensionKind": ["ui"]```
```"extensionKind": ["ui", "workspace"]```
or 
```"extensionKind": ["workspace", "ui"]``` 
are valid options.

Therefore the extensionKind: kind which results in a string ("ui" or "workspace") should be a list of strings. (File: src/Exthost/Extension/Manifest.re)

If you need any additional information do not hesitate to ask for them. I would be glad to help =)

*PS: I also have a working patch which I will upload and link to this issue in a few minutes*

reinies:
Change property extensionKind to list of strings as described in
https://code.visualstudio.com/api/advanced-topics/remote-extensions#:~:text=%22extensionKind%22%3A%20%5B%22workspace,extensions%20fall%20into%20this%20category

The property extensionKind have only
  "extensionKind": ["workspace"]
  "extensionKind": ["ui"]
  "extensionKind": ["ui", "workspace"]
  "extensionKind": ["workspace", "ui"]

Therefore the old way "extensionKind":"workspace" or "extensionKind": "ui"
is not valid anymore and has also be updated in InitData.re

Fixes #2097

### DIFF
--- a/src/Exthost/Extension/Exthost_Extension.rei
+++ b/src/Exthost/Extension/Exthost_Extension.rei
@@ -126,6 +126,12 @@ module Contributions: {
 };
 
 module Manifest: {
+  module Kind: {
+    [@deriving show]
+    type t =
+      | Ui
+      | Workspace;
+  };
   [@deriving show]
   type t = {
     name: string,
@@ -142,16 +148,15 @@ module Manifest: {
     activationEvents: list(string),
     extensionDependencies: list(string),
     extensionPack: list(string),
-    extensionKind: list(string),
+    extensionKind: list(Kind.t),
     contributes: Contributions.t,
     enableProposedApi: bool,
-  }
+  };
 
   let decode: Oni_Core.Json.decoder(t);
 
   let identifier: t => string;
   let getDisplayName: t => string;
-
 };
 
 module Scanner: {
--- a/src/Exthost/Extension/Exthost_Extension.rei
+++ b/src/Exthost/Extension/Exthost_Extension.rei
@@ -142,21 +142,16 @@ module Manifest: {
     activationEvents: list(string),
     extensionDependencies: list(string),
     extensionPack: list(string),
-    extensionKind: kind,
+    extensionKind: list(string),
     contributes: Contributions.t,
     enableProposedApi: bool,
   }
-
-  and kind =
-    | Ui
-    | Workspace;
 
   let decode: Oni_Core.Json.decoder(t);
 
   let identifier: t => string;
   let getDisplayName: t => string;
 
-  module Encode: {let kind: Oni_Core.Json.encoder(kind);};
 };
 
 module Scanner: {
@@ -199,7 +194,7 @@ module InitData: {
       engines: string,
       activationEvents: list(string),
       extensionDependencies: list(string),
-      extensionKind: string,
+      extensionKind: list(string),
       enableProposedApi: bool,
     };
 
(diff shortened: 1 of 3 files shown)


tristil:
@reinies This has to be backward compatible though, so it has to support both list and a string.

reinies:
Hey @tristil,
thanks for your response. Do you have an example which is using this property as a string. So I could test it while developing.
Thanks you =)

reinies:
Just for testing purposes I can change it by hand in the flutter package and try to install it. But is there an extension which have it by default? Maybe from the beginning it was never a string but a list.

reinies:
Ah, no.. prior version 1.4 of the visual studio code release it was only a string. Found on https://code.visualstudio.com/api/advanced-topics/remote-extensions#:~:text=%22extensionKind%22%3A%20%5B%22workspace,extensions%20fall%20into%20this%20category 
``` Note: Prior releases allowed an extension to specify single location as a string and it is deprecated in favor of multiple location support (array).``` 

Therefore you are right. I should change it to support both, list and string.

tristil:
@reinies Here is one that uses the old format https://marketplace.visualstudio.com/items?itemName=Shan.code-settings-sync

reinies:
Thanks @tristil . So I go back to the drawing board and search a way to support both.

bryphe:
Thanks @reinies for the help investigating and fixing this, and @tristil for the review. Really appreciate it!

> Thanks @tristil . So I go back to the drawing board and search a way to support both.

Indeed - this was a little confusing for me too (took me a while to wrap my head around the 'decoder' strategy - the approach we use is inspired by decoders in Elm: https://guide.elm-lang.org/effects/json.html)

There's a `one_of` decoder that takes a bunch of decoders and tries them to see if one works - so it's a good way to handle parsing different kinds of JSON (either a string, or a list of strings) - and then in the single-string case, we can just map it back to a list. I gave this a go here: https://github.com/onivim/oni2/pull/2098/commits/78aaad6f97d5e90e64bcbd530752eb01dc0c7615

bryphe:
Looks like the extension at least gets past parsing now - thanks @reinies and @tristil for the help!

reinies:
@bryphe great work. Love the solution you build *thumbs up*

Issue #1820, opened 2020-05-25
Search the yanked string crashes the app.


dlagg:
If you try to yank a selection, `/` for search, and `Ctrl` + `R` + `"` for paste the yanked string. The app crashes. I don't know if `Ctrl` + `R` keybind is implemented yet.

mabuchner:
Just entering insert mode and pressing `<C-R>` will also freeze Oni2.

1. `oni2 foo.txt`
2. `i`
3. `<C-R>`

In Vim I'm using this feature a lot, because it allows to paste a register without having to leave insert mode.

bryphe:
Fixed with #2101

PR #2101, opened 2020-07-13
fix(vim/#566): Insert register crashes

### LINKED ISSUE #566: Crash: Insert register in insert mode
__Repro:__
- Go to insert mode `i`
- `<C-r>a`

Expected: Contents of a register should be insert
Actual: Crash

This is hitting a blocking input mode in `libvim` - we'll have to fix it there

bryphe:
__Issue:__ `<C-R>` (insert register) in insert mode or command line would crash the editor.

__Defect:__ In libvim, we'd enter a blocking input call, which would cause Onivim to hang. 

__Fix:__ We already have an API exposed on libvim to get register contents - we'll use that to get the contents of the register and handle insertion in a similar way to paste. This adds a `Feature_Registers` to handle the state management of the register, along with some effects / subscriptions for vim to get registers and to subscribe to evaluation results.

Fixes #566 #1820 

Evaluation of registers (`=`) will be improved with https://github.com/onivim/libvim/pull/214

__TODO:__
- [x] Verify #1820 (`/` and `?`) are also fixed


mabuchner:
Does this also fix #1820?

bryphe:
Oh, good catch @mabuchner - it should! I'll double-check and make sure it's also addressed.

Issue #941, opened 2019-11-18
Can't apply normal commands across range


instantaphex:
OS: Windows 10
Build Info:
Version: 2.0.2.0
Date: 11/16/2019, 8:35:36 PM
Commit: ad3d45b

I'm not able to apply normal commands across a range of lines.

Steps to reproduce:

1.) Example fille:

```
console.log('line1');
console.log('line2');
console.log('line3');
```

2.) Enter ex mode commands:

`
:1,3 normal ^i//
`

3.) This shows error:

`
libvim
E35: No previous regular expression
`

3.) Expected output:

```
//console.log('line1');
//console.log('line2');
//console.log('line3');
```

PR #2113, opened 2020-07-16
refactor(dependencies): upgrade revery to 7ccd20c

zbaylin:
This is a fairly big migration because the text/font API changed pretty widely, but nothing too breaking.


src/Service/Font/FontLoader.re
```diff
@@ -5,150 +5,248 @@
...
+
+let isFontMonospace = font => {
+  let metrics = Skia.FontMetrics.make();
+  Skia.Paint.setTypeface(paint, font |> Revery_Font.getSkiaTypeface);
+  Skia.Paint.getFontMetrics(paint, metrics, 1.0) |> (ignore: float => unit);
+  let avgCharWidth = Skia.FontMetrics.getAvgCharacterWidth(metrics);
+  let maxCharWidth = Skia.FontMetrics.getMaxCharacterWidth(metrics);
+
+  maxCharWidth -. avgCharWidth < 0.1;
 };
```
bryphe, line 153: Nice!

src/Service/Font/FontLoader.re
```diff
@@ -5,150 +5,248 @@
...
-        r,
-        font => {
-          let (isMono, c1width, _) =
-            isMonospace(~font, ~smoothing, ~fontSize);
-          if (!isMono) {
-            Error("Not a monospace font");
-          } else {
-            let measuredWidth = c1width;
-            let {lineHeight, descent, _}: Revery.Font.FontMetrics.t =
-              Revery.Font.getMetrics(font, fontSize);
```
bryphe, line 61: Great to see all this go away - thank you for cleaning it up!

src/Core/Kernel/dune
```diff
@@ -5,7 +5,7 @@
         yojson
         ppx_deriving.runtime
         ppx_deriving_yojson.runtime
-        zed_oni
+        Revery.zed
```
bryphe, line 8: Nice 👍

src/Service/Font/FontLoader.re
```diff
@@ -5,150 +5,248 @@
...
+       (
+         ~requestId: int,
+         ~smoothing: Revery.Font.Smoothing.t,
+         ~fontFamily: Revery_Font.Family.t,
+         fullPath,
+         fontSize: float,
+       ) => {
+     Log.tracef(m =>
+       m("loadAndValidateEditorFont path: %s | size: %f", fullPath, fontSize)
+     );
```
bryphe, line 33: Once we verify everything is working - we can just remove this

bryphe:
It looks like the inline tests are failing: `esy '@test' inline`:
https://dev.azure.com/onivim/oni2/_build/results?buildId=12535&view=logs&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7&t=2a220312-72c6-56b5-f19f-728b78c47903&l=24

I suspect it might be due to assuming an asset is available - the inline tests are run from a directory that might not have access to the font files. Is there a place in the top-level / global scope where we are trying to load a font?

zbaylin:
@bryphe Hmm, I dont remember changing any of the font loading paths, but I can go through the diff. I can also run the inline tests locally to see if I get any more information

bryphe:
> @bryphe Hmm, I dont remember changing any of the font loading paths, but I can go through the diff. I can also run the inline tests locally to see if I get any more information

Could there be a change in Revery that depends on asset paths? Something like this? https://github.com/revery-ui/revery/blob/b05d4a97842a571ca1f07b18616aab313abb46e0/src/Font/FontFamily.re#L93

I was a bit concerned with how big the delta of changes in Revery is - it would be easier to isolate the failures if we upgrade incrementally (ie, merge #937, then #941, then #932 , then #944) . It's still an option for us if we get stuck on what's causing this failure - it doesn't necessarily have to be all-or-nothing upgrade.

zbaylin:
Im getting this error locally, which is even weirder:

```
   generator src/Core/BuildInfo.re (exit 2)
(cd /Users/zbaylin/Development/oni2/_esy/default/store/b/oni2-534abfdd/default/src/Core && ../gen_buildinfo/generator.exe)
Fatal error: exception Sys_error("../../package.json: No such file or directory")
```

I dont know what I could have changed to trigger that -- it might not even be the same error as the one on CI

bryphe:
I've seen that error intermittently, I'm not sure what causes it - you can try deleting your `_esy` folder and rebuilding.

bryphe:
It looks like the monospace detection isn't working properly, as I'm able to set the font family with Arial with this change:
image

In addition, if I set to an invalid font - like `"editor.fontFamily": "Some-Invalid-Font"` - it defaults to a non-monospaced font.

Here's a set of test cases we need to verify cross-platform:
---------------------------------------------------------
| Test Case | OSX | Linux | Windows |
| --- | --- | --- | --- |
| No font set - should default to JetBrains Mono | [ ] | [ ] | [ ] |
| Set to Arial at runtime (modifying existing config), should default to JetBrains Mono | [ ] | [ ] | [ ] |
| Startup with Arial already set in config, should default to JetBrains Mono | [ ] | [ ] | [ ]|
| Set to per-platform monospace font (Consoles/Menlo/Liberation/etc) modifying existing config | [ ] | [ ] | [ ] |
| Start up with per-platform monospace font(Consoles/Menlo/Liberation/etc) | [ ] | [ ] | [ ] |
| Set to an invalid font at runtime (`"editor.fontFamily": "An-Invalid-Font"`) | [ ] | [ ] | [ ] |
| Startup with an invalid font (`"editor.fontFamily": "An-Invalid-Font"`)  | [ ] | [ ] | [ ] |

zbaylin:
@bryphe Thanks for testing that! I probably just have the range set too high for the difference between the average and maximum character widths. Sorry about that!

bryphe:
> I probably just have the range set too high for the difference between the average and maximum character widths.

Maybe we should just use `Float.epsilon`? Is that too strict?

zbaylin:
> > I probably just have the range set too high for the difference between the average and maximum character widths.
> 
> Maybe we should just use `Float.epsilon`? Is that too strict?

Good question. I'll try that out.

zbaylin:
Looks like both average and max char widths are always being returned as the same:
```
[DEBUG]    +0ms Oni2.Service.FontLoader : Testing for monospace-ness: Arial, maxCharWidth: 31.535156, avgCharWidth: 31.535156
```
For now I'll revert to the "old" way of testing for monospace fonts and try to investigate why that is happening on the Revery side.

zbaylin:
@bryphe I can get all those tests done on Linux and macOS -- Windows might be a bit harder for me. If you or @CrossR would be able to help out there that would be super helpful!

zbaylin:
The startup tests are hard to show, but they work the same way as these runtime ones do:

## macOS

- Nothing set
Screen Shot 2020-07-16 at 2 00 11 PM

- "Arial" set
Screen Shot 2020-07-16 at 2 00 18 PM

- "Incorrect-Font-Name" set
Screen Shot 2020-07-16 at 2 00 41 PM

- Raw path set
Screen Shot 2020-07-16 at 2 02 10 PM

- Incorrect raw path set
Screen Shot 2020-07-16 at 2 02 21 PM

bryphe:
Thanks @zbaylin for investigating! I'll run the tests on Windows 👍

zbaylin:
Looks like there's a memory leak issue with Skia -- @bryphe what do you think we should do about this? I can't find an `sk_string_deref` anywhere, it looks like it should be taken care of with the garbage collector/`free`.

zbaylin:
Linux seems to work as well!

## Linux
- Nothing set
Screen Shot 2020-07-16 at 3 27 07 PM

- "Arial" set
Screen Shot 2020-07-16 at 3 27 45 PM

- "Random-Font-Name" set
Screen Shot 2020-07-16 at 3 27 59 PM

- Raw path set
Screen Shot 2020-07-16 at 3 29 47 PM

- Garbage raw path set
Screen Shot 2020-07-16 at 3 30 14 PM

CrossR:
## Windows
- Nothing set
nothing


- "Arial" set
arial

- "Random-Font-Name" set
Fake

- Raw path set
Path

- Garbage raw path set
BadFontPath

bryphe:
Thanks for running those tests, @CrossR ! Appreciate it 👍 

I double-checked perf too, and it looks like we still have the same 'hot-paths':
image

- `drawText`
- Canvas flushing
- Garbage collection

so doesn't look like any perf regressions

bryphe:
Nice catch on the memory leak @zbaylin - once we're merged up with that, seems like this is good to go 👍

PR #2119, opened 2020-07-16
feat(syntax-highlighting/#1829): update syntax highlight lookup to use additional information

CrossR:
In progress at the mo, but this should fix some of the holes in our syntax highlighting. VSCode gives us a bunch of ways to work out the language for a file and right now we only use the extension (well...not really I think we use that badly as well), so this adds loading the `"filenames"` field that the extensions gives which fixes extensionless files (say a Gemfile in Ruby etc) and files that maybe have different names than expected.

I've also populated two of the other fields (firstline and patterns) but I/someone else can deal with that in a follow up, since I haven't looked at our regex compilation stuff really. EDIT: This was actually a lot easier than I thought, so I've added that too.

Need to figure out the best way to do this yet and may need a discussion around loading order/investigate what VSCode does.


src/Store/Features.re
```diff
@@ -10,6 +10,16 @@ module Internal = {
...
+  /* TODO: I always use the fileType from Core.Buffer.getFileType, which is bad!
+   *
+   *       This should now take into account the full filename, or the extension.
+   *       Should be future proof enough to do "ext or filename or pattern match or
+   *       first line or default."
+   *
+   *       I think this actually might be causing the current issues with .bashrc etc
+   *       now as well. .bashrc should work, but I expect the filetype is empty,
+   *       and this doesn't use the extension (which would work).
+   */
```
CrossR, line 22: This is the main open question really.

With this PR we have three ways of getting a scope back:

 - File Type
 - Extension (already here but not used as far as I can see?)
 - Filename

So this should be extracted into a suitable "try a, try b, try c" etc, but ordering there obviously matters... In say a `CMakelists.txt` from the extension its a `txt` file, but from the filename its a `cmake` file. Gets worse later with the two regex options which will be more expensive to run, so I'm a bit confused about the best ordering.

src/Exthost/Extension/Contributions.re
```diff
@@ -371,6 +374,9 @@ module Language = {
         {
           id: field.required("id", string),
           extensions: field.withDefault("extensions", [], list(string)),
+          filenames: field.withDefault("filenames", [], list(string)),
```
CrossR, line 377: I want to look up what vscode does when loading these. The Ruby file defines `gemfile` as the filename, but most projects seem to use `Gemfile` (i.e. capitalised) so a straight look up won't work. Then the shell one defines them in all caps.

Should we just lower case/upper case all file names and do the same to the filenames? Or is there a reason to not do that?

CrossR:
Tests also to come, but I think the important tests are around what I've just commented so we know the order is what we want expect and the filenames match as expected.

src/Exthost/LanguageInfo.re
```diff
@@ -159,15 +172,19 @@ let getTreesitterPathFromScope = (li: t, scope: string) => {
   li.scopeToTreesitterPath |> StringMap.find_opt(scope) |> Option.join;
 };
 
-let _getLanguageTuples = (lang: Contributions.Language.t) => {
+let getLanguageTuples = (lang: Contributions.Language.t) => {
   List.map(extension => (extension, lang.id), lang.extensions);
 };
 
-let _getGrammars = (extensions: list(Scanner.ScanResult.t)) => {
```
CrossR, line 166: (Dropped the `_` from the other functions since this file has an interface file so)

src/Exthost/Extension/Contributions.re
bryphe: Thanks for looking at this, @CrossR !

> The Ruby file defines gemfile as the filename, but most projects seem to use Gemfile (i.e. capitalised) so a straight look up won't work. 

Good catch.

> Should we just lower case/upper case all file names on loading and do the same to the filenames? Or is there a reason to not do that?

I don't think we want to change the actual filename / path - as some filesystems are case sensitive, so it would change the file location. But we should do a case-insensitive compare when we check these fields - we could normalize both the 'filenames' entry and the file we check to lowercase-ascii when we check to see if this filenames field matches.

src/Store/Features.re
bryphe: I guess we could even use the first line, too!

We actually are setting the buffer file type here: https://github.com/onivim/oni2/blob/d9efc5b08ba63caa0bcd15f0c7e8d569d5a034de/src/Store/VimStoreConnector.re#L297 - we could potentially revise that logic, to use our new filetype resolution logic here - so that we can trust the filetype.

bryphe: 1) We could update our `getLanguageFromBuffer` method in LanguageInfo to take into account the extension, filename, or even the first line (since we could check the first line with the `Buffer.t` / `Buffer.getLine`)

2) It looks like the precendence order is:
- Extensions
- Exact file name match
- First line match

https://github.com/microsoft/vscode/blob/4c99db1a05687a67c1d214361946b18e5f65e0cf/src/vs/base/common/mime.ts#L136

For now - I think we could apply that logic. Given a buffer, we'll run each of those rules, and see if there is a match (first, for extension, then for file name, then for first line) - if so, we'll apply that filetype.

src/Core/Utility/Path.re
```diff
@@ -38,3 +38,15 @@ let trimTrailingSeparator = path => {
...
+  let fileName = Rench.Path.filename(path);
+
+  if (String.length(fileName) == 0) {
+    "";
+  } else if (fileName.[0] == '.') {
+    fileName;
+  } else {
+    Rench.Path.extname(fileName);
+  };
+};
```
CrossR, line 52: So I was wrong! The reason that the `.bashrc` and a few other things weren't working was because of a conflict between `Rench.Path.extname` and what VSCode considers an extension.

VSCode considers `.bashrc` a file extension (to an unnamed file I guess, I don't follow the logic and assumed that `.bashrc` would have been a `filename` 🤷 Do people have files with it as an extension? ), whereas Rench would return an empty string for `.bashrc`.

Now I just return the full filename if the first character is a `.`, which fixes the issue but not sure if there is some edge case I'm missing.

bryphe: > VSCode considers .bashrc a file extension (to an unnamed file I guess, I don't follow the logic and assumed that .bashrc would have been a filename 🤷 Do people have files with it as an extension? ), whereas Rench would return an empty string for .bashrc.

Oh, nice catch @CrossR ! That would indeed be a problem. It does seem counter-intuitive (I wouldn't expect '.bashrc' to be considered an extension either) - but being consistent with how VSCode evaluates would be the correct behavior here. This logic looks reasonable to me

src/Exthost/LanguageInfo.re
```diff
@@ -78,14 +90,34 @@ let getLanguageFromFilePath = (li: t, fp: string) => {
...
-  | Some(v) => getLanguageFromFilePath(li, v)
-  };
+  let filePath =
+    switch (Buffer.getFilePath(buffer)) {
+    | None => ""
+    | Some(v) => v
+    };
+
+  getLanguageFromFilePath(li, filePath);
+  // TODO: this should then flow into updateIfDefault(firstLineCheck);
```
bryphe, line 120: Perfect! We can check if `Buffer.getNumberOfLines` is greater than 0 - and if it is, we can check `Buffer.getLine(0, buffer) |> BufferLine.raw` to get the raw string, and validate against the regular expression.

We can create a test case too, where we could create some buffer like:
```
let testBuffer = Buffer.ofLines([|"A random first line to test against"|])
|> Buffer.setFilePath(Some("file-path.txt"));
```

And validate that `getLanguageFromBuffer` returns the correct thing. 
Probably the trickiest part of testing this is getting a valid `LanguageInfo.t` to use with it - we could load one of our JSON files like:
```
let languageInfo =
Exthost.Extension.Scanner.load(~category=Exthost.Extension.Scanner.Default, "extensions/shellscript/package.json")
|> Option.to_list
|> Exthost.LanguageInfo.ofExtensions
```
Probably the easiest way, right now!

bryphe: Oh, and I forgot to mention, if you are adding tests to `test/Exthost` to validate this - probably the quickest way is to run:
```
esy '@test' exthost --filter LanguageInfo
```
The exthost tests take a while for the ones that actually start up the extension host - this would filter those out and scope it to just the language tests, which would be much quicker and easier to iterate on.

src/Store/Features.re
CrossR: Based on this: https://github.com/microsoft/vscode/blob/4c99db1a05687a67c1d214361946b18e5f65e0cf/src/vs/base/common/mime.ts#L194-L207

I've gone with the following order, where the first of these to match is used:

 - Oni2 filename Regex (I don't really know what this is...I can't seem to see it used anywhere?)
 - Exact file name match
 - Pattern file name match (i.e. a file name regex)
 - Extension match
 - First line match

That seems to work for the more odd cases like `CMakeLists.txt` or `tsconfig.json` (which is a JSON file, defined as a `jsonc` file in the TS `package.json`).

CrossR:
The actual logic here is all sorted out now (except some confusion/commented out tests around the `oni://` filenames).

I need to look over the actual VimStore bits now, since we will still fail for buffers with no extension and only a first line, as we make the initial buffer only based on the file name function, not the buffer function (so we do not have access to the first line at that point). There is I think a few other places that also default to using the file name based function, which should now be swapped to the buffer one (i.e. stuff like `onFileNameChanged` in the Vim store, which if left as is would swap to `plaintext` even if there is a matching first line).

The other is just a gap in my knowledge I think...I've made `patternLanguagePair` for the `pattern`/`language` tuples we get from the file name/first line regexs. The `pattern` is currently a string, but really I think it should just be compiled into a Oniguruma regexp straight away and stored as `Result(Onigurma.OnigRegExp.t)`/`string` pair, not a `string`/`string` pair. This is being run quite a lot (every open file, every visible file in the explorer etc) so we probably don't want to be doing that `create` every single time versus the once. I was getting an error when I tried to use that though, I think due to me not fully understanding what I need to give `ppx_deriving`. It was an Invalid Arg error due to a Custom datatype I believe.

src/Core/Utility/Path.re
```diff
@@ -38,3 +38,27 @@ let trimTrailingSeparator = path => {
...
+
+let%test_module "getExtension" =
+  (module
+   {
+     let%test "Simple file" = getExtension("/home/oni/test.md") == ".md";
+     let%test "No file name, only extension" =
+       getExtension("/home/oni/.bashrc") == ".bashrc";
```
bryphe, line 62: Nice!

src/Store/VimStoreConnector.re
```diff
@@ -173,7 +173,7 @@ let start =
       let fileType =
         switch (meta.filePath) {
         | Some(v) =>
-          Some(Exthost.LanguageInfo.getLanguageFromFilePath(languageInfo, v))
+          Some(Exthost.LanguageInfo.getLanguageFromFilePath(languageInfo, v)) // TODO: I should be fromBuffer..but I use metadata right now.
```
bryphe, line 176: I think this is reasonable, for now - this code is hit when the filename changes

src/Store/VimStoreConnector.re
```diff
@@ -294,6 +294,8 @@ let start =
         let fileType =
           switch (metadata.filePath) {
           | Some(v) =>
+            // TODO: Can't be changed, as buffer is only made a few lines below.
+            // But we probably could re-run once the buffer is made, in BufferEnter?
             Some(
               Exthost.LanguageInfo.getLanguageFromFilePath(languageInfo, v),
             )
```
bryphe, line 301: I believe we could move this to after we create the buffer, but before we `setFileType` - I think that would work

test/Exthost/LanguageInfoTest.re
```diff
@@ -0,0 +1,104 @@
+open Oni_Core;
+open TestFramework;
+
+// Language info, with multiple loaded extensions to check order of loading
+// works.
+let li =
```
bryphe, line 6: Excellent!

bryphe:
> I need to look over the actual VimStore bits now, since we will still fail for buffers with no extension and only a first line, as we make the initial buffer only based on the file name function, not the buffer function (so we do not have access to the first line at that point). There is I think a few other places that also default to using the file name based function, which should now be swapped to the buffer one (i.e. stuff like onFileNameChanged in the Vim store, which if left as is would swap to plaintext even if there is a matching first line).

Cool! I left a comment about the initial buffer case - I think we can shim in the file type detection after we create the buffer, but before setting `setFileType`.

CrossR:
Waiting on #2159 now.

I've added a test for the Oni2 file name now, though it seems to only be used to parse the `oni://terminal/...` into  a `terminal` filetype, is that correct? All other file names of that format fail the regex, like `oni://Welcome`, as the regex is lowercase only and expects a `/` at the end. That might be by design though.

Once the linked PR is merged, I also should do a quick test of the 4 file types as well again. I did notice a few odd results where the filetype was set correctly, but the syntax highlights weren't showing (`dockerfile` is where I noticed it with the official extension for it installed, but I don't think that is a blocker as I expect that is linked to the same issues in HTML files etc that the current release has and its probably linked to the `tree-sitter` scopes issue too.)

src/Exthost/LanguageInfo.rei
```diff
@@ -12,12 +12,12 @@ let toString: t => string;
...
 let defaultLanguage: string;
 
-let getLanguageFromExtension: (t, string) => string;
 let getLanguageFromFilePath: (t, string) => string;
 let getLanguageFromBuffer: (t, Buffer.t) => string;
 
 let getScopeFromLanguage: (t, string) => option(string);
-let getScopeFromExtension: (t, string) => option(string);
+let getScopeFromFileName: (t, string) => option(string);
+let getScopeFromBuffer: (t, Buffer.t) => option(string);
```
CrossR, line 20: Dropped the extension based functions from this interface file, as they aren't used and using just the extension gives us less info that the filename.

The functions still exist or are easy to add so if we do end up needing it, its no issue, but I think having them accessible will mean that they get used more readily when the filename itself gives us 3 ways of determining the language vs 1 way with the extension.

CrossR:
Few other comments:

 - It is currently failing to pick up the first line files (testing with a shellscript with no ext), despite the first line looking by eye the same as the test I've added...which is odd. Looking into now.
 - We may want to cache these results? Moving my mouse in a buffer causes every open buffer to go through `getLanguageFromBuffer`. I'm not sure its necessary that it recalculates it every time I move my mouse, so maybe we should stop it being polled so much, otherwise add a lookup for it since hitting regexes etc on every move of the mouse is a bit pointless (especially for things like `oni://Welcome` or the changelog which fails all 5 tests repeatedly).

src/Exthost/LanguageInfo.rei
bryphe: Cool! Ya, good to get rid of these - less to think about, and it's likely that using them won't give the complete picture with the new logic 👍

bryphe:
> Looks like the buffer isn't correctly populated where I've moved the filetype detection, must need to be a bit later again?
The first run is using "" as the first line for this shell file:

Ya, I think initially we get an empty buffer - and then it gets populated. Maybe we could check every buffer update?

> We may want to cache these results? Moving my mouse in a buffer causes every open buffer to go through getLanguageFromBuffer. 

Caching would be ideal.. it seems like w're calling this a few places:
https://github.com/onivim/oni2/blob/5dfaf36793dc006d267636cc330a7ab876f655b0/src/Feature/Hover/Feature_Hover.re
https://github.com/onivim/oni2/blob/5dfaf36793dc006d267636cc330a7ab876f655b0/src/Model/Selectors.re#L22

as well as in the editor view.

Previously - we had it cached on the `Buffer`: https://github.com/onivim/oni2/blob/5fe5cd718dcedd2735498765311487e8940ad449/src/Core/Buffer.re#L14

I think the ideal flow would be:
- On buffer enter, check the name / path - and call `Buffer.setFileType` to set the inferred type
- On buffer update, if the first line changes, call `getLanguageInfoForBuffer` and see if it's different than before.

But the source of truth for the language would be `Buffer.getFileType`. Would that work?

CrossR:
> I think the ideal flow would be:
>
>   On buffer enter, check the name / path - and call Buffer.setFileType to set the inferred type
>   On buffer update, if the first line changes, call getLanguageInfoForBuffer and see if it's different than before.
>
> But the source of truth for the language would be Buffer.getFileType. Would that work?

That works great! There is still 2 issues, but I think it can be left until a follow up. I've tested:

 - Extensions that worked already (Bunch of random files in Oni2)
 - Extensions that didn't work (`.bashrc` and `.bash_profile` tested)
 - Filetype matches (`Gemfile` and `Dockerfile`)
 - File pattern (`tsconfig-base.json` which is a jsonc file, despite the ext)
 - First line matches (`shell` and `python` checked)

These all work great! The two issues I've found still are:

 - Changes whilst you are in the buffer don't work. I.e. open a `.txt` file with a shebang at the top and it'll be correctly picked up. Change that from `#!/bin/bash` to `#!/bin/python` and the icon and filetype will change correctly, but not the syntax highlights. Leave the buffer and re-enter and its okay again.
 - The `TODO` I've put in relation to filename change. If you have that same `.txt` file with a shebang at the top, you open it and its a `shellscript` file. If you were to `:saveas my_new_file.txt` the filetype would change to `plaintext` as it goes through that code and updates it.

I'm not sure they are a big enough blocker to stop this when the rest is working well. The second shouldn't be too bad I don't think, so I can follow up with that later.

bryphe:
This is awesome - picks up a lot of places where we had syntax highlighting from our bundled extensions, but weren't using - like the git merge/commit files!

PR #2155, opened 2020-07-21
feat(editor): folder drag and drop

zbaylin:
As of right now, we only support dropping _files_ onto the editor. This adds the behavior that if you drop a folder/directory onto the editor, it will change the directory in the same way that `:cd x/y/z` changes the directory.

### DIFF
--- a/src/Model/Actions.re
+++ b/src/Model/Actions.re
@@ -183,7 +183,7 @@ type t =
   | Pane(Feature_Pane.msg)
   | PaneTabClicked(Feature_Pane.pane)
   | PaneCloseButtonClicked
-  | VimDirectoryChanged(string)
+  | DirectoryChanged(string)
   | VimExecuteCommand(string)
   | VimMessageReceived({
       priority: [@opaque] Vim.Types.msgPriority,
--- a/src/Store/Features.re
+++ b/src/Store/Features.re
@@ -561,10 +561,10 @@ let update =
   | FilesDropped({paths}) =>
     let eff =
       Service_OS.Effect.statMultiple(paths, (path, stats) =>
-        if (stats.st_kind == S_REG) {
-          OpenFileByPath(path, None, None);
-        } else {
-          Noop;
+        switch (stats.st_kind) {
+        | S_REG => OpenFileByPath(path, None, None)
+        | S_DIR => DirectoryChanged(path)
+        | _ => Noop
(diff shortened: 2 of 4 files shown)


src/Model/Actions.re
```diff
@@ -183,7 +183,7 @@ type t =
   | Pane(Feature_Pane.msg)
   | PaneTabClicked(Feature_Pane.pane)
   | PaneCloseButtonClicked
-  | VimDirectoryChanged(string)
+  | DirectoryChanged(string)
```
zbaylin, line 186: I debated whether or not I should add a duplicate variant to distinguish the vim  `:cd` from other possible directory changes, but looking where this is used, it doesn't seem super specific to vim. Let me know if you think I should change it.

bryphe: Seems reasonable - it is more general than Vim!

src/Store/Features.re
```diff
@@ -561,10 +561,10 @@ let update =
   | FilesDropped({paths}) =>
     let eff =
       Service_OS.Effect.statMultiple(paths, (path, stats) =>
-        if (stats.st_kind == S_REG) {
-          OpenFileByPath(path, None, None);
-        } else {
-          Noop;
+        switch (stats.st_kind) {
+        | S_REG => OpenFileByPath(path, None, None)
+        | S_DIR => DirectoryChanged(path)
```
bryphe, line 566: One thing I'd expect in this case is that we'd `chdir` to change the working directory of the process, too. Dispatching the `DirectoryChanged` event tells our internal state the directory is changed, but doesn't actually change the working directory of the process.

We'd probably want to use `Luv.Path.chdir` to change the work directory, and then - https://github.com/aantron/luv/blob/bb11498de65a2acb56a7269801c1d2ba41672179/src/path.mli#L18 - if that is successful, dispatch the `DirectoryChanged` event.

bryphe: If the working directory isn't in sync - our file explorer and stuff will show the 'correct' directory, but when running vim commands like `:e some/file` it would be relative to the previous directory.

zbaylin: That's a great point, I hadnt thought of that. Thanks! Will fix now.

zbaylin: Just fixed in fe2985c!

PR #1372, opened 2020-02-26
Fix #1364 - Characters overlapping, cursor not positioned correctly

bryphe:
__Issue:__ On some platforms, the text rendering for the editor surface was not correct - characters would be overlapping, or the cursor position would be off.

__Defect:__ The way we measured the text, and the rendering strategy, could be out of sync. We relied on some serendipity for the same subpixel antialias settings to be used. The anti-alias settings can impact the measured width of character glyphs, therefore, we need to be sure we're measuring with the same anti-alias settings as we use to render.

__Fix:__ Wire up a `Font.Smoothing.t` parameter in Revery, and ensure we're using the same setting for rendering and for measuring. In addition, a configuration parameter - `editor.fontSmoothing` - has been exposed so that it can be customized or modified if the user would like a different text rendering strategy.

I also made an opportunistic refactoring to move our `FontStoreConnector` to a subscription - this will make it easier to re-use for other scenarios (ie, terminal fonts).

Fixes #1364 , at least on all the platforms I tested (Windows, OSX, CentOS 6)

Issue #1364, opened 2020-02-23
Weird font display - characters overlapping


pkariz:
Here's a video of the issue: https://easyupload.io/aqkume

I didn't have this issue around 2 months ago when i last used onivim (not sure which version was that).

After running `oni2 --debug -f --log-filter "*Font*"`:
```
[INFO]   +181ms Revery.FontCache : Loaded : /tmp/.mount_OnivimJPlEpk/usr/bin/FiraCode-Regular.ttf
[DEBUG]    +8ms Oni2.Store.Font : Measured width: 8.000000 
[DEBUG]    +0ms Oni2.Store.Font : Line height: 16.799999 
[DEBUG]   +11ms Oni2.Store.Font : Measured width: 8.000000 
[DEBUG]    +0ms Oni2.Store.Font : Line height: 16.799999 
[INFO]    +56ms Revery.FontCache : Loaded : FontAwesome5FreeSolid.otf
[INFO]     +8ms Revery.FontCache : Loaded : Inter-UI-Medium.ttf
[INFO]     +7ms Revery.FontCache : Loaded : Inter-UI-Regular.ttf
[INFO]     +5ms Revery.FontCache : Loaded : FiraCode-Regular.ttf
[INFO]    +10ms Revery.FontCache : Loaded : seti.ttf
[INFO]    +10ms Revery.FontCache : Loaded : Inter-UI-Italic.ttf
```
After setting `"editor.fontFamily": "DejaVu Sans Mono",`:
```
[INFO]   +197ms Revery.FontCache : Loaded : /tmp/.mount_OnivimjeZkVv/usr/bin/FiraCode-Regular.ttf
[DEBUG]    +8ms Oni2.Store.Font : Measured width: 8.000000 
[DEBUG]    +0ms Oni2.Store.Font : Line height: 16.799999 
[DEBUG]    +8ms Oni2.Store.Font : Discovering font: DejaVu Sans Mono
[DEBUG]   +19ms Oni2.Store.Font :   at path: /usr/share/fonts/truetype/dejavu/DejaVuSansMono.ttf
[INFO]     +0ms Revery.FontCache : Loaded : /usr/share/fonts/truetype/dejavu/DejaVuSansMono.ttf
[DEBUG]    +0ms Oni2.Store.Font : Measured width: 8.000000 
[DEBUG]    +0ms Oni2.Store.Font : Line height: 16.296875 
[...]

PR #2171, opened 2020-07-23
fix(editor): correctly render variable width characters

zbaylin:
This is important for something like CJK fallback, where the fallback font is not guaranteed to be (and is often not) monospace.

__Demo__
Current (`master`):
image

This branch:
image

Issues:
- [x] Incorrect spacing with non-default font sizes
    - I was forgetting to update the size and the smoothing 😳 
- [x] Crash on startup
    - I was trying to traverse past the list. Now the while loop has a condition to make sure this doesn't happen.
- [x] Click coordinates incorrect
    - I needed to update `Editor.bufferLineToPixel`
- [x] Incorrect space character size
    - We were measuring two spaces rather than one
- [x] Indentation guides not lining up
    - Use space character rather than `'x'`
- [x] Characters disappearing before Canvas end
    - Calculate the characters to render within based on pixels
- [x] Buffer doesn't update with the new font size
    - Update the Buffer's font when the Editor receives a font update
- [x] Line numbers overflowing
    - Use the width of an underscore rather than the width of a space


bryphe:
Awesome @zbaylin ! This is really exciting... did not expect to support arbitrary-width fonts already 🎉 

I tried out the branch and I'm seeing some issues, though:

__Issue 1:__
image

Looks like some of the measurements are off / not working correctly - and the cursor position isn't consistent.

__Issue 2:__

After trying to switch my font to Comic Sans and Arial, I'm hitting a crash on startup:
```
[DEBUG]    +0ms Oni2.Core.BufferLine : MeasurementCache : Hit! Typeface : Helvetica, Uchar:   (32)
[DEBUG]    +0ms Oni2.Core.BufferLine : resolveTo loop: glyphStringByte : 102, pixelPosition : 253.535156, glyphNumber : 64
[DEBUG]    +0ms Oni2.Core.BufferLine : Continuing on current glyphString
[DEBUG]    +0ms Oni2.Core.BufferLine : MeasurementCache : Hit! Typeface : Helvetica, Uchar: ] (93)
[DEBUG]    +0ms Oni2.Core.BufferLine : resolveTo loop: glyphStringByte : 104, pixelPosition : 256.869141, glyphNumber : 15
[DEBUG]    +0ms Oni2.Core.BufferLine : Reached end of current glyphString
[ERROR]    +0ms Oni2.Exception : Exception (Failure hd):
Raised at file "stdlib.ml", line 29, characters 17-33
Called from file "list.ml" (inlined), line 30, characters 10-23
Called from file "src/Core/BufferLine.re", line 165, characters 35-57
Called from file "src/Core/BufferLine.re" (inlined), line 277, characters 2-44
Called from file "src/Feature/Editor/BracketMatch.re", line 51, characters 8-76
Called from file "src/Feature/Editor/BracketMatch.re", line 150, characters 7-190
Called from file "src/Feature/Editor/BracketMatch.re", line 176, characters 10-155
Called from file "list.ml", line 121, characters 24-34
Called from file "src/Feature/Editor/Editor.re", line 164, characters 2-149
Called from file "src/Feature/Editor/EditorSurface.re", line 217, characters 8-186
Called from file "lib/Brisk_reconciler_internal.re", line 520, characters 10-109
Called from file "lib/Brisk_reconciler_internal.re" (inlined), line 559, characters 6-45
Called from file "lib/Brisk_reconciler_internal.re", line 767, characters 12-59
Called from file "lib/Brisk_reconciler_internal.re", line 1346, characters 14-401
```

I think we'll want to come up with a test matrix for this, and verify it works cross-platform - like we did for https://github.com/onivim/oni2/pull/2113

zbaylin:
Sounds great @bryphe. I wasn't sure if this was ready yet, so thanks for putting it through the wringer! 

What font were you using for the first issue? The second one I got once but I couldn't reproduce it, so I'll look into trying.

bryphe:
> Sounds great @bryphe. I wasn't sure if this was ready yet, so thanks for putting it through the wringer!

I was excited to try it out 😄 

> What font were you using for the first issue? 

I actually hadn't set an `editor.fontFamily` at all - so it should be using JetBrains Mono unless that changed.

zbaylin:
@bryphe I fixed the second issue in d7509b7 by adding a condition to the while loop that the glyphString list must not be empty -- thanks for bringing it to my attention. That first one is really strange, I haven't been able to replicate it locally. I'm wondering if it has something to do with a scaling factor? Can you try it out on a line that has multiple spaces to see if the error "grows", i.e. the overlap increases? If so I think scaling is the issue.

bryphe:
Lots of progress! Nice work @zbaylin . The editor-text rendering and positioning is very complicated -really impressive how far you've gotten!

The previous issues look fixed!

I'm still seeing some issues here:

__Issue 1:__

Changing the font-size doesn't seem to take effect immediately, causing the editor text to be laid out incorrectly:



__Issue 2:__

I think the way we are deciding what 'token' to render to is assuming monospace width - so it means we're not always rendering all the way to the edge for the tokens. If you watch Line `8` when I scroll to the right, you see text to start to appear - it seems we're stopping rendering 'early'



__Issue 3:__

The click coordinates are not quite correct - 



This suggests the `pixelPositionToBufferLineByte` might not be correct anymore: https://github.com/onivim/oni2/blob/8cc7c10832b31ea36731923b9c7e287e9190188d/src/Feature/Editor/Editor.rei#L142 - which could be problematic for other things, too.

__Issue 4:__

The indentation lines aren't lined up as expected:
Screen Shot 2020-07-24 at 5 17 44 PM

This might suggest that the `Editor.bufferLineByteToPixel` methods might be incorrect: https://github.com/onivim/oni2/blob/8cc7c10832b31ea36731923b9c7e287e9190188d/src/Feature/Editor/IndentLineRenderer.re#L90

__Issue 5:__

The bracket highlights are larger than the character - maybe still assuming monospace? Could be related to issue 4 - we might need to call `bufferLineCharacterToPixel`/`bufferLineByteToPixel` to get the `width`, which is returned b y those methods: https://github.com/onivim/oni2/blob/8cc7c10832b31ea36731923b9c7e287e9190188d/src/Feature/Editor/Editor.rei#L109

image

src/Core/BufferLine.re
```diff
@@ -33,24 +37,130 @@ type t = {
...
   mutable nextByte: int,
-  // nextIndex is the nextIndex to work from
+  // [nextIndex] is the nextIndex to work from
   mutable nextIndex: int,
-  // nextPosition is the graphical position (based on character width)
-  mutable nextPosition: int,
+  // [nextGlyphStringIndex] is the next nth glyph string in shapes
+  mutable nextGlyphStringByte: int,
+  // [nextCharacterPosition] is the graphical position (based on character width)
+  mutable nextCharacterPosition: int,
```
bryphe, line 57: I'm wondering why we need `nextCharacterPosition` with the pixel position?

src/Core/BufferLine.re
```diff
@@ -33,24 +37,130 @@ type t = {
...
+    Lru.M.Make(SkiaTypefaceUcharHashable, MeasureResult);
+
+  let measurementsCache =
+    MeasurementsCache.create(~initialSize=1024, 128 * 1024);
+
+  // Create a paint to measure the character with
+  let paint = Skia.Paint.make();
+  Skia.Paint.setTextEncoding(paint, GlyphId);
+  Skia.Paint.setLcdRenderText(paint, true);
+  Skia.Paint.setAntiAlias(paint, true);
```
bryphe, line 96: We'll need to make sure these stay in sync with the user settings - in particular, the `setSubpixelText` can impact rendering and cause sizes to be off - and can be changed with `editor.fontSmoothing` configuration. 

This inconsistency caused problems like #1364 (fixed in https://github.com/onivim/oni2/pull/1372)

src/Core/BufferLine.re
```diff
@@ -33,24 +37,130 @@ type t = {
...
+  let getCharacterWidth = (~indentation: IndentationSettings.t, uchar) =>
+    if (Uchar.equal(uchar, tab)) {
+      indentation.tabSize;
     } else {
-      Uucp.Break.tty_width_hint(c);
+      // Some unicode codepoints are double width, like many CJK characters
+      Uucp.Break.tty_width_hint(
+        uchar,
+      );
+    };
```
bryphe, line 106: I wonder if we still need this? This is really just a shortcut when we assume the font is monospace, and we try to figure out if the character is 1x or 2x the monospace width - but if we're actually using pixel widths, I think we can remove this - the actual text measurements are more accurate

src/Core/BufferLine.re
```diff
@@ -33,24 +37,130 @@ type t = {
...
+      let characterWidth =
+        getCharacterWidth(~indentation=cache.indentation, uchar);
+      Skia.Paint.setTypeface(paint, typeface);
+
+      // When the character is a tab, we have to make sure
+      // we offset the correct amount.
+      let pixelWidth =
+        if (Uchar.equal(uchar, tab)) {
+          Skia.Paint.setTextEncoding(paint, Utf8);
+          let spaceSize = Skia.Paint.measureText(paint, "  ", None);
```
bryphe, line 151: Maybe it makes sense to just cache this as part of the `Font.t` object? Seems like we'd measure this a lot!

src/Core/BufferLine.re
```diff
@@ -33,24 +37,130 @@ type t = {
...
+          Skia.Paint.setTextEncoding(paint, GlyphId);
+          float(cache.indentation.size) *. spaceSize;
```
bryphe, line 153: Cool 👍

zbaylin:
@bryphe looks like I fixed Issue 4, though it exposes a different bug I think
image
The indentation lines (not sure the correct word) look to be rendered on top of the characters, which is usually fine in a monospace context since the characters are centered rather than left aligned and always aligned in a grid, but with variable-width fonts, the glyphs are usually left aligned and obviously dont conform to a uniform grid.

Do you think this would be easy to fix?

src/Core/BufferLine.re
zbaylin: Ah thanks! I set the smoothing every time I measure, but it looks like I should get rid of that `antiAlias` call:

https://github.com/revery-ui/revery/blob/1531d5feebb894830010b14dadf89e09db9b648b/src/Font/Smoothing.re#L12

Will do now!

src/Core/BufferLine.re
zbaylin: Great idea. I just renamed `Font.t`'s `measuredWidth` to `spaceWidth` and set it to the width of a space character. Since fonts aren't always monospaced anymore, I think `measuredWidth` is slightly misleading.

src/Core/BufferLine.re
zbaylin: Fixed in 857b541e05b1018f561985ad8917e3fa546a8cdf!

bryphe:
@zbaylin - cool! 

Switching the indent line renderer to render underneath the text would be straightforward (and makes sense - we don't really want to render over text!). Just need to move this block:
https://github.com/onivim/oni2/blob/2c169569ff6286549095119096eb6cdbfd6efdf1/src/Feature/Editor/SurfaceView.re#L228

Before this block:
https://github.com/onivim/oni2/blob/2c169569ff6286549095119096eb6cdbfd6efdf1/src/Feature/Editor/SurfaceView.re#L210

It seems like there is an additional issue though - it doesn't look like we're using the correct size for a space character in the non-monospace case. For a case like this:
image

It seems like the 'dots' should be closer together to represent the actual size of the space character

zbaylin:
@bryphe I updated the first comment with all the bugfixes I implemented so far. Let me know if any of them don't work for you!

src/Core/BufferLine.re
zbaylin: I got rid of all the character measurements after our discussion 👍

src/Feature/Editor/ContentView.re
```diff
@@ -170,6 +170,17 @@ let renderText =
           | _ => Some(List.hd(v))
           }
         };
+      let bufferLine = Editor.viewLine(editor, item).contents;
+      let startPixel =
```
bryphe, line 174: I think we might just want to use the editor `scrollX` for `startPixel` - may simplify things.

src/Feature/Editor/ContentView.re
```diff
@@ -154,7 +154,7 @@ let renderText =
       ~bufferSyntaxHighlights,
       ~leftVisibleColumn,
       ~shouldRenderWhitespace,
-      ~bufferWidthInCharacters,
+      ~bufferWidthInPixels,
```
bryphe, line 157: Good idea to switch this out to use pixels instead of characters, now!

src/Feature/Editor/ContentView.re
zbaylin: Sounds good!

zbaylin: Changed in 0b6a796!

zbaylin:
@bryphe I think I fixed all the bugs you found (all are listed in the first comment). I'm going to keep testing things and fixing up the integration tests that are broken now due to API changes, but let me know if you find any more issues!

CrossR:
I'll give this a run through on Windows with scaling in a few hours, and can also test with scaling on Linux I believe (I think between Ubuntu/CentOS/Manjaro, one of them uses scaling but I never remember which it is), if that is useful.

zbaylin:
Awesome, thanks @CrossR!

bryphe:
Awesome, great progress @zbaylin ! I'm testing it out while I do the rest of my work for today - see if there are any show stoppers.

I noticed one issue still with the sizing getting out of sync when changing a configuration file:


I'll update if I find anything else. Thanks for your help testing as well, @CrossR !

zbaylin:
@bryphe I tried out your suggestion about tying the font to the buffer itself, but unless Im missing something obvious, it still seems to result in the same error. Im wondering if it's a race condition bug since we update the font asynchronously? Trying to chase down what else it could be!

src/Core/BufferLine.re
```diff
@@ -73,60 +162,128 @@ module Internal = {
...
+  // We assume that the whoever is requesting the buffer has verified
+  // that the font is resolvable.
+  // TODO (maybe): we will probably eventually support having a different
+  // default weight. This will probably be part of the font record, so
+  // this will need to be updated if/when this is added.
+  let loadedFont =
+    fontFamily
+    |> Revery.Font.Family.toSkia(Revery.Font.Weight.Normal)
+    |> Revery.Font.load
+    |> Result.get_ok;
```
bryphe, line 251: I think this is a bit risky - it relies on the consumer verifying the font, but this isn't enforced by the type system, or obvious without diving into the code. And if done incorrectly - it could cause the app to crash (`Result.get_ok` is a red flag in production code!).

Some ideas:
- We could enforce that the font is 'pre-validated' by having the consumer pass in a loaded font.
- We could create a default fallback to our jetbrains font

src/Store/Features.re
```diff
@@ -662,18 +657,32 @@ let update =
...
+              let bufferId = Feature_Editor.Editor.getBufferId(editor);
+              let updatedBuffer =
+                buffers'
+                |> Buffers.getBufferExn(bufferId)
+                |> Feature_Editor.EditorBuffer.ofBuffer;
+              Feature_Editor.Editor.updateBuffer(
+                ~buffer=updatedBuffer,
+                editor,
+              );
+            },
```
bryphe, line 680: As we discussed - we don't want to throw here. It would mean a crash for the user.

src/Core/BufferLine.re
zbaylin: I fixed this in 1a9b6cd -- we have to use `Result.get_ok` for our default JetBrains font, but I think that's unavoidable. Let me know what you think!

src/Store/Features.re
zbaylin: Fixed in e8391a8!

src/Service/Font/FontLoader.re
```diff
@@ -14,145 +14,22 @@ type t = {
...
-         ~smoothing: Revery.Font.Smoothing.t,
-         ~fontFamily: Revery_Font.Family.t,
-         fullPath,
-         fontSize: float,
-       ) => {
-     Log.tracef(m =>
-       m("loadAndValidateEditorFont path: %s | size: %f", fullPath, fontSize)
-     );
-
-     fullPath
```
bryphe, line 35: We can just remove this code, since it's no longer used! 🎉

src/Feature/Editor/Editor.re
```diff
@@ -41,7 +41,6 @@ type t = {
   viewLines: int,
   cursors: [@opaque] list(Vim.Cursor.t),
   selection: [@opaque] VisualRange.t,
-  font: [@opaque] Service_Font.font,
```
bryphe, line 44: Cool! Thanks for removing this 👍

bryphe:
Code change is looking great - thanks for all the work on this, @zbaylin ! 

Pretty amazing that we can support variable width fonts now 🎉  You're a rockstar - way to go above and beyond!

I would like to see a comprehensive set of test cases, so that we can reduce our regression risk and have confidence that this will be solid for the 0.5.0 release

zbaylin:
Awesome @bryphe -- thanks for the review and the help along the way! I'm converting the inline BracketMatch tests to Rely-based tests right now since the new buffer changes require assets which aren't discoverable from where the inline tests are run. After I finish that, CI should be passing, and all that's left is coming up with that testing matrix!

Super excited about this!

zbaylin:
I came up with this testing matrix to make sure this is robust before we merge. Let me know if you have any other ideas for things to test!

| Test                                        | OS      | Result |
|---------------------------------------------|---------|--------|
| Switching font in Configuration file        | macOS   |        |
|                                             | Linux   |        |
|                                             | Windows |        |
| Switching font while other buffers are open | macOS   |        |
|                                             | Linux   |        |
|                                             | Windows |        |
| Opening UTF-8 test file                     | macOS   |        |
|                                             | Linux   |        |
|                                             | Windows |        |
| Opening UTF-8 test file and click           | macOS   |        |
|                                             | Linux   |        |
|                                             | Windows |        |
| Opening UTF-8 test file and trigger hover   | macOS   |        |
|                                             | Linux   |        |
|                                             | Windows |        |

bryphe:
Thanks for the test matrix, @zbaylin !

Some cases I think we should add:
- Scrolling horizontal for `tofu.md` renders as expected (no clipping on edges)
- Validate zoom / scaling (You can use `Control+` and `Control-` to exercise this)
- Validate both tabs and spaces rendering
- Font Smoothing: Test with `none`, `antialiased`, and `subpixel-antialiased`
- Invalid font specified

Could you add these to the table and run through them on OSX / Linux?

I'll help run through the Windows tests shortly!

zbaylin:
@bryphe great ideas! I'll test them out now!

bryphe:
On Windows, if I change the font - it doesn't take effect until I restart. Trying to see why...

zbaylin:
I think @CrossR had the same issue, but it also appeared to happen on `master` as well. Not sure why that would be the case, though.

CrossR:
I noticed that seemingly randomly, sometimes I would get correctly updating font size and/or family, and other times I'd have to restart the editor. I didn't manage to pin down what the issue was, but a quick test on master gave me no change at all.

bryphe:
Thanks @zbaylin and @CrossR ! I opened #2206 to track - since it reproduces on `master`, it shouldn't block bringing this PR in 👍

zbaylin:
It seems like on macOS a tabbed-file crashes Oni -- not good! Trying to figure out why that would be the case...

zbaylin:
Ah actually it seems like it's a fallback bug that would be fixed by #2204. If I use JB Mono it's fine, so I think it's unrelated to this PR

zbaylin:
@bryphe I finished the tests on macOS and Linux -- all looks good!

bryphe:
Thanks @zbaylin ! Looking at Windows - how did you trigger hover in `tofu.md`? I'm not seeing any hover UI pop-up - is there part of text you were trying it on?

bryphe:
Tabs don't render correctly on windows:

image

Note the discrepancy between the indent lines and the tab character spacing

zbaylin:
Ah sorry @bryphe -- for the hover popup I was using the "fake" extensions we have for hover. I had them installed back when I was implementing hover, so I forgot they were there!

Thanks for the tabs bug -- it's possible I was looking at the wrong thing on macOS and Linux. I'll look again now!

bryphe:
Thanks @zbaylin !

zbaylin:
@bryphe it's possible I misunderstood the fields of `IndentationSettings.t` -- what is the difference between `size` and `tabSize`? And I guess a follow up to that would be how many spaces should a tab take up?

Sorry for that bug!

bryphe:
what is the difference between size and tabSize? And I guess a follow up to that would be how many spaces should a tab take up?

- `size` is how many spaces an indentation should be, when indenting and `insertSpaces` is `true`.
- `tabSize` is how large the tab character is (in number-of-spaces). So if `tabSize` is `3`, it should be rendered `3 *. pixelSizeOfSpace` wide.

Note that they _can_ be different values - I could have indentation settings like:
- `insertSpaces: true` - use spaces for indentation
- `size: 4` - use 4 spaces for each indent level
- `tabSize: 3` - render a tab character as 3-spaces-wide

This could make a difference, for example, if I had some text like:
```
abc
....def
........ghi<TAB>jkl
```
(Where the `.` are spaces and the `<TAB>` is the tab characteR)

zbaylin:
Ah interesting -- I guess that complicates things a bit with calculating the indentation size at the beginning of the line, since we have to know if we're at the beginning when we get a tab or somewhere else in the middle of the line.

zbaylin:
Never mind, I think just using `tabSize *. spaceWidth` should be sufficient, since `size` is only ever used with spaces. Pushing up a fix now.

zbaylin:
@bryphe just pushed a fix, let me know if it works on Windows!

bryphe:
> @bryphe just pushed a fix, let me know if it works on Windows!

Looks great! Thanks @zbaylin

zbaylin:
This keeps failing on Windows for random reasons each time -- not sure if it's just bad luck or if it's something in the PR itself causing that... @bryphe any ideas?

bryphe:
@zbaylin - I re-ran and green this time 🎉  

Thanks for your work on this!

PR #2208, opened 2020-07-31
fix(signature-help): Crash on negative index

bryphe:
__Issue:__ Typing `Console.wri(` in a C# project with the C# extension would crash.

__Defect:__ We were getting an exception - an `Invalid_argument List.nth`:
```
[ERROR]    +0ms Oni2.Exception : Exception (Invalid_argument List.nth):
Called from file "src/Feature/SignatureHelp/Feature_SignatureHelp.re", line 476, characters 6-52
Called from file "src/Feature/SignatureHelp/Feature_SignatureHelp.re", line 639, characters 6-251
Called from file "src/UI/EditorView.re", line 122, characters 10-432
```

The code hitting this is: 
```
    let maybeSignature: option(Signature.t) =
      List.nth_opt(model.signatures, signatureIndex);
    let maybeParameter: option(ParameterInformation.t) =
      Option.bind(maybeSignature, signature =>
        List.nth_opt(signature.parameters, parameterIndex)
      );
```

A bit confusing - because we're using `List.nth_opt`, which I'd assume wouldn't throw, and would just return `None` if the index is invalid. However - looking at `List.nth_opt` - it's implemented like this:
```
let nth l n =
  if n < 0 then invalid_arg "List.nth" else
  let rec nth_aux l n =
    match l with
    | [] -> failwith "nth"
    | a::l -> if n = 0 then a else nth_aux l (n-1)
  in nth_aux l n
```
https://github.com/ocaml/ocaml/blob/8b6241c64c8e02a02daa7c4aa6c2181238023834/stdlib/list.ml#L46

If the index is less than 0 - it still throws.... I wonder if this is intentional? It seems confusing. 

__Fix:__
In any case - I switched to use `Base.List.nth` which has the behavior we expect.


bryphe:
Opened a PR here: https://github.com/ocaml/ocaml/pull/9816

PR #2222, opened 2020-08-03
Fix #2205: Control+R working in terminal

### LINKED ISSUE #2205: Ctrl+R in :terminal
I wasn't able to find this mentioned before, so I'm opening an issue here. Please link me if there are previous discussions around it.

My current pre-Oni2 setup is nvim in a tmux pane with terminals on the bottom, and I heavily heavily rely on Ctrl-r to run previous commands. I use fzf in my terminal to get fuzzy searching through the command history, and it's really nice for running old tests or ssh commands and the like.  With the Oni2 term I've run into a problem. <C-r> in the terminal doesn't work since it's a vim binding and it waits for me to paste something from a register instead of searching history.

Is there a way to "passthrough" C-r so that vim doesn't handle it and it goes to the terminal when it's focused? Here's my current, unsuccessful attempt

    { "key": "<C-r>", "command": "", "when": "terminalFocus" }

With this mapping <C-r> will no longer check the register, but it also doesn't open the history, just does nothing. 

Is there a way to "substract" the mapping and let the terminal handle it? Or, conversely, to "add" a mapping that will go to the shell fzf command when it's pressed? 

Note, <Ctrl+t> does work fine in the :term, so I'm guessing it's not a problem with terminal or fzf itself.

bryphe:
__Issue:__ In terminal, the `Control+R` triggers 'insert register' instead of the default terminal behavior (running previous commands).

__Defect:__ The default keybinding was setup to override the terminal action.

__Fix:__ Make the default keybinding inactive when the terminal is focused.

Fixes #2205


ftlc:
Thanks for the quick turnaround!

Issue #1933, opened 2020-06-13
no syntax highlighting


gour:
Hello,

today, after installing #5d000137 AppImage I noticed that Oni2 does nor do syntax highlighting for any filetype I tried. THought it's maybe problem with some extensions, but even after rm-ing all installed extension and opening simple *.html file, still it does not work.

Here is is the debug log file:
[oni2.log](https://github.com/onivim/oni2/files/4775276/oni2.log)

and here is is output of `--checkhealth` as suggested by Ryan:

[oni3.log](https://github.com/onivim/oni2/files/4775277/oni3.log)

I'm on Fedora fc32 Linux.

Sincerely,  
Gour

bryphe:
Thanks for logging the issue and for the logs, @gour ! The health-check looks good (

Do you happen to have any custom configuration, or are using a custom theme or font? The configuration file would be at `~/.config/oni2/configuration.json`. 

Possibly related:
- #1836
- #1879

gour:
> Do you happen to have any custom configuration, or are using a custom theme or font?

Nope, just a pristine install...

Also, strange thing is that Explore/Sidebar are not visible:

![oni2](https://user-images.githubusercontent.com/114592/84586857-13566300-ae1b-11ea-962c-1656ead2eca3.png)

CrossR:
The sidebar/explorer not being shown is part of Zen mode (You can toggle it off in the Ctrl-Shift-P menu).

What file types are you missing syntax highlights in?  The original post says none, but the screenshot has it for JSON files at least.

gour:
> The sidebar/explorer not being shown is part of Zen mode (You can toggle it off in the Ctrl-Shift-P menu).

Ahh, OK. I wondered why it was disabled by default..

> What file types are you missing syntax highlights in? 

E.g. HTML.

bryphe:
Thanks for the update, @gour ! 

The HTML issue may be related to https://github.com/onivim/oni2/issues/1879 (a bug with our default theme, `OneDark Pro`... interestingly, other themes seem to work...)

gour:
> The HTML issue may be related to #1879 (a bug with our default theme, `OneDark Pro`... interestingly, other themes seem to work...)

Yes, I can confirm that with few other themes I've tried, everything is fine. :smile:

thismat:
The official Dracula theme is also having this issue. I'm adding what I've pin-pointed in the theme here for reference, maybe it will help in pin-pointing what's going on with theme compatibility.

[...]

Issue #1879, opened 2020-06-06
HTML syntax highlighting is not displayed


thecodejedi:
I'm opening a partial .html file and there is no syntax highlighting. Full html files with <html> tag do work, correctly.

Highlighting for .ts files work, also highlighting of CSS or JS in an html work correctly.

Here is a screenshot of my problem:
![Screenshot from 2020-06-06 11-45-27](https://user-images.githubusercontent.com/1226172/83941544-85f88a80-a7ec-11ea-9035-e15fad2b59df.png)

here is a [log](https://github.com/onivim/oni2/files/4739795/log.txt) of just opening the file with `-f --debug`

this is the [source](https://gist.github.com/thecodejedi/dfebb12915de718e7314e62313bf5c51) of the file I tried to open.

I'm on Fedora 32 with GDM on Wayland. (5.6.15-300.fc32.x86_64)

bryphe:
This looks like a similar issue to https://github.com/onivim/oni2/issues/1899 - a problem with our default `OneDark-Pro` theme. I'm able to reproduce with that theme:
![image](https://user-images.githubusercontent.com/13532591/84720741-bd291300-af33-11ea-865b-ef0b3b1484f9.png)

But other themes, like `Monokai`, seem to work correctly:
![image](https://user-images.githubusercontent.com/13532591/84720776-cc0fc580-af33-11ea-9036-da5aba13a658.png)

yunti:
Just came across this issue today, seems I get it with full html files.  
I'm on the onedark-pro theme as that was the earlier default theme.
What's the default/most tested theme now, so I can switch to that?

<img width="524" alt="image" src="https://user-images.githubusercontent.com/1826271/87945811-29e67e80-ca99-11ea-9793-1d5c0c034755.png">

yunti:
Note syntax highlighting support varies per theme quite a bit eg for jsx in js files. (possibly related)

Dark + theme.
<img width="528" alt="image" src="https://user-images.githubusercontent.com/1826271/87946540-0ff96b80-ca9a-11ea-861a-8bb914d08e8c.png">

one dark pro
<img width="764" alt="image" src="https://user-images.githubusercontent.com/1826271/87946626-2e5f6700-ca9a-11ea-9dd9-15a6a6423a0c.png">

bryphe:
This looks better now with #2301 👍 

[...]

Issue #2220, opened 2020-08-03
Syntax Highlighting: Terraform - Not getting full highlights


bryphe:
Plugin: https://github.com/hashicorp/vscode-terraform (downloaded from: https://github.com/hashicorp/vscode-terraform/releases/download/v2.1.1/terraform-2.1.1.vsix)

Example file: https://github.com/terraform-providers/terraform-provider-aws/blob/master/examples/alexa/main.tf

Some syntax highlights are not present in Onivim for this syntax / theme pair -
VSCode:
![image](https://user-images.githubusercontent.com/13532591/89229194-d8a8b580-d595-11ea-8dcc-e3dbd7f1dce8.png)

Onivim: 
![image](https://user-images.githubusercontent.com/13532591/89229225-e6f6d180-d595-11ea-9871-f34a3a84fd9e.png)

Need to investigate and see if it's a problem with our handling of the textmate grammar - or if it is a problem with theme matching

LeonT-A:
Just ran into this myself and see that it's already mentioned. Thanks for your effort!

Extension: [Ayu Dark Lighter](https://marketplace.visualstudio.com/items?itemName=divola.ayu-dark-lighter)

VSCode
![](https://i.imgur.com/IuW8RtB.png?raw=true)

Oni
![](https://i.imgur.com/ZfCdBCc.png?raw=true)

lainor:
+1 I ran into this problem with certain themes and elixir-ls

E.g. Nord theme `arcticicestudio.nord-visual-studio-code`

VSCode:
![image](https://user-images.githubusercontent.com/828651/96334620-631b9400-1072-11eb-9cab-53acdc5aad20.png)

Oni:
![image](https://user-images.githubusercontent.com/828651/96334651-9a8a4080-1072-11eb-8925-3eff8a8b713a.png)

Notice how the module name are in the standard foreground color.
Note: I am using a bracket pair colorizer in VSCode but not in Oni

Issue #2144, opened 2020-07-20
Typescript syntax highlighting broken after constructors


jisensee:
Using constructors in classes seems to break syntax highlighting after it.
### Reproduction
`:e test.ts`
```typescript
class C {
  constructor() {
    
  }

  private f() {
    console.log(2)
  }
}
```
results in:

![Capture](https://user-images.githubusercontent.com/17804319/87932657-28fb2000-ca8c-11ea-9447-a3d08597fd57.PNG)

I am on Windows, using the current nightly build.

jisensee:
This appears to only be a problem with the LaserWave theme. All other themes work fine.

PR #2255, opened 2020-08-06
test(textmate): Show used theme colour

CrossR:
Just a test to help look into stuff like #1933 / #1879.

We can look at the colours being produced more easily now and see the difference between themes:

image

I.e. OneDark always produces the same `#bbbbbb` colour for everything here. This was just a quick test to check it was reproducible, so I'll now extend it to `tokens : (___), colour: (__) from (__)` so we know which scope is being used to give us that colour.

### DIFF
--- a/test/reason-textmate/bin/TextmateCli.re
+++ b/test/reason-textmate/bin/TextmateCli.re
@@ -4,7 +4,7 @@
 // Tool for testing the textmate grammars
 //
 // Can be run via:
-// `esy @test x TextmateCli --file=package.json --grammar=test/reason-textmate/json.json --line=25
+// `esy @test x TextmateCli --file=package.json --grammar=extensions/json/JSON.tmLanguage.json --theme=extensions/onedark-pro/themes/OneDark-Pro.json --line=25
 
 open Textmate;
 
@@ -12,12 +12,14 @@ module Cli = {
   type t = {
     sourceFile: string,
     grammarFile: string,
+    themeFile: string,
     lineNumber: int,
   };
 
   let parse = args => {
     let sourceFile = ref(None);
     let grammarFile = ref(None);
+    let themeFile = ref(None);
     let lineNumber = ref(0);
     let () =
       Arg.parse_argv(
@@ -35,6 +37,11 @@ module Cli = {
             "Grammar file to use",
           ),
           ("--line", Int(i => lineNumber := i), "Line number to show"),
+          (
+            "--theme",
+            String(s => themeFile := Some(s)),
+            "Theme file to use",
+          ),
         ],
         _ => (),
         "",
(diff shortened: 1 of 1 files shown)


CrossR:
Other point of interest/nice comparison here is the VSCode Ctrl-Shift-P Inspect Editor tokens/scopes:

image
image

The scopes are the exact same*, so they are all valid and complete (ignore the `text.html.derivative` vs `basic`, that is just the syntax file I loaded). 

Gruvbox managed to match against `text.html entity.name.tag`, which does exist as a HTML/XML specific part and is the same colour shown above.
OneDark however matches against `entity.name.tag`, and should have a colour....but we don't have that.

Looking into it and with help from the comment over in #1933, it looks like a lower priority scope is causing the higher priority ones to be ignored. In that case it was a general "scope" one, but for OneDark and HTML, its the "meta.tag" one. Removing that gives me slightly better results.

image

Its then possible that is causing further issues like the text to not be matched correctly etc. But removing that option does fix the `div` to start reporting correctly.

Still trying to get my head around what that means need changing though!

(* The scopes loaded aren't quite the exact same...we have `$1` etc in ours. I haven't done enough digging to check if that is just a result of the CLI tool, or if we are infact not replacing those variables with what is needed).

bryphe:
Awesome, thanks for investigating this @CrossR ! Huge help.

This is the module where we do the matching: https://github.com/onivim/oni2/blob/master/src/reason-textmate/TokenTheme.re

And we have some basic test cases around the matching here:
https://github.com/onivim/oni2/blob/c6b7772e18696ac033a4ec3d0e19e6fa4e0249b5/test/reason-textmate/TokenThemeTests.re#L68

>  The scopes loaded aren't quite the exact same...we have $1 etc in ours. 

This is an interesting observation - I think it might be a bug with the textmate parsing - we might be missing a case where we need to replace a part of the token name with a match group in the regex. 

So it looks like two separate issues - these are possible the root cause of several highlighting issues, as well:
- https://github.com/onivim/oni2/issues/2220
- https://github.com/onivim/oni2/issues/2006
- https://github.com/onivim/oni2/issues/2144

CrossR:
I've had a poke around the HTML grammar, and this is the part we are looking at: https://github.com/onivim/oni2/blob/94e9fa21fb9ef63022ee2514eb9bfd108c4b0910/extensions/html/syntaxes/html.tmLanguage.json#L2150-L2171

However, I don't really see anywhere that the capture groups are really dealt with as part of the regex creation, it looks like its a TODO that might have gotten lost/an open issue?

https://github.com/onivim/oni2/blob/94e9fa21fb9ef63022ee2514eb9bfd108c4b0910/src/reason-textmate/RegExpFactory.re#L134-L143

`captureGroups` is only referenced twice, that `None` and the type def.

I would guess, since we are missing those capture groups, we then can't replace the $2 with the contents of that capture group.

I'll loop back to that later, and try and pin down what is also causing the scope matching issue as well, since this is two separate issues of making scopes and matching scopes. I've added a failing test for this specific part at least now.

PR #2271, opened 2020-08-08
refactor(editor): pass buffer to Editor component

SirFurness:
Makes things more consistent and simple by passing the buffer to the `Editor` component instead of finding it again.


bryphe:
Sorry for the slow reply, @SirFurness ! 

Thanks for the PR - this looks good 👍  Bringing in now.

PR #2272, opened 2020-08-09
docs: correct AppImage name

mkenigs:
The name of the AppImage available at the early access portal is
Onivim2-x86_64-master.AppImage


bryphe:
Thanks for the contribution! I've updated the docs to handle either the `Onivim2-x86_64-master.AppImage` or `Onivim2-x86_64-stable.AppImage`

PR #2273, opened 2020-08-10
docs: fix hover keybinding

iryan2:
It was previously `gd` (the mapping for go-to-definition)


bryphe:
Thanks for correcting this, @iryan2 !

PR #2275, opened 2020-08-10
Correct hover shortcut in docs

cwstra:
Hover shortcut in Language Features documentation section is the same as open definition (`g`-`d`), rather than what it actually is (`g`-`h`).


bryphe:
Thanks @cwstra for catching this and fixing it! 

@iryan2 actually just fixed this as well in #2273 - I've merged it already, so I'll close this one out.

Issue #1842, opened 2020-05-28
Windows 10 on bare metal - just a white screen


kevintynfron:
This is an update from #850 - I have upgraded from windows 7 to Windows 10 Pro.  I am getting the same result as #1809. The underlying UI seems to be there, but I can't see it. 

I did the debug start and attached the result:
[onivim_Debug_log.txt](https://github.com/onivim/oni2/files/4697921/onivim_Debug_log.txt)
It is basically looping around 

```
[INFO]    +27ms Revery.CanvasContext : Creating Skia context...
[ERROR]    +0ms Revery.CanvasContext : Unable to create skia context
```

You mention in #1809 that Onivim is using OpenGL 4+ .  I am using the latest Intel HD3000 driver but it doesn't have any Windows 10 drivers.

I used OpenGL Extension Viewer 6.1.0.0 and it says that 4.0 is supported (21% - 3/4) and no later version. I have added the report generated to this ticket in 
[GLExtensionViewer.zip](https://github.com/onivim/oni2/files/4697985/GLExtensionViewer.zip)

Here's the highlights of what the report on v4.0 supported features.
![gl4_4_support](https://user-images.githubusercontent.com/11351294/83188286-5d56fd80-a127-11ea-8dca-5671604b0b86.png)

bryphe:
Thanks for logging the issue and your detailed investigation, @kevintynfron ! Sorry that it's just one problem (Windows 7 not working) to another...

The log you shared is very helpful - this section here gives some info:
```
[INFO]    +49ms Revery.Core.Window : Window created successfully.
[INFO]     +2ms Revery.Core.Window : OpenGL hardware info:
[INFO]     +0ms Revery.Core.Window :   renderer: GDI Generic
[INFO]     +1ms Revery.Core.Window :   version: 1.1.0
[INFO]     +1ms Revery.Core.Window :   vendor: Microsoft Corporation
[INFO]     +1ms Revery.Core.Window :   shadingLanguageVersion: (null)
```

The OpenGL version is coming across as `1.1.0` for us when we initialize Skia. 

It's unfortunate but the OpenGL support just doesn't seem as great on Windows. I think the best for us to better support broader Windows hardware is to switch to DirectX. Will require us to build ANGLE and change the way we initialize SDL 2 on Windows.

[...]

Issue #1809, opened 2020-05-23
Windows 10 Pro - No usable display


dlbradford:
Hi,
Please see screencap. 

My Windows 10 Pro is running as a VM in Virtualbox 6.1 if that matters.

![Onivim2-Windows10-2020-05-23_084154](https://user-images.githubusercontent.com/72617/82731049-27062200-9cd2-11ea-9be8-98523e1d51f8.png)

bryphe:
Thanks @dlbradford for logging the issue! Currently, the editor relies on OpenGL 4+ on Windows - which [doesn't seem to be well-supported in VirtualBox](https://forums.virtualbox.org/viewtopic.php?f=2&t=82614) There might be a way to configure VirtualBox for it, though.

We're also looking at supporting DirectX render strategy (via `ANGLE`), which it does seem like virtualbox has better support for (and more broadly, better Windows support in general).

dlbradford:
@bryphe, thank you for the reply and likely explanation. I will mess about with it to see if I can sort it out re OpenGL 4+ in VirtualBox.

bryphe:
Hi @dlbradford - this should be fixed now w/ latest builds, as we use ANGLE as the renderer strategy on Windows (which uses the better-supported D3D instead of OpenGL). I've tested on Hyper-V and Parallels which both had similar issues.

Will close as fixed with #2283 , but please let me know if you still have issues opening Onivim

Issue #1522, opened 2020-03-29
Crash on startup: Virtualized Windows 10 Pro 


pa3k-lu:
OS: Windows 10 Pro x64, 1909 build 18363.752
Oni2: Onivim 2.0.3.0 (97eb4fea) downloaded from EA

Environment:  Windows 10 Pro running in Proxmox 6.1, the same symptoms are in Windows 10 Pro virtualized in bhyve in FreeBSD 12.1-stable
  - with Microsoft Basic Display Driver and Microsoft Remote Display Driver
  - connected to admin console through RDP

Oni2 running from command line briefly opens white onivim window that after while closes, command:
```oni2 -f --debug --trace --log-file oni2.log```

oni2.log contains only:
```--
Log started at 2020-03-29T19:14:32Z
--

[INFO]  Oni2_editor Starting Onivim 2.0.3.0 (97eb4fea)
[DEBUG] Oni2_editor Calling App.start
[DEBUG] Oni2_editor Init
[DEBUG] Revery.Core.Window Starting window creation...
[INFO]  Revery.Core.Window Creating window Oni2 width: 800 height: 600
[INFO]  Revery.Core.Window Window created successfully.
[DEBUG] Revery.Core.Window Window id: 1
[DEBUG] Revery.Core.Window Setting window context
[DEBUG] Revery.Core.Window GL setup. Checking GL version...
[DEBUG] Revery.Core.Window Checking GL vendor...
[DEBUG] Revery.Core.Window Checking GL shading language version...
```

Windows Event Log contains error entries for Onivim2 crash:

```Faulting application name: Oni2_editor.exe, version: 0.0.0.0, time stamp: 0x5e80191d
Faulting module name: Oni2_editor.exe, version: 0.0.0.0, time stamp: 0x5e80191d
Exception code: 0xc0000005
Fault offset: 0x00000000006d460d
Faulting process id: 0x25c
Faulting application start time: 0x01d605ff8b8fec2d
Faulting application path: C:\Users\pa3k\Desktop\win32\Oni2_editor.exe
Faulting module path: C:\Users\pa3k\Desktop\win32\Oni2_editor.exe
Report Id: 37d63d3b-576b-4d9f-81b1-ea757344e448
Faulting package full name: 
Faulting package-relative application ID: 
```

According to glennsl at discord:
[...]

PR #2283, opened 2020-08-11
feat(windows): ANGLE Support (Use Direct3D)

bryphe:
__Issue:__ Several windows environments are unable to open Onivim, due to not having a recent enough OpenGL support (and some hardware doesn't even have drivers for recent OpenGL...)

__Fix:__ In terms of driver / hardware support, Direct3D is much better supported on Windows than OpenGL.

Google created a library called [ANGLE](https://github.com/google/angle) - which is an OpenGL ES implementation that actually uses Direct3D as the back-end. By using ANGLE, we can better support Windows environments - both in the case where OpenGL is not available (in which case Direct3D should be!), and also by virtue of the fact that D3D tends to better supported / performant on Windows.

This picks up https://github.com/revery-ui/revery/pull/983 to bring in ANGLE support to Onivim - which should unblock several cases where Onivim would not start (ie, on VMs):
- Fixes https://github.com/onivim/oni2/issues/1522
- Fixes https://github.com/onivim/oni2/issues/1809
- Fixes https://github.com/onivim/oni2/issues/1842

Going to the `:version` screen validates that we use ANGLE with D3D back-end on WindowS:
image


threeforksp:
Hi,

Saw the merge and downloaded the latest nightly build for Windows. 

The same issue as filed in #1809  still exists for me.

If there is something else I need to do other than install the nightly build, please let me know.

If I can be of any use for your testing, again, let me know. 

Thanks,
threeforks

bryphe:
Thanks @threeforksp ! 

It looks like it hasn't made it into the nightly's yet (there was a blocking issue for the nightly build - https://github.com/revery-ui/revery/pull/985) - I'll update this once I've confirmed its in 👍

PR #2295, opened 2020-08-13
fixed WARNS (struct declaration style, type casts)

mrillusi0n:
Had to cast into proper types and change the way the struct was declared.


bryphe:
Thanks @mrillusi0n , this is great! So nice to see those warnings disappear.

I added a small commit just to add `-Werror` to prevent any warnings from coming back.

Will bring this in when the next round of CI passes

PR #2296, opened 2020-08-13
feat(workbench) adds configurable sidebar position

kyleshevlin:
This PR adds a configuration setting allowing the user to put the sidebar on the right side. This is accomplished by creating a list of the "surface" components, and reversing their order if the setting is set to `right`. The default CSS lays out the components correctly.

wootGotIt

### DIFF
--- a/src/Core/ConfigurationParser.re
+++ b/src/Core/ConfigurationParser.re
@@ -381,10 +381,10 @@ let configurationParsers: list(configurationTuple) = [
     (config, json) => {...config, workbenchEditorShowTabs: parseBool(json)},
   ),
   (
-    "workbench.sideBar.position",
+    "workbench.sideBar.location",
     (config, json) => {
       ...config,
-      workbenchSideBarPosition: parseString(json),
+      workbenchSideBarLocation: parseString(json),
     },
   ),
   (
--- a/src/Core/ConfigurationParser.re
+++ b/src/Core/ConfigurationParser.re
@@ -381,6 +381,13 @@ let configurationParsers: list(configurationTuple) = [
     (config, json) => {...config, workbenchEditorShowTabs: parseBool(json)},
   ),
   (
+    "workbench.sideBar.position",
+    (config, json) => {
+      ...config,
+      workbenchSideBarPosition: parseString(json),
+    },
+  ),
+  (
     "workbench.sideBar.visible",
     (config, json) => {...config, workbenchSideBarVisible: parseBool(json)},
   ),
--- a/src/Feature/SideBar/Feature_SideBar.re
+++ b/src/Feature/SideBar/Feature_SideBar.re
@@ -100,13 +100,7 @@ let toggle = (pane, state) =>
   };
 
 let setDefaultLocation = (state, setting) => {
-  let location =
-    switch (setting) {
-    | "right" => Right
-    | "left" => Left
-    | _ => Left
-    };
-
+  let location = setting == "right" ? Right : Left;
   {...state, location};
 };
 
--- a/src/Feature/SideBar/Feature_SideBar.re
+++ b/src/Feature/SideBar/Feature_SideBar.re
@@ -44,11 +44,16 @@ let initial = {
   location: Left,
 };
 
-let width = ({width, resizeDelta, isOpen, shouldSnapShut, _}) =>
+let width = ({width, resizeDelta, isOpen, location, shouldSnapShut, _}) =>
   if (!isOpen) {
     0;
   } else {
-    let candidate = width + resizeDelta;
+    let candidate =
+      switch (location) {
+      | Left => width + resizeDelta
+      | Right => width - resizeDelta
+      };
+
     if (candidate < Constants.minWidth && shouldSnapShut) {
       0;
     } else if (candidate > Constants.maxWidth) {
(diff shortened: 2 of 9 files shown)


src/Core/ConfigurationParser.re
```diff
@@ -380,6 +380,13 @@ let configurationParsers: list(configurationTuple) = [
     "workbench.editor.showTabs",
     (config, json) => {...config, workbenchEditorShowTabs: parseBool(json)},
   ),
+  (
+    "workbench.sideBar.position",
```
CrossR, line 384: Small thing, but could this be swapped to `"workbench.sideBar.location"`? I think that `"position"` makes more sense really, its not a layout...but it would keep us in sync with the vscode config option I believe, which we attempt to do where it makes sense.

kyleshevlin: Yeah, I'm happy to do whatever's best for the project. Not torn up about that. I'll make an update later to this. Do I need to squash commits when I do for this repo?

CrossR: Nah, we will do a squash commit at the end, so don't worry about that.

kyleshevlin: Resolved.

bryphe:
Awesome PR @kyleshevlin - thank you!

bryphe:
I wonder if we should flip the 'highlight' in the activity bar:

Screen Shot 2020-08-13 at 4 11 36 PM

there is a highlight line to show the active selected sidebar pane - I believe it would make sense to move it to the right side when `workbench.sideBar.location` is `right`.

kyleshevlin:
Sure, @bryphe, that's how it looks in VSCode, so I can give that a shot
Screen Shot 2020-08-13 at 4 13 52 PM

bryphe:
Sweet, thanks @kyleshevlin !

That border is here: https://github.com/onivim/oni2/blob/5e385db6ce03635bd390d4429d25ebd83af2ea7a/src/UI/Dock.re#L33

so we'll need to add a parameter to `<Dock />`, and then plumb it through `<activityBar />`:

https://github.com/onivim/oni2/blob/5e385db6ce03635bd390d4429d25ebd83af2ea7a/src/UI/Root.re#L126

Pretty close to where we query currently in `Root.re` to get the sidebar location.

src/Feature/SideBar/Feature_SideBar.re
```diff
@@ -91,3 +98,25 @@ let toggle = (pane, state) =>
   } else {
     {...state, shouldSnapShut: true, isOpen: true, selected: pane};
   };
+
+let setDefaultLocation = (state, setting) => {
+  let location =
+    switch (setting) {
```
yawaramin, line 104: You may have intentionally made the `"left"` case explicit, but it can be reduced to:

```reason
switch (setting) {
| "right" => Right
| _ => Left
}
```

Or indeed even just `let location = setting == "right" ? Right : Left;`

kyleshevlin: I did purposefully do this explicitly. I'm a big fan of state machines, enums, etc, and while it's probably true that "not right" will always mean "left", they aren't truly opposites. There's nothing to say there couldn't be a future where there's "top" or "bottom" variants (though very, very unlikely).

All that to say, I did this on purpose, but can change if it's bothersome to project owners.

kyleshevlin:
I've discovered a bug that should be addressed before merging this involving resizing the sidebar.

SmallBug

kyleshevlin:
I have the resizer on the correct side already. Working on the pane resizing as expected.

bryphe:
Nice catch @kyleshevlin !

I'm thinking we'll need to change the way the `delta` is applied as the resize in progress, depending on the sideBar location: 
https://github.com/onivim/oni2/blob/5e385db6ce03635bd390d4429d25ebd83af2ea7a/src/Feature/SideBar/Feature_SideBar.re#L44
(If it's on the right, we'd want `width - resizeDelta` instead)

src/Feature/SideBar/Feature_SideBar.re
bryphe: I don't have a strong feeling on this particular case - thanks for checking!

In general, we try to use exhaustive checks - when possible - but that's certainly not the case for strings.

kyleshevlin:
And got that, too.

fixed

kyleshevlin:
FWIW, Ran unit tests locally and they all pass.

src/UI/Dock.re
```diff
@@ -24,20 +24,28 @@ module Styles = {
...
-    backgroundColor(
-      theme |> (isHovered ? Colors.activeBackground : Colors.background).from,
-    ),
-  ];
+  let item = (~isHovered, ~isActive, ~theme, ~sideBar) => {
+    let borderFn =
+      switch (Feature_SideBar.location(sideBar)) {
+      | Feature_SideBar.Right => borderRight
+      | Feature_SideBar.Left => borderLeft
+      };
```
bryphe, line 32: Nice, thanks for plumbing this through 👍

bryphe:
Just pulled the latest and tested it out - looks great, @kyleshevlin !

 Thanks for fixing the highlight position and catching the resize issue 👍  Bringing in now!

PR #2310, opened 2020-08-17
Updated doc to describe more test checks that we do on CI

fanantoxa:
- How to run inline tests
- How to run duno build check
- How to run esy formatter


bryphe:
Awesome! Thanks for adding these details @fanantoxa , looks great

PR #2333, opened 2020-08-20
fix(crash/#2331): Remove debug prints

### LINKED ISSUE #2331: Running the command `:set filetype` closes editor [fix(buffer/#782)]
On WIndows 10, build 18362, attempting to run the command `:set filetype=` immediately and unexpectedly closes the editor, regardless of the filetype specified. 

However, if the application is initiated with the  `--nofork` and `--debug` flags, the command functions as expected.

Clicking on the filetype in the status bar works without issue.

--- #782

CrossR:
Fixes #2331.

I think this is the cause of #2331, as I can't repro it on mac and Windows I believe this can cause a crash as a terminal isn't allocated when launched normally, which would explain why it is fine with `-f` on Windows.

### DIFF
--- a/src/reason-libvim/BufferInternal.re
+++ b/src/reason-libvim/BufferInternal.re
@@ -71,8 +71,6 @@ let checkBufferForUpdate = buffer => {
         };
 
         if (!String.equal(string_opt(lastFt), string_opt(newFiletype))) {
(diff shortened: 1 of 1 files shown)


bryphe:
Ouch, thanks for catching this @CrossR ! This would cause a crash on Windows... I'll see if we can add a check for this on CI so these don't sneak in in the future.

PR #2336, opened 2020-08-20
chore: JetBrains Mono Upgrade 1.03 → 2.001

zbaylin:
Not sure how we were on such an old version, but this updates us to the most recent one. I've been using Oni with it and it works as expected.


bryphe:
Thanks @zbaylin !

PR #2348, opened 2020-08-22
refactor: Byte/Character index differentation in the type system

bryphe:
A common point of confusion that's come up when working with the buffer lines or editor surface is handling _byte index_ vs _character index_ (UTF-8 index) conversions. We took a step forward by adding an `Index.t` type - but unfortunately it's not granular enough to describe whether an API expects a _byte_ or _character_ index. When looking at an API that uses an `Index.t` - it's never clear if it is supposed to be a byte, line number, or character.

This change splits the `Index` concept into 3 separate concepts:
- `ByteIndex` -> a 0-based index, representing a byte offset
- `CharacterIndex` -> a 0-based index, represeting a UTF-8 character offset
- `LineNumber` -> represents a line number

On top of that, there is now `BytePosition`, which is a `{lineNumber, byteIndex}` and a `CharacterPosition`, which is a `{lineNumber, characterIndex}`.

In addition, instead of `Range`, which likewise was confusing in terms of whether it represented a range of bytes, or a range of characters, we now have `ByteRange` and `CharacterRange`.

Hopefully this will make it easier to see whether an API expects a byte or character, and help enforce it at the type-level. There are still some dubious places in terms of converting the old `Range` into `CharacterRange`, but this should make it easier to improve and help ensure we don't regress it as easily.


zbaylin:
This is awesome -- should prevent a lot of the bugs in the text-rendering pipeline!

PR #2352, opened 2020-08-22
fix(editor/#2335): Add tooltip for path

### LINKED ISSUE #2335: Editor hover over tab should reveal full path
In VSCode when you have over a tab it reveals the absolute path of the file.

This is useful in scenarios when you are editing multiple files of the same number.
There are many projects where you will need to edit several of the same filenames 
i.e. `index.tsx` and `dune` files.

bryphe:
This adds a tooltip to show the fullpath of items in the editor tabs:

Fixes #2335


chinwobble:
nice!

PR #2354, opened 2020-08-24
Update modal-editing-101.md

rogererens:
Fix a typo


Apostolique:
I think instead of `positions` it should be `characters`.

bryphe:
Thanks @rogererens and @Apostolique ! Glad we're fixing the `lines` - the `characters` change sounds good to me too while we're here, so I just merged that in.

PR #2355, opened 2020-08-24
feat(codelens): Initial CodeLens UI

bryphe:
Prior to the 0.5.0 release, I was pretty close to having codelens hooked up - figured I should bring in what I have, gated by an experimental feature, and clean it up a bit:

This adds a codelens subscription to pick up latest codelens, and adds the concept of an 'inline element' to the editor surface - basically a non-code-custom-rendered-block that takes up vertical space, as well as adds a component for handling rendering and tracking the dimensions.

This is turned off by default, and gated by an `"experimental.editor.codeLens": true` configuration flag.

There are still some issues that need to be fixed:
- Buffer updates / code lens state changing needs to be handling more gracefully - there are some stale lenses being shown briefly, and animations occurring when they aren't required.
- The `getVisibleTopLine`, `getVisibleBottomLine`, and `pixelPositionToBufferLine` need to be updated to account for the inline elements - these don't take inline elements into consideration, and thus the calculation for these (and `zz`/`zb`/`zt` behavior is not correct). 

In addition, we need to better handle the case where inline elements pop-in - a degenerate case right now is using `zb` to move a line to the bottom, and then a codelens gets rendered - pushing that line further down. We might need to maintain some additional state (ie, what line do we guarantee is shown - the bottom line / centered line / top line)? And maintain that invariant as inline elements pop in.


bryphe:
Merged with master, updated to account for wrapping in #2646

PR #2367, opened 2020-08-27
refactor: Run DCE tool

bryphe:


bryphe:
Looks like a really cool idea!

Unfortunately there a few false positives, like: 
```
  Warning Dead Module
  File "./src/reason-textmate/TextMate.rei", lines 172-209, characters 0-1
  TextMate.Trie is a dead module as all its items are dead.
```

which is definitely used in the codebase.

And some errors which aren't actionable, like missing optional arguments from brisk components:
```
  Warning Unused Argument
  File "./src/Feature/Search/Feature_Search.re", lines 187-195, characters 4-11
  optional argument ref of function +brisk-component is never used
```

Would love to have dead code analysis though - I'm sure there is quite a bit we can remove!

PR #986, opened 2019-11-30
[WIP] [Experiment] feat(editor): UI animation: buffer highlights on yank

bryphe:

PR #2377, opened 2020-08-29
feat(ux): Yank highlights

bryphe:
This implements a visualization when yanking text (a revamp of #986, using our new editor pixel conversion APIs on the editor surface):

__TODO:__

- [x] Configuration setting: `editor.yankHighlightAnimation`
- [x] Add linewise visualization on minimap

### DIFF
--- a/docs/docs/configuration/settings.md
+++ b/docs/docs/configuration/settings.md
@@ -76,6 +76,8 @@ The configuration file, `configuration.json` is in the Oni2 directory, whose loc
 
 - `editor.smoothScroll` __(_bool_ default: `true`)__ - When `true`, smoothly scroll the editor when the viewport is adjusted due to a cursor motion.
 
+- `editor.yankHighlightAnimation` __(_bool_ default: `true`)__ - When `true`, briefly highlight yanks on the editor surface.
+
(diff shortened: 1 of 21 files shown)


pikachunakapika:
That's perfect!
Do you think something similar could be done to give a little indication where I will end up when pressing a number?
E.g.: I want do delete 3 lines with `3dd`. After i hit 3 onivim does jot know yet what will follow so it highlights 3rd line up and down, and 3 columns left and right for a second or until command sequence finishes.

bryphe:
> Do you think something similar could be done to give a little indication where I will end up when pressing a number?

I'm definitely interested in these sorts of visualizations to help give immediate feedback!

> E.g.: I want do delete 3 lines with 3dd. After i hit 3 onivim does not know yet what will follow so it highlights 3rd line up and down, and 3 columns left and right for a second or until command sequence finishes.

I think the challenge here is to just figure out what sets of motions we'd show a hint for (there could be many - `3dW`, `3dG`, `3dj`, `3dl`, etc) - and decided what the visualization should look like.

yunti:
Love this feature, thanks for adding it.  It dissapears a little too quick for me and is probably a little too faint.  It would be useful to have a setting to amend this to make it appear a litter bit longer (and a bit brighter/less transparent - but that might depend on the theme?)

PR #2397, opened 2020-09-03
feat(keybindings): execute keybinding starting with : as ex command

### LINKED ISSUE #807: Feature Request: let an Oni2 command that sends vimscript be writable in the keybindings file
Hi. Exciting project. I was happy to buy the lifetime license to support future development.
It's not yet ready for me to use for actual work so I don't have extensive feedback but I do have two big feature requests which I'm posting in two tickets. I might note some of the glitches etc I've encountered later but I'm sure you're overwhelmed with those now.

Feature Request: let an Oni2 command that sends vimscript be writable in the keybindings file

I picture myself then doing something like this:

```
{
    "bindings": [
        { "key": "<⌘-`>", "command": "sendVimscript", "args": [[":bnext"]] },
        { "key": "<S-⌘-`>", "command": "sendVimscript", "args": [[":bprev"]] },
        { "key": "<⌘-W>", "command": "sendVimscript", "args": [[":q!"]] },
    ]
}
```

glennsl:
A simple change that allows keybindings to invoke ex commands directly, and could address quite a few of the requests in #1423. The downside of doing this instead of implementing proper commands is of course that they won't be discoverable through the command palette, but this doesn't prevent proper commands to be added either, so I think it might be a nice convenient stop-gap solution to unblock people at least.

Example:

```
  {key: "kk", command: ":split", when: "editorTextFocus"}
```

Let me know if you think this is a viable approach @bryphe, then I'll add a few integration tests before it's merged.

Fixes #807

### DIFF
--- a/integration_test/ExCommandKeybindingTest.re
+++ b/integration_test/ExCommandKeybindingTest.re
@@ -1,0 +1,49 @@
+open Oni_Model;
+open Oni_IntegrationTestLib;
+open Actions;
+
+let keybindings =
+  Some(
+    {|
+[
+  {"key": "kk", "command": ":split", "when": "editorTextFocus"}
+]
+|},
+  );
+
+runTest(
+  ~keybindings,
+  ~name="ExCommandKeybindingTest",
+  (dispatch, wait, _) => {
+    let input = key => {
+      let keyPress =
+        EditorInput.KeyPress.{
+          scancode: Sdl2.Scancode.ofName(key),
+          keycode: Sdl2.Keycode.ofName(key),
+          modifiers: EditorInput.Modifiers.none,
+        };
+      let time = Revery.Time.now();
+
+      dispatch(KeyDown(keyPress, time));
+      //dispatch(TextInput(key));
+      dispatch(KeyUp(keyPress, time));
+    };
+
+    wait(~name="Initial sanity check", (state: State.t) => {
+      let splitCount =
+        state.layout |> Feature_Layout.visibleEditors |> List.length;
+
+      splitCount == 1;
+    });
+
+    input("k");
+    input("k");
+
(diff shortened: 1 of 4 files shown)


bryphe:
Thanks for implementing this, @glennsl ! Approach looks great to me.

> I think it might be a nice convenient stop-gap solution to unblock people at least.

I agree - and definitely fits in to the 0.6.0 milestone work. 👍  

This fixes https://github.com/onivim/oni2/issues/807 as well

bryphe (approved):

glennsl:
Oh wow, exactly what you proposed. And here I was thinking I had novel ideas! Probably had that one stored away in the subconscious somewhere :D I'll add some integration tests then.

glennsl:
I'm unfortunately not able to run any integration tests locally because I get a sanitation error:

```
==63020==ERROR: AddressSanitizer failed to deallocate 0x2000 (8192) bytes at address 0x6250001e0100
==63020==AddressSanitizer CHECK failed: /build/gcc/src/gcc/libsanitizer/sanitizer_common/sanitizer_posix.cpp:61 "(("unable to unmap" && 0)) != (0)" (0x0, 0x0)
```

So I'm just pushing and hoping they'll work...

glennsl:
CI is being very difficult, but it's just Validate Windows Release that times out now.

bryphe:
Thanks for adding those tests, @glennsl - sorry about the CI difficulties. I bumped the Windows machine and it's green now. Looks great!

bryphe:
Oh, and with [license key bounty](https://onivim.github.io/docs/for-developers/free-license) - this would definitely be a qualifier for it. LMK if you want an extra license key, @glennsl

glennsl:
Thanks! But no need to worry about bounties on my account 🙂

Issue #2231, opened 2020-08-05
Allow to disable automatic file explorer scrolling


mabuchner:
When switching between multiple files from different directory structures, then the automatic scrolling inside the file explorer gets really distracting.

![ezgif-6-82c52e708c04](https://user-images.githubusercontent.com/6006874/89368960-5d6b0f00-d70f-11ea-9f7e-ad01d965b945.gif)

Sublime Text also highlights the current file in the side bar, but it will not automatically scroll to the highlighted file. This results in a much more unobtrusive experience.

It would be nice, if Oni2 could at least provide an option to disable the automatic scrolling.

bryphe:
> It would be nice, if Oni2 could at least provide an option to disable the automatic scrolling.

Definitely reasonable 👍  There's an equivalent VSCode setting for it - `explorer.autoReveal`. We should implement that.

marcagba:
Do you think this one is a good first issue for someone eager to contribute? 🤔

bryphe:
Hi @marcagba ,

Thanks for offering to help! I do think this would be a pretty good first issue, actually 👍 

There's a couple things we have to do for it:

1) __Add a configuration setting:__ - the VSCode-equivalent setting is `explorer.autoReveal`. When set to `true`, we'd use the existing behavior (automatically select and focus the file), and when set to `false`, we wouldn't scroll / focus a file.

We'd need to add the setting here: https://github.com/onivim/oni2/blob/master/src/Core/ConfigurationValues.re and then add a parser for it here: https://github.com/onivim/oni2/blob/68ca4d93947b5a74c93453776a9a281fd526c2a6/src/Core/ConfigurationParser.re#L232

2) __Turn off the auto-focus when the configuraiton setting is `false`:__ - This behavior is controlled by a subscription, here: https://github.com/onivim/oni2/blob/68ca4d93947b5a74c93453776a9a281fd526c2a6/src/Store/StoreThread.re#L313

We could gate this by doing something like: `let fileExplorerSub = autoReveal ? Model.Sub.activeFile(...) : Isolinear.Sub.none` (this would turn the subscription to a no-op when the configuration is off). And then the configuration value is pretty straightforward to query here once 1) is implemented: https://github.com/onivim/oni2/blob/68ca4d93947b5a74c93453776a9a281fd526c2a6/src/Store/StoreThread.re#L239

Would be awesome to have help! Let me know if you have any questions.

marcagba:
[...]

PR #2414, opened 2020-09-06
Auto scroll option

caleywoods:
Attempts to address #2231 

@bryphe - One behavior that I noticed after these changes were that in addition to no longer scrolling (this is expected), the file explorer does not highlight the file that currently has focus. The reporting user mentions that Sublime Text does continue to highlight the file, this seems like a better approach in terms of UX than leaving the last file that had focus while `editor.autoReveal` was true highlighted (or none in terms of a clean launch without change).

My question then becomes - maybe a method needs added that is similar to `revealAndFocusPath` in `FileExplorerStore.re` that only highlights a file without affecting the scrolling position of the file explorer? I wasn't sure if that approach was favorable to growing the arity of `revealAndFocusPath` and having it acccept another argument which would be the `autoReveal` bool.

Also - this PR still needs tests added, those are a WIP.


mabuchner:
As @bryphe mentioned in #2231, VSCode’s ‘explorer.autoReveal’ setting actually supports three values:
- true: Files will be revealed and selected.
- false: Files will not be revealed and selected.
- focusNoScroll: Files will not be scrolled into

“focusNoScroll” corresponds to the behavior of Sublime Text.

caleywoods:
@mabuchner Thanks! I'm working on changes to have it support those three options. 

Does anyone have a link for some examples on how I would handle pattern matching the new `autoReveal` in `StoreThread.re`? Previously it was a ternary because the type was `Bool` but now the type can be `Bool` or string. I'm thinking it should look something close to this:

```
switch (autoReveal) {
| true => Do sub for highlight and scroll
| false => Isolinear.Sub.none
| 'focusNoScroll` => Do sub for highlight only
};
```

bryphe:
Thanks for the work on this, @caleywoods ! Looks like a great start. And thanks @mabuchner for the feedback

We could define a type for this in `ConfigurationValues.re`, something like:
```
type autoReveal = [| `NoReveal | `HighlightOnly | `HighlightAndScroll];
```
sort of like the one we have here: https://github.com/onivim/oni2/blob/master/src/Core/ConfigurationValues.re#L26 - except a bit simpler because we don't need any arguments. (The syntax is for [Polymorphic Variants](https://blog.klipse.tech/reason/2018/03/12/blog-reason-types.html) which are convenient since we can use them without a type annotation)

Then, in our configuration parser, we could have a helper like `parseAutoReveal` (kind of like the ligatures one again: https://github.com/onivim/oni2/blob/e051efd80af9cbdf8e7ceb936d39c3628f861384/src/Core/ConfigurationParser.re#L141) - and we could use pattern matching like:
```
let parseAutoReveal = json => switch(json) {
| Bool(true) => `HighlightAndScroll
| Bool(false) => `NoReveal
| String("focusNoScroll") => `HighlightOnly
| _ => `NoReveal 
};
```

The strategy you called out for handling these different options sounds great!

The `revealAndFocus` is happening here: https://github.com/onivim/oni2/blob/e051efd80af9cbdf8e7ceb936d39c3628f861384/src/Store/FileExplorerStore.re#L152

We could check the configuration value - kind of like this code: https://github.com/onivim/oni2/blob/e051efd80af9cbdf8e7ceb936d39c3628f861384/src/Store/FileExplorerStore.re#L14 - and based on that, we could either call `revealAndFocus` (highlights+scrolls), or just `setFocus` (just sets the highlight).

caleywoods:
@bryphe The last bit of your comment was a little confusing. Based on what I have now - I think it looks like I should move the `autoReveal` config query and switch statement that I have in `StoreThread.re` into the `Some(path) =>` area inside `FileExplorerStore.re`. This would revert `StoreThread.re` `fileExplorerActiveFileSub` to always end up calling `ActiveFilePathChanged` (where my changes would pick it up).

caleywoods:
@bryphe I've moved the `autoReveal` stuff into `FileExplorerStore.re` and it seems to be working as I think it should. Going to do some further testing now.

docs/docs/configuration/settings.md
```diff
@@ -72,6 +72,8 @@ The configuration file, `configuration.json` is in the Oni2 directory, whose loc
 
 - `editor.rulers` __(_list of int_ default: `[]`)__ - Render vertical rulers at given columns.
 
+- `explorer.autoReveal` __(_string|bool_ default: `true`)__  - When `true`, the file explorer will jump to highlight the file current focused. When `false` the file explorer will remain static. If a string is entered it must be `"focusNoScroll"` which will still highlight the currently focused file in the file explorer but the file explorer will not scroll to it. Any other string supplied will be treated as if `false` was entered and the file explorer will remain static and not highlight the currently focused file. 
+
```
bryphe, line 76: Thanks for the detailed description here! I made a small tweak (there was an extra back-tick around true), but looks great.

src/Store/FileExplorerStore.re
```diff
@@ -149,7 +149,17 @@ let start = () => {
...
+        | Some(path) =>
+          let autoReveal =
+            Oni_Core.Configuration.getValue(
+              c => c.explorerAutoReveal,
+              state.configuration,
+            )
+            switch (autoReveal) {
+            | `HighlightAndScroll => revealAndFocusPath(path, state)
+            | `HighlightOnly => (setFocus(Some(path), state), Isolinear.Effect.none)
+            | `NoReveal => (state, Isolinear.Effect.none)
```
caleywoods, line 161: @bryphe This is still causing files to be highlighted (but not scrolled through) with the false setting/`NoReveal` type. Any idea why that might be? Everything else seems to work as expected.

docs/docs/configuration/settings.md
caleywoods: Thanks!

src/Store/FileExplorerStore.re
bryphe: Ah, was just looking at this - I was testing out your branch, and I saw the same behavior as you (if I set `explorer.autoReveal` to `false` - the `NoReveal` is set, but we still get a highlight).

I think it's related to this line: https://github.com/onivim/oni2/blob/46bde3b1820efee54e5d77884e662962357c7b7c/src/Store/FileExplorerStore.re#L150

Confusingly - there is a concept of 'active' _and_ 'focus'. I think we may need to refactor those to a single concept - I'm not sure why we need them both, or what case active != focus. In the meantime - if we move that `setActive` call into the `HighlightScroll`/`HighlightOnly` cases, it seems to work as expected.

bryphe:
> The last bit of your comment was a little confusing. Based on what I have now - I think it looks like I should move the autoReveal config query and switch statement that I have in StoreThread.re into the Some(path) => area inside FileExplorerStore.re. 

Sorry about the confusion! The latest logic you have looks good to me 👍

bryphe:
Oh - and it looks like `esy format` will need to be run to auto-format pass the hygiene check

src/Store/FileExplorerStore.re
glennsl: I believe "active" is the file in the currently active editor and "focus" is the manually selected one. If you unify them, you'd have to change the active editor when you navigate up and down the tree using the keyboard, for example. Or you could of course not specifically highlight the active editor, but just change the focus when the active editor changes. FWIW, VS Code makes this same distinction.

bryphe:
Looks good to me - thanks for all your work on this, @caleywoods !

Somehow our automated CI is not being triggered - I'll try and kick it off manually. Once that's green, we'll bring it in 👍

CrossR:
For this, I think just a configuration parser test is needed. We don't have any form of UI based testing at least, so unless you wanted and were easily able to add a test that can check the state of the `focus`, then I think the config parser is the best we can do currently (and none of the other Stores are tested like that I believe).

caleywoods:
> Looks good to me - thanks for all your work on this, @caleywoods !
> 
> Somehow our automated CI is not being triggered - I'll try and kick it off manually. Once that's green, we'll bring it in 👍

Thank  you. I've added some tests and barring any changes needed with those I think it's ready to go.

test/Core/ConfigurationParserTests.re
```diff
@@ -271,4 +271,68 @@ describe("ConfigurationParser", ({test, describe, _}) => {
       cases,
     );
   });
+
+  test("autoReveal bool(true) setting", ({expect, _}) => {
```
bryphe, line 275: Thanks for adding these tests!

caleywoods:
@bryphe I see one of the checks is failing and it looks like the azure stuff had some kind of connection issue - not sure if I need to do anything to resolve that?

bryphe:
Thanks for catching the formatting, @caleywoods ! 

Indeed, looks like the connection issue isn't related to your change - I'll bring this in. Thanks for the contribution!

bryphe:
This qualified for our [free-license-key-bounty](https://onivim.github.io/docs/for-developers/free-license) - so I e-mailed you a coupon codes via your GitHub e-mail, @caleywoods . Cheers!

caleywoods:
@bryphe Thanks! I appreciate your time on this. Have a great weekend.

PR #2416, opened 2020-09-07
chore(dune): upgrade project lang to 2.5

zbaylin:
- Allows us to tie native libs to `bin_editor` (mostly important for Sparkle)
- The docs are more aligned with what we're using
- Automatic formatting of `dune` files

### DIFF
--- a/assets/dune
+++ b/assets/dune
@@ -1,4 +1,4 @@
 (install
-    (section bin)
-    (package Oni2)
-    (files changelog.xml))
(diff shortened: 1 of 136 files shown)


bryphe:
> Allows us to tie native libs to bin_editor (mostly important for Sparkle)

Ah, nice - thanks @zbaylin ! That's great 👍

bryphe (approved):

PR #2430, opened 2020-09-10
feat(macos): integrate Sparkle autoupdate framework

zbaylin:
This does a couple things:
- adds an `oni2_sparkle` library
- brings in `Sparkle.framework`
- separates out the release-based scripts to an `@release` namespace
- add both `Feature_AutoUpdate` and `Service_AutoUpdate` projects
- probably other things I'm forgetting

This is still WIP, but I'm making good progress so hopefully it should be done (relatively) soon!


scripts/windows/publish.ps1
```diff
@@ -1,38 +1,76 @@
...

-}

-

-mv _publish/Onivim2-win.exe _publish/Onivim2-$SHORT_COMMIT_ID-x64.exe

+$SHORT_COMMIT_ID_UNTRIMMED = (git rev-parse --short HEAD) | Out-String
+
+$SHORT_COMMIT_ID = $SHORT_COMMIT_ID_UNTRIMMED.Trim()
```
bryphe, line 3: This is the windows code-signing script - was anything changed here?

scripts/release.js
```diff
@@ -236,6 +257,10 @@ if (process.platform == "linux") {
     }
     fs.writeFileSync(entitlementsPath, require("plist").build(entitlementsContents))
 
+    if (codesign) {
+        shell(`codesign -fs "${certName}" --deep ${appDirectory}`);
```
bryphe, line 261: Unfortunately the `--deep` strategy doesn't work for Onivim, because there's some nested dependencies it misses. We have a custom codesign script in: scripts/osx/codesign.sh

We'll probably  need to add a step to code-sign the sparkle framework there.

src/Feature/AutoUpdate/Feature_AutoUpdate.re
```diff
@@ -0,0 +1,142 @@
+open Oni_Core;
```
bryphe, line 1: Awesome! Thanks for adding a feature for this 🎉

src/Service/AutoUpdate/Service_AutoUpdate.re
```diff
@@ -0,0 +1,106 @@
...
+
+module Sub = {
+  type params = {
+    automaticallyChecksForUpdates: bool,
+    licenseKey: string,
+    releaseChannel: [ | `Nightly | `Stable],
+    uniqueId: string,
+  };
+
+  module AutoUpdateSubscription =
```
bryphe, line 21: Nice 👍

scripts/windows/publish.ps1
zbaylin: Ah yes I moved `esy create-windows-setup` to `esy @release create-windows-setup` to facilitate Sparkle linking when I add WinSparkle

src/Store/StoreThread.re
```diff
@@ -345,6 +345,25 @@ let start =
...
+      Service_AutoUpdate.Sub.autoUpdate(
+        ~uniqueId="autoUpdate",
+        ~licenseKey,
+        ~releaseChannel,
+        ~automaticallyChecksForUpdates,
+      )
+      |> Isolinear.Sub.map(msg =>
+           Model.Actions.AutoUpdate(Feature_AutoUpdate.Subscription(msg))
+         );
+
```
bryphe, line 366: We could probably consolidate this by having a 

`Feature_AutoUpdate.sub(~config)` - it could handle internally resolving the configuration settings, and then call `Service_AutoUpdate.sub` under the hood. This follows the pattern we have for several features, where the `Feature_XXX.sub` calls into maybe a bunch of `Service_XXX.sub`s

scripts/release.js
zbaylin: Definitely! This is just for local debugging purposes, where it seems to work "okay". It definitely shouldn't be used for actual release codesigning, but I think it should be "good enough" for local debugging.

src/Store/StoreThread.re
zbaylin: Ah that works! I can definitely consolidate that. I based it off the `Service_Font` subscription, which I guess doesn't have a corresponding feature project.

zbaylin: Ah that works! I can definitely consolidate that. I based it off the `Service_Font` subscription, which I guess doesn't have a corresponding feature project.

zbaylin:
@bryphe I tried merging master, but I got the following dependency error (even after deleting _esy & .esy):
```
Conflicting constraints:
  Oni2 -> ocaml@=4.10.1002
  Oni2 -> revery -> flex -> ocaml@>=4.6.1 <=4.10.0
```

Were you able to get around this somehow? Should I just pin `ocaml` to `4.10.1002` in `resolutions`?

scripts/release.js
bryphe: I see - I think it's confusing that we have two code-signing strategies (it's already very complicated as-is 😬 ) - is it possible to use our existing code-sign strategy with a custom cert?

It seems like we could use the `OSX_P12_CERTIFICATE` and `CODESIGN_PASSWORD` environment variables with a self-signed certificate as well - this gets the benefit that we can verify that our production code-signing would work, too.

I'm also thinking about testing - I wonder if it would make sense to check-in a self-signed certificate, and then we could code-sign builds locally with that as part of release. If we had this - I could publish up some sort of dummy-update builds to a `test-self-signed` channel, that are signed with that same certificate - and we could have a test case to validate the update flow.

bryphe: It's tough to test the production flow right now - since it relies on both the current build and the updated build to be fully code-signed with our production certificate - but if we had a self-signed certificate available, we could test most of the flow with that!

zbaylin: > I see - I think it's confusing that we have two code-signing strategies (it's already very complicated as-is 😬 ) - is it possible to use our existing code-sign strategy with a custom cert?

Sure!

> It's tough to test the production flow right now - since it relies on both the current build and the updated build to be fully code-signed with our production certificate - but if we had a self-signed certificate available, we could test most of the flow with that!

Sounds great -- I can check one in or you can, just let me know!

bryphe: Cool! Did you already generate one? If so - you could check-in that one, if you don't mind. We could add it in `scripts/osx` like `test-certificate` (just named clearly so people don't think we accidentally checked-in a production cert 😄 )

bryphe:
> @bryphe I tried merging master, but I got the following dependency error (even after deleting _esy & .esy):

Hmm, ya, I'm seeing this too merging up your branch - looks like the version constraint is tight on `flex` - anything over `4.10.0` will be rejected. I opened a PR to loosen this up, and added it as a resolution: https://github.com/jordwalke/flex/pull/28

scripts/release.js
zbaylin: Just checked in `test-certificate.p12` in `scripts/osx` with the password `OutrunLabs` (I also added a txt file with that information)!

src/Store/StoreThread.re
zbaylin: Added in bb7b802!

scripts/release.js
zbaylin: I just changed `release.js` to use `codesign.sh` with the new test certificate -- let me know if that's in line with what you were thinking!

scripts/release.js
```diff
@@ -236,6 +255,14 @@ if (process.platform == "linux") {
     }
     fs.writeFileSync(entitlementsPath, require("plist").build(entitlementsContents))
 
+    if (codesign) {
+        const cert = fs.readFileSync("./scripts/osx/test-certificate.p12", { encoding: "base64" })
+
+        shell(
```
bryphe, line 261: Nice, thanks @zbaylin !

release.json
```diff
@@ -0,0 +1,16 @@
+{
+  "source": "./package.json",
+  "scripts": {
+    "run": "esy @release x Oni2",
```
bryphe, line 4: When I try to run this command - I hit a failure: 

```
dyld: Library not loaded: @rpath/Sparkle.framework/Versions/A/Sparkle
  Referenced from: /Users/bryphe/oni2/_esy/release/store/i/oni2-8c9f41d8/bin/Oni2_editor
  Reason: image not found
```

.ci/release-and-publish.yml
```diff
@@ -1,10 +1,10 @@
 # Cross-platform set of build steps for building esy projects
 
 steps:
-  - script: esy x Oni2 --help
+  - script: esy @release x Oni2 --help
     displayName: "esy x Oni2 --help"
```
bryphe, line 5: We should update this comment here too - `esy @release x Oni2 --help`

.ci/release-and-publish.yml
```diff
@@ -1,10 +1,10 @@
...
-  - script: esy create-release
-    displayName: "esy create-release"
+  - script: esy @release create
+    displayName: "esy @release create"
```
bryphe, line 7: We could use the `--codesign` flag here to get even more validation than we have today

bryphe (approved):
This is looking great to me, @zbaylin - thanks for all the work to integrate Sparkle!

Still need to troubleshoot the CI failure... but the code change looks great.

zbaylin:
Finally got a green CI build! @bryphe ready to merge when you are!

bryphe:
Thanks for all your hard work on this, @zbaylin ! Very exciting to have an auto-update story for OSX 🚀

PR #2480, opened 2020-09-19
Update why-onivim.md

eleijonmarck:
Allow for editing directly from the docs.

This could be set inplace for all documentation.


bryphe:
Thanks for looking into this, @eleijonmarck ! I updated our `siteConfig.js` in #2483 to add an edit button across all the docs - so I think this is resolved now with that change.

eleijonmarck:
Amazing. Now I can edit the docs right away.

🦄🦄🦄

bryphe:
Awesome, thanks for the help @eleijonmarck , they could definitely use it 👍

Issue #2473, opened 2020-09-18
Add documentation to git


eleijonmarck:
Hey,

I find most of the documentation that is on par is updated through their users by being able to submit a PR to a repo where the documentation for the site relies.

much like this site:
https://www.graphile.org/postgraphile/introduction/

suggest improvements on the page:
https://github.com/graphile/graphile.github.io/edit/develop/src/pages/postgraphile/introduction.md

glennsl:
The docs are already in the repository and accepting contributions: https://github.com/onivim/oni2/tree/master/docs/docs

I'm sure they could be easier to find though, and that would be a good contribution opportunity in itself, I'd say!

glennsl:
This was resolved by #2483, no?

eleijonmarck:
@glennsl yes

PR #2483, opened 2020-09-19
docs: Add 'Edit' button on all pages

bryphe:
I really like @eleijonmarck 's idea in #2473 / #2480 to have an Edit button - this just implements it in a slightly different way - apparently Docusaurus has an `editUrl` property: https://docusaurus.io/docs/en/site-config#editurl-string

When populated, you get a nice `Edit` button at the top of the page:

image


eleijonmarck:
@bryphe nice work!

PR #2492, opened 2020-09-21
feat(ux): Add a vim-navigable tree component

bryphe:
More work on the path to #528 - this expands our vim-navigable-list to handle trees, too.

This PR adds a navigable-tree component, adds the concept of a tree view model (a flattened array of tree nodes, more amenable to the index-based operations we need for vim-style nav), and hooks up to search and diagnostics, to improve the UX (aka, make search a little more useful...):

This lays the scaffolding to actually address #528 - next step will be to update the file explorer model to match up with the `Component_VimTree`. Getting closer! Will also be useful to finally finish #1181 and potentially hook up document symbols in the explorer.

__TODO (blocking check-in):__
- [x] Fix warnings / remove unused code
- [x] UX: Fix hover / focus background (should cover entire row)
- [x] Diagnostic: Add line number back to result
- [x] Wire up selection - this was pulled out temporarily
- [x] Wire up expand / collapse

__Next steps:__
- Handle scroll commands (Control+D, Control+E, etc)
- Implement search functionality across vimlist/vimtree
- Hook up to file explorer


fanantoxa:
@bryphe I see you already working on making search on right side. Nice to see that.
I'll drop my update then (don't worry that was not to much)

bryphe:
@fanantoxa - sorry this interfered with work you're doing! 

I think there may be an issue with the sidebar-navigation when it's on the right side - not sure we handle the window motion to it correctly. I'll check it out shortly.

PR #2497, opened 2020-09-22
feat(Windows): integrate WinSparkle autoupdate framework

zbaylin:
This extends #2430 to use WinSparkle on Windows.


scripts/release.js
```diff
@@ -313,6 +331,20 @@ if (process.platform == "linux") {
...
+            '"C:\\Program Files\\Git\\cmd\\git.exe" rev-list --count origin/master',
+        ).replace(/(\r\n|\n|\r)/gm, "")
+        const semvers = package.version.split(".")
+        const bundleVersion = `${semvers[0]}.${semvers[1]}.${numCommits}`
+
+        const oni2Ini = `
+        [Application]
+        Version = ${bundleVersion}
+        `
+        fs.writeFileSync(path.join(platformReleaseDirectory, "Oni2.ini"), oni2Ini)
```
bryphe, line 345: Thanks for figuring out how to set this up for us 👍

src/oni2-sparkle/stubs/winsparkle.c
```diff
@@ -0,0 +1,126 @@
...
+  CAMLparam1(vData);
+  
+  CAMLreturn(Val_unit);
+}
+
+CAMLprim value oni2_SparkleSetFeedURL(value vUpdater, value vUrlStr) {
+  CAMLparam2(vUpdater, vUrlStr);
+  static int initialized = 0;
+
+
```
bryphe, line 87: Is this `initialized` variable needed? Or just for debugging?

src/oni2-sparkle/stubs/winsparkle.c
```diff
@@ -0,0 +1,126 @@
...
+  win_sparkle_set_appcast_url(String_val(vUrlStr));
```
bryphe, line 88: Does sparkle make a copy of this string? If not - we might need to. The GC can move strings around, if we just keep the pointer.

src/reason-oniguruma/config/discover.ml
```diff
@@ -1,54 +1,54 @@
...
-;;
-
-let cxx_flags = c_flags
-;;
-
-Configurator.V1.Flags.write_sexp "c_flags.sexp" c_flags;
-Configurator.V1.Flags.write_sexp "cxx_flags.sexp" cxx_flags;
-Configurator.V1.Flags.write_sexp "flags.sexp" flags;
-Configurator.V1.Flags.write_sexp "flags_with_sanitize.sexp" flags_with_sanitize;
+
```
bryphe, line 1: Looks like some crazy formatting here! Maybe a bug in our format scripts on Windows?

We should revert these changes, though

src/bin_editor/Oni2_editor.re
```diff
@@ -395,6 +396,8 @@ switch (eff) {
       App.quit(~askNicely=true, app);
     };
 
+    Callback.register("oni2_close", close);
```
bryphe, line 399: Cool, looks like this is wired up now!

src/Core/Diff.mli
```diff
@@ -1,47 +1,47 @@
...
-      [d1] has the same size as [a1].
-      [d2] has the same size as [a2].
-      [d1.(i)] is [True] if [a1.(i)] has no corresponding value in [a2].
-      [d2.(i)] is [True] if [a2.(i)] has no corresponding value in [a1].
-      [d1] and [d2] have the same number of values equal to [False].
-    Can be used to write the [diff] program (comparison of two files),
-    the input arrays being the array of lines of each file.
-    Can be used also to compare two strings (they must have been exploded
-    into arrays of chars), or two DNA strings, and so on.
+(* This file has been adapted from camlp5:
```
bryphe, line 1: We should revert these whitespace-only changes to `Diff.ml`/`Diff.mli`

scripts/release.js
```diff
@@ -27,6 +27,14 @@ const extensionsSourceDirectory = path.join(process.cwd(), "extensions")
 const eulaFile = path.join(process.cwd(), "Outrun-Labs-EULA-v1.1.md")
 const thirdPartyFile = path.join(process.cwd(), "ThirdPartyLicenses.txt")
```
bryphe, line 28: I just realized reading through we're missing the licenses for sparkle-osx and winsparkle in our `ThirdPartyLicenses.txt` - would you mind adding them?

src/oni2-sparkle/stubs/winsparkle.c
zbaylin: You're right, it doesn't look like they make a copy of the string: 
https://github.com/vslavik/winsparkle/blob/91bb56ea20eb787c0cbb1f00277f98a4cbef594a/src/settings.h#L181
I'll make a copy of the string to pass to WinSparkle!

src/reason-oniguruma/config/discover.ml
zbaylin: Yeah it's very strange -- looks like an ocamlformat bug. I'll revert the changes.

scripts/release.js
zbaylin: Sure!

bryphe: Thanks Zach!

src/Core/Diff.mli
zbaylin: Fixed!

src/reason-oniguruma/config/discover.ml
zbaylin: Fixed!

src/oni2-sparkle/stubs/winsparkle.c
zbaylin: Fixed in 1532cee!

scripts/release.js
zbaylin: Added in ea1363a!

src/oni2-sparkle/stubs/winsparkle.c
```diff
@@ -0,0 +1,127 @@
...
+CAMLprim value oni2_SparkleDebugLog(value vData) {
+  CAMLparam1(vData);
+  
+  CAMLreturn(Val_unit);
+}
+
+CAMLprim value oni2_SparkleSetFeedURL(value vUpdater, value vUrlStr) {
+  CAMLparam2(vUpdater, vUrlStr);
+
+  char *urlStr = malloc(caml_string_length(vUrlStr) * sizeof(char));
```
bryphe, line 86: I believe this would actually need to be `caml_string_length(vUrlStr) + 1`, to leave an extra byte for the null-terminating character.

At least, the examples seem to do this too:
```
  /* Copy the string argument to a C string, allocated outside the
     OCaml heap. */
  name = caml_stat_alloc(caml_string_length(vname) + 1);
  name = caml_strdup(String_val(vname));
```
http://caml.inria.fr/pub/distrib/ocaml-4.05/ocaml-4.05-refman.html

I believe we could just use those same functions here

ThirdPartyLicenses.txt
```diff
@@ -36,6 +36,8 @@ This project incorporates components from the projects listed below. The origina
 30. uucp (https://github.com/dbuenzli/uucp)
 31. vscode-codicons (https://github.com/microsoft/vscode-codicons)
 32. openssl (https://github.com/openssl/openssl)
+33. Sparkle (https://sparkle-project.org/)
+34. WinSparkle (https://winsparkle.org/)
```
bryphe, line 40: Thanks for adding these 💯

src/oni2-sparkle/stubs/winsparkle.c
zbaylin: Interesting! I just changed it to this in 3302915, but the way those functions work is a little unclear to me -- what is the point of allocating the pointer in the first line, only to discard it in the second? I know you didn't write the example, I'm just confused haha.

Here I see a line with the exact same comment, but done in a more understandable (at least to me!) one-line solution: https://caml.inria.fr/pub/docs/manual-ocaml/intfc.html
```c
 /* Copy the string argument to a C string, allocated outside the
     OCaml heap. */
  name = caml_stat_strdup(String_val(vname));
```

I wonder which is "better" -- the link you provided is for OCaml 4.05, so it might be out of date? I'll look into it and see if anything's changed!

zbaylin: Looks like it may be out of date, actually -- the OCaml 4.10 manual has the same line as the one in the FFI docs:
http://caml.inria.fr/pub/distrib/ocaml-4.10/ocaml-4.10-refman.html

I'll update it to that one!

zbaylin: Fixed in ce55580!

bryphe (approved):
Looks good to me, @zbaylin ! Thanks for making the changes :+1:

PR #2500, opened 2020-09-23
docs: fix release steps

CrossR:
Just updates the release step, since the release is now in its own sandbox, you need to run Oni2 in that sandbox once for the binary to exist, to be copied for the release.

Arguably, the scripts should be a little smarter about this (i.e. check, error, say "Run `esy ....`).


bryphe:
Thanks for fixing this, @CrossR ! Nice catch

PR #2503, opened 2020-09-23
feat(ux): Use opacity to make the focused pane more obvious

bryphe:
With all the vim-navigation pieces we're adding, it can be hard to follow where focus is going, and get nice visual feedback to confirm the gesture behaved as the user intended.

This uses opacity to help draw the eye, in a subtle way, to changes in focus:

Depends on https://github.com/revery-ui/revery/pull/1002


Miziak:
Hi @bryphe! Awesome feat. I shoot idea of config value for opacity strength would be very nice 👍

bryphe:
Hi @Miziak ! Sure, that's a good idea - I added an `oni.inactiveWindowOpacity` setting:

Issue #2517, opened 2020-09-27
Readline shortcuts in Ctrl-P and vim :commands


fratajczak:
Ctrl-W (delete previous word) is missing from both the Ctrl-P and vim commands text boxes.
Ctrl-K works in the Ctrl-P text box but not in the :command box, it prints some kind of unicode char? (see screenshot below)
![Ctrlk](https://user-images.githubusercontent.com/33835579/94370759-11768e00-00f2-11eb-8b11-a12516ca570f.png)

PR #2518, opened 2020-09-27
feat(input): add some readline shortcuts in InputText

fratajczak:
This adds the Ctrl-U, Ctrl-K, Ctrl-C shortcuts from readline (bash) to InputText components (like the Find in Files search bar).

I also modified the previous implementation of removeWord (Ctrl-W / Ctrl-Backspace) to delete multiple spaces between words like the original readline implementation.

By the way, Ctrl-W no longer works in text boxes because it's used to switch between panes, so maybe it should be removed, or maybe make it so you need to press escape then Ctrl-W to be able to switch panes in a text box?

There are also some other issues with readline shortcuts in Ctrl-P and :command text boxes, I didn't find where to fix them. (the vim command text box is handled by libvim, I think?) See #2517

### DIFF
--- a/src/Components/InputText/Model.re
+++ b/src/Components/InputText/Model.re
@@ -63,6 +63,9 @@ module Internal = {
     let finalIndex = 0;
     let index = ref(max(focus - 1, finalIndex));
 
+    while (index^ > finalIndex && separatorOnIndexExn(index^, text)) {
+      index := index^ - 1;
+    };
     while (index^ > finalIndex && !separatorOnIndexExn(index^ - 1, text)) {
       index := index^ - 1;
     };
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for the contribution @fratajczak - this is great!

> By the way, Ctrl-W no longer works in text boxes because it's used to switch between panes, so maybe it should be removed, or maybe make it so you need to press escape then Ctrl-W to be able to switch panes in a text box?

Right - we're relying a lot on `Control-W` shortcuts to navigate around text areas.

I like your idea to potentially have a pseudo-normal-mode to use the Ctrl-W shortcuts - the same experience you'd have in the editor. Another idea I was tossing around was having the equivalent of a [`termwinkey`](https://vimhelp.org/options.txt.html#%27termwinkey%27) or `guiwinkey`

PR #2537, opened 2020-10-02
fix(ux/#2404): support macos shortcuts in inputs

fiddler:
**Issue:** Shortcuts in the text inputs are not handled on macOS.

**Fix:** Added handling for most common macOS keyboard shortcuts.

**NOTE:** While testing all the keyboard shortcuts by hand I noticed that for some reason `<C-w>`, `<C-LEFT>` and `<C-RIGHT>` don't do anything on my mac. It's quite mysterious since `<C-h>` and `<S-C-LEFT>` for example work just fine. These issues were repeatable on the master branch as well, so not a result of this PR. @bryphe perhaps you could point me to where the keyboard input handling happens - I could do some digging and see if I can figure out why those shortcuts don't work my end?


bryphe:
Hi @fiddler - thanks for adding these! Very nice improvement to the text input on OSX.

> While testing all the keyboard shortcuts by hand I noticed that for some reason <C-w>, <C-LEFT> and <C-RIGHT> don't do anything on my mac. It's quite mysterious since <C-h> and <S-C-LEFT> for example work just fine.

Good question, thanks for the thorough testing. It looks like two different issues here -

__Issue 1: `<C-LEFT>` & `<C-RIGHT>` not working__

This is due to OSX assigning these keybindings to mission control / change workspaces.

You can check if these keybindings are set by going to System Preferences -> Keyboard -> Shortcuts -> Mission Control (they're set by default, I believe):
Screen Shot 2020-10-02 at 5 05 27 PM

If these shortcuts are set - the key presses don't even make it to Onivim to process. I turned off these shortcuts, and I saw `<C-LEFT>` and `<C-RIGHT>` input be processed after that.

__Issue 2: `<C-w>` not working__

The `<C-w>` issue is because, in some parts of the UI with text inputs, we allow window navigation - so the `<C-w>` is being picked up by bindings like `<C-w><C-j>` to move to a different window.

There's a couple of possibilities - discussed a bit here too: https://github.com/onivim/oni2/pull/2518 - we either can disable the window-navigation bindings in text input, and set it up the input text to be 'modal' (ie, you'd need to press `<esc>` to enter normal mode, and then use window navigation) - or we could have an alternate key for initiating windows navigation (like the `termwinkey` setting in vim). I need to think a bit more about the user experience around that. Let me know if you have ideas.

> These issues were repeatable on the master branch as well, so not a result of this PR.

Definitely not related to your PR 👍  We can go ahead and merge this, and leave in the `<C-w>` binding - and depending on the solution for issue 2, it'll eventually be picked up.

fiddler:
Thanks for the in-depth explanations @bryphe!

Disabling those Mission Control shortcuts indeed made the `<C-LEFT/RIGHT>` work. For the `<C-w>` handling I do not have an informed opinion to offer since I've not used the window navigation bindings.

Seems there is a bunch of checks failing - is there anything I should do regarding this PR to get it to pass? I checked some of the errors, but not really sure what those are about.

bryphe:
Looks like CI is green now - thanks @CrossR for kicking off a new build, and thanks @fiddler for the contribution!

Issue #2535, opened 2020-10-01
Filename does not appear in search results


BenoitPingris:
**OS:** Manjaro
**Oni version:** Onivim 2 (0.5.1-nightly)

When I use the search in project feature, in the results I do not have the filename displayed, only its icon as you can see on the image.

![2020-10-01-183207_611x564_scrot](https://user-images.githubusercontent.com/29386109/94839540-6f1f1900-0417-11eb-93d8-6d40ea0b3249.png)

I tried on many folders and I still had the same behavior.

I also give you a log file if it can help
[oni.log](https://github.com/onivim/oni2/files/5313640/oni.log)

Thanks!

BenoitPingris:
This issue seems to have been fixed by this [change](https://github.com/onivim/oni2/commit/b86ccae4be926a9399fdeeed8a76505340c46935)
Now I have the filename in the results like we would expect !

PR #2538, opened 2020-10-02
fix(ux): Missing fontFamily setting on new <Text /> components

bryphe:
__Issue:__ In some of the new UI - like the `<FileItemView />` used in search and diagnostics, and the symbol view - a `fontFamily` wasn't set, so this would fall back to a default system font in those new pieces of UI, which isn't desirable - which can either cause a mismatching font to show, or not display at all.

__Defect:__ Missed setting the `fontFamily` parameter on these new elements.

__Fix:__ Set the missing properties.

As a next step, it may make sense to have an `<OniText font />` component, that takes a `UiFont.t` as a required parameter, so it isn't missed elsewhere.

### DIFF
--- a/src/Components/FileItemView.re
+++ b/src/Components/FileItemView.re
@@ -101,6 +101,7 @@ module View = {
         <Text
           style={Styles.fileText(~theme)}
           text=filename
+          fontFamily={uiFont.family}
           fontSize={uiFont.size}
           fontWeight=Revery.Font.Weight.SemiBold
         />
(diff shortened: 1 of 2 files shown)


CrossR:
Could this be causing https://github.com/onivim/oni2/issues/2535?

bryphe:
I think it could be, @CrossR ! 

Although it is surprising that the default font we pick up in Revery wouldn't render. I need to try and see if I can reproduce this to be sure - but given other elements seem to render OK from the screenshot, I suspect this will fix #2535

PR #2547, opened 2020-10-03
fix(editor/#2421): Reopen after closing is not working

### LINKED ISSUE #2421: Can't reopen file after closing
Using the `Ctrl+p`/`Cmd+p` dialog to open a file works the first time, but if I open the file, close it with `:q`, and then attempt to reopen it with `Ctrl+p`/`Cmd+p`, the file does not open. This seems semi-related to #803.

Screen-Recording-2020-09-08-at-1

I tried using `--log-file` but I couldn't get it to write anything to the file.

Using 0.5.1 nightly as of a few hours ago.
macOS 10.15.6

bryphe:
__Issue:__ As described in #2421 - when closing a buffer, and then reopening, the buffer would not open the second time.

__Defect:__ There was an effectful depending on a 'buffer enter' event - this was used when we used the [`msgpack`](https://neovim.io/doc/user/api.html#nvim_buf_attach()) API - we'd get notified that a buffer was loaded via a buffer enter event, and then have a separate update event.

This flow was emulated in `reason-libvim` - having a split between an enter and update event. However, this added some complexity - Onivim relied on a buffer enter event to populate a new buffer, but it was a globably stateful - and sometimes that state wouldn't match up. For #2421 , we wouldn't get that buffer enter event, and an editor wouldn't be opened.

__Fix:__ Remove that stateful buffer-entered event, which involves some refactoring around the pipeline... but it gives us much more flexibility in how we integrate buffers (ie, loading a buffer without an editor).

Fixes #2421


src/Feature/Buffers/Feature_Buffers.re
```diff
@@ -53,30 +55,55 @@ let isModifiedByPath = (buffers: model, filePath: string) => {
 let setIndentation = indent =>
   Option.map(buffer => Buffer.setIndentation(indent, buffer));
 
-let disableSyntaxHighlighting =
-  Option.map(buffer => Buffer.disableSyntaxHighlighting(buffer));
+// TODO: When do we use this?
+//let disableSyntaxHighlighting =
+//  Option.map(buffer => Buffer.disableSyntaxHighlighting(buffer));
```
bryphe, line 60: Hook this back up, if currently unused.

PR #2551, opened 2020-10-06
fix(extensions/#2484): Searching for extensions w/o network crashes Onivim

### LINKED ISSUE #2428: ContextKeys are missing and not properly defined
Current implementation of `src\Model\ContextKeys.re` not fully correct

It missing both `textInputFocus` and `menuFocus`
The were removed here https://github.com/onivim/oni2/pull/786 but wasn't added back anywhere.

Also after looking in `ContextKeys.re` I've noticed that other key also not propely defined.
For example `editorTextFocus`
```
| TerminalInsert
| TerminalNormal
| TerminalVisual(_) => false
| _ => true
```

Basically it say that it on focus all the time but not terminal.
And there kinda wrong because you might be in `Menu`, `Search`, `WildMenu`, and other places at this moment.

That why `paste into text input` currently work in `insertMode` and also work when I add `editorTextFocus`

Same goes for `terminalFocus`
is active when
```
| TerminalInsert
| TerminalNormal
| TerminalVisual(_) => true
| _ => false
```
But still can open menu or smth.

Can we refine and refactor this to be more persistent and clear about different `ContextKeys`?
It already blocking me from further improving `textInputs` and few other things.
In addition we have [Keybinding Suggestions and Contribution Opportunities](https://github.com/onivim/oni2/issues/1423)
as `good-first-issue` and people going to be confused while working on it.

### LINKED ISSUE #2484: Searching for extensions while network disconnected crashes onivim2
## MacOS:

 - 10.15.6

## Onivim2:

 - 0.5.1 nightly

## Steps:

 - With wifi disconnected from network or turned off
 - Launch Onivim2 (can already be launched before or after the disconnect)
 - Change to extensions page with mouse and click in search field
 - search "go" and press return key
 - Onivim2 crashes every time I attempt this.

### LINKED ISSUE #2545: Service_Net: node errors not being handled properly in Lwt
I discovered this when writing #2520. If I try to make a web request with my NIC disabled, I get the following node error (expected):
```
[ERROR]    +0ms Oni2.Core.NodeTask : events.js:292
      throw er; // Unhandled 'error' event
      ^

Error: getaddrinfo ENOTFOUND v2.onivim.io
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26)
Emitted 'error' event on ClientRequest instance at:
    at TLSSocket.socketErrorListener (_http_client.js:426:9)
    at TLSSocket.emit (events.js:315:20)
    at emitErrorNT (internal/streams/destroy.js:92:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:60:3)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  errno: 'ENOTFOUND',
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'v2.onivim.io'
}
```

However, this in turn causes an Lwt error which causes a full-on crash:
```
(Invalid_argument Lwt.wakeup_exn)
Called from file "src/process.ml", line 107, characters 12-54
```

I can confirm that this also happens with the extension search menu, so it's not a lack of error handling in #2520. I wish the backtrace was a bit more illuminating, but it probably has to do with the external `node` integration.

bryphe:
__Issue:__ When searching for extensions without a network connection, the editor would crash with:

```
(Invalid_argument Lwt.wakeup_exn)
Called from file "src/process.ml", line 107, characters 12-54
```

__Defect:__ There is a race condition here - we are waiting for `EOF` on `stdout` to return the promise - and claim success. However, in the error case, we'd get EOF on stdout, run `Lwt.wakeup` to complete the promise, and _then_ get a call that the process has exited - and we'd call `Lwt.wakeup_exn` on the same promise. This triggers an invalid argument exception - we're trying to complete and then fail the same promise.

__Fix:__ Wait until both the process has exited, and stdout has given the `EOF` - then we'll complete the promise. At that point, we know if the process was successful or failed, and we could do the appropriate action on the promise.

In addition, while debugging this, I saw that parsing was failing frequently when querying `open-vsx` - our parsing code was making the assumption that an `icon` field was always available, when that's not the case. 

Finally, I also added an error message when the search query fails.

Fixes #2484 , and should fix #2545 as well

### DIFF
--- a/node/success.js
+++ b/node/success.js
@@ -1,0 +1,1 @@
+process.stdout.write("Successful node task!");
--- a/src/Core/NodeTask.re
+++ b/src/Core/NodeTask.re
@@ -25,10 +25,7 @@ let run = (~name="Anonymous", ~args=[], ~setup: Setup.t, script: string) => {
 
     let success = () => {
       let allOutput =
-        buffers^
-        |> List.rev
-        |> List.map(Luv.Buffer.to_string)
-        |> String.concat("");
+        buffers^ |> List.rev_map(Luv.Buffer.to_string) |> String.concat("");
       Log.debugf(m => m("Got output: %s", allOutput));
       Lwt.wakeup(resolver, allOutput);
     };
--- a/src/Core/NodeTask.re
+++ b/src/Core/NodeTask.re
@@ -11,7 +11,7 @@ let run = (~name="Anonymous", ~args=[], ~setup: Setup.t, script: string) => {
   let (promise, resolver) = Lwt.task();
 
   let scriptPath = Setup.getNodeScriptPath(~script, setup);
-  Log.info("Using node path: " ++ scriptPath);
+  prerr_endline("Using node path: " ++ scriptPath);
 
   let result = {
     open Base.Result.Let_syntax;
@@ -53,6 +53,7 @@ let run = (~name="Anonymous", ~args=[], ~setup: Setup.t, script: string) => {
         stdoutPipe,
         fun
         | Error(`EOF) => {
+            prerr_endline ("Got EOF");
             Log.info("Got EOF on stdout");
             Luv.Handle.close(stdoutPipe, ignore);
             let allOutput =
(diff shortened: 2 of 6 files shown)


bryphe:
Fixed branch name in #2552

Issue #2550, opened 2020-10-06
Latest Build: Download has cursor issues, building dies on macOS.


raguay:
I’m trying to get the latest build of oni2 after I had to wipe my full system and start over. So, I download the pre-built binaries for macOS. For it, the cursor isn’t in the right location after a line hits around the 20th character mark. It looks like a variable font is being used as the blinking box shrinks and grows as you move around. Look awful.

So, I’m trying to build from the repository like I used to. It is constantly failing as well with the following errors:

```bash
    error: include location '/usr/local/include' is unsafe for cross-compilation [-Werror,-Wpoison-system-directories]
    1 error generated.
    ninja: build stopped: subcommand failed.
```

You get this message for every file compiled. I never had that happen on my system before the wipe, but since then I’ve turned on SIP. Not sure if that is related or not. I’m not to good at the Revary build system and language yet. So, I’m not sure where to debug this one.

raguay:
My building issue seems to be the same as `onivim2 build failure on macOS Catalina with macports #2508`. I’m on Catalina latest but I use homebrew and not macports. Therefore, I doubt it’s an issue with the installers being used. I would of posted under that one, but I just noticed it after posting (I must of overlooked it).

bryphe:
> I’m trying to get the latest build of oni2 after I had to wipe my full system and start over

Ouch, sorry to hear you had to wipe your system! Always a pain.

I assume that you don't have any custom configuration, then? My first thought is maybe it's not finding a font, and falling back to a non-monospace font.

And I'm assuming it's Catalina?

> . So, I download the pre-built binaries for macOS. For it, the cursor isn’t in the right location after a line hits around the 20th character mark.

This is very strange! Would it be possible to see a screenshot? What sort of file is it?

I tried to repro on a retina display device (Catalina) w/ latest master - but I'm not seeing it. Might help to see a screenshot.

Another thing that might help is to see the logs - can tell us what font is being loaded and if it is falling back to a non-monospace font.

raguay:
Your right. I forgot to install my fonts! 🤦 Looks and works much better with the right font. 

Therefore, Oni2 doesn't work well for non-monospaced fonts. It also doesn't compile on my system right now. Yes, I'm using the latest Catalina, but I don't have a retina display. I have a 2017 Macbook Air.

raguay:
Also, where are logs kept? I've looked using Console and in the `~/.config/oni2` directories and can't find logs.

raguay:
[...]

PR #2582, opened 2020-10-14
chore(deps): update Skia to fix XCode 12 bug

### LINKED ISSUE #2508: onivim2 build failure on macOS Catalina with macports
On a current macOS Catalina installation using macports, I am unable to build oni2. The build bombs out with:

https://gist.github.com/raw-bin/4290e6b75bf65df09d467dcf8f23e542

Something (perhaps a CMake include directive or some esy equivalent) appears to reference /usr/local/include as an include path which I believe is incorrect for a macports installation (which uses /opt/local/include).

For some reason, this results in a warning due to the compiler being built with '-Wpoison-system-directories'. That then bombs out because of '-Werror'.

As a data point: I managed to build one of the objects above in isolation by cutting and pasting the compiler invocation on the command line while taking care to remove '-Werror'. Not ideal of course. I couldn't figure out where to override the CFLAGS that are used for building objects in esy_libjpeg_turbo to remove '-Werror' globally. That would have given me a stop-gap build.

Any pointers appreciated. I'd rather not switch to Homebrew!

Thanks

zbaylin:
Picks up revery-ui/esy-skia#25

Fixes #2508 and #2550

### DIFF
--- a/bench.esy.lock/index.json
+++ b/bench.esy.lock/index.json
@@ -1,5 +1,5 @@
 {
-  "checksum": "6c75cff7f6f58bccf8419b7648733903",
+  "checksum": "486019be99d740bdc109fbe369502a25",
   "root": "Oni2@link-dev:./package.json",
   "node": {
     "yarn-pkg-config@github:esy-ocaml/yarn-pkg-config#db3a0b63883606dd57c54a7158d560d6cba8cd79@d41d8cd9": {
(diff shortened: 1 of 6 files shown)


bryphe:
Awesome - thanks for investigating and unblocking this, @zbaylin ! 💯

bryphe (approved):

PR #2592, opened 2020-10-16
Make example color valid Hex color

bartduisters:


bartduisters:
> The other color was also valid. The hex represents the red, green, blue, and alpha values.

My bad. I'm only used to the RGB hex codes. I'll close the PR.

PR #2593, opened 2020-10-16
refactor(vim): Synchronize vim modes

bryphe:
This change is the next step from #2584 - when the mode provided by the `Editor.t` is different from the 'real' Vim state - synchronize it. 

Brings in https://github.com/onivim/libvim/pull/231 so that we can change the visual / select easily as part of that synchronization.

### DIFF
--- a/src/reason-libvim/Mode.re
+++ b/src/reason-libvim/Mode.re
@@ -5,12 +5,12 @@ type t =
   | Insert({cursors: list(BytePosition.t)})
   | CommandLine
   | Replace({cursor: BytePosition.t})
-  | Visual({range: VisualRange.t})
+  | Visual(VisualRange.t)
   | Operator({
       cursor: BytePosition.t,
       pending: Operator.pending,
     })
-  | Select({range: VisualRange.t});
+  | Select(VisualRange.t);
 
 let show = (mode: t) => {
   switch (mode) {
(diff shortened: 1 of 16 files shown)


src/reason-libvim/Mode.re
```diff
@@ -82,3 +82,65 @@ let isOperatorPending =
   fun
   | Operator(_) => true
   | _ => false;
+
+module Internal = {
+  let ensureNormal = () =>
+    if (!isNormal(current())) {
+      Native.vimKey("<ESC>");
+      Native.vimKey("<ESC>");
+    };
```
CrossR, line 91: Out of curiosity, how does this play with keybindings? I.e. if for whatever reason, a user has rebound any of these core keys, is that an issue?

I've not looked at the implementation of `Native.vimKey`, so perhaps that just ignores user binds, but if not was wondering if it could cause some strange behaviours later on.

bryphe: Good question @CrossR ! `Native.vimKey` basically calls [`vimKey`](https://github.com/onivim/libvim/blob/d5f023c4f976dde93bacc20cd72823a9a4df1e60/src/libvim.h#L173) in libvim - and it does get impacted by `nnoremap` bindings, currently.

However - my plan for the bindings is to wire up all the mapping to be handled through Onivim. So when `nnoremap` is run - it'll be Onivim handling it (same for all the other mapping / abbrev operations). Once that's in place - all of the `Native.vimKey` / `Native.vimInput` will be handle outside the keybindings - so it won't interfere. Planning on starting this work once the word wrap / scrolling stuff is done (it's the next thing that needs to be done to unblock VimL)

PR #2596, opened 2020-10-19
Add file preview

fanantoxa:
- Preview works for file tree
- Preview for for Search
- Preview works for SCM
- Preview works for Problems
- Double click on tab to discard preview
- Double click on File (in file tree, SMC, etc) to discard preview
- Works with layout splits (groups?) properly
- On buffer modification preview get discarded

Here some example:
preview


src/Components/VimList/Component_VimList.re
```diff
@@ -117,7 +118,8 @@ type msg =
 
 type outmsg =
   | Nothing
-  | Selected({index: int});
```
CrossR, line 120: I'm not 100% sure on the changing/removal of this name. `Clicked` and `DoubleClicked` don't have a vim equivalent, if I hit enter, I'm not clicking at all, so the name seems strange to me.

src/Components/FileExplorer/Component_FileExplorer.re
```diff
@@ -83,7 +83,10 @@ module Effects = {
 type outmsg =
   | Nothing
   | Effect(Isolinear.Effect.t(msg))
-  | OpenFile(string)
+  | OpenFile({
+      filePath: string,
+      preview: bool,
+    })
```
CrossR, line 89: Would it be better to have `OpenFile(str)` and `PreviewFile(str)`, instead of `OpenFile(str, bool)`? (I don't know the issues that would cause, but making it explicit that it is a preview could be nice?)

src/Components/FileExplorer/Component_FileExplorer.re
```diff
@@ -242,19 +245,28 @@ let update = (~configuration, msg, model) => {
...
-    | Selected(node) =>
+    | Component_VimTree.Collapsed(_) => (model, Nothing)
+    | Component_VimTree.Clicked(node) =>
       // Set active here to avoid scrolling in BufferEnter
-      (model |> setActive(Some(node.path)), OpenFile(node.path))
-    | Nothing => (model, Nothing)
+      (
+        model |> setActive(Some(node.path)),
+        OpenFile({filePath: node.path, preview: true}),
+      )
```
CrossR, line 262: I think another consequence of removing `Selected` is that I believe this will make `Enter` in the file explorer open a preview.

Not sure what the desired approach would be there actually.....If I go into the explorer and hit enter on a file, what should happen? Preview or fully open?

Not sure, though I guess the act of double clicking doesn't really exist with an enter key (plus when you hit enter, focus jumps to the buffer, so right now there is no way of not previewing a file via the vim binds).

src/Core/Diff.ml
```diff
@@ -1,192 +1,192 @@
-(* This file has been adapted from camlp5:
```
CrossR, line 1: Apparently Windows really didn't like the line endings of these files? Very strange!

src/Feature/Buffers/Feature_Buffers.re
```diff
@@ -441,6 +512,47 @@ module Effects = {
...
+        ~font: Service_Font.font,
+        ~languageInfo: Exthost.LanguageInfo.t,
+        ~split=`Current,
+        ~position=None,
+        ~grabFocus=true,
+        ~filePath,
+        model,
+      ) => {
+    Isolinear.Effect.createWithDispatch(
+      ~name="Feature_Buffers.openFileInEditor", dispatch => {
```
CrossR, line 527: "openFileInEditor" -> "previewFileInEditor"

Also if its just the name and bool that are different here, maybe remove the duplicated logic?

CrossR (commented):
Looks good! Feature works well and really helps with some of the language features that open up new buffers for you. Excited to have this feature in!

Main question for me is around the Vim bindings, and what/how selecting stuff with those should work.

After this, can add some config options around this as well, to enable/disable the previews (I believe VSCode has the option to turn them off, and that may be needed with "vim style" tabs, since this feature gets very confusing when you can't see tab names).

The only other question I had, which I don't really have an answer to either, is should these files show up in the `Ctrl-Tab` (i.e. recent files) menu? I can see arguments for yes and for no, so not 100%.

bryphe:
Thanks for looking at this, @fanantoxa ! Very exciting

src/Components/VimList/Component_VimList.re
bryphe: I agree! I'd view the List as having events like 'Click', 'Double-Click', 'Enter' as input, and then `outmsg` as the produced effect.

Good point about the vim equivalent - I think to be consistent with NerdTree/Ex/etc, we'd want to have the keyboard behavior be to open a file.

src/Components/FileExplorer/Component_FileExplorer.re
bryphe: Good point @CrossR  - I think the `Enter` key should behave similar to NerdTree/Ex, so open a file here.

src/Core/Diff.ml
bryphe: Ouch, ya, we should revert these for sure (might fail the format check anyway)

docs/docs/configuration/settings.md
```diff
@@ -43,6 +43,8 @@ The configuration file, `configuration.json` is in the Oni2 directory, whose loc
 
 - `editor.largeFileOptimizations` __(_bool_ default: `true`)__ - When `true`, Onivim will turn off certain settings like syntax highlighting for large files.
 
+- `editor.enablePreview` __(_bool_ default: `true`)__ - When `true`, Onivim will opend files in preview mode unless change is made or DoubleClick on tab or file
```
bryphe, line 46: Thanks for updating the documentation @fanantoxa

docs/docs/configuration/settings.md
```diff
@@ -43,6 +43,8 @@ The configuration file, `configuration.json` is in the Oni2 directory, whose loc
 
 - `editor.largeFileOptimizations` __(_bool_ default: `true`)__ - When `true`, Onivim will turn off certain settings like syntax highlighting for large files.
 
+- `editor.enablePreview` __(_bool_ default: `true`)__ - When `true`, Onivim will open files in _preview mode_ unless a change is made or the tab is double-clicked. In _preview mode_, the editor tab will be re-used.
```
bryphe, line 46: We should switch this to `workbench.editor.enablePreview` to be consistent with VSCode

src/Components/FileExplorer/Component_FileExplorer.re
```diff
@@ -242,19 +243,29 @@ let update = (~configuration, msg, model) => {
...
         Effect(
           Effects.load(node.path, configuration, ~onComplete=newNode =>
             NodeLoaded(newNode)
           ),
         ),
       )
-    | Collapsed(_) => (model, Nothing)
-    | Selected(node) =>
+    | Component_VimTree.Collapsed(_) => (model, Nothing)
+    | Component_VimTree.Touched(node) =>
```
bryphe, line 255: I like the way you divided up the responsibility here - `Touched` vs `Selected`. I wonder if `Selected` should potentially preview as well - this would occur if the user selected a file via keyboard?

src/Components/VimList/Component_VimList.re
```diff
@@ -346,6 +348,15 @@ let update = (msg, model) => {
   | MouseClicked({index}) =>
     let isValidIndex = index >= 0 && index < Array.length(model.items);
 
+    if (isValidIndex) {
+      (model |> setSelected(~selected=index), Touched({index: index}));
+    } else {
+      (model, Nothing);
+    };
+
+  | MouseDoubleClicked({index}) =>
```
bryphe, line 357: Thanks for adding a double-clicked handler here :+1:

src/Components/VimTree/Component_VimTree.re
```diff
@@ -271,9 +272,28 @@ let update = (msg, model) => {
...
+          expanded
+            ? model.expansionContext
+              |> ExpansionContext.collapse(data.uniqueId)
+            : model.expansionContext |> ExpansionContext.expand(data.uniqueId);
+
+        (
+          updateTreeList(model.trees, expansionContext, model),
+          expanded ? Collapsed(data.inner) : Expanded(data.inner),
+        );
+
```
bryphe, line 290: Is there a reason we can't use the `toggleExpanded` function here? Seems like it would remove some duplication

src/Feature/LanguageSupport/Common.re
```diff
@@ -5,4 +5,8 @@ type outmsg =
   | OpenFile({
       filePath: string,
       location: option(Location.t),
+    })
+  | PreviewFile({
+      filePath: string,
+      location: option(Location.t),
```
bryphe, line 11: It looks like this isn't actually used by any of the LanguageSupport features - seems like we could simplify and remove this in `Feature_LanguageSupport` (and remove the handler in `Features.re`).

src/Components/FileExplorer/Component_FileExplorer.re
fanantoxa: I didn't find in a NerdTree any preview mode. So to match API it'll be just `Enter` to select.
To add preview capabilities we'd need to introduce some another keybinding lie `Shift+Enter` that can be mapped to `Touched`

src/Components/VimTree/Component_VimTree.re
fanantoxa: We could if it fit here. I missed that we already have this function.

src/Store/Features.re
```diff
@@ -775,18 +823,54 @@ let update =
...
+            preview,
+            Editor.create(
+              ~config=config(~fileType),
+              ~buffer=editorBuffer,
+              ~preview,
+              (),
+            ),
+          )
+        };
+
```
bryphe, line 873: It seems like this is breaking the split functionality:

broken-split

In particular - note that when I split using `:vsp` - it seems like a common editor is being used. When I scroll, both editors scroll, and likewise when I move the screen. So this looks like a regression.

bryphe:
Thanks for the update here, @fanantoxa ! The high-level approach looks good to me - I like the design you chose.

I do see a regression around splits - mentioned above - it seems like a unique editor is not created anymore. That will need to be fixed prior to merging in.

src/Feature/LanguageSupport/Common.re
fanantoxa: @bryphe Are you sure that we won't need that in future?
For example in vscode `Go To Definition` also opens in `preview` mode.
We could use it in future

src/Components/FileExplorer/Component_FileExplorer.re
bryphe: Ah, right, I think you mentioned this before - thanks for the details!

src/Store/Features.re
fanantoxa: Got you, will check it later.

src/Feature/Search/Feature_Search.re
```diff
@@ -85,10 +85,14 @@ type outmsg =
...
+  | PreviewFile({
+      filePath: string,
+      location: CharacterPosition.t,
+    })
   | Focus
   | UnhandledWindowMovement(Component_VimWindows.outmsg);
 
-let update = (model, msg) => {
+let update = (~previewEnabled, model, msg) => {
   switch (msg) {
```
bryphe, line 96: One thing I'm thinking about - we could remove injecting this parameter, and instead update the `OpenFile` outmsg to something like:
```
OpenFile({
   filePath: string,
   location: CharacterPosition.t,
   preferToPreview: bool,  // Let the consumer know that, if the configuration allows for previewing, we should preview!
})
```

This way - the feature can communicate the intent - that we should try to preview - without needing to inject additional parameters. Then, when the `OpenFile` outmsg is handled in `Features` - we could decide whether it should be an `openFile` effect or `previewFile` effect based on that property and the user's configuration.

src/Feature/LanguageSupport/Common.re
bryphe: I'd rather see it added when it's hooked up - so the usage is clearer. Until we're actually using it, it would just add a maintenance burden to `Features.re` and `Feature_LanguageSupport` (in other words - it's something else we have to refactor!). 

In addition, until it is actually used, we might not have the interface totally right - once we have a use case, that connection will be more crystallized and we can make sure we have it right. We also don't have a way to test this functionality w/o some connection.

So to me, this seems like a case of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), at least until we have a usage. Perhaps a subsequent change could wire up 'Go to Definition' to preview, and add this plumbing as part of it.

src/Store/Features.re
bryphe: Looks like this is fixed now in the latest - thanks @fanantoxa !

docs/docs/configuration/settings.md
bryphe: Saw this was fixed in latest - thanks!

bryphe:
Changes look good to me, @fanantoxa ! Thanks for addressing the feedback and fixing the `:vsp` bug. Seems solid!

I just pushed up a few minor tweaks:
- Add a CHANGES entry
- Add a simple manual test case (mainly a reminder for me to make sure I don't break this)
- Merged `master`

I'm ready to merge when this latest build is green 💪

bryphe:
Woohoo, green! 🎉 🎉 

Thanks for all your work on this, @fanantoxa !

PR #2598, opened 2020-10-19
feat(diagnostics): Runtime trace logging

bryphe:
This adds a switch during runtime to turn on trace logging to a file:

- Accessible via the `Command+Shift+P`/`Control+Shift+P` command palette - `Debug: Log to trace file`

When this option is selected, an `onivim-trace.log` file will be created in the current working directory, and the trace log level will be enabled. 

The use case for this is to acquire more details around issues like https://github.com/onivim/oni2/issues/2407 - cases where the editor may not have been started with logging, but having the log would be helpful in diagnosing.


CrossR:
I see this is specifically for `Trace` logging, is there any benefit to adding a similar command for debug logging, for the cases where we don't need the extra info?

Just thinking that may be an easier way to recommend enabling logging for some issues, but right now that could be a pain for us if there is a LOT more info than we need for just LSP issues say.

bryphe:
Sure, we could definitely add options for other sorts of logging - just don't want it to be too confusing - the nice thing about trace logging is it will have the most information.

Some ideas:
1) We could add options for other log levels
2) We could add an option to turn-off trace logging, so that once the issue is reproduced, it could be stopped.

I'm open to either/or - the only case I worry about is if a user goes to the trouble get logs, but it doesn't have quite enough info for us to diagnose.

There could be some other ways we could bubble this up too... I'd like to have an 'Output' pane at the bottom, that could show output from both Vim commands (ie `:!ls`) as well as from the extension host (there's a whole [Output Channel](https://code.visualstudio.com/api/references/vscode-api#OutputChannel) API that we're missing out on, at the moment)

PR #2609, opened 2020-10-21
fix(terminal/#1548): use login shell for macOS

### LINKED ISSUE #1161: $PATH not set correctly when onivim not launched from command line
Version 0.3.0 (macos catalina)

When I load up a .re file in a bucklescript project I get a 'Could not run bsb" error:

image

Reason completion and laguage server features don't work because of this.

### LINKED ISSUE #1548: Terminal doesn't include `/usr/local/bin` in PATH
Oni2's terminal does not include `/usr/local/bin` in PATH by default. This means on OSX, many programs, including software installed via brew is not accessible.

I could workaround this by simply adding it to my .zshrc, but that shouldn't be necessary as `/usr/local/bin` is automatically added to PATH by other terminals on OSX.

CrossR:
Tiny change to bring in a modification that I've been running locally for a bit. (Seeing the other terminal PR made me actually spend 2 mins and look it up!).

Apparently, shells on macOS should be login shells by default, at least according to the stackoverflow answer I've linked. VSCode also defaults to `["-l"]` on macOS, and links the to the same answer in their code, so I would say it should be fine.

Fixes #1548
Fixes #1161

### DIFF
--- a/docs/docs/using-onivim/integrated-terminal.md
+++ b/docs/docs/using-onivim/integrated-terminal.md
@@ -71,7 +71,7 @@ The terminal font can be configured independently from the editor font, via the 
 
 - `terminal.integrated.fontFamily` __(_string_)__ - The font family used by the editor surface. This must be a monospace font. The font may be specified by either the name of the font, or an absolute path to the font file.
 - `terminal.integrated.fontSize` __(_int_ default: `12`)__ - The font size used by the editor surface.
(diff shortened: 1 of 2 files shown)


bryphe (approved):
Sweet! Nice catch @CrossR - thanks for the research and the stackoverflow link. Looks great 👍

PR #2618, opened 2020-10-23
docs(infrastructure/esy#1099) fix docs to describe reality

AkechiShiro:
- Mention issue [esy/esy#1099](https://github.com/esy/esy/issues/1099) for Linux users.
- Update root binary folder created by `esy '@release' create`.
- Explain where the binary of Oni2 is actually generated, current documentation is false since no `_release` folder is created in the root folder of the Oni2 repo.
- I ran into an issue when running `esy '@release' create`, it was saying it failed since `esy '@release' install` had not been run prior to running it, so I added this line in the documentation.

Hope theses changes will help people to build Oni2 more easily.

### DIFF
--- a/docs/docs/for-developers/building.md
+++ b/docs/docs/for-developers/building.md
@@ -11,7 +11,7 @@ sidebar_label: Building from Source
 - Install [Git](https://git-scm.com/)
 - Install [Node](https://nodejs.org/en)
 - Install [Esy](https://esy.sh) (__0.6.2__ or above is required, but the latest version is recommened: `npm install -g esy@latest`)
-> __NOTE:__ **Linux-only**: if you need to install using `sudo npm -g esy@latest` then you might run into this issue [esy/esy#1099](https://github.com/esy/esy/issues/1099), the actual workaround is the following :`sudo npm install -g --unsafe-perm esy`.
+> __NOTE:__ **Linux-only**: if you need to install using `sudo npm -g esy@latest` then your NPM installation **might be broken** follow [the instruction here to fix it](https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally) this is related to this issue [esy/esy#1099.](https://github.com/esy/esy/issues/1099)
 
 - __Windows-only__: Run `npm install -g windows-build-tools` (this installs some build tools that aren't included by default on Windows)
 - Install any other system packages required by Oni2 dependencies, as outlined below.
--- a/docs/docs/for-developers/building.md
+++ b/docs/docs/for-developers/building.md
@@ -10,7 +10,9 @@ sidebar_label: Building from Source
 
 - Install [Git](https://git-scm.com/)
 - Install [Node](https://nodejs.org/en)
-- Install [Esy](https://esy.sh) (__0.6.2__ or above is required, but the latest version is recommended: `npm install -g esy@latest`)
+- Install [Esy](https://esy.sh) (__0.6.2__ or above is required, but the latest version is recommened: `npm install -g esy@latest`)
+> __NOTE:__ **Linux-only**: if you need to install using `sudo npm -g esy@latest` then you might run into this issue [esy/esy#1099](https://github.com/esy/esy/issues/1099), the actual workaround is the following :`sudo npm install -g --unsafe-perm esy`.
+
 - __Windows-only__: Run `npm install -g windows-build-tools` (this installs some build tools that aren't included by default on Windows)
 - Install any other system packages required by Oni2 dependencies, as outlined below.
 
@@ -109,9 +111,12 @@ We use auto formatting tool, you might want to run it before you commit changes
 To create a release build, run:
 
 - `esy '@release' run -f --checkhealth`
+- `esy '@release' install`
 - `esy '@release' create`
 
-This will create a `_release` folder at the root with the application bundle inside.
+This will create a `_esy/release` folder at the root with the application bundle inside that folder there will be folders for built binaries, in `_esy/release/install/bin` the `Oni2` binary resides along `Oni2_editor`.
+
+Mind that these are actually symbolic links to `oni2/_esy/release/store/b/oni2-<hashvalue>/install/default/bin/Oni2`, the same is true for the `Oni2_editor` binary.
 
 ### Windows
 


docs/docs/for-developers/building.md
```diff
@@ -11,6 +11,8 @@ sidebar_label: Building from Source
 - Install [Git](https://git-scm.com/)
 - Install [Node](https://nodejs.org/en)
 - Install [Esy](https://esy.sh) (__0.6.2__ or above is required, but the latest version is recommened: `npm install -g esy@latest`)
+> __NOTE:__ **Linux-only**: if you need to install using `sudo npm -g esy@latest` then you might run into this issue [esy/esy#1099](https://github.com/esy/esy/issues/1099), the actual workaround is the following :`sudo npm install -g --unsafe-perm esy`.
+
```
glennsl, line 15: If you have to use `sudo` then your npm installation is broken<sup>1</sup>. The proper solution is to fix your installation according to [the official documentation](https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally) or [this tep-by-step guide](https://github.com/sindresorhus/guides/blob/master/npm-global-without-sudo.md), not to invite the entire internet to a party with full access to your computer. The flag is called "unsafe" for good reason.

<sup>1</sup> There are historical reasons why this is still so prevalent: NPM wasn't originally designed for installing packages globally, and most Linux package managers don't like other package managers meddling in their business.

AkechiShiro: I see I understand, the thing is the unsafe flag being used here is only needed for the `esy` package other packages installations works fine without that additionnal flag but since my NPM install is kinda broken  I guess these changes are useless.

Sorry for having made this PR I should've known better. I understand that this is related to security concerns.

glennsl: No need to apologize. There's unfortunately a lot of material on this that's pointing in the wrong direction, and not a lot pointing in the right direction.

The problem is the use of `sudo` in the first place, not so much the use of `--unsafe-perm`. As I understand it, `--unsafe-perm` was added to try to mitigate some of the harm of using `sudo` by trying to install as the user invoking `npm` instead of the root user. Unfortunately that doesn't always work, since the boundaries are so fuzzy, and completely unenforced, some `postinstall` go outside them. Therefore, to enable the legitimate use case of installing npm packages globally for all users on a machine (as opposed to globally for _one_ user, which is usually the case), this flag was added. It's a real mess, for historical reasons as I said.

In any case, it's not my place to say what's acceptable or not around here, but I would at least avoid offering insecure advice directly. I think it's OK to link to the issue on the `esy` repo though, and if the `esy` developers want to offer insecure advice then that's on them. I've warned them about it before, and I do think Manas offers good advice there, though it's less convenient than the bad advice in the short term and what I'm advising here in the long-term, without offering much explanation of why it's worth it.

AkechiShiro: I've updated the documentation note with a more accurate description of what should be done (I suppose).

Can you confirm to me that this looks good to you as of now ?

glennsl: Nitpicking here, but it's a bit wordy. I'd cut the last part after pointing to the esy issue. The recommendation given earlier should be sufficient. Also, technically it might not be broken. As I said earlier there are legitimate uses cases for installing packages globally for all users, where `sudo` and `--unsafe-perm` are needed, but then the admin certainly ought to know what they're doing. For single-user systems I'd consider it broken.

AkechiShiro: It's fine to nitpick, no problem I understand that they are some grey cases where it may be okay to allow someone to do this when they know what they are doing, I updated the documentation following your comment and also said that the NPM install might be broken since this probably doesn't always hold true I guess.

glennsl (approved):
:+1: LGTM!

AkechiShiro:
I've just solved the conflict (done a rebase based on upstream/master I think what's left is to merge this PR).

AkechiShiro:
Is there any reason why the CI has failed ? Maybe the current branch doesn't build successfully I'm not too sure, how do I go about relaunching the Azure DevOps to try and verify if the tests passes fully ??

docs/docs/for-developers/building.md
```diff
@@ -11,6 +11,8 @@ sidebar_label: Building from Source
 - Install [Git](https://git-scm.com/)
 - Install [Node](https://nodejs.org/en)
 - Install [Esy](https://esy.sh) (__0.6.2__ or above is required, but the latest version is recommened: `npm install -g esy@latest`)
+> __NOTE:__ **Linux-only**: if you need to install using `sudo npm -g esy@latest` then your NPM installation **might be broken** follow [the instruction here to fix it](https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally) this is related to this issue [esy/esy#1099.](https://github.com/esy/esy/issues/1099)
```
bryphe, line 14: Nice tip! I didn't know about this

bryphe:
Thanks @AkechiShiro for the updates - great improvements! And thanks for the review @glennsl 

> Is there any reason why the CI has failed ? Maybe the current branch doesn't build successfully I'm not too sure, how do I go about relaunching the Azure DevOps to try and verify if the tests passes fully ??

We unfortunately have some flaky tests that need to be fixed - failure doesn't look related to the changes. I do see some extra changes in the diff though - it'd be helpful to rebase so that the change is just scoped to the documentation changes. LMK if you want help with this. Otherwise, doc changes look good to me :+1:

src/Exthost/ShellLaunchConfig.re
```diff
@@ -1,14 +1,40 @@
+open Oni_Core;
+
+type environment =
+  | Inherit
+  | Additive(StringMap.t(string))
+  | Strict(StringMap.t(string));
+
```
bryphe, line 7: These parts of the diff don't look related to the documentation change - I believe it's just a rebase issue. Rebasing to the latest master commit should remove these from the diff view here.

AkechiShiro: Yep you're right, no problem will do a rebase soon and let you know when it's done.
Thanks for the review !

bryphe: Awesome, thank you @AkechiShiro!

AkechiShiro: Rebase done, I think you can safely merge this PR now.

AkechiShiro: Actually @Dark-Matter7232 which has some latest commits merged in master did not sign the CLA it seems so until he signs it I'm not sure merging is possible.
I'll let you handle that, if another rebase is needed in the future I will do them as soon as I can.

bryphe: Thanks @AkechiShiro for your help! Looks like those extra commits were being picked up inadvertently - I revised the history a bit. Bringing it in now :+1:

PR #2623, opened 2020-10-23
docs(core/#2611): improve build docs and dependencies 

### LINKED ISSUE #2611: Building on Fedora 32
missing dependencies fontconfig-devel and libasan

```
/usr/bin/ld: cannot find /usr/lib64/libasan.so.6.0.0
    collect2: error: ld returned 1 exit status
    File "caml_startup", line 1:
    Error: Error during linking
```
```
checking for tgetent()... configure: error: NOT FOUND!
          You need to install a terminal library; for example ncurses.
          Or specify the name of the library with --with-tlib.
    Makefile:302: auto/config.mk: No such file or directory
    make: *** [Makefile:1672: auto/config.mk] Error 1
    error: command failed: 'bash' '-c' 'build/build-posix.sh' (exited with 2)
    
  building libvim@8.10869.65
```
Package ncurses-6.1-15.20191109.fc32.x86_64 is already installed.

CrossR:
Small change to include all the builds deps for Fedora 32. I've updated the Revery deps already, since that is just in the Wiki there and doesn't need a PR.

Also just updated to explain what the 3 build steps do, a common question I see is "The bootstrapping is taking ages! The build is going to take even longer..." when really the build step is a few seconds since 99% of the work happens in the bootstrap bit. (We can't swap the steps around as eventually you hit an error as the bootstrap files are needed and getting and error during the build would cause even more questions).

Fixes #2611.

### DIFF
--- a/docs/docs/configuration/extensions.md
+++ b/docs/docs/configuration/extensions.md
@@ -49,7 +49,7 @@ If your favorite extension is missing - you can help us out by publishing it to 
 1) Register for an account using the [Open VSX GitHub OAuth](https://open-vsx.org/oauth2/authorization/github) provider
 2) Create a [personal access token](https://open-vsx.org/user-settings/tokens) 
 3) Install the `ovsx` tool - `npm install -g ovsx`
-4) Create a namespace corrresponding to your extension: `ovsx create-namespace <publisher> --pat <token>`
+4) Create a namespace corresponding to your extension: `ovsx create-namespace <publisher> --pat <token>`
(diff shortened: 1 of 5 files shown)


CrossR:
(Made a typo in my commit....Probably shouldn't edit in the GitHub UI. Took that as an opportunity to do a speedy spellcheck pass over the doc pages).

bryphe:
Awesome, thanks @CrossR !

bryphe (approved):

Issue #2624, opened 2020-10-24
command tab-complete in non-first path argument deletes all text past first path argument and 'tab-completes' the first argument


MichaelBelousov:
command tab-complete in non-first path argument deletes all text past first path argument and 'tab-completes' the first argument.

Try for instance the command: `:!cp /some/path /another/incomplete/pat`
then without entering the command, press tab to tab complete the second path.
the tab completion for the second path will be applied to the first path, and all text after the first path obliterated.
this means tab completion can mess up multi-path argument commands and make you have to retype them and remember
to not accidentally try completing.

MichaelBelousov:
Put up a PR fixing it with a one character change :smile: + a test

bryphe:
Looks like this is fixed now in #2630 - thanks @MichaelBelousov !

PR #2630, opened 2020-10-26
fix(): fix tab completion in quick menu for multiple argument commands

MichaelBelousov:
- fixed issue #2624 
- fixed by searching from the back of the input instead of front
- added an accompanying `let%test`. I'm 100% new to ReasonML so please let me know if I missed something which I likely did. I don't precisely know what `let%test` does

### DIFF
--- a/src/Feature/Vim/Feature_Vim.re
+++ b/src/Feature/Vim/Feature_Vim.re
@@ -63,14 +63,18 @@ let update = (msg, model: model) => {
 module CommandLine = {
   let getCompletionMeet = commandLine => {
     let len = String.length(commandLine);
-
     if (len == 0) {
       None;
     } else {
-      String.rindex_opt(commandLine, ' ')
-      |> Option.map(idx => idx + 1)  // Advance past space
-      |> OptionEx.or_(Some(0));
-    };
+      let last_unescaped_space_index = ref(Some(0)); // default result
+      String.iteri(
+        (i, c) => if (i > 0 && commandLine.[i-1] != '\\' && c == ' ') {
+          last_unescaped_space_index := Some(i+1); // Advance past space
+        },
+        commandLine
+      );
+      last_unescaped_space_index^;
+    }
   };
 
   let%test "empty command line returns None" = {
--- a/src/Feature/Vim/Feature_Vim.re
+++ b/src/Feature/Vim/Feature_Vim.re
@@ -81,11 +81,15 @@ module CommandLine = {
   };
 
   let%test "meet with a path, spaces" = {
-    getCompletionMeet("vsp /path with spaces/") == Some(4);
+    getCompletionMeet("vsp /path\\ with\\ spaces/") == Some(4);
   };
 
   let%test "meet multiple paths" = {
     getCompletionMeet("!cp /path1 /path2") == Some(11);
+  };
+
+  let%test "meet multiple paths with spaces" = {
+    getCompletionMeet("!cp /path\\ 1 /path\\ 2") == Some(11);
   };
 };
 
--- a/src/Feature/Vim/Feature_Vim.re
+++ b/src/Feature/Vim/Feature_Vim.re
@@ -61,14 +61,8 @@ let update = (msg, model: model) => {
 };
 
 module CommandLine = {
-  let getCompletionMeet = commandLine => {
-    let len = String.length(commandLine);
-    if (len == 0) {
-      None;
-    } else {
-      StringEx.findUnescapedFromEnd(commandLine, ' ');
-    };
-  };
+  let getCompletionMeet = commandLine =>
+    StringEx.findUnescapedFromEnd(commandLine, ' ');
 
   let%test "empty command line returns None" = {
     getCompletionMeet("") == None;
--- a/src/Feature/Vim/Feature_Vim.re
+++ b/src/Feature/Vim/Feature_Vim.re
@@ -67,7 +67,7 @@ module CommandLine = {
     if (len == 0) {
       None;
     } else {
-      String.index_opt(commandLine, ' ')
+      String.rindex_opt(commandLine, ' ')
       |> Option.map(idx => idx + 1)  // Advance past space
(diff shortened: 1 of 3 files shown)


MichaelBelousov:
I realized this certainly breaks space-including path support, I'll fix it tomorrow

MichaelBelousov:
So vim does this by escaping spaces with backslashes, so I'll just check the behavior in oni2 and check for that

bryphe:
Thanks for the PR, @MichaelBelousov !

> added an accompanying let%test. I'm 100% new to ReasonML so please let me know if I missed something which I likely did. I don't precisely know what let%test does

Awesome - having a test case to document this behavior is perfect. The `let%test` syntax comes from this library: [`ppx_inline_test`](https://github.com/janestreet/ppx_inline_test). They're nice because they are inline with the source code but don't get compiled in the built app. You can run them with `esy '@test' inline`.

bryphe:
> So vim does this by escaping spaces with backslashes, so I'll just check the behavior in oni2 and replicate that

Ya, I think we'll need this behavior too - otherwise we can't distinguish between a space-in-a-path vs a space-between-arguments.

src/Feature/Vim/Feature_Vim.re
```diff
@@ -92,6 +92,10 @@ module CommandLine = {
   let%test "meet with a path, spaces" = {
     getCompletionMeet("vsp /path with spaces/") == Some(4);
   };
```
bryphe, line 94: Looking at CI - it seems like this test case is failing now. Must be picking up the wrong space - I think your proposal of escaping spaces, and then reversing the logic to find the first non-escaped space, should get both this test and your new one green.

MichaelBelousov: yup actually just started working on it, should have a something soon once I figure out how to do stuff resonably in ocaml

bryphe: Cool! We have some utilities that here might be helpful / give some ideas: https://github.com/onivim/oni2/blob/master/src/Core/Utility/StringEx.re

Also the OCaml [`String`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/String.html) and [`Str`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/String.html) might be useful. 

> should have a something soon once I figure out how to do stuff reasonably in ocaml

Seems like you're picking it up fast :smile:

MichaelBelousov:
so I added a commit with a potentially OK way of doing it. I'll yield to anyone with more OCaml/Reason experience than me. I've done fully functional stuff like this in `elm` and a *tiny* bit of Haskell but it's been a bit and I'm not used to the OCaml libraries. A better way to do it might right_fold or reverse iterate and stop on the first unescaped space that is found, but I couldn't find an easy way to do that with String, maybe I should add to `StringEx`.

I also noticed the following issues while testing the result:
- spaces are not escaped when tab completion occurs (so hitting tab will yield a path like `'/my path'` when it should be `'/my\ path'`. I will try to include a fix for that in this PR unless you consider it out of scope. I can't test without it really anyway
- I think when doing `:!mkdir 'Test Me', the quotes were removed and separate directories `Test` and `Me` were created. I'll file a separate issue if this is the case.

MichaelBelousov:
alright @bryphe I finished the functionality
- tab-completed paths now have spaces escaped
- multiple path arguments with spaces are now supported"
I would expect maybe you might want me to change some of the code or move some of the `StringEx` utils to `Path` utils, just let me know, I can do so. I personally thought the string manipulation library functions were a bit lacking so I think it could be done better if some of those utilities were written into StringEx.

src/Feature/Vim/Feature_Vim.re
```diff
@@ -90,7 +90,15 @@ module CommandLine = {
...
-    getCompletionMeet("vsp /path with spaces/") == Some(4);
+    getCompletionMeet("vsp /path\\ with\\ spaces/") == Some(4);
+  };
+
+  let%test "meet multiple paths" = {
+    getCompletionMeet("!cp /path1 /path2") == Some(11);
+  };
+
+  let%test "meet multiple paths with spaces" = {
+    getCompletionMeet("!cp /path\\ 1 /path\\ 2") == Some(13);
```
bryphe, line 101: Thanks for adding this!

src/Feature/Vim/Feature_Vim.re
```diff
@@ -61,17 +61,17 @@ let update = (msg, model: model) => {
...
     } else {
-      String.index_opt(commandLine, ' ')
-      |> Option.map(idx => idx + 1)  // Advance past space
-      |> OptionEx.or_(Some(0));
+      let meet = StringEx.findUnescapedFromEnd(commandLine, ' ');
+      if (meet == None) {
+        Some(0);
+      } else {
+        meet;
+      };
```
bryphe, line 73: This conditional could be streamlined a bit:
```
let maybeMeet = StringEx.findUnescapedFromEnd(commandLine, ' ')
|> OptionEx.or_(Some(0));
```

bryphe:
> alright @bryphe I finished the functionality

Excellent! Thanks for adding the test cases and new utilities in `StringEx`. Looks good overall to me :+1:

bryphe (approved):

src/Feature/Vim/Feature_Vim.re
MichaelBelousov: I was looking for Option.or, should have looked for OptionEx, will fix tonight

MichaelBelousov: done, resolving

MichaelBelousov:
the CLA wasn't working until I switched my email address, but that seems to be done now

bryphe:
Excellent, thanks for incorporating the feedback @MichaelBelousov and signing the CLA. Impressive how quickly you got up-to-speed on ReasonML/OCaml!

I've e-mailed you a code for a license key - let me know if you have any issues :+1:

PR #2665, opened 2020-11-03
docs: remove experimental.editor.smoothScroll

glennsl:
This seems to have been forgotten when the setting was promoted to default in #2237.


bryphe:
Nice catch, thanks @glennsl !

PR #2667, opened 2020-11-03
fix(packaging): esy @release create fails on linux

glennsl:
This reverts ad136eef1867e14e67a2a0e0c31a3945d02d83f1/#2254 

The workaround is broken since the server hosting the binary seems to be down entirely, and the issue seems to have been fixed.


bryphe:
Awesome, thanks @glennsl !

PR #2687, opened 2020-11-09
feat(input): implement a Reason alternative to atom/keyboard-layout

zbaylin:
@bryphe mentioned to me the other day that he would appreciate a ReasonML fork of the atom/keyboard-layout JS library.

At first I thought I could just bind directly to the C++ the Atom folks wrote, but their code is fairly heavily integrated with the Node FFI, so I think it makes more sense to take inspiration from it and write a separate version here.

As of the first commit, there is an API to subscribe to keyboard layout changes that works on macOS. It works something like this:

```reason
Oni2_KeyboardLayout.subscribe((~layout, ~language) => printf("Layout: %s, Language: %s\n", layout, language));
```

I will implement more features of the JS lib as I go along!


test/oni2-keyboard-layout/KeymapTest.re
```diff
@@ -0,0 +1,22 @@
...
+    let entry = Keymap.entryOfKey(keymap, "KeyS");
+
+    expect.option(entry).toBeSome();
+
+    let entry = Option.get(entry);
+
+    expect.string(Keymap.(entry.unmodified)).toEqual("s");
+    expect.string(Keymap.(entry.withShift)).toEqual("S");
+  });
+});
```
zbaylin, line 22: @bryphe I got the keymaps working on macOS, which I think is the feature you most wanted. This test will need to be improved since it obviously will only work when the S key is mapped to S, but for now it works as a small test

bryphe:
This is great @zbaylin - thanks for taking this on!

> At first I thought I could just bind directly to the C++ the Atom folks wrote, but their code is fairly heavily integrated with the Node FFI, so I think it makes more sense to take inspiration from it and write a separate version here.

I agree - it's very coupled with the Node FFI. Extracting the core pieces and re-wrapping them makes sense.

src/oni2-keyboard-layout/config/discover.re
```diff
@@ -0,0 +1,42 @@
+module Configurator = Configurator.V1;
+
```
bryphe, line 2: Thanks for setting up all the plumbing for this :+1:

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
```diff
@@ -0,0 +1,243 @@
...
+      NSLog(@"Unable to acquire callback list reference!");
+      return;
+    }
+  }
+
+  NSString *nsLayout = oni2_priv_GetCurrentKeyboardLayout();
+  NSString *nsLanguage = oni2_priv_GetCurrentKeyboardLanguage();
+
+  value vLayout = caml_copy_string([nsLayout UTF8String]);
+  value vLanguage = caml_copy_string([nsLanguage UTF8String]);
```
bryphe, line 65: I believe these will need to be wrapped with `CAMLlocal` - I wonder how this will work with the `caml_acquire_runtime_system` below?

bryphe: We might have to change the flow a bit here so that these are properly tracked by the GC, to avoid crashes - we could have a function get the current layout, and current language, and have the notification be a simple callback that just basically says "Go ask for the latest layout and language"

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  if (status == noErr && !iscntrl(characters[0])) {
+    return charCount;
+  } else {
+    return 0;
+  }
+}
+
+CAMLprim value oni2_KeyboardLayoutPopulateCurrentKeymap(value keymap, value Hashtbl_replace) {
```
bryphe, line 165: Clever - so the consumer passes a function to collect the keys and mappings?

src/oni2-keyboard-layout/stubs/keyboard-layout.c
```diff
@@ -0,0 +1,23 @@
...
+  Simple helper method to construct a keymap
+  entry from strings
+ */
+value createKeymapEntry(
+  const char *unmodified,
+  const char *withShift,
+  const char *withAltGraph,
+  const char *withAltGraphShift
+) {
+  value keymapEntry = caml_alloc(4, 0);
```
bryphe, line 17: I believe this will need to be wrapped with `CAMLparam0()`, `CAMLlocal1(keymapEntry)`, and `CAMLreturn(keymapEntry)`

test/oni2-keyboard-layout/KeymapTest.re
bryphe: Awesome! Are you planning on looking at the Linux / Windows bindings too? Or should I check those out?

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
zbaylin: Yep! I was trying to think of the best way to modify an OCaml `Hashtbl.t` in C and that was the idea that I came up with -- the consumer creates a `Hashtbl`, passes it into our `populate` function alongside the `Hashtbl.replace` OCaml callback so that we can modify the table in place from C -- it seemed like a good idea to me, but let me know if you think there's anything wrong with it!

test/oni2-keyboard-layout/KeymapTest.re
zbaylin: I'll look at those as well -- macOS is always the easiest first step for me since it's what my host is, but I will definitely look into Windows and Linux as well!

src/oni2-keyboard-layout/stubs/keyboard-layout.c
zbaylin: Fixed in 0641b93!

src/oni2-keyboard-layout/stubs/keyboard-layout.c
```diff
@@ -0,0 +1,25 @@
...
+value createKeymapEntry(
+  const char *unmodified,
+  const char *withShift,
+  const char *withAltGraph,
+  const char *withAltGraphShift
+) {
+  CAMLparam0();
+  CAMLlocal1(keymapEntry);
+  keymapEntry = caml_alloc(4, 0);
+  Store_field(keymapEntry, 0, caml_copy_string(unmodified));
```
bryphe, line 20: Are there cases where these strings can be `NULL`? I'm wondering if we need to make these optional.

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
bryphe: I think the idea is great! My only thought is that it seems more general than a `Hashtbl.t` - could potentially use other values, too. I'll add a comment on the signature with more details.

src/oni2-keyboard-layout/Oni2_KeyboardLayout.re
```diff
@@ -0,0 +1,45 @@
...
+    unmodified: string,
+    withShift: string,
+    withAltGraph: string,
+    withAltGraphShift: string,
+  };
+
+  type t = Hashtbl.t(string, entry);
+
+  external populateCurrentKeymap: (t, (t, string, entry) => unit) => unit =
+    "oni2_KeyboardLayoutPopulateCurrentKeymap";
```
bryphe, line 33: It seems like this is more general than `Hashtbl.t` - it could really work for any data structure + a function to 'accumulate' values, like:

```
external populateCurrentKeyMap: ('keymap, ('keymap, string, entry) => unit) => unit
```

Where the `'keymap` type is polymorphic. Just means it is a bit more loosely coupled. The `getCurrent` below could remain unchanged, though.

src/oni2-keyboard-layout/stubs/keyboard-layout.c
zbaylin: The strings can be null, but the string would have length 0 -- we can make empty strings `None` and non-empty strings `Some(key)`!

src/oni2-keyboard-layout/Oni2_KeyboardLayout.re
zbaylin: That's a great idea! Added in 216398e!

src/oni2-keyboard-layout/stubs/keyboard-layout.c
zbaylin: I ended up doing this in e531d4c:
https://github.com/onivim/oni2/blob/e531d4c8d4da96bed33a919569831673aa347492/src/oni2-keyboard-layout/stubs/keyboard-layout.c#L17-L26

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
zbaylin: Changed in e8e0b09!

test/oni2-keyboard-layout/KeymapTest.re
bryphe: Cool! Looks like CI is failing due to missing symbols on Windows / Linux - I guess we'll need at least stubs for this to get green

src/oni2-keyboard-layout/Oni2_KeyboardLayout.re
```diff
@@ -0,0 +1,45 @@
...
+};
+
+external init: unit => unit = "oni2_KeyboardLayoutInit";
+
+external getCurrentLayout: unit => string =
+  "oni2_KeyboardLayoutGetCurrentLayout";
+external getCurrentLanguage: unit => string =
+  "oni2_KeyboardLayoutGetCurrentLanguage";
+
+Callback.register("oni2_KeyboardLayoutCallbackListRef", callbackListRef);
```
bryphe, line 19: Here, the C binding could be made simpler if we have a wrapper function like:

```
let callAllKeyboardCallbacks = () => {
 callbackListRef^ |> List.iter(f => f());
};
Callback.register("oni2_KeyboardLayoutCallbackListRef", callAllKeyboardCallbacks);
```
Means less juggling to do in the C bindings, which is nice

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
```diff
@@ -0,0 +1,240 @@
...
+  // We let OCaml manage the list since they have better abstractions than we do
+  static const value *callbackListRef = NULL;
+  // If we haven't gotten the reference yet, get it.
+  if (callbackListRef == NULL) {
+    callbackListRef = caml_named_value("oni2_KeyboardLayoutCallbackListRef");
+    // If we didn't get anything back from the OCaml runtime, return early.
+    if (callbackListRef == NULL) {
+      NSLog(@"Unable to acquire callback list reference!");
+      return;
+    }
```
bryphe, line 60: This `return` will actually cause problems, since we would need a `CAMLreturn` to work properly with the garbage collector. But if we streamline the callback strategy as mentioned in an earlier comment - we shouldn't even need this logic here.

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
```diff
@@ -0,0 +1,240 @@
...
+  caml_acquire_runtime_system();
+
+  // Loop through the callback list and call all of them with the langusage and layout
+  while (callbackList != Val_emptylist) {
+    value callback = Field(callbackList, 0);
+    caml_callbackN(callback, 1, args);
+    callbackList = Field(callbackList, 1);
+  }
+
+  caml_release_runtime_system();
```
bryphe, line 77: We'll need a `CAMLreturn0` here as well, to match with the `CAMLparam` and `CAMLlocal`

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
```diff
@@ -0,0 +1,240 @@
...
+
+  if (layoutData == NULL) {
+    CAMLreturn(Val_unit);
+  }
+
+  const UCKeyboardLayout *keyboardLayout = (UCKeyboardLayout *)CFDataGetBytePtr(layoutData);
+
+  size_t keyCodeMapSize = sizeof(keyCodeMap) / sizeof(keyCodeMap[0]);
+  for (size_t i = 0; i < keyCodeMapSize; i++) {
+    const char *dom3Code = keyCodeMap[i].dom3Code;
```
bryphe, line 189: One thing I need to explore is how to map the SDL keys to the keymap here - I'm not sure there actually are values corresponding to the `dom3Code`.

So I'll need to either:
1) Have a function that maps from SDL key <-> dom3code
2) Use another value available on the key (either keycode or scancode) to map to these entries

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
zbaylin: Fixed in dc53ea4!

src/oni2-keyboard-layout/Oni2_KeyboardLayout.re
zbaylin: That's a much better idea! I changed it to this in dc53ea4. Let me know if that's better!

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
zbaylin: I might be mistaken, but I think I still need to check if the pointer returned by `caml_named_value` is null, or else I would get a segfault when I de-reference it in `caml_callbackN`. Please let me know if I'm wrong!

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
zbaylin: That's a good point. I looked into this and I couldn't find if there's a way to convert native scancodes/keycodes to SDL2 ones. The internal representation of SDL's scancodes/keycodes seem to be decoupled from the raw USB codes that `keyboard-layout` uses. This means we would have to write a mapping either way -- either `SDL2 Keycode => DOM3 Keycode` or `SDL2 Keycode => Native Keycode`. I think the first option is easier since we wouldn't have to write a mapping for each OS since they all have different keycodes:
https://github.com/onivim/oni2/blob/dc53ea4a3075d7ea267b0ec509d4c2581be4bed8/src/oni2-keyboard-layout/stubs/keycode_converter_data.h#L125-L133

zbaylin: I actually realized we could add the mapping in that header file, so in 384d465 I painstakingly added all the SDL2 scancodes and keycodes to the mapping. That should let us construct a `Hashtbl.t` that maps `SDL2 Scancode => Keymap Entry`

zbaylin: > That should let us construct a `Hashtbl.t` that maps `SDL2 Scancode => Keymap Entry`

I did this in 0787940 -- let me know what you think! The tests now use the new API:
https://github.com/onivim/oni2/blob/0787940af366828aea4c381f9180454debb0e46a/test/oni2-keyboard-layout/KeymapTest.re#L14-L15

bryphe:
Just tried to build the latest - got this error:
```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: src/oni2-sparkle/libOni2_Sparkle_stubs.a(winsparkle.o) has no symbols
      ocamlc src/oni2-keyboard-layout/keyboard-layout-mac.o (exit 2)
(cd /Users/bryphe/oni2/_esy/default/store/b/oni2-85f9e939/default/src/oni2-keyboard-layout && /Users/bryphe/.esy/3/i/ocaml-4.10.0-f28f0ffe/bin/ocamlc.opt -g -I /Users/bryphe/.esy/3/i/revery-8a0f608b/lib/reason-sdl2 -ccopt -x -ccopt objective-c -ccopt -g -o keyboard-layout-mac.o keyboard-layout-mac.c)
keyboard-layout-mac.c:12:10: fatal error: 'SDL2/SDL.h' file not found
#include <SDL2/SDL.h>
         ^~~~~~~~~~~~
1 error generated.

```

I wonder if we need the `esy-sdl2` include path in the `discover` script for `keyboard-layout`?

src/oni2-keyboard-layout/Oni2_KeyboardLayout.rei
```diff
@@ -0,0 +1,23 @@
...
+    withShift: option(string),
+    withAltGraph: option(string),
+    withAltGraphShift: option(string),
+  };
+
+  type t;
+
+  let getCurrent: unit => t;
+
+  let entryOfScancode: (t, Sdl2.Scancode.t) => option(entry);
```
bryphe, line 20: Awesome! This looks like it'll be easy to map from Sdl2 keys -> keymaps now

bryphe: I just added a usage of the library in `InputStoreConnector` as a proof-of-concept - right now, it just logs the keys, but this is where I'll need to resolve the scancode to improve matching.

src/oni2-keyboard-layout/Oni2_KeyboardLayout.re
```diff
@@ -0,0 +1,63 @@
...
+let callbackListRef: ref(list(callback)) = ref([]);
+
+let subscribe = (callback: callback) => {
+  callbackListRef := [callback, ...callbackListRef^];
+
+  () =>
+    callbackListRef := List.filter(cb => cb !== callback, callbackListRef^);
+};
+
+let callKeyboardCallbacks = () => callbackListRef^ |> List.iter(cb => cb());
```
bryphe, line 12: Thanks for simplifying this!

bryphe:
Latest changes look good to me, @zbaylin - thanks for the work on this!

The only item I see remaining to bring this in is to stub or implement the remaining platforms (Win/Linux)

zbaylin:
@bryphe I just implemented the existing functions on Windows -- let me know what you think!

zbaylin:
@bryphe Linux should now be working too!

bryphe:
Just looking at the build, seems like Linux build is hitting a segfault:
```
[INFO]  [ 0.178s] IntegrationTest : --- Starting test: ExCommandKeybindingTest
[INFO]  [ 0.178s] IntegrationTest :  WAITER (Initial sanity check): Starting
[INFO]  [ 0.178s] IntegrationTest :  WAITER (Initial sanity check): Finished - result: true
[INFO]  [ 0.178s] Oni2.Store.dispatch : dispatch: (KeyDown (
   (KeyPress.PhysicalKey
      { PhysicalKey.scancode = 14; keycode = 107;
        modifiers =
        { Modifiers.control = false; alt = false; altGr = false;
          shift = false; meta = false }
        }),
   <opaque>))
ASAN:DEADLYSIGNAL
=================================================================
==48351==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000968 (pc 0x7fd2a31051f4 bp 0x55ed8a969e68 sp 0x7fff567664d0 T0)
==48351==The signal is caused by a READ memory access.
==48351==Hint: address points to the zero page.
    #0 0x7fd2a31051f3 in _XRefreshKeyboardMapping (/usr/lib/x86_64-linux-gnu/libX11.so.6+0x2a1f3)
    #1 0x7fd2a3164cc5 in XRefreshKeyboardMapping (/usr/lib/x86_64-linux-gnu/libX11.so.6+0x89cc5)
    #2 0x55ed89b1bc49 in oni2_KeyboardLayoutPopulateCurrentKeymap /home/vsts/work/1/s/_esy/integrationtest/store/b/oni2-ba37a797/default/src/oni2-keyboard-layout/keyboard-layout-linux.c:180
    #3 0x55ed895f78cc in camlOni2_KeyboardLayout__getCurrent_603 (/home/vsts/work/1/s/_esy/integrationtest/store/b/oni2-ba37a797/default/integration_test/ExCommandKeybindingTest.exe+0x89d8cc)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/usr/lib/x86_64-linux-gnu/libX11.so.6+0x2a1f3) in _XRefreshKeyboardMapping
```
https://dev.azure.com/onivim/oni2/_build/results?buildId=14382&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=a5f1f371-5fa6-5cca-6ed4-f1b753b00277&l=52988

src/oni2-keyboard-layout/stubs/keyboard-layout-linux.c
```diff
@@ -0,0 +1,222 @@
...
+
+  if (xInputContext) {
+    int count = Xutf8LookupString(xInputContext, xKeyEvent, buffer, 2, NULL, NULL);
+    if (count > 0 && !iswcntrl(buffer[0])) {
+      buffer[count] = '\0';
+    } else {
+      buffer[0] = '\0';
+    }
+  } else {
+    char *cBuffer = (char *)buffer;
```
bryphe, line 160: I wonder why this is needed - can we just use `buffer` here? Or does it need to be copied?

src/oni2-keyboard-layout/stubs/keyboard-layout-linux.c
```diff
@@ -0,0 +1,222 @@
...
+    keyboardBaseState = 0x2000;
+  } else if (xkbState.group == 2) {
+    keyboardBaseState = 0x4000;
+  }
+
+  XEvent xEvent;
+  memset(&xEvent, 0, sizeof(XEvent));
+  XKeyEvent *xKeyEvent = &xEvent.xkey;
+  xKeyEvent->display = xDisplay;
+  xKeyEvent->type = KeyPress;
```
bryphe, line 195: Does this `xKeyEvent` need to be allocated?

src/oni2-keyboard-layout/stubs/keyboard-layout-linux.c
```diff
@@ -0,0 +1,222 @@
...
+    }
+  }
+}
+
+CAMLprim value oni2_KeyboardLayoutPopulateCurrentKeymap(value keymap, value Hashtbl_replace) {
+  CAMLparam2(keymap, Hashtbl_replace);
+  CAMLlocal2(keymapEntry, vSDLScancode);
+
+  // Allocate UTF-8 Buffers to be populated
+  char unmodified[4];
```
bryphe, line 175: Do these get initialized to `0` automatically? It might be worth being on the safe side:
```
char unmodified[4] = { 0 };
```
https://stackoverflow.com/questions/2589749/how-to-initialize-array-to-0-in-c

test.json
```diff
@@ -2,7 +2,7 @@
   "source": "./package.json",
   "scripts": {
     "run": "esy '@test' x OniUnitTestRunner",
-    "run-ci": "esy '@test' x OniUnitTestRunnerCI",
+    "run-ci": "#{os == 'linux' ? 'xvfb-run' : ''} esy '@test' x OniUnitTestRunnerCI",
```
bryphe, line 5: Thanks for figuring this out!

Windows CI is failing with this:
```
========================== Starting Command Output ===========================
"C:\windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\5df81a8d-b1b5-4fb0-b381-158024b29307.cmd""
error: Exec format error
  
```
https://dev.azure.com/onivim/oni2/_build/results?buildId=14382&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=d6c700ea-12df-53fd-bd57-c3cacedc9ab9&l=11


I wonder if it is due to the leading whitespace here?

zbaylin: Ah that would make sense -- I can fix that!

src/oni2-keyboard-layout/stubs/keyboard-layout-linux.c
zbaylin: Oh my bad this is leftover from when I was using `wchar_t` for the buffers. I'll remove that line now.

test.json
bryphe: Maybe we can try: 
```
"run-ci": "#{os == 'linux' ? 'xvfb-run esy' : 'esy'} '@test' x OniUnitTestRunnerCI",
```

src/oni2-keyboard-layout/stubs/keyboard-layout-linux.c
zbaylin: Hmm this is just an alias to the `xEvent` member -- it's not allocating anything on the heap or anything, unless you mean allocating a pointer on the stack. I might be misunderstanding what you mean though

bryphe:
Looks like the CentOS build is failing here:
```
(cd /oni2/_esy/release/store/b/oni2-f540fb70/default/src/oni2-keyboard-layout && /esy/store/3__________________________________________________________________________/i/ocaml-4.10.0-e4568572/bin/ocamlc.opt -g -I /esy/store/3__________________________________________________________________________/i/revery-b078cbca/lib/reason-sdl2 -ccopt -I -ccopt /esy/store/3__________________________________________________________________________/i/esy_sdl2-2.0.10008-c1b32b3d/include -ccopt -g -o keyboard-layout-linux.o keyboard-layout-linux.c)
keyboard-layout-linux.c:9:37: fatal error: X11/extensions/XKBrules.h: No such file or directory
 #include <X11/extensions/XKBrules.h>
                                     ^
compilation terminated.
error: command failed: 'refmterr' 'dune' 'build' '-p' 'libvim,textmate,treesitter,Oni2' '-j4' (exited with 1)
error: build failed with exit code: 1
  
      ocamlc src/oni2-keyboard-layout/keyboard-layout-linux.o (exit 2)
(cd /oni2/_esy/release/store/b/oni2-f540fb70/default/src/oni2-keyboard-layout && /esy/store/3__________________________________________________________________________/i/ocaml-4.10.0-e4568572/bin/ocamlc.opt -g -I /esy/store/3__________________________________________________________________________/i/revery-b078cbca/lib/reason-sdl2 -ccopt -I -ccopt /esy/store/3__________________________________________________________________________/i/esy_sdl2-2.0.10008-c1b32b3d/include -ccopt -g -o keyboard-layout-linux.o keyboard-layout-linux.c)
keyboard-layout-linux.c:9:37: fatal error: X11/extensions/XKBrules.h: No such file or directory
 #include <X11/extensions/XKBrules.h>
                                     ^
compilation terminated.
```
(https://dev.azure.com/onivim/oni2/_build/results?buildId=14402&view=logs&j=e3ae384d-e199-50c7-2cb4-9ce0785e7448&t=575416f9-89fc-5369-7933-ebe25f02131f&l=154)

bryphe:
Looks like we may need this package: https://centos.pkgs.org/7/centos-x86_64/libxkbfile-devel-1.0.9-3.el7.x86_64.rpm.html

I'll push up the change to try it out :crossed_fingers:

zbaylin:
Oh thanks @bryphe -- I was very confused why it was failing with the "file not found" error when everything before it was green. Looks like we got a "false positive" haha

zbaylin:
Looks like this is finally passing CI! @bryphe anything else needed before we merge?

bryphe:
Just trying out the linux version now - it seems like it might be not be picking up my keyboard variant on Manjaro Linux.

I did the following:
- Run `setxkbmap us -variant intl` so I could test out AltGr characters (Alt+C makes ©, for example)
- Entered the © character (AltGr+c)

But the output shows:
```
[INFO]  [148.323s] Oni2.Store.Input : Key info: Unmodified: c
 WithShift: C
 WithAltGr: (none)
 WithAltGrShift: (none)
```

So somehow it isn't recognized the key - I wonder if it knows my layout changed?

zbaylin:
@bryphe it actually looks like the atom team didn't implement the Alt-key variants on Linux, interestingly:
https://github.com/atom/keyboard-layout/blob/a325ddfc9070592c2eafa29b9bd57f2f739ba561/src/keyboard-layout-manager-linux.cc#L136-L140
I wonder why -- I can look into what it would take to get those implemented.

bryphe:
Oh, interesting - didn't realize that - thanks @zbaylin !

Saw some info here: https://github.com/atom/keyboard-layout/pull/18

Given that it's not a bug - I'm OK bringing in what we have, and then we can look to see if we need those. Maybe we can get away without them, too!

bryphe (approved):
Thanks for the awesome work on this! Not easy to figure out how to revamp the FFI from the node approach, or get the automation working on Linux - nicely done :smile: 

Will be a big help for the next round of input work!

zbaylin:
@bryphe sounds good! I'm actually not sure X supports alt-codes in general like macOS and Windows do: https://www.linux.org/threads/alt-keys-and-linux.11517/ -- this info could be outdated, though!

bryphe:
> @bryphe sounds good! I'm actually not sure X supports alt-codes in general like macOS and Windows do: https://www.linux.org/threads/alt-keys-and-linux.11517/ -- this info could be outdated, though!

Interesting, sounds like we'll need SDL2 for those cases still.

> I should also add I couldn't get Alt+C working anywhere in Ubuntu after that setxkbmap command as well

It's actually the [`AltGr`](https://en.wikipedia.org/wiki/AltGr_key) key which is usually the _right_ alt key. So pressing LeftAlt+c won't work on my keyboard - you have to use RightAlt+c (after setting to the international variant).

zbaylin:
@bryphe after doing some testing with the various bitmasks here: https://tronche.com/gui/x/xlib/events/keyboard-pointer/keyboard-pointer.html I was able to determine that (at least on Ubuntu 18.04), `Mod5Mask` corresponds to the `AltGr` key. I've read on StackOverflow that these used to be inconsistent across different distros, so if you could test it out on your Manjaro install, that would definitely help!

Let me know if that fixes it! 🤞

src/oni2-keyboard-layout/stubs/keyboard-layout-linux.c
```diff
@@ -210,12 +211,14 @@ CAMLprim value oni2_KeyboardLayoutPopulateCurrentKeymap(value keymap, value Hash
     if (sdlScancode && xkbKeycode > 0x0000) {
       characterForNativeCode(xInputContext, xKeyEvent, xkbKeycode, keyboardBaseState, unmodified);
       characterForNativeCode(xInputContext, xKeyEvent, xkbKeycode, keyboardBaseState | ShiftMask, withShift);
+      characterForNativeCode(xInputContext, xKeyEvent, xkbKeycode, keyboardBaseState | Mod5Mask, withAltGraph);
+      characterForNativeCode(xInputContext, xKeyEvent, xkbKeycode, keyboardBaseState | Mod5Mask | ShiftMask, withAltGraphShift);
```
bryphe, line 215: Nicely done! This works for me us-intl keyboard :tada:

src/oni2-keyboard-layout/stubs/keyboard-layout-mac.c
```diff
@@ -0,0 +1,233 @@
...
+      // The NSStrings we allocate are simply mechanisms to convert the UTF-16 code points to UTF-8
+      // so they should be autoreleased.
+      @autoreleasepool {
+        unmodified = [[NSString stringWithCharacters:uniUnmodified length:ccUnmodified] UTF8String];
+        withShift = [[NSString stringWithCharacters:uniWithShift length:ccWithShift] UTF8String];
+        withAltGraph = [[NSString stringWithCharacters:uniWithAltGraph length:ccWithAltGraph] UTF8String];
+        withAltGraphShift = [[NSString stringWithCharacters:uniWithAltGraphShift length:ccWithAltGraphShift] UTF8String];
+      }
+      
+      keymapEntry = createKeymapEntry(unmodified, withShift, withAltGraph, withAltGraphShift);
```
bryphe, line 222: Hi @zbaylin , I just thought of something - is it possible for the `unmodified`, `withShift`, etc keys to be released _prior_ to them being used in `createKeymapEntry`? Just worried about a possible race condition here, where the strings could be released prior to being copied to OCaml objects.

zbaylin: That's a good point. According to the `UTF8String` selector documentation:
> This C string is a pointer to a structure inside the string object, which may have a lifetime shorter than the string object and will certainly not have a longer lifetime.

So I think it makes sense to move that call to `createKeymapEntry` into the `autoreleasepool`, or manually create and release the `NSString`s. So sorry I didn't think of that before merging!

PR #2714, opened 2020-11-14
fix(configuration): hook up missing render whitespace option

CrossR:
Possibly I'm missing something here, but this just looks like it wasn't hooked up?


bryphe:
Ah, ya, nice catch @CrossR - looks like this was broken in https://github.com/onivim/oni2/commit/a89fc696473b7a25dce43d2954cdbf396224a1ce when migrating the configuration

Thanks for fixing it!

bryphe (approved):

bryphe:
The CentOS hang isn't related to your change - investigating here #2715 & here https://github.com/esy/esy/issues/1227

PR #2719, opened 2020-11-15
fix(buffers/#2605): buffer metadata not getting updated on switch to existing buffer

### LINKED ISSUE #2605: Ctr+Tab menu is no longer ordered by most recently used
...and hence a quick tap of Ctrl+Tab no longer works to switch to the previous editor either.

This was implemented in #1025 and used to work, but the menu now seems to ordered by last opened.

Possibly caused either by `lastUsed` on each buffer not being updated properly, or the menu no longer being sorted by `lastUsed`. From a quick glance through the code I can't see definite evidence of either, but there has been a major refactor since #1025 with the `BufferEnter` action previously used to update it being replaced by several different mechanisms in `Feature_Buffers`.

glennsl:
Fixes #2605


bryphe:
Thanks for the fix, @glennsl ! 

CentOS failure is unrelated to the change - working on fixing here: https://github.com/onivim/oni2/pull/2715 - I'll override the CI check.

PR #2728, opened 2020-11-17
fix(extensions/#2723): Searching for "theme" fails with an error

### LINKED ISSUE #2723: Extension search fails
When I search for extensions with a very common keyword (such as "theme"), results initially show up and fill the screen, but then they are quickly replaced with a message saying "There was an error searching for extensions. Please check your network connection and try again."

This was a problem in a nightly version downloaded on 2020-09-22, and still is.

bryphe:
__Issue:__ When searching for extensions, some queries - like `"theme"` - will end up with an error message (Please check your n etwork connection). 

__Defect:__ Our JSON parser expects that the `description` field is required - but this is not a valid assumption. In the case of the `"theme"` query, there are some extensions that do not have a description field populated, causing the search results to break.

__Fix:__ Make `description` optional, add test case.

Fixes #2723


iwikal:
Perhaps one could also change the error handling so that parsing errors show some more accurate error message? "Internal error - please file an issue" perhaps.

PR #2756, opened 2020-11-23
fix(KeyboardLayout): manually allocate and release NSStrings

zbaylin:
@bryphe pointed out in #2687 that it was possible we would get stale strings in `createKeymapEntry` since we didn't know when the `autoreleasepool` would release the `NSString`s. This removes the `autoreleasepool` in favor of manually allocating and releasing the `NSStrings`


bryphe (approved):
Nice, thanks for fixing this so quickly @zbaylin !

PR #2759, opened 2020-11-24
fix(ci): remove unneeded failure in release creation

zbaylin:
This changes `esy @release run --version` → `esy @release run --help`

As of right now, when we try to run the binary to create it for release, we get a linking error on macOS since we try to dynlink `Sparkle.framework` in a scenario that is explicitly disallowed by the kernel.

However, since the `--help` CLI flag means the program is terminated before we try to link the framework, we don't get this failure.

It's not a huge deal, but I think it makes more sense to have a failure that actually means failure rather than a failure that means success and we allow it to proceed.

### DIFF
--- a/.ci/esy-build-steps.yml
+++ b/.ci/esy-build-steps.yml
@@ -31,9 +31,8 @@ steps:
     displayName: 'Release: install'
   - script: 'esy @release build'
     displayName: 'esy @release build'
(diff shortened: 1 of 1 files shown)


bryphe (approved):
Great idea! Thanks @zbaylin :+1:

PR #2846, opened 2020-12-15
fix: Tab bug

SeitaHigashi:
Fixed the problem that the left and right scroll directions are reversed on the trackpad.

When I was using onivim on my Mac, I noticed that the scrolling of the tabs was reversed to the left and right of the trackpad. This means that you are not aware that you are using a mouse that only scrolls up and down.
I can't check it with OS other than mac, so it would be helpful if you could check it.

### DIFF
--- a/CHANGES_CURRENT.md
+++ b/CHANGES_CURRENT.md
@@ -6,6 +6,7 @@
 - #2842 - CLI: AppImage - fix argument parsing in `AppRun` (fixes #2803)
 - #2839 - Extensions: CodeLens - fix lens persisting when line is deleted
 - #2844 - Vim: Fix `:tabnew`/`:new`/`:vnew` behavior (fixes #1455, #2753, #2843)
+- #2846 - UX: Editor Tabs - horizontal scrolling on trackpad is reversed (thanks @SeitaHigashi!)
 
 ### Performance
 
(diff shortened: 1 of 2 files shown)


bryphe:
Thanks @SeitaHigashi , nice catch on the bug, and well done on the fix 👍 

I believe this should be a safe change on other platforms, as the `deltaX`/track-pad case is primarily OSX. I'll double-check on a Windows machine though.

bryphe:
I confirmed on a windows machine w/ trackpad - same bug exists there :+1:

bryphe (approved):

goetzc:
I have had this bug where the horizontal scrolling direction using trackpad is reversed, and after this fix, it's still reversed.

My system is Linux (libinput 1.16.4) and running Onivim 2 (0.5.9-nightly, f8b4933f). Vertical scrolling direction is correct, but horizontal is reversed. I'm using "natural scrolling" settings, but disabling this makes vertical scrolling direction incorrect and horizontal correct.

despairblue:
> I have had this bug where the horizontal scrolling direction using trackpad is reversed, and after this fix, it's still reversed.
> 
> My system is Linux (libinput 1.16.4) and running Onivim 2 (0.5.9-nightly, f8b4933f). Vertical scrolling direction is correct, but horizontal is reversed. I'm using "natural scrolling" settings, but disabling this makes vertical scrolling direction incorrect and horizontal correct.

Is this for scrolling the buffers or the tab bar at the top? Because I think this fix is only for the tab bar.

For file buffers I also still have this.

goetzc:
Oh, this is right, I was referring to the file buffers. Although the tab bar is scrolling in reverse horizontal direction too.

PR #2876, opened 2020-12-24
fix(titlebar): various UI fixes on macOS

zbaylin:
This attempts to make the titlebar more consistent with the standard one on macOS.
- Centers the content vertically in the titlebar
- Makes the font bold (is this just a Big Sur thing?)
- Makes the font size match the titlebar height

On Big Sur (11.0/10.16), here is Oni compared to iTerm2 now (which uses the default Cocoa titlebar):
Screen Shot 2020-12-24 at 11 21 51 AM

To my eyes, they look nearly identical. However, I'm not sure how it would look on a version of macOS earlier than Big Sur. @bryphe @CrossR, are either of you able to test this out on Mojave || Catalina etc. to make sure I didn't mess anything up there? No rush, obviously -- happy holidays!


bryphe:
Thanks for fixing this up @zbaylin ! Much appreciated

I tested on Catalina:

image

It looks a bit small compared to `iTerm2` - I think due to the proportional font size. On Catalina, for me, the titlebar is `22.` pixels - so ends up being < 10px. 

Maybe we could add a cut-off - if the titlebar is <= `22` pixels, we use the `12px` font size, otherwise a larger one? I think a semi-bold font would be better than a bold on Catalina, as well.

zbaylin:
@bryphe I just used the new version info from revery to determine the font size and weight -- let me know if this looks correct on Mojave/Catalina/etc!

bryphe (approved):
Looks great @zbaylin ! 

Thanks for the lengths you went through to get this set in a robust way 💯

PR #2896, opened 2020-12-29
fix(dependency): esy-skia -> 91c98f6

glennsl:
Brings in fixes to esy-skia that allows it to build with clang 10.2 on Arch Linux.


bryphe (approved):
Thanks @glennsl !

PR #2903, opened 2020-12-30
refactor(dependencies): update revery → f2da66e

zbaylin:
This brings in all the OS versioning information from the recent revery PRs. I wanted to bring this in separate from #2876 since it requires a lot of changes to switch statements across the codebase.


bryphe (approved):
Thanks for splitting this out into a separate change, @zbaylin !

PR #2904, opened 2020-12-30
chore(build): remove macOS build target on non-release builds

zbaylin:
As far as I can tell, the only time we need to specify the build target (10.12) is on release builds. As of right now, building Oni on anything other than 10.12 will spam the console with the following errors:
```
ld: warning: object file (foobar.o) was built for newer macOS version (10.15) than being linked (10.12)
```
Since as of adding auto-update we have a separate sandbox for release builds, we don't need to specify the release target in the debug builds.

@bryphe let me know if any of this is incorrect!


bryphe:
> Since as of adding auto-update we have a separate sandbox for release builds, we don't need to specify the release target in the debug builds.

Oh, nice catch! These errors were always annoying, but I thought they were unavoidable. I think you're right, though.

I just wanted to double-check and make sure we have CI coverage of building on OSX 10.15 and running the health-check on OSX 10.14 ... and I realized the output for the health-check was broken, so added a fix here: https://github.com/onivim/oni2/pull/2905

Once that's in, I just want to double-check this change with it - if it works, then we're in good shape.

PR #2920, opened 2021-01-02
feat(ux): add "window.menuBarVisibility" setting

glennsl:


bryphe (approved):
Thanks for adding this, @glennsl !

PR #2944, opened 2021-01-06
fix(components): remove overscroll in `VimList`

zbaylin:
I'm actually not sure if there are historical reasons that we allow overscroll up to 1 element, but if not, I find it slightly annoying sometimes when I scroll with the mousewheel and get something like this (especially considering the displayed scroll bar doesn't allow the same and disappears):
image

This is just a simple math change so that we bound the scroll to remove overscroll.

### DIFF
--- a/CHANGES_CURRENT.md
+++ b/CHANGES_CURRENT.md
@@ -33,6 +33,7 @@
 - #2927 - Input: Windows - fix crash in entering Unicode character (fixes #2926)
 - #2938 - Input: OSX - Modifier keys not working on Romaji keyboard (fixes #2924)
 - #2941 - Input: Fix handling of `<space>` as leader key (fixes #2935)
+- #2944 - Components: Remove overscroll in `VimList`
 
 ### Performance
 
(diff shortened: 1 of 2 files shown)


bryphe:
Thanks for looking into this and fixing it, @zbaylin ! Agree - that's a good point with the mouse interaction, clamping it to the bottom makes sense.

> I'm actually not sure if there are historical reasons that we allow overscroll up to 1 element, 

The reason is some vim commands (`zz`, `zt`) could overscroll - for example, if you press `G` and `zt`, it would scroll such that the last element is at the very top. This is just parity with how it would behave in a text buffer.

I think we could have both, though - we could pass an `~allowOverscroll=false` to `setScrollY`. Depending on whether that is set, we'd use the original calculation or the new calculation for the max scroll.

By default, `allowOverscroll` would be false - but we could override in `scrollSelectedToTop` and `scrollSelectedToBottom` - these are the functions that are relevant for `zz`/`zt`.

zbaylin:
> The reason is some vim commands (`zz`, `zt`)

Ah that makes total sense -- I totally missed those in my head.

> I think we could have both, though - we could pass an `~allowOverscroll=false` to `setScrollY`

Sounds great! I just added this in 6a476a7 -- let me know what you think!

bryphe (approved):
Just tested and it looked like `zz` wasn't working quite right - but was an easy fix :+1: 

Thanks for implementing this @zbaylin - looks great to me!

PR #2954, opened 2021-01-07
fix(markdown/#2953): explicitly set code block font size in Markdown

### LINKED ISSUE #2953: Extensions: codeblocks use incorrect font in extension descriptions
I noticed when looking at extensions recently that (on macOS, at least), code blocks in extension descriptions are rendered with the wrong font at the wrong size. Take the Ruby extension:
image
The inline code is rendered fine (as in "`ruby.useLanguageServer`") but the code block seems to use the default system font which is not monospace and is also the wrong size.

I'm not sure if this is a bug in the Markdown component or if there is something we're not passing to the component that we should be.

Also @bryphe are the Uninstall/Update button colors inherited from the theme at all? It seems a little strange to have Uninstall be red in the sidebar but green on the extension page -- should be an easy fix though!

zbaylin:
For some reason, explicitly setting the code block font size fixes the code block issue in extension descriptions.

This should just be a temporary fix until I can figure out *why* this is happening, but it works well for now.

Fixes #2953

### DIFF
--- a/CHANGES_CURRENT.md
+++ b/CHANGES_CURRENT.md
@@ -37,6 +37,7 @@
 - #2942 - TextMate: Fix infinite loop with vala grammar (fixes #2933)
 - #2937 - Editor: Fix bugs around horizontal scrolling (fixes #1544, #2914)
 - #2946 - SCM: Fix StackOverflow when retrieving original content for large files
+- #2954 - Markdown: explicitly set code block font size (fixes #2953)
 
 ### Performance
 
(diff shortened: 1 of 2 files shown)


bryphe (approved):
Thanks @zbaylin , nice catch! It is weird that setting the size gets the font picked up 🤔 

Looks so much better... can't believe I missed that default font being used.

bryphe:
I'll merge this in so you don't have to deal with any merge conflicts from other changes - thanks for investigating this, @zbaylin !

PR #2990, opened 2021-01-14
fix(editor): Signature Help - Fix escape key behavior

bryphe:
__Issue:__ When the signature help / parameter hints are visible, the `<esc>` key closes the signature help UI, but does not exit insert mode. This breaks muscle memory - requiring extra `<esc>` presses when that UI is visible.

__Fix:__ Don't block the `<esc>` key - instead, map `shift`+`esc` to close parameter hints, and keep `esc` reserved for going from insert -> normal mode. In addition, move the `SignatureHelp` logic from its own feature to `Feature_LanguageSupport`, so some of the shared machinery can be re-used.


fiddler:
Big thanks for fixing this! Was one of the main annoyances now that I'm trying to shift to using Oni2 as my daily editor. 🙇

bryphe:
Definitely, glad to hear @fiddler - this was a major annoyance for me too - this sort of bug takes you out of the 'flow' when writing code. Let me know if you're hitting any other annoyances like this :+1:

PR #2996, opened 2021-01-15
docs(building): reorder instruction

jakubbaron:
The first step on [Building from Source](https://onivim.github.io/docs/for-developers/building) should be cloning the repo and cd'ing into it. 

I had my friend asking me where can they get `install-node-deps.js` from. If we were to clone repo firstly, cd into it, all the following commands in the instructions work out of the box.

Removing clone from down the line, moving it as one of the top steps.

### DIFF
--- a/docs/docs/for-developers/building.md
+++ b/docs/docs/for-developers/building.md
@@ -40,6 +40,13 @@ Some Linux distributions may need other packages:
 
 ## Build and Run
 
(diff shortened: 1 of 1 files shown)


bryphe:
Good catch - thanks for improving this, @jakubbaron !

bryphe (approved):

PR #2997, opened 2021-01-15
fix(syntax/#2985): PHP syntax highlights not showing

### LINKED ISSUE #2985: PHP: Syntax highlighting does not appear to work at all
I'm having trouble getting any syntax highlighting to work at all. I am trying to open primarily PHP and JS files, but they appear completely white/gray. I've also tried opening various other files like JSON.

Full disclosure: I have only been trying onivim out every so often to see how it has been progressing, so I may not know what I'm doing. I thought that syntax highlighting used to work for these out of the box, but I may be mistaken?

I've previously loaded vscode extensions and tinkered around, so I tried to fully reset onivim by removing `~/.config/oni2` and starting it fresh but get the same result.

I'm using Onivim 2 (0.5.9-nightly) on (Arch) Linux.

bryphe:
__Issue:__ PHP syntax highlights are no longer showing by default

__Defect:__ https://github.com/onivim/oni2/pull/2889 upgraded the extensions, but the PHP extension has a new dependency on `embeddedLanguages`, which our current textmate implementation does not handle correctly

__Fix:__ Remove the `embeddedLanguages` for now - add a regression / canary test to ensure we don't break this in the future.

Fixes #2985 
Tracking `embeddedLanguage` support in #2998

### DIFF
--- a/CHANGES_CURRENT.md
+++ b/CHANGES_CURRENT.md
@@ -57,6 +57,7 @@
 - #2990 - Signature Help: Fix blocking `esc` key press back to normal mode
 - #2991 - OSX: Fix shortcut keys double-triggering events
 - #2993 - CodeLens: Handle null command id & label icons
+- #2997 - Syntax: Fix regression in syntax highlighting for PHP (fixes #2985)
 
 ### Performance
 
--- a/extensions/php/package.json
+++ b/extensions/php/package.json
@@ -35,20 +35,6 @@
 				"language": "php",
 				"scopeName": "source.php",
 				"path": "./syntaxes/php.tmLanguage.json"
-			},
-			{
-				"language": "php",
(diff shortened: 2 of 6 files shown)


nerdo:
There still seems to be an issue with this fix.

I was able to open a single PHP file and see syntax highlighting, but when I open more PHP files in my project, they do not have syntax highlighting.

nerdo:
I just did a little bit more tinkering and figured out what seemed to break it.

1. Open any number of PHP files
2. Open a file without syntax highlighting (in my case, I opened a Laravel log file)
3. Subsequent files no longer seem have syntax highlighting applied. It doesn't seem to matter if they are PHP files or other types in my case.

OniVim v0.5.9-nightly ad850337

PR #3000, opened 2021-01-16
fix(extensions/#2979): fix order of extensions

jakubbaron:
With the current implementation, subsequent subquery was prepending the
results to the previous results while we should be appending and thus
keeping the sought match at the top

Previously:
ExtensionSearch

New:
ExtensionSearchNew


bryphe (approved):
Thanks for the fix @jakubbaron and for showing the gifs - definitely a huge improvement in the after vs before! :tada: Looks great

PR #3023, opened 2021-01-21
Quote the when keyword (when -> "when")

noahhaasis:


bryphe:
Thanks for catching this and fixing it, @noahhaasis !

bryphe (approved):
Looks great 👍

Issue #2115, opened 2020-07-16
Add search.exclude option


CrossR:
As far as I can see, we are missing the `search.exclude` option, which is the same as `files.exclude` but specific to search.
It should also inherit all the values from `files.exclude`.

Technically this is what we should be using for the quick open as well according to the VSCode docs: 

```json
 // Configure glob patterns for excluding files and folders in fulltext searches and quick open. Inherits all glob patterns from the `files.exclude` setting.
  "search.exclude": {
    "**/node_modules": true,
    "**/bower_components": true,
    "**/*.code-search": true
  },
```

I'll tag this as a good first issue, since I don't think it should be too much work? We have the same functionality under a different name already, its just a case of adding this option, and also hooking it up to be used in `findInFiles`. https://github.com/onivim/oni2/pull/1099 hooked this up, but does not include the addition of the config option itself, and this option would need to apply to both `search` and `files.exclude`.

At a later date, we may want to update them both since right now we have a list of strings, rather than the key/value pairs that VSCode uses (not sure when that was changed).

JonathanLorimer:
Hello, I would love to help out with this. I am new to OCaml / ReasonML, but I have a fair bit of experience with Haskell and Javascript / Typescript. If you could point me to the relevant files that I should start looking at I will try and get up to speed on this!

CrossR:
Most of the files you'll be interested in are over in the `src/Core/` folder. You'll see the few related to the configuration, with `ConfigurationParser` being where the actual JSON option gets transformed. I'm not sure if there is another config option that uses the `{"option": bool, "option2": bool}` format, so it could be implemented as a list of strings to start most likely, which should be easier (and there are other examples). The other files will contain the default values and types etc.

`Ripgrep` in that same folder is where we do the actual calls into `ripgrep`, so you'd want to add `searchExclude` to `findInFiles`, and you can probably follow the linked PR in the OP to see how the `filesExclude` arg was hooked up, which is very similar to this. The only real difference is that you need to add the mentioned config option as well, and then concat `search.exclude` and `files.exclude` before sending.

Hopefully that makes sense, and there isn't something massively difficult hiding that I've not spotted from a quick poke around. The discord in the README can be a nice way to get some quick help (either specifically to Oni2 in our discord, or for OCaml/Reason in general in their respective discords).

JonathanLorimer:
@CrossR Thanks for the quick response. I will start looking into this!

bryphe:
[...]

PR #3044, opened 2021-01-25
fix(search): Implement search.exclude configuration option 

joseeMDS:
Hi, im working at #2115, this is still WIP, and i have some doubts, like how can i acess the configuration state at `Feature_Search` and in what file should i edit to `findInFiles` dont search the excluded paths


bryphe:
Hi @joseeMDS ,

This looks like a great start - thank you for taking this on!

I'll take a look through and leave some feedback / address your questions.

src/Feature/Search/SearchSubscription.re
```diff
@@ -36,6 +37,7 @@ module Make = (Config: {type action;}) => {
 
       let dispose =
         ripgrep.Ripgrep.findInFiles(
+          ~searchExclude=[],
```
bryphe, line 40: Looks like we'll need to plumb the `searchExclude` params. 

The `params` is destructured in line 26 in this file - we could add `searchExclude` to that destructuring.

Then, we can just use `~searchExclude,`, which is a short-hand for `~searchExclude=searchExclude,`

src/Core/ConfigurationParser.re
```diff
@@ -205,6 +206,11 @@ let configurationParsers: list(configurationTuple) = [
     "files.exclude",
     (config, json) => {...config, filesExclude: parseStringList(json)},
   ),
+  (
+    "search.exclude",
+    (config, json) => { ...config, searchExclude: parseStringList(json)}
+
```
bryphe, line 212: This isn't documented very well in our developer docs, but we're starting to move away from this strategy of configuration - and instead specifying configuration per-feature.

A good example of this is here: https://github.com/onivim/oni2/blob/0f1bf1ea4b2da4590e7c5cd5d67d2cd994f09fd5/src/Feature/Zoom/Feature_Zoom.re#L93 (just recently added in #3021 )

Instead of implementing this here - we could add this to `Feature_Search`:
```
module Configuration = {
  open Config.Schema;

  let exclude = setting("search.exclude", list(string), ~default=[]);
};
```

We could follow a similar strategy for the `search.exclude` parameter, and add something like this in `Feature_Search`.

The reason for doing this, is that this will make it easier to plumb the configuration setting into the subscription runner here: https://github.com/onivim/oni2/blob/0f1bf1ea4b2da4590e7c5cd5d67d2cd994f09fd5/src/Feature/Search/Feature_Search.re#L212

Once we have that configuration setting in `Feature_Search`, we could modify the `let subscriptions = ...` function so that we can get the configuration setting:

```
let subscriptions = (~config: Oni_Core.Config.resolver, ~workingDirectory, ripgrep, dispatch) => {
  let searchExclude = Configuration.exclude.get(config);
  let search = query => {
    SearchSubscription.create(
      ~id="workspace-search",
      ~query,
      ~directory=workingDirectory,
      ~ripgrep, 
      ~searchExclude,
      ~onUpdate=items => dispatch(Update(items)),
      ~onCompleted=() => Complete,
      ~onError=msg => SearchError(msg),
    );
  };
  
   ...
};
```
(We'd also need to update `Feature_Search.rei` to add this new parameter).

The next thing we need to do, though, is get that `config` parameter - we could wire that in here: https://github.com/onivim/oni2/blob/0f1bf1ea4b2da4590e7c5cd5d67d2cd994f09fd5/src/Store/Features.re#L1883

something like:
```
(state: State.t, dispatch) => {
    ...
    let searchDispatch = msg => dispatch(Search(msg));
    let config = Selectors.configResolver(state);
    searchSubscriptions(~config, ~workingDirectory, searchDispatch, state.searchPane)
    |> SearchSubscriptionRunner.run(~dispatch=searchDispatch);
```

Hope that helps in terms of wiring up the configuration setting to the subscription. Let me know if you have questions!

src/Core/Ripgrep.re
```diff
@@ -271,18 +272,33 @@ let search =
...
+    |> List.concat
+    |> List.append([
+         "--fixed-strings",
+         "--smart-case",
+         "--hidden",
+         "--json",
+         "--",
+         query,
+         directory,
+       ]);
```
bryphe, line 297: We might be able to simplify this a bit with:
```
let excludeArgs = searchExclude
   |> List.concat_map(x => ["-g", "!" ++ x]);

let args = excludeArgs @ [
	 "--fixed-strings",
         "--smart-case",
         "--hidden",
         "--json",
         "--",
         query,
         directory,
]
```

The `@` is just a short-hand for append, and the `concat_map` is a more efficient combination of `map` and `concat`: https://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html

joseeMDS:
Thanks for all comments @bryphe, i thinks it is all okay now, except that `search.exclude` should inherit from `files.exclude`, i think we can do it at `let subscriptions = ...` using it `~config` param, but im not sure what function should i use, and i if you leave `search.exclude: [""]` it will just ignore all the paths and show nothing, so now im thinking in how to fix this, any suggestion in welcome

src/Feature/Search/Feature_Search.re
```diff
@@ -31,8 +31,8 @@ let initial = {
 
 module Configuration = {
   open Config.Schema;
-
   let searchExclude = setting("search.exclude", list(string), ~default=[]);
+  let filesExclude = setting("files.exclude", list(string), ~default=["_esy",".git","node_modules"]);
```
joseeMDS, line 35: I dont know if this the right thing to do, but its working

bryphe: Great idea! Makes sense to define both of these settings here.

src/Feature/Search/Feature_Search.re
```diff
@@ -209,11 +215,16 @@ module SearchSubscription =
     type action = msg;
   });
 
-let subscriptions = (~workingDirectory, ripgrep, dispatch) => {
+let subscriptions =
+    (~config: Oni_Core.Config.resolver, ~workingDirectory, ripgrep, dispatch) => {
+  let searchExclude = Configuration.searchExclude.get(config) 
+  |> List.append(Configuration.filesExclude.get(config))
```
bryphe, line 221: Looks good 👍

docs/docs/configuration/settings.md
```diff
@@ -8,162 +8,171 @@ Onivim's settings are designed to be mostly compatible with [VSCode's User Setti
...
+-   `editor.zenMode.hideTabs` **(_bool_ default: `true`)** - When `true`, the Onivim will hide the buffer tabs from the user whilst in zen mode. Zen mode can be toggled in the command pallette, or automatically enabled with the `editor.zenMode.singleFile` configuration option.
 
-- `editor.zenMode.singleFile` __(_bool_ default: `true`)__ - When `true`, the Onivim will automatically enter zen mode when started up with a single file. Zen mode hides most of the UI until disabled via the command pallette.
+-   `files.exclude` **(_list of string_ default: `[".git", "_esy", "node_modules"]`)** - When using `Quick Open` or `Find in files`, Onivim will ignore the files inside the directories listed here
 
-- `editor.zenMode.hideTabs` __(_bool_ default: `true`)__ - When `true`, the Onivim will hide the buffer tabs from the user whilst in zen mode. Zen mode can be toggled in the command pallette, or automatically enabled with the `editor.zenMode.singleFile` configuration option.
+-   `search.exclude` **(_list of string_ default: `[]`)** - When using `Find in files` Onivim will not look at files located at the directories listed here, this inherit all the values from `files.exclude`
```
bryphe, line 114: Thanks for adding the docs here!

Are the extra changes just formatting changes? Would be helpful to minimize the changes just to the settings being added.

bryphe:
Thanks for all your work here, @joseeMDS ! The changes are looking good to me - I like your approach for merging the two settings for search 👍 

Just had a question about the documentation, but after that, ready to bring this in!

docs/docs/configuration/settings.md
joseeMDS: Hi, i didnt notice this, it was the editor format autosave, but i will revert this and add just the necessary

src/Core/Ripgrep.re
```diff
@@ -271,18 +272,35 @@ let search =
...
+      ~query,
+      ~onUpdate,
+      ~onComplete,
+      ~onError,
+    ) => {
+  let excludeArgs =
+    switch (searchExclude) {
+    | [""] => []
+    | searchExclude => searchExclude |> List.concat_map(x => ["-g", "!" ++ x])
+    };
```
bryphe, line 288: We could potentially replace this with:
```
  let excludeArgs = searchExclude
    |> List.filter(str => !StringEx.isEmpty(str))
    |> List.concat_map(x => ["-g", "!" ++ x]);
```

This would have the added benefit of handling some other cases like `["", ""]`, or `["valid-path", ""]`

bryphe (approved):
Looks great - thanks for your help with this, @joseeMDS !

Picked up ReasonML fast 😄 Not easy to figure out how to wire up all the different pieces, too... nice work!

bryphe:
Latest iteration looks good; just pushed up a fix for the hygiene checks (running `esy format`) - I'll merge when it's green.

@joseeMDS - I assume you're `joseeMDS` on twitter that DM'd me - is it ok if I DM you your license key?

joseeMDS:
Hey, Yes its me,I will try to contribute more, will keep learning and come back soon, and yes you can DM me with the key

PR #3117, opened 2021-02-09
fix(terminal/#3098): Fix windows terminal input

### LINKED ISSUE #3098: Powershell terminal doesn't take keyboard input
:terminal opens a new powershell window, but it doesn't register any keyboard input. Trying 'i' to enter insert mode, doesn't work. 
Env: Windows 10 up to date. OniVim nightly 0.5.9. 
Note: this could be a duplicate, didn't check all previous issues.

bryphe:
- Add test case to verify terminal input is round-tripped back to the client
- Fix #3098 (may require reverting the node 14 LTS upgrade)


haaland (approved):
Sad to see a revert..

bryphe:
For sure, bummer to have to revert the node upgrade... but at least there is a test case in place to make it safer to upgrade in the future, now (will keep an eye on https://github.com/microsoft/node-pty/issues/457 - once that's fixed, can un-revert the revert 😄 )

bryphe:
Note that users building from source will need to run `node install-node-deps.js` to pick up the right node depencies

Eugeny:
@bryphe I see you've run into the same `node-pty` issue as I did. I've published a fix as `@terminus-term/node-pty@0.10.0-terminus.3` - would appreciate it if you could give it a quick test!

bryphe:
Hi @Eugeny ,

Indeed - thanks for logging the issue (and fixing it). I've tried out your fix, and it addresses the bug for me as well! :tada:

PR #3123, opened 2021-02-09
fix(extensions): Large VSIX extensions are failing to install

bryphe:
__Issue:__ Reported on Discord - `redhat.java` extension was all of a sudden failing to install.

__Defect:__ Onivim uses the [`follow-redirects`](https://www.npmjs.com/package/follow-redirects) library to handle `302` redirects. The URL `https://open-vsx.org/api/redhat/java/0.74.0/file/redhat.java-0.74.0.vsix` actually gets redirected. `follow-redirects` has a default limit of 10MB for downloads, otherwise it will fail.

__Fix:__ Increase size limit to accommodate extensions on open-vsx, so that the download will complete.

__Next steps:__
- The download speed using `follow-redirects` seems to be slow. Hopefully, we can switch to @zbaylin 's libcurl work eventually: https://github.com/zbaylin/ocurl-luv-test

### DIFF
--- a/CHANGES_CURRENT.md
+++ b/CHANGES_CURRENT.md
@@ -31,6 +31,7 @@
 - #3085 - Snippets: Fix drop-shadow calculation at end of buffer
 - #3091 - Snippets: Fix auto-closing pairs when placeholders are on same line
 - #3102 - Vim / Input: Implement mapping timeout (fixes #2850)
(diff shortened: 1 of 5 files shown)


bryphe:
Workaround in the meantime would be to manually download the extension: https://open-vsx.org/api/redhat/java/0.74.0/file/redhat.java-0.74.0.vsix and then run `oni2 --install-extension /path/to/redhat.java-0.74.0.vsix`

PR #3143, opened 2021-02-16
chore(ci): Add validation step for MacOS Big Sur since that VM image is available

bryphe:


bryphe:
Unfortunately this isn't available in Azure DevOps yet-

PR #3144, opened 2021-02-16
fix(build): support libintl discovery on OSX ARM

ilyaglow:
```changelog
Homebrew's Cellar has new location on ARM Macs: /opt/homebrew/Cellar.
This change adds this path in order to find libintl library.

Related to #2708.
```


ilyaglow:
Better approach is implemented in #3806

PR #3160, opened 2021-02-18
fix(windows): Fix configuration reloading on Windows

bryphe:
__Issue:__ On Windows, saving the configuration file does not immediately reflect changes - the editor needs to be re-opened to see the impact of configuration changes.

__Defect:__ There were a few issues contributing to this:
- Another windows-style-pathing comparison issue (comparing `\` to `/` path separators) - we were comparing a 'normalized' path with an 'un-normalized' path, so the comparison failed.
- There were separate configuration-loading pathways for some configuration settings, which added to the confusion above

__Fix:__ This ended up being a larger fix than I expected - first, the configuration was wired to use strongly-typed paths, which avoid the pitfalls of string-based path comparison. Then, a common loading path was implemented for both the old and new configuration model - and several configuration settings (like Zen mode) were migrated to the new model. Finally - once the configuration issue was addressed, there were still some issues with the async-font-loading. I've moved the asynchronous font-loading back to synchronous, as the new skia loading path is much faster than the previous one.


src/Feature/Configuration/ConfigurationLoader.re
```diff
@@ -0,0 +1,99 @@
...
+);
+
+let loadConfiguration = (loadResult: result(Yojson.Safe.t, string)) => {
+  // let parseResult =
+  // loadResult
+  //   |> Utility.ResultEx.flatMap(Keybindings.of_yojson_with_errors);
+  // switch (parseResult) {
+  // | Ok((bindings, errors)) => (bindings, errors)
+  // | Error(msg) => ([], [msg])
+  //};
```
bryphe, line 17: Remove

src/Exthost/Client.re
```diff
@@ -51,10 +51,13 @@ let start =
   let dispatch = msg => {
     Protocol.Message.(
       switch (msg) {
-      | Incoming.Connected => Log.info("Connected")
+      | Incoming.Connected =>
+        Log.info("Connected");
+        prerr_endline("Connected");
```
bryphe, line 56: Remove debug logging

src/Feature/Configuration/ConfigurationLoader.re
```diff
@@ -0,0 +1,99 @@
...
+  loadResult
+  |> ResultEx.flatMap(json => {
+       prerr_endline("Reading new JSON: " ++ Yojson.Safe.to_string(json));
```
bryphe, line 20: Remove debug logging

src/Feature/Configuration/Feature_Configuration.re
```diff
@@ -62,30 +87,108 @@ let toExtensionConfiguration = (config, extensions, setup: Setup.t) => {
...
+  let task = ConfigurationLoader.transformTask(~transformer, model.loader);
+
+  {...model, transformTasks: Task.queueTask(~task, model.transformTasks)};
+};
+
+let update = (model, msg) =>
   switch (msg) {
-  | UserSettingsChanged =>
+  | UserSettingsChanged({config: user, legacyConfiguration}) =>
+    //prerr_endline ("!!USER SETTINGS CHANGED");
```
bryphe, line 132: Remove

PR #3259, opened 2021-03-11
fix(quickmenu): Fix quick menu smart case

amiralies:
This fixes issue with smart casing while filtering items in quickmenu (theme picker, formatter picker, snippets..).


bryphe:
Great catch - thanks @amiralies !

bryphe (approved):

PR #3246, opened 2021-03-08
feat(formatting/#2464): Initial format-on-save implementation

bryphe:
This implements format-on-save, adding an `"editor.formatOnSave"` configuration setting that can be set per-filetype, and then actually implementing the format-on-save behavior.

__TODO:__
- [x] Fix up notification message (when applying edits due to save, update to say something like "Saved with X formatting edits from <format-provider>")
- [x] Move the write command to `Feature_Vim.Effects.save(~bufferId)`
- [x] Investigate adjusting cursor position in response to formatting edits (#3263)
- [x] Investigate limits for large files
- [x] Add test case

Fixes #2464

Issue #1581, opened 2020-04-08
Syntax highlighting has visible latency for basic operations


zeux:
When editing files of moderate size - the example below is https://github.com/zeux/meshoptimizer/blob/master/src/simplifier.cpp which is ~1500 lines - syntax highlighting tends to have a visible delay. For some operations, like deleting lines, the delay manifests as flashing highlights for the rest of the file, which is visually distracting. Unsure if this is limited to C++ (which is what I mostly edit) or a more general problem.

[_simplifier.cpp - meshoptimizer - Onivim 2 2020-04-08 09-27-37.zip](https://github.com/onivim/oni2/files/4451636/_simplifier.cpp.-.meshoptimizer.-.Onivim.2.2020-04-08.09-27-37.zip)

As a comparison, Sublime on the same file has rock-solid syntax highlighting, any editing operations seem to immediately highlight the newly inserted text and don't disturb the highlighting for the rest of the file.

(by the way, I love the work you guys are doing - I've been using Vim exclusively since 2010, and I increasingly gravitate towards Sublime or Visual Studio with Vim emulation plugins to get more IDE functionality, so I'm excited about the future of onivim - just filing issues that come up as I start to explore the editor)

pikachunakapika:
Hi, I have this with files that are only like 100 lines long when deleteting lines or adding lines.

Raikiri:
I'd really like to see OP contributing to this project's performance optimization because he's pretty good at it.

PR #3260, opened 2021-03-11
fix(formatting): Compute minimal updates for buffer updates / shifting

bryphe:
In cases where there is a scoped buffer update, like a few lines, Onivim is smart enough to shift 'markers' (syntax highlighting, codelens in #3255 , search highlights).

However, there are some cases where we get a full update - the entire buffer contents is flagged as updated. A couple of these cases are:
- Running `:e!` to re-load the file from disk - this comes through as a refresh of the entire buffer
- Some format providers provide scoped updates (ie, line-by-line). Others, however, like `ocamllsp` just send the whole buffer contents.

The problem with processing the full update is that, today, those don't shift markers - we just clear out syntax highlighting, codelens, etc. This is a poor user experience, because it causes 'flashes' of syntax highlighting disappearing, or codelens disappearing and then re-appearing.

In order to have a better experience - this PR adds the computation of a set of minimal updates from these full updates. After computing minimal updates, Onivim can then properly shift markers, codelens, syntax highlighting, etc.

This is important for #3246 , as this same logic will be used to shift and preserve the cursor position.

### DIFF
--- a/CHANGES_CURRENT.md
+++ b/CHANGES_CURRENT.md
@@ -50,6 +50,7 @@
 
 - #3148 - Large Files: Improve performance & fix crash when opening large files (related #1670)
 - #3139 - Large Files: Fix hang when using `/` search (fixes #1670)
+- #3260 - Editor: Compute minimal updates, fix flash when reloading file
 
 ### Documentation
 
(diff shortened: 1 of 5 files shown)


bryphe:
This actually helps with some of the cases for https://github.com/onivim/oni2/issues/1581 - the flashes of the existing lines that are shifted by the add / delete are now gone.

The case that still needs to be addressed is immediately applying syntax highlights for the pasted text to eliminate that flash.

PR #3352, opened 2021-03-31
feat(keybindings): add binding for opening config

LiHRaM:
Listed in #1423 as a good place to get started as adding a keybinding for the "Open Settings" shortcut. Hope it's okay!

(I accidentally force pushed to master when tracking upstream, so setting this to a different branch now. Unfortunately GitHub doesn't seem to allow me to change the tracking downstream branch after opening an issue...)


bryphe:
Hi @LiHRaM ,

Sorry for the slow reply! Thank you for the PR, really appreciate the contribution 🎉 

Made a couple small tweaks; but when CI is green, will merge.

src/Feature/Configuration/Feature_Configuration.re
```diff
@@ -333,8 +333,22 @@ module Commands = {
     );
 };
 
+module Keybindings = {
+  open Feature_Input.Schema;
+
+  let command = Commands.openConfigurationFile.id;
+
+  let openConfigOnMac =
+    bind(~key="<D-,>", ~condition="isMac" |> WhenExpr.parse, ~command);
```
bryphe, line 342: I believe the condition is case-sensitive (which may be a bug itself) - flipped from `IsMac` -> `isMac`.

src/Feature/Configuration/Feature_Configuration.re
```diff
@@ -333,8 +333,22 @@ module Commands = {
...
+
+  let openConfigOnOther =
+    bind(~key="<C-,>", ~condition="!isMac" |> WhenExpr.parse, ~command);
```
bryphe, line 345: I was worried this might conflict with a default Vim keybinding, but doesn't appear that it does 👍

src/Model/State.re
```diff
@@ -31,6 +31,7 @@ let defaultKeyBindings =
     ),
   ]
   @ Feature_SideBar.Contributions.keybindings
+  @ Feature_Configuration.Contributions.keybindings
```
bryphe, line 34: Thanks for wiring this up!

bryphe:
Looks like the CI failure isn't related to this change; it's a bug with loading the cache. So I'll bring it now.

Thanks for the contribution, @LiHRaM - much appreciated!

PR #3368, opened 2021-04-05
experiment(ci): Test esy@next

bryphe:
Check `esy@next` - cc @EduardoRFS


EduardoRFS:
Hoping that this works, because I already released it, it was blocking some other stuff ;/

bryphe:
Looks like there was an error on CentOS, but not related to `esy` -> seeing the same issue in #3365

PR #3398, opened 2021-04-10
fix(ux/#3362): handle text overflow in FileExplorer

### LINKED ISSUE #3362: No ellipsis in sidebar, selected items have cut-off background
Sidebar visuals are somewhat off right now:
- the scrollbar cuts off selection colors for items
- text is overflowing without ellipsis

Screenshot from 2021-04-02 14-41-43

For comparison, this is how it looks in VSCode, which feels more natural to me:
Screenshot from 2021-04-02 14-43-20

I will be glad to help fix this if someone can point me in the right direction. I have already found the revery example where ellipsis are applied. Does this work in a `flexGrow` context or do we need to set an explicit width? 
Also, would these changes need to be made in this repo or are they affecting the general revery elements as well?

z0al:
An attempt to fix #3362 

Edit: this doesn't work actually. I trying to figure out what is going on

## Open questions

I see similar cut-off issues all over the place. Is there a proper way to fix the _root_ cause? 

In an ideal world, I'd imagine the `Text` component to _somehow calculates the available width internally_ so setting `textOverflow(Ellipsis)` would work out of the box.

### DIFF
--- a/src/Components/FileExplorer/FileTreeView.re
+++ b/src/Components/FileExplorer/FileTreeView.re
@@ -27,7 +27,7 @@ module Styles = {
 
   let item = [flexDirection(`Row), flexGrow(1), alignItems(`Center)];
 
-  let text = (~isFocus, ~isActive, ~decoration, ~theme) => [
+  let text = (~isFocus, ~isActive, ~decoration, ~theme, ~maxWidth) => [
     color(
       switch (
         Option.bind(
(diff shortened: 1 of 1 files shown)


z0al:
I tried everything I could today and nothing seems to work. Closing for now as it's pointless to keep the PR open.

bryphe:
Hi @z0al ,

Thanks for investigating this! Indeed, it does seem like there is a revery bug - opened a PR here: https://github.com/revery-ui/revery/pull/1056

PR #3441, opened 2021-04-21
lint spelling on EULA

taithethai:
Corrected some misspelled words.


zbaylin:
Thanks @taithethai! @bryphe does this require a new version of the EULA to be published (v1.2 or something?)

bryphe:
Thanks @taithethai , really appreciate the fixes! Changes look good.

Good question @zbaylin - since these are just spelling corrections, and there aren't material changes to the EULA, we don't need to version bump.

PR #3444, opened 2021-04-22
re-instate clobbered changelog entry after merge

jesstelford:


jesstelford:
nvm, I see it's listed under "fixes".

PR #3445, opened 2021-04-22
Move workbench.action.quickOpenBuffer feature under "Feature" in changelog

jesstelford:
Also corrects miss-named action (was "action.workbench.quickOpenBuffer", changed to "workbench.action.quickOpenBuffer")


bryphe (approved):
Thanks @jesstelford , nice catch!

PR #3448, opened 2021-04-22
chore(ci): Validate big sur

bryphe:
Looks like a `macos-11.0` vm image might be available, now? https://github.com/actions/virtual-environments/commit/98743bcf9cd51a429767516f8291c43bdd89ba4f


bryphe:
Unfortunately not available yet

PR #3451, opened 2021-04-22
fix(terminal/#2297): 'oni2' command in integrated terminal should open files in current editor

### LINKED ISSUE #2297: Integrated Terminal: 'oni2' command should open file in current editor
__Expected:__ When running the `oni2` command to open a file, ie, `oni2 some-file.txt`, the expectation is that the file would open in the current_session.

__Actual:__ A new Oni2 window is created, which isn't very useful behavior.

I thought we had an issue logged for this already - but couldn't find it.

Related to #1250 - solving that issue would likely solve this one, too. We can potentially pass an environment variable / some context so that a 'nested' Onivim 2 process knows there is already an instance running, and can let it know to open a file.

bryphe:
__Issue:__ Running `oni2 some-file.txt` in the Onivim integrated terminal opens `some-file.txt` in a new instance of Onivim, rather than the current instance.

__Defect:__ There's no context sent to the `oni2` command in the terminal to know to send to the active editor.

__Fix:__ Create a named pipe that multiple instances of Onivim can communicate through, and send the information for that pipe to the terminal.

__TODO:__
- [x] Fix CLI-specified files from opening in same split as terminal
- [x] Test on Windows
- [x] Test on Linux

Fixes #2297


zbaylin:
FYI @bryphe if you use `open -a Onivim2.app <file>` on macOS it already opens in the current editor because of the work we did with opening files with right-click, etc. Not sure if that makes it any easier for you on macOS.

bryphe:
Thanks @zbaylin ! I basically used a new (cross-platform) setup to facilitate sending remote commands from the CLI -> a target process (and, for the terminal, sending context about the target process) - same strategy on all platforms (create a named pipe, send the named pipe info to the terminal, send command over named pipe, etc)

PR #3458, opened 2021-04-25
docs: change search motion from // to /

Minnozz:


bryphe (approved):
Nice catch @Minnozz , thanks for the fix!

Issue #3467, opened 2021-04-28
Build issue related to node versions


dglanzman:
Hello,

Following the source build instructions here: https://onivim.github.io/docs/for-developers/building on a clean install of windows server 2019, I get a build error where I think node-gyp is looking in the wrong directory for `node_api.h`

From the build instructions, I install git, node and esy. To get node, I go to https://nodejs.org and download the latest version, 16.0.0.

Later on in the build process, when running `node install-node-deps.js`, I get an error message:
```
Building the projects in this solution one at a time. To enable parallel build, please add the "-m" switch.
  watchdog.cc
C:\Users\Administrator\oni2\node\node_modules\native-watchdog\src\common.h(1,10): fatal error C1083: Cannot open include file: 'node_api.h': No such file or directory [C:\Users\Administ
rator\oni2\node\node_modules\native-watchdog\build\watchdog.vcxproj]
```

Checking the mentioned .vcxproj file, I see it has directories:
```
  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
    <ClCompile>
      <AdditionalIncludeDirectories>C:\Users\Administrator\AppData\Local\node-gyp\Cache\14.15.4\include\node;C:\Users\Administrator\AppData\Local\node-gyp\Cache\14.15.4\src;C:\Users\Administrator\AppData\Local\node-gyp\Cache\14.15.4\deps\openssl\config;C:\Users\Administrator\AppData\Local\node-gyp\Cache\14.15.4\deps\openssl\openssl\include;C:\Users\Administrator\AppData\Local\node-gyp\Cache\14.15.4\deps\uv\include;C:\Users\Administrator\AppData\Local\node-gyp\Cache\14.15.4\deps\zlib;C:\Users\Administrator\AppData\Local\node-gyp\Cache\14.15.4\deps\v8\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```

Looks like it's looking for node_api.h in AppData\Local\node-gyp\Cache\14.15.4, but I've never installed that version. When I check the same path but with 16.0.0, the expected file is there.

Any idea what's supposed to be happening here? I didn't encounter this specific issue when building on Linux, so it may be a windows-specific thing. Also I see the 14.15.4 version number hardcoded in a few places in the repo, but don't have enough node experience to know if it's significant or not.

bryphe:
Hi @dglanzman ,

Looks like we were missing a step in the build instructions.

[...]

PR #3490, opened 2021-05-01
Fix onivim#3467 documentation

dglanzman:
Hi, thanks for the investigation in #3467. I found that node-gyp didn't understand the "add" command, but "install" seemed to do the right thing and fixed the issue for me.


bryphe (approved):
Ah, thanks @dglanzman for correcting this!

PR #3491, opened 2021-05-03
Update CLI instructions for new --new-window requirement

jesstelford:
The changes in #3469 mean using `--nofork` is no longer enough to get a new instance of `oni2` when one is already running.

This is notable when setting `oni2` as your `$EDITOR` or git editor.

To replicate the issue:

1. Open onivim2
2. On the cli, run `oni2 --nofork`
3. Note the existing instance of `oni2` is attached to

This documentation change ensures future users will have the correct setup.


bryphe:
Ah, great catch @jesstelford  - thank you !

Issue #2708, opened 2020-11-12
OSX - ARM Processor Support


bryphe:
Onivim doesn't yet support the new Mac ARM processors - a few things need to happen:

- [ ] Update to OCaml 4.12
- [ ] Update Revery to build with ARM: https://github.com/revery-ui/revery/issues/998
- [ ] Add an ARM OSX machine to Azure Pipeline 
- [ ] Update OSX release / packaging script to support multiple platforms

kolombet:
Using onivim on apple silicon via Rosetta and it works extremely fast, comparing to electron based editors and java IDE's. It will be amazing to have native version 👍

bryphe:
That's very exciting @kolombet ! Testament to Rosetta... amazing that it performs so well under emulation. 

Can't wait until we have an ARM build to take full advantage of the hardware :sunglasses:

seanchen1991:
Wanted to chime in as well and mention that onivim runs just as well on my M1 as it does on my Intel-based Mac 😄

syaiful6:
hi all, the initial support for M1 will be available in this commit: https://github.com/onivim/oni2/pull/3806

We will also need to wait my PR is esy-skia merged: https://github.com/revery-ui/esy-skia/pull/38

You can also test that in your M1 machine by following comment in https://github.com/onivim/oni2/pull/3806 and let me know if there are problems.

thiagodebastos:
Hi all! Has Apple ARM support been abandoned 🥺

brasilikum:
Seems like the whole proyect has been abandoned

PR #3531, opened 2021-05-11
refactor(dependency): ocaml -> 4.11

bryphe:
Gotta get Onivim caught up with: https://github.com/revery-ui/revery/pull/1066

This also will let us start using `memtrace` (since 4.11 has the `Gc.Memprof` apis) which will be really helpful for perf: https://github.com/janestreet/memtrace


EduardoRFS:
@bryphe why not a 4.12? 4.11 doesn't support macOS ARM64 while 4.12 and 4.10.2 do.

And at least for building OCaml the differences between native ARM64 and Rosetta emulation is huge.

bryphe:
> @bryphe why not a 4.12? 4.11 doesn't support macOS ARM64 while 4.12 and 4.10.2 do.

Good question @EduardoRFS - primarily to minimize risk - I like to take baby steps with critical dependencies like OCaml. Sometimes there can be surprises (a perf regression... a change in GC behavior that exposes a bug in our C FFI interop... packaging gets broken... etc).

I'll start a PR for 4.12 after this is in. Having the ARM build (https://github.com/onivim/oni2/issues/2708) is a good reason to not wait too long to bring in 4.12 :+1: I didn't realize 4.11 doesn't support macOS ARM64...

PR #3601, opened 2021-05-26
Update visual-mode.md

rogererens:
I think for Gnome and MS Windows users, it is good to know the alternative IMO


bryphe (approved):
For sure - thank you for adding this, @rogererens !

PR #3613, opened 2021-05-28
feat(input): add Emoji & Symbols panel on macOS

zbaylin:
This uses revery-ui/revery#1071 to add a command to open the <kbd>Emoji & Symbols</kbd> panel on macOS


src/Feature/Keyboard/Feature_Keyboard.re
```diff
@@ -0,0 +1,77 @@
+open Oni_Core;
```
zbaylin, line 1: I had to create this separate feature project because `Feature_MenuBar` depends on `Feature_Input`, so putting this there created a cyclic dependency. Maybe it makes sense to refactor some of the stuff #3546 on top of this if it gets merged?

src/Feature/Keyboard/Feature_Keyboard.re
```diff
@@ -0,0 +1,77 @@
...
+let update = (model, msg) =>
+  switch (msg) {
+  | Command(ShowEmojiAndSymbols) => (
+      model,
+      Effect(
+        Service_Keyboard.Effects.showEmojiAndSymbols
+        |> Isolinear.Effect.map(msg => Empty(msg)),
+      ),
+    )
+  | Empty () => (model, Nothing)
```
zbaylin, line 76: Is this the best way to handle effects with type `unit` (or, should I use something other than `unit`)?

src/Feature/Keyboard/Feature_Keyboard.re
```diff
@@ -0,0 +1,77 @@
...
+      command(~title="Emoji & Symbols", Commands.showEmojiAndSymbols);
+  };
+};
+
+module Keybindings = {
+  open Feature_Input.Schema;
+
+  let emojiAndSymbols =
+    bind(
+      ~key="<C-D-Space>",
```
zbaylin, line 46: @bryphe here's the keybind definition that doesn't seem to be working. I'll try to debug it

bryphe: Oh, interesting - did your change here address this? https://github.com/revery-ui/revery/pull/1079

src/Feature/Keyboard/Feature_Keyboard.re
```diff
@@ -0,0 +1,77 @@
...
+type model = int;
+let initial: model = 0;
+
+[@deriving show]
+type command =
+  | ShowEmojiAndSymbols;
+
+[@deriving show]
+type msg =
+  | Empty(unit)
```
bryphe, line 13: We could change this to `Empty` with no args - that would simplify it below (We use `Noop` though in some other parts of the code).

src/Feature/Keyboard/Feature_Keyboard.re
bryphe: Commented above - we could actually remove the argument entirely 👍

src/Feature/Keyboard/Feature_Keyboard.re
bryphe: Just tried the latest and saw that it is working - nice!

PR #3622, opened 2021-05-29
refactor(dock): use SVG icons in dock

zbaylin:
Now that we have SVG support in Revery, I think it makes sense to swap out some of the old Font Awesome icons we have for parameterize-able SVG icons. 

The Font Awesome icons are great, but they aren't meant to be rendered at the size we are currently, leading to stroke widths that look too thick, etc.

I chose to use [Feather Icons](https://feathericons.com) to replace the ones in the dock since they are configurable and MIT-licensed, but if anyone has any better ideas, let me know!

**Before:**
image

**After:**
image

The beautiful thing about the SVGs is that they're configurable -- I created an `SVGIcon.t` which takes a size, stroke-width, and color and re-writes the SVG accordingly.

### DIFF
--- a/src/Components/FeatherIcons.re
+++ b/src/Components/FeatherIcons.re
@@ -36,7 +36,7 @@ let folder: SVGIcon.t =
           }
     />;
 
-let codeSandbox: SVGIcon.t =
+let package: SVGIcon.t =
   (~size=24, ~strokeWidth=2, ~color=Colors.white, ()) =>
     <SVG
       src={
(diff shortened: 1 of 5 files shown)


src/Components/FeatherIcons.re
```diff
@@ -0,0 +1,148 @@
+open Revery;
+open Revery.UI.Components;
+
+let colorToCSS = color => {
+  let (r, g, b, _) = Color.toRgba(color);
+  Printf.sprintf(
+    "#%02X%02X%02X",
+    int_of_float(r *. 255.),
+    int_of_float(g *. 255.),
+    int_of_float(b *. 255.),
```
bryphe, line 10: Can we include alpha here? I think some themes leverage it

src/Components/FeatherIcons.re
```diff
@@ -0,0 +1,148 @@
...
+                size,
+                size,
+                colorToCSS(color),
+                strokeWidth,
+              ),
+            )
+          }
+    />;
+
+let codeSandbox: SVGIcon.t =
```
bryphe, line 46: We'll need to include the license in the `ThirdPartyLicenses.txt` - https://github.com/feathericons/feather/blob/master/LICENSE

(We should put a link the source file here too)

src/Components/FeatherIcons.re
```diff
@@ -0,0 +1,148 @@
...
+                colorToCSS(color),
+                strokeWidth,
+              ),
+            )
+          }
+    />;
+
+let gitPullRequest: SVGIcon.t =
+  (~size=24, ~strokeWidth=2, ~color=Colors.white, ()) =>
+    <SVG
```
bryphe, line 117: So cool to see the SVGs in use in Onivim 😎

src/Components/FeatherIcons.re
zbaylin: Unfortunately the SVG backend doesn't seem to like it when you specify an alpha value, either with 8 digit hex or `rgba(...)`. It just doesn't render the stroke, so nothing appears.

src/Components/FeatherIcons.rei
```diff
@@ -0,0 +1,4 @@
+let folder: SVGIcon.t;
+let codeSandbox: SVGIcon.t;
```
bryphe, line 2: I realized this was the branded codesandbox icon (https://codesandbox.io/) - I worry this would be confusing. 

We could try the package icon instead:
https://github.com/feathericons/feather/blob/master/icons/package.svg

bryphe: (CodeSandbox sure has a cool logo, though!)

zbaylin: Oh that's my bad, I didn't realize `codesandbox` was a proper noun 🤦 -- I'll change that out for the package one now!

src/Components/FeatherIcons.re
bryphe: Ah, gotcha - thanks for trying out!

src/Components/FeatherIcons.re
bryphe: Thanks!

src/Components/FeatherIcons.rei
bryphe: New icon looks great :+1:

bryphe (approved):

PR #3624, opened 2021-05-29
refactor(dependency): SDL2 -> 2.0.14

### LINKED ISSUE #2191: Key-repeat doesn't work under Wayland/Sway
It seems that OniVim2 under Wayland (Sway v1.5 in my case) doesn't support key-repeat, meaning that if you press and hold a key, it's typed only once.

bryphe:
This picks up @zbaylin 's work to upgrade SDL2: https://github.com/revery-ui/esy-sdl2/pull/19

Fixes #2191

### DIFF
--- a/CHANGES_CURRENT.md
+++ b/CHANGES_CURRENT.md
@@ -18,5 +18,6 @@
 ### Refactoring
 
 - #3595 - Dependency: OCaml -> 4.12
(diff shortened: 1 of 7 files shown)


zbaylin:
So excited to see all the new stuff we get from this upgrade!

Issue #2962, opened 2021-01-08
keybinding for next/previous tab


dfroger:
The [doc](https://onivim.github.io/docs/configuration/key-bindings#commands) seems outdated, `Control+Tab` and `Shift+Control+Tab` seems to be for the next/previous *used* tab.

What are the bindings for the next/previous tab in the order they are displayed?

Also, is there a way to display all default key binding in onvim2?

thanks!

dfroger:
Ok, just found it's [C-PAGEDOWN](https://github.com/onivim/oni2/blob/ea24875c7abd8fbc73f6c3f6a43174662fd8f989/src/Model/State.re#L215).

Just found also the tab bar is in the bottom (I did'nt saw it first...), and `gt`, `gT` navigate in it like in `vim`.

Should I do a `PR` to update the doc, or is there already a work on the doc?

dfroger:
By the way, I would have expect the inverse: DOWN for previous and UP for next

PR #3626, opened 2021-05-29
docs(A-input): add next/previous tab keybinding

paul-louyot:
Fix https://github.com/onivim/oni2/issues/2962


bryphe (approved):
Thanks @paul-louyot for adding this, looks great!

PR #3639, opened 2021-06-01
refactor(dependencies): upgrade Sparkle to 1.26.0

zbaylin:
This moves us from Sparkle 1.23.0 to 1.26.0. It also adds the Sparkle version to the `:version` page (ignore my bad theme)

When Sparkle is inactive:
image

When Sparkle is active:
image


bryphe (approved):
Thanks for upgrading and adding the version string, @zbaylin ! Looks good to me 👍

PR #3641, opened 2021-06-02
feat(search): optionally enable regex

### LINKED ISSUE #2599: Search: Add option for RegEx search
If I search for `log` I get 78 results.  If I search for `log(` I get 0 results.  This is because it's actually doing a regex search and not a simple text search.  The regex was invalid so the search didn't run at all.

If I try `log()` it finds 78 results again because while the regex was valid, the empty `()` has no effect so it really only searched for `log`.  If I escape the `\(` then it finally searches for the text string `log(`.
regex_find

I find this behavior unintuitive.  When first entering this screen, I had no idea it was doing a regex search.  If there was a red error  message saying "invalid regex" when I searched for `log(` I would've understood it wasn't just a basic text search.  Instead, I spent way too much time wondering why I couldn't search for an open parenthesis `(`.

Ideally, it'd be best to have a checkbox below the text input field which allowed me to enable/disable Regex Search.  That way, I would know the search supports regex but I could also choose whether or not I want that behavior.

zbaylin:
In the first part of my attempts to improve the global search pane, I added an option to enable or disable searching with a RegEx. A while ago, RegExs were the default but that was switched in #2826. This brings back the ability to search with a RegEx if you click the button like in VSCode. Here is a demo:

https://user-images.githubusercontent.com/4527949/120423367-4483b800-c338-11eb-886c-69799a71af7e.mov

I still need to figure out the Vim-style navigation, but the functionality is there.

Fixes #2599


zbaylin:
Forgot to mention this will fix #2599

bryphe:
This is awesome @zbaylin - thanks for adding this!

> I still need to figure out the Vim-style navigation, but the functionality is there.

For sure, we can think about in the scope of another PR though - I have to figure out the right gesture for this sort of navigation. Some things of the top of my head:
- The ability to switch to a 'normal-mode' and use `L` to navigate over to those buttons
- A key command, like `Alt+L` to switch to that button strip

A good compromise in the meantime, though, that we could add to this PR, would be to make the regexp button 'sneakable'

bryphe:
Just made a few minor tweaks:
- Made the new button 'sneakable' so that it can be accessed via keyboard-only, while figuring out a more robust way to navigate to it
- Added a tooltip
- Just tweaked the `msg` - moved from a prescriptive `EnableRegex` to descriptive `ToggleRegexButtonClicked` (following the Elm-style model of describing what happened, and letting the model update accordingly, vs prescribing how the model should update)

Already worked great though! :sunglasses: Feel free to tweak if you aren't happy with any of the changes I made

bryphe (approved):

zbaylin:
Awesome @bryphe! I just merged your tweaks with the changes I made locally. When I was on my flight I decided to add this to the vim navigation in the updater, so now we have that + sneaking! Very excited to get this in :)

bryphe:
> When I was on my flight I decided to add this to the vim navigation in the updater

Nice! Sorry to add a merge on your plate. Thanks for the exploring the vim navigation - I'm curious if there will be users who are confused with the extra horizontal window nav to get to the search input, but there's a few options for dealing with that (ie, we could move the button below the input text - might make sense if we have more buttons / options later).

Code change looks good, feel free to merge when you're ready 👍

PR #3684, opened 2021-06-14
feat(assets): update macOS icon to be more inline with the new Big Sur UI

### LINKED ISSUE #3679: Updated icon to match macOS Big Sur HIG
The current app icon on macOS Big Sur feels out of place. Apps are steadily transitioning to the new rounded square shape, so it would be a good idea to update oni2's icon to feel more platform-native.

Dock including oni2

There's already a different icon for macOS, and because of the shadow it actually looks pretty decent on the Oni2-purple background (although I can't help but notice the circle isn't a perfect circle).

Background: `#171633`

Screenshot of Sketch demonstrating the new icon shape

Apple's design kits (includes icon templates) can be found on [this page](https://developer.apple.com/design/resources/).

zbaylin:
This uses the template provided by Apple here (thanks @SeparateRecords!): https://developer.apple.com/design/resources/

newOnivim

Fixes #3679


bryphe (approved):
Looks great @zbaylin - thank you for updating this icon :100:

PR #3699, opened 2021-06-23
fix(vim/#3680): Respect silent flag for output command

### LINKED ISSUE #3680: Ability to run commands in silent mode
It would be cool to reproduce the behavior of vim for silent commands execution:

### Detailed steps to reproduce the problem:

- In Onivim, run `:silent execute '! echo onivim'`

### What happened:

The Output buffer opens

### What should have happened:

The Output buffer should not open

bryphe:
Fix #3680 - respect the `isSilent` flag for executing output commands

Pick up https://github.com/onivim/libvim/pull/264


paul-louyot:
Awesome!

PR #3705, opened 2021-06-27
feat(editor): add right click menus

zbaylin:
This adds right click menus based on the menu bar ones @bryphe created. 

Right now the only available option is to paste in the editor surface, but this is easily extendable to other areas and actions of the editor.

Demo:

https://user-images.githubusercontent.com/4527949/123551195-c05a0000-d73e-11eb-8f27-25dbff6ab30a.mov

Todo:
- [x]  fix accidental menu items clicks
  - fixed by offsetting the menu by one pixel on right click
- [x] fix cancelling behavior
  - stop displaying a menu when there is a right mouse down or a left mouse up
- [x] prevent visual mode selection on right click
  - only set `isMouseDown` when the mouse button is not `BUTTON_RIGHT`
- [x] fix pasting in normal mode
  - use context that's passed in rather than `Vim.Context.current()`
- [x] use native menus on macOS


src/Components/ContextMenu/Component_ContextMenu.re
```diff
@@ -143,19 +136,6 @@ module View = {
...
-          //     IconTheme.IconDefinition.(
-          //       <Text
-          //         style={Styles.icon(icon.fontColor)}
-          //         text={FontIcon.codeToIcon(icon.fontCharacter)}
-          //       />
-          //     )
-
-          //   | None => <Text style={Styles.icon(Colors.transparentWhite)} text="" />
-          //   };
-
```
bryphe, line 158: Thanks for cleaning this up 👍

src/Feature/MenuBar/NativeMenu.re
```diff
@@ -1,46 +1,17 @@
...
-    list(list(EditorInput.KeyPress.t)) =>
-    option(Revery.Native.Menu.KeyEquivalent.t) =
-    candidates => {
-      candidates
-      // Filter to single key press
-      |> List.filter_map(
-           fun
-           | [hd] => Some(hd)
-           | _ => None,
-         )
```
bryphe, line 14: Thanks for refactoring this to a common place 👍

bryphe (approved):
This is awesome @zbaylin - can't wait to wire this up to the explorer and other pieces of the UI. I like the approach you took to factor out Feature_ContextMenu from Feature_MenuBar. Looks great!

bryphe:
Oh, one minor thing I missed - could you please add an entry in `CHANGES_CURRENT.md` @zbaylin ?

Issue #2308, opened 2020-08-16
Paste into quick menu works only in insert mode or command mode


fanantoxa:
Steps to reproduce:
- Open Oni
- `Ctrl+P`
- `Ctrl+v` (need to have something in clipboard)
-  Nothing happen  < ------
- Press  `Esc` (close menu)
- press i (to enter insert mode)
- Press `Ctrl+P`
- Press `Ctrl+v`
- Text from clipboard pasted into input

Same for `Ctrl+Shift+p`, `Ctrl+Shift+f`

Comment from `Ryan C`
> I expect its this
> https://github.com/onivim/oni2/blob/749848d028340f789587ebaf81b9ff81499a5f79/src/Store/KeyBindingsStoreConnector.re#L71-L75
> Since it only runs on insertMode or commandLineFocus (i.e. the vim command line)
> Adding `textInputFocus` should fix that bug, I believe its a mistake its only `insertMode` and `commandLineFocus`
> It was originally always active, but that obviously clashes with the vim bind so it was changed and must have been missed

I'll fix it after https://github.com/onivim/oni2/pull/1716

fanantoxa:
Fixed by PRs related to https://github.com/onivim/oni2/issues/2428

PR #3733, opened 2021-07-03
feat(quickopen): Add binds to open in splits to Quick Open

### LINKED ISSUE #564: Commands: QuickOpen - open in vertical split / open in horizontal split
For the QuickOpen, it is often convenient to be able to open directly in a split.

We should have shortcut keys out-of-the-box for this, as well as have them be bindable commands.

Proposal:
- `quickOpen.verticalSplit` - opens the currently selected file in a vertical split (Default: `<c-v>`)
- `quickOpen.horizontalSplit` - opens the currently selected file in a horizontal split (Default: `<c-s>`)

These commands would only be enabled when the quickopen menu is up

### LINKED ISSUE #2232: QuickOpen: Open in new (layout) tab
When working with layout tabs

```json
{
  "oni.layout.layoutTabPosition": "top",
  "oni.layout.showLayoutTabs": "always",
  "workbench.editor.showTabs": false
}
```

it would be super useful to have a  `<c-t>` key binding in QuickOpen to open the selected file in a new tab. Otherwise, files will get opened in the same tab and opening a file with QuickOpen becomes a cumbersome operation.

CrossR:
Fix #564
Fix #2232

A rough first pass, but adds default bindings for selecting items from the quick open to open in current/vertical/horizontal/new tab.

Binds aren't finalised, so some input there would be nice (Ctrl-s and Ctrl-v are the defaults in Fzf in vim which is why I went with them).

### DIFF
--- a/src/Model/GlobalCommands.re
+++ b/src/Model/GlobalCommands.re
@@ -51,9 +51,7 @@ module List = {
   let selectHorizontal =
     register(
       "list.selectHorizontal",
-      ListSelect({
-        direction: Oni_Core.SplitDirection.Horizontal,
-      }),
+      ListSelect({direction: Oni_Core.SplitDirection.Horizontal}),
     );
   let selectVertical =
     register(
@@ -65,9 +63,7 @@ module List = {
   let selectNewTab =
     register(
       "list.selectNewTab",
-      ListSelect({
-        direction: Oni_Core.SplitDirection.NewTab,
-      }),
+      ListSelect({direction: Oni_Core.SplitDirection.NewTab}),
     );
 
   let selectBackground =
--- a/src/Model/GlobalCommands.re
+++ b/src/Model/GlobalCommands.re
@@ -48,26 +48,29 @@ module List = {
       "list.select",
       ListSelect({direction: Oni_Core.SplitDirection.Current}),
     );
+
+  let selectBackground =
+    register("list.selectBackground", ListSelectBackground);
+
   let selectHorizontal =
     register(
-      "list.selectHorizontal",
+      "oni.list.selectHorizontal",
       ListSelect({direction: Oni_Core.SplitDirection.Horizontal}),
     );
+
   let selectVertical =
     register(
-      "list.selectVertical",
+      "oni.list.selectVertical",
       ListSelect({
         direction: Oni_Core.SplitDirection.Vertical({shouldReuse: false}),
       }),
     );
+
   let selectNewTab =
     register(
-      "list.selectNewTab",
+      "oni.list.selectNewTab",
       ListSelect({direction: Oni_Core.SplitDirection.NewTab}),
     );
-
-  let selectBackground =
-    register("list.selectBackground", ListSelectBackground);
 };
 
 module Oni = {
--- a/src/Model/GlobalCommands.re
+++ b/src/Model/GlobalCommands.re
@@ -43,7 +43,33 @@ module List = {
   let focusDown = register("list.focusDown", ListFocusDown);
   let focusUp = register("list.focusUp", ListFocusUp);
 
-  let select = register("list.select", ListSelect);
+  let select =
+    register(
+      "list.select",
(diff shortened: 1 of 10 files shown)


src/Model/GlobalCommands.re
```diff
@@ -43,7 +43,33 @@ module List = {
...
+        direction: Oni_Core.SplitDirection.Vertical({shouldReuse: false}),
+      }),
+    );
+  let selectNewTab =
+    register(
+      "list.selectNewTab",
+      ListSelect({
+        direction: Oni_Core.SplitDirection.NewTab,
+      }),
+    );
```
bryphe, line 71: Awesome! Thanks for adding these 🎉

src/Model/State.re
```diff
@@ -125,6 +125,21 @@ let defaultKeyBindings =
         ~command=Commands.List.select.id,
         ~condition="listFocus" |> WhenExpr.parse,
       ),
+      bind(
+        ~key="<C-v>",
+        ~command=Commands.List.selectVertical.id,
+        ~condition="listFocus" |> WhenExpr.parse,
+      ),
```
bryphe, line 132: The only concern I have here is that the `Control+V` keybinding would conflict with paste on Windows/Linux (ie, regressing https://github.com/onivim/oni2/issues/2308)

A couple ideas:
- We could use `Control+Shift+V` for opening in vertical tab
- We could use the `Shift+CR` keybinding for vertical, and `Control+T` for new tab

src/Model/GlobalCommands.re
```diff
@@ -43,7 +43,33 @@ module List = {
...
+      "list.selectVertical",
+      ListSelect({
+        direction: Oni_Core.SplitDirection.Vertical({shouldReuse: false}),
+      }),
+    );
+  let selectNewTab =
+    register(
+      "list.selectNewTab",
+      ListSelect({
+        direction: Oni_Core.SplitDirection.NewTab,
```
bryphe, line 69: Looks like this will address #2232 🎉

bryphe (changes requested):
Thanks for taking this on, @CrossR ! Approach looks great to me. Just had a comment about the `Control+V` binding -

src/Model/State.re
```diff
@@ -125,6 +125,21 @@ let defaultKeyBindings =
...
+      bind(
+        ~key="<C-s>",
+        ~command=Commands.List.selectHorizontal.id,
+        ~condition="listFocus" |> WhenExpr.parse,
+      ),
```
bryphe, line 137: This also conflicts with the `save` keybinding, but I think this one is OK - understandable that save wouldn't occur when the quickmenu is up.

CrossR:
Did a quick survey of some common binds for this sort of thing and updated them accordingly, but with Shift-enter as vertical open by default since we use c-v as copy.

That said, obviously if a user wanted to swap they could just do something like:

```json
  { "key": "<C-s>", "command": "list.selectHorizontal", "when": "listFocus" }
```

The only other thing I wanted to check here is if the command names are okay. VSCode has no equivalent, so wasn't sure if they should be prefaced with `oni` or something.

bryphe:
> Did a quick survey of some common binds for this sort of thing and updated them accordingly, but with Shift-enter as vertical open by default since we use c-v as copy.

Awesome! Thanks for looking into this @CrossR . Looks good to me.

> The only other thing I wanted to check here is if the command names are okay. VSCode has no equivalent, so wasn't sure if they should be prefaced with oni or something.

Hmm, good point - VSCode does have a `list.*` set of commands, so prefixing it with `oni.` will make it clear these are Onivim-specific, and keep us safe if VSCode does introduce keybindings with the same name.

bryphe (approved):

CrossR:
Swapped the names, added some basic docs, just waiting for the CI to be happy then I'll squash/merge.

bryphe:
Sorry about the build issues; the CentOS build failure looks unrelated. I'll override and merge. 

Thanks for the awesome contribution @CrossR !

PR #3740, opened 2021-07-07
Remove dead link from getting started doc.

Doerge:


bryphe:
Thanks @Doerge for catching this!

PR #3746, opened 2021-07-07
fix(extension): Fix edit application from trailing spaces plugin

bryphe:
__Issue:__ The [trailing spaces](https://open-vsx.org/extension/shardulm94/trailing-spaces) plugin isn't working as expected - there is a statusbar message reporting that trailing spaces have been removed... but they aren't actually removed.

__Defect:__ There were two problems:
1) The `$tryApplyEdits` API wasn't parsing single-edits correctly in the case where the text property is null as opposed to a string
2) In the `null` case, if there was no text on the line, the line would be completely removed

__Fix:__
1) Correctly parse the text property - handling either null or a string
2) When text is null, treat it as an empty string


manaskarekar:
Thanks for the quick turnaround!

PR #3767, opened 2021-07-15
fix(minimap/#3648): Fix minimap diff marker position

### LINKED ISSUE #3648: EditorDiffMarkers Offset
Looks like there is a slight miscalculation somewhere (I expect in `EditorDiffMarkers.renderMinimap`?) such that the diff markers are sliding when the scroll becomes too large.

https://user-images.githubusercontent.com/10038688/120556902-11ebc500-c3f5-11eb-9609-d5de0d951e58.mov

Here I have a change on the 2nd line of my function (i.e. the line just under the line with red syntax hls). If I scroll, that slides around rather than remaining locked to the correct line.

Its more noticable at the end of the video where I scroll to the bottom of my file and the markers aren't lined up with the comments. This was a 20-30 line file that I added 200 lines of comments on at the bottom, so the entire bottom part of the buffer is new.

CrossR:
Fix #3648.

Very tiny change! Perhaps I'm missing something here, but it looks like the diff markers don't need to include this offset, at least in the calc for the y pos.

Before / After:

https://user-images.githubusercontent.com/10038688/125708244-36845971-c357-4b89-8e53-a0fa3e24775e.mov

If you look at the two add markers, in the before on the left they move about when scrolled.
On the right, they stay locked on the actual lines that changed.
They should be attached to the three comment lines (above and below the long comment line).
(Sorry for the short video, only allowed 10MB here!).

Assuming this is sensible, just needs the changelog updating.

### DIFF
--- a/CHANGES_CURRENT.md
+++ b/CHANGES_CURRENT.md
@@ -43,6 +43,7 @@
 - #3753 - Extension: Don't bubble up extension runtime errors to notifications
 - #3756 - Search: Follow symlinks with ripgrep by default (fixes #1588)
 - #3757 - Buffers: Fix welcome buffer appearing when splitting (fixes #3631)
+- #3767 - Minimap: Fix diff marker positions on scroll (fixes #3648)
 
 ### Performance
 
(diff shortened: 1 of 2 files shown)


bryphe (approved):
Nice catch, @CrossR ! You're right - the `-. scrollY` normalization isn't needed, because that's already accounted for with `ImmediateList.render` - the `y` value provided already has the scroll baked in.

Fix looks good to me - thank you!

Issue #2525, opened 2020-09-30
Git checkout does nothing


danielo515:
OX: macosx
Onivim version: downloaded yesterday

When I open the command palette and I select git checkout from the list of options nothing happes. I was expecting it to ask me a branch to checkout to, but that didn't happened.
Not sure what else can I report, but just let me know.

edwinclement08:
Same here, git checkout does nothing

edwinclement08:
This is a BUG, not an enhancement

PR #3785, opened 2021-07-24
fix(A-scm/#2525): see git branches when using git checkout

Jakkestt:
See the git branches when using git checkout

### DIFF
--- a/extensions/git/out/commands.js
+++ b/extensions/git/out/commands.js
@@ -1330,13 +1330,9 @@ class CommandCenter {
             picks.push(createBranch, createBranchFrom, checkoutDetached);
         }
         picks.push(...createCheckoutItems(repository));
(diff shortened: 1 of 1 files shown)


bryphe:
Hi @Jakkestt ,

Thanks for the PR and for investigating #2525 !

There's a couple of concerns that I have with the fix:
1) Modifying the git extension will make it difficult to merge updates from the upstream VSCode git extension
2) There are likely other extensions that create quickpick menus in a similar way, that are also broken.

Ideally, we could fix the root cause - that we're not handling the API surface the Git extension is using. I'm testing this out in #3791 , which implements some of the missing APIs. It was definitely helpful to have this PR to see the broken APIs!

Jakkestt:
Well shit. I followed all the guides like a good boy and in the end I don't get shit.

PR #3822, opened 2021-09-20
ci: Update path to delete

Et7f3:
The suffix was removed https://github.com/actions/virtual-environments/blob/9fe940e8a3cb67b6ecf0dbbcc9cb7efcc68aa008/images/win/scripts/Installers/Install-MysqlCli.ps1#L29


Et7f3:
merge of this PR https://github.com/onivim/esy-oniguruma/pull/7 can help for CI check. here timestamp aren't preserved by git on checkout so configure think it need to regenerate. We trick it by setting the same time for all files.

bryphe:
Thanks for investigating this and catching where it was removed in the VMs :100:

PR #3825, opened 2021-09-29
Fix linguist tool configuration

prekel:
After these changes: 
image

Up to that moment, C is main, mostly because of [vendored sqlite](https://github.com/onivim/oni2/blob/fda6bf7f3bb8caa6bc66225d36886a07a7a1579a/bench/reason-treesitter/sqlite3.c)

image


bryphe:
Thank you @prekel for adding this - much more accurate now!

PR #3829, opened 2021-10-05
chore: CI - Use common step for installing esy (upgrade esy to 0.6.10)

bryphe:
The hygiene check was using esy@0.6.7, but the rest of the build pipeline uses esy@0.6.10 - bring it back in sync


bryphe:
Back to green builds, hopefully

